PR #2, opened 2022-01-14
Add structure

kmdeck:
This PR defines the concept of an AbstractModel (consisting of a domain, parameters, prognostic variables, auxiliary variables) and introduces a number of core methods that every concrete type of AbstractModel  must have, including a way to create the state vector structure (assumed to be a ClimaCore FieldVector with a NamedTuple like format) and advance this state vector using the OrdinaryDiffEq package. 

This should allow for an LSM, contructed from many individual AbstractModels and itself a type of AbstractModel, to run using the same interface as each of the components when run in standalone mode.

### DIFF
--- a/src/ClimaLSM.jl
+++ b/src/ClimaLSM.jl
@@ -1,5 +1,204 @@
 module ClimaLSM
 
-greet() = print("Hello World!")
+using ClimaCore
+import ClimaCore: Fields
+include("Domains.jl")
+using .Domains
+
+export AbstractModel,
+    make_rhs,
+    make_ode_function,
+    make_update_aux,
+    initialize_prognostic,
+    initialize_auxiliary,
+    initialize,
+    LandModel,
+    prognostic_vars,
+    auxiliary_vars
+## Default methods for all models - to be in a seperate module at some point.
+
+abstract type AbstractModel{FT <: AbstractFloat} end
+
+"""
+   prognostic_vars(m::AbstractModel)
+
+Returns the prognostic variable symbols for the model in the form of a tuple.
+"""
+prognostic_vars(m::AbstractModel) = ()
+
+"""
+   auxiliary_vars(m::AbstractModel)
+
+Returns the auxiliary variable symbols for the model in the form of a tuple.
+"""
+auxiliary_vars(m::AbstractModel) = ()
+
+"""
+    make_rhs(model::AbstractModel)
+
+Return a `rhs!` function that updates state variables.
+
+`rhs!` should be compatible with OrdinaryDiffEq.jl solvers.
+"""
+function make_rhs(model::AbstractModel)
+    function rhs!(dY, Y, p, t) end
+    return rhs!
+end
+
+"""
+    make_update_aux(model::AbstractModel)
+
+Return an `update_aux!` function that updates auxiliary parameters `p`.
+"""
+function make_update_aux(model::AbstractModel)
+    function update_aux!(p, Y, t) end
+    return update_aux!
+end
+
+"""
+    make_ode_function(model::AbstractModel)
+
+Returns an `ode_function` that updates auxiliary variables and
+updates the prognostic state.
+
+`ode_function!` should be compatible with OrdinaryDiffEq.jl solvers.
+"""
+function make_ode_function(model::AbstractModel)
+    rhs! = make_rhs(model)
+    update_aux! = make_update_aux(model)
+    function ode_function!(dY, Y, p, t)
+        update_aux!(p, Y, t)
+        rhs!(dY, Y, p, t)
+    end
+    return ode_function!
+end
+
+"""
+    initialize_prognostic(model::AbstractModel, state::Union{ClimaCore.Fields.Field, Vector{FT}, FT})
+
+Returns a FieldVector of prognostic variables for `model` with the required
+structure, with values equal to `state`. This assumes that all prognostic
+variables are defined over the entire domain. 
+
+The default is a model with no prognostic variables, in which case 
+the returned FieldVector contains only an empty array.
+
+Adjustments to this - for example because different prognostic variables
+have different dimensions - require defining a new method.
+"""
+function initialize_prognostic(
+    model::AbstractModel{FT},
+    state::Union{ClimaCore.Fields.Field, Vector{FT}, FT},
+) where {FT}
+    keys = prognostic_vars(model)
+    if length(keys) == 0
+        return ClimaCore.Fields.FieldVector(; model.model_name => FT[])
+    else
+        values = map((x) -> similar(state), keys)
+        return ClimaCore.Fields.FieldVector(;
+            model.model_name => (; zip(keys, values)...),
+        )
+    end
+
+end
+
+"""
+    initialize_auxiliary(model::AbstractModel,state::Union{ClimaCore.Fields.Field, Vector{FT}, FT})
+
+Returns a FieldVector of auxiliary variables for `model` with the required
+structure, with values equal to `state`. This assumes that all auxiliary
+variables are defined over the entire domain. 
+
+The default is a model with no auxiliary variables, in which case 
+the returned FieldVector contains only an empty array.
+
+Adjustments to this - for example because different auxiliary variables
+have different dimensions - require defining a new method.
+"""
+function initialize_auxiliary(
+    model::AbstractModel{FT},
+    state::Union{ClimaCore.Fields.Field, Vector{FT}, FT},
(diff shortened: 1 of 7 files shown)


src/ClimaLSM.jl
```diff
@@ -1,5 +1,164 @@
...
+
+"""
+    initialize_prognostic(model::AbstractModel,_)
+
+Returns a FieldVector of prognostic variables for `model` with the required
+structure.  This is a stub. We will change it, if possible,
+ as we determine what components require.
+An idea is to use something like `similar(coords)`.
+"""
+function initialize_prognostic(model::AbstractModel, _)
```
kmdeck, line 84: @jb-mackay in the end, it may be best for each submodel to define this method and not have all submodels be using the same default - but we can revisit. It actually might be nice to have the fallback/default be not adding *any* variables, because some submodels may have no aux vars, or even no prognostic vars (if a certain submodel involves only algebraic equations, e.g., not a differential equation).

src/Soil.jl
```diff
@@ -0,0 +1,69 @@
...
+        domain,
+        :soil,
+    )
+end
+
+prognostic_vars(model::RichardsModel) = (:ϑ_l,)
+
+auxiliary_vars(model::RichardsModel) = (:z,)
+
+function initialize_states(model::RichardsModel{FT}) where {FT}
```
kmdeck, line 42: I kept this "initialize_states" wrapper for now. The issue of only creating the space once is somewhat limiting. What we had sketched out today for just the soil model worked (first get the coordinates, then initialize prognostic by passing in coordinates and `model` to some init_prognostic(model, coords), and same for aux), but it doesnt nicely generalize to an LSM with multiple components. (get the coordinates for all components? then pass all to init_prognostic(land, all_coordinates) which calls init_prognostic(soil, coordinates_soil), etc?).

src/Roots.jl
```diff
@@ -0,0 +1,52 @@
...
+    initialize_states,
+    initialize_auxiliary
+export RootsModel
+
+"""
+    RootsModel
+
+A standin for the roots model.
+"""
+struct RootsModel{FT, ps} <: AbstractModel{FT}
```
kmdeck, line 22: not sure if this needs a domain at this stage. can add in when it is clearer!

src/ClimaLSM.jl
```diff
@@ -1,5 +1,164 @@
...
+
+"""
+    initialize_auxiliary(model::AbstractModel,_)
+
+Returns a FieldVector of auxiliary variables for `model` with the required
+structure.  This is a stub. We will change it, if possible,
+ as we determine what components require.
+An idea is to use something like `similar(coords)`."""
+function initialize_auxiliary(model::AbstractModel, _)
+    return ClimaCore.Fields.FieldVector(; model.model_name => [])
```
kmdeck, line 97: is this the best way to not add variables? It would imply that in this case, p.model_name is Any[]. Is it better not to even have `model_name` as part of `p`?

src/ClimaLSM.jl
```diff
@@ -1,5 +1,164 @@
...
+    Y_roots, p_roots, coords_roots = initialize_states(land.roots)
+    Y = ClimaCore.Fields.FieldVector(;
+        soil = Y_soil.soil,
+        roots = Y_roots.roots,
+    )
+    p = ClimaCore.Fields.FieldVector(;
+        soil = p_soil.soil,
+        roots = p_roots.roots,
+    )
+    coords =
```
kmdeck, line 133: not sure if we really need coords, but it might be useful to pass them back to the user so they have them handy for plotting, e.g.. These would be (x,y,z) locations (z=z_srf for surface variables) or (z) alone - column mode.

src/ClimaLSM.jl
jb-mackay: Makes sense to me, I think we should definitely start there and then see if we can generalize once we have a few examples.

src/ClimaLSM.jl
jb-mackay: My instinct is to not have `model_name` as part of `p`, since a particular `p` already belongs to a particular model, right? The only case were I would want `p.model_name` is a multi-component LSM run.

src/ClimaLSM.jl
jb-mackay: Is there a better place to return `coords`? Not sure that returning them is necessarily a responsibility of `initialize_states`.

kmdeck: Agreed that this might not be the right place..but I kept running into the block of computing the space 1x. We could split coordinates out like:
```
coordinates = compute_coordinates(soil) # computes the space first
Y, p = initialize_states(soil, coordinates) # or else store the coordinates or space in soil rather than the domain
```
But then for the LSM, we would have to do
```
coordinates = compute_coordinates(land) # what form would this take given distinct domains for each component?
Y,p = initialize_states(land, coordinates)
```

or
```
function initialize_states(land)
    coordinates_soil = compute_coordinates(land.soil)
    Y,p = initialize_states(land.soil, coordinates_soil)
    #etc
end
```
but maybe there is a better way?

src/ClimaLSM.jl
kmdeck: Agreed - we dont really want the hierarchy in a single component run. But I thought we had worked out that we need to pass the entire `p` in the LSM run to each component RHS (due to shared variables). and so for consistency in the RHS function, we also need to pass the entire `p` in single component mode. that is, the arguments to rhs_component! can be `dY,Y,p,t` and we always store variables in Y and p as Y.component, p.component, or we can pass
`dY.component, Y.component,p,t` and just unpack `p`....

Maybe we can give the `p` we pass additional structure? like pass p_component (with no hierarchy) and p_interactions as separate pieces, and in single component mode p_interactions could be `nothing`.

test/runtests.jl
```diff
@@ -1,5 +1,30 @@
 using Test
 
-@testset "Stand-in test" begin
-    @test 1 == 1
-end
+using ClimaLSM
+using ClimaLSM.Domains: Column
+include(joinpath("../src", "Soil.jl"))
```
jb-mackay, line 5: Can do `using ClimaLSM.Soil`, unless we'd rather not include the subcomponent models in the `ClimaLSM` module in anticipation of splitting the repo later.

kmdeck: im good with whatever you think makes it easier to develop now!

src/Roots.jl
jb-mackay: Yes, we can develop this in a follow up PR where we adapt Anna's code to the interface.

kmdeck:
some of the ?s we are running into are just hard to evaluate also without concrete examples...I think we can expect to modify a lot of this as we introduce models, different domains, Simulations, etc, but that isnt a negative. Im OK with merging something that seems to do the basic job and creating an issue to flag things we arent super happy with for later?

src/ClimaLSM.jl
kmdeck: see https://github.com/CliMA/ClimaLSM.jl/issues/3, will leave as is.

src/Soil.jl
kmdeck: https://github.com/CliMA/ClimaLSM.jl/issues/3

src/ClimaLSM.jl
kmdeck: https://github.com/CliMA/ClimaLSM.jl/issues/4 - will revisit

src/ClimaLSM.jl
kmdeck: https://github.com/CliMA/ClimaLSM.jl/issues/3 to be revisited once more is in place. what we have is a place to start from.

kmdeck:
> some of the ?s we are running into are just hard to evaluate also without concrete examples...I think we can expect to modify a lot of this as we introduce models, different domains, Simulations, etc, but that isnt a negative. Im OK with merging something that seems to do the basic job and creating an issue to flag things we arent super happy with for later?

See https://github.com/CliMA/ClimaLSM.jl/issues/3 and https://github.com/CliMA/ClimaLSM.jl/issues/4.

kmdeck:
@jb-mackay let me know if you think this is good to merge! I made it so that both Roots and Soil can use the default initialize methods, and these defaults now use the `prognostic_vars(model)` and `auxiliary_vars(model)` functions. What is in there requires that each component has a domain from which we can derive coordinates, and the default assumes that all variables are ~similar(coordinates). Models that don't satisfy this need their own methods.

jb-mackay (approved):

PR #5, opened 2022-01-19
Add root model

kmdeck:
This PR adds in the standalone root model, based off of the script @gagnelandmanna wrote here: https://github.com/CliMA/LandHydrology.jl/blob/bulk_roots/src/plant_hydraulics.jl .

This model is used to simulate water flow within a bulk plant, with roots of different depths, and stem compartment, and a leaf compartment. A soil pressure at the root tips and a transpiration rate from the leaf are prescribed. 

This model uses the AbstractModel functionality laid out in ClimaLSM.jl.

### DIFF
--- a/src/Roots.jl
+++ b/src/Roots.jl
@@ -1,8 +1,11 @@
 module Roots
 using ClimaLSM
 using ClimaCore
+using UnPack
 import ClimaCore: Fields
-using ClimaCore.Domains
+using ClimaLSM.ComponentExchanges: AbstractComponentExchange
+using ClimaLSM.Domains: AbstractPlantDomain, RootDomain
+
 import ClimaLSM:
     AbstractModel,
     initialize_prognostic,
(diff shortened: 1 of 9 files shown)


kmdeck:
@gagnelandmanna Do these plots look correct to you? It seems good in that starting at equilibrium, nothing changes until T changes. In the test, I also check that it returns to a new equilibrium after T stops changing. But let me know what you think :)

These were generated with `test/root_test.jl`, which does the same run as in the script you sent Ben and I via email. 

absolute_water_content
pressure
relative_water_content

gagnelandmanna:
Hey!! This looks fantastic! I just compared plots you sent to plots I had run on my end before sending to you and Ben and they are the same indeed.

gagnelandmanna (approved):
I read through the code, it looks great and easy to use. The descriptions are awesome too. I had a minor doubt about the description of z_up and z_do but think what you have is correct. Thanks for putting this together, I look forward to using it.

src/Roots.jl
```diff
@@ -13,30 +16,312 @@ import ClimaLSM:
...
+    RootsParameters
+
+"""
+    RootsParameters{ft <: AbstractFloat}
+
+
+A struct for holding parameters of the Root Model. Eventually to be used with ClimaParameters.
+"""
+struct RootsParameters{ft <: AbstractFloat}
+    a_root::ft
```
jb-mackay, line 33: It's standard in CliMA and, from what I've seen, a general Julia convention, to capitalize the parametric type identifiers e.g. `FT` rather than `ft`. We do this elsewhere in ClimaLSM, too

jb-mackay: Same comment for `RootsModel` struct definition.

src/Roots.jl
```diff
@@ -13,30 +16,312 @@ import ClimaLSM:
...
     domain::d
+    boundary_exchanges::b
     model_name::Symbol
 end
 
 function RootsModel{FT}(;
     param_set = param_set,
-    domain = domain <: AbstractDomain{FT},
+    domain::AbstractPlantDomain{FT} = domain,
+    boundary_exchanges::AbstractComponentExchange{FT} = boundary_exchanges,
```
jb-mackay, line 68: Another small style thing - for each keyword argument it's not necessary to have, e.g. `param_set = param_set`, we can just have `param_set` as a keyword and there is no default.

src/Roots.jl
kmdeck: got it! will fix the other instances too.

jb-mackay (approved):
Looks good! Just a few style notes :D

src/Roots.jl
```diff
@@ -13,30 +16,312 @@ import ClimaLSM:
...
+"""
+struct RootsParameters{ft <: AbstractFloat}
+    a_root::ft
+    b_root::ft
+    a_stem::ft
+    b_stem::ft
+    size_reservoir_stem_moles::ft
+    size_reservoir_leaf_moles::ft
+    K_max_root_moles::ft
+    K_max_stem_moles::ft
```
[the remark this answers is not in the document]
kmdeck: good idea! adding some notes

PR #10, opened 2022-01-26
Pond model

kmdeck:
Add a pond model. this mimicks pieces of both the river and snow model in terms of domain (surface only) and how it will eventually interact with the soil model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,38 @@
 # ClimaLSM
-an in-progress prototype interface for the land model
+an in-progress prototype interface for running land models in integrated (multi-
+component) or standalone (single component) modes.
+
+## Models
+```
+Model <: AbstractModel
+```
+
+All models have the same abstract supertype, and have shared functionality.
+There is the the option to define new methods particular to your model, or to
+fall back on defaults. The functions each new Model type can define are:
+- make_rhs()
+- make_update_aux()
+- initialize_prognostic()
+- initialize_auxiliary()
+- initialize()
+- prognostic_vars()
+- auxiliary_vars()
+    
+Each model will also have some notion of a domain with coordinates, parameter sets,
+and boundary conditions or other prescribed drivers.
+
+Examples:
+
+Component Models:
+RichardsModel <: AbstractSoilModel <: AbstractModel [runnable w/o LandModel wrapper as well]
+
+RootsModel <: AbstractVegetationModel <: AbstractModel  [runnable w/o LandModel wrapper as well]
+
+PondModel <: AbstractSurfaceWaterModel  <: AbstractModel  [runnable w/o LandModel wrapper as well]
+
+Combined Models:
+
+RootSoilModel <: AbstractModel (constructs the individual ComponentModels based on arguments)
 
 |||
 |---------------------:|:----------------------------------------------|
--- a/src/SurfaceWater/Pond.jl
+++ b/src/SurfaceWater/Pond.jl
@@ -1,0 +1,67 @@
+module Pond
+using ClimaLSM
+using ClimaCore
+using UnPack
+using DocStringExtensions
+import ClimaCore: Fields
+
+import ClimaLSM: AbstractModel, make_rhs, prognostic_vars, name
+using ClimaLSM.Domains
+import ClimaLSM.Domains: coordinates
+export PondModel, PrescribedRunoff, surface_runoff
+
+abstract type AbstractSurfaceWaterModel{FT} <: AbstractModel{FT} end
+abstract type AbstractSurfaceRunoff{FT <: AbstractFloat} end
+"""
+    PondModel{FT, R} <: AbstractSurfaceWaterModel{FT}
+
+A stand-in model for models like the snow or river model. In 
+standalone mode, a prescribed soil infiltration rate
+ and precipitation rate
+control the rate of change of the pond height variable `η` via an ODE.
+
+In integrated LSM mode, the infiltration into the soil will be computed
+via a different method, and also be applied as a flux boundary condition
+for the soil model. 
+"""
+struct PondModel{FT, R} <: AbstractSurfaceWaterModel{FT}
+    runoff::R
+end
+
+function PondModel{FT}(; runoff::AbstractSurfaceRunoff{FT}) where {FT}
+    return PondModel{FT, typeof(runoff)}(runoff)
+end
+
+
+"""
+    PrescribedRunoff <:  AbstractSurfaceRunoff
+
+The required input for driving the simple pond model: precipitation, as a
+function of time, soil effective saturation at a depth `Δz` below the surface,
+as a function of time, and soil parameters, which affect infiltration.
+"""
+struct PrescribedRunoff{FT} <: AbstractSurfaceRunoff{FT}
+    "Time dependent precipitation magnitude, given in m/s. Negative is into the soil"
+    precip::Function
+    "Time dependent infiltration magnitude, given in m/s. Negative is into the soil."
+    infil::Function
+end
+
+ClimaLSM.prognostic_vars(model::PondModel) = (:η,)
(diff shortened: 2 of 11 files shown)


src/Pond.jl
```diff
@@ -0,0 +1,119 @@
...
+    S = boundary_exchanges.S(t)
+    Ksat, α, n, Δz = boundary_exchanges.soil_params
+    m = FT(1) - FT(1) / n
+    i_c =
+        hydraulic_conductivity(Ksat, m, S) .*
+        (max.(FT(0.0), Y.pond.η) .- matric_potential(α, n, m, S) .+ Δz) ./ Δz
+    println(typeof(i_c))
+    println(typeof(Y.pond.η))
+    println(typeof(boundary_exchanges.P(t)))
+    infiltration =
```
kmdeck, line 89: @jb-mackay This was an example where I had to use `Y.pond.eta[1]` a lot originally. 
```
 i_c =
        hydraulic_conductivity(Ksat, m, S) *
        (max(FT(0.0), Y.pond.η[1]) - matric_potential(α, n, m, S) + Δz) / Δz
    if Y.pond.η[1] > eps(FT) # when there is a pond, infiltration is pond/soil limited
        infiltration = i_c
    else # when there is no pond, infiltration is precip/soil limited
        infiltration = min(i_c, boundary_exchanges.P(t))
    end
```

But I tried to rewrite it so we do not...it seems like it should work, but this line is leading to errors - only when run within the function.  calling `infiltration_at_point` by itself with objects of the same type `infiltration_at_point.(a::Vector{FT}, b::Vector{FT}, c::FT)` does not produce an error.

kmdeck: fixed this! all set

src/Domains.jl
```diff
@@ -48,9 +47,19 @@ function coordinates(domain::RootDomain{FT}) where {FT}
...
+"""
+   SurfaceDomain{FT} <: AbstractDomain{FT}
 
-### Another example, but a ClimaCore FD Column space/domain may be general enough
-### that we keep it as a concrete type of AbstractDomain, rather than an AbstractSoilDomain
+A placeholder for a 2d domain that will coincide with the 
+surface of the soil domain.
+"""
+struct SurfaceDomain{FT} <: AbstractDomain{FT}
+    height::Vector{FT}
```
jb-mackay, line 57: Can you explain this a bit and its use cases? I'm a little confused that a 2D surface domain has a `height` field.

kmdeck: hm, so surface variables are defined on the 2d surface `(x,y, z = z_sfc(x,y))`, where `height = z_sfc` (I can rename `height` if we keep this).  There will be topography eventually. 

I think there are two questions here.
1. Does the model require this `z` info to compute RHSs? For example,  the soil model needs topography, `z_sfc(x,y)`, and should have a warped surface to the domain, but Im not sure the actual surface models need it because I dont think `z` shows up in the RHS terms for them. *As long as the correct surface variables are linked up to the right column of soil (in terms of horizontal location - I dont know how we will do this in 3D LSMs), we may not need the surface coordinates at all in compute RHS's.*

2. We initialize variables by computing `coordinates(domain)` and (`similar(coords)`). So we need something here...whether it is `(x,y)` or `(x,y, z_sfc(x,y))`. But right now we are only working with columns, so we only have a notion of `z`?

kmdeck: for the first question - we can use a Plane Domain, and give it a single element and single nodal point -> single set of (x,y) is returned by coordinates(model)

The second question is still pertinent! It makes sense to return coordinates as (x,y) pair, but then we cant do similar(coordinates), because coordinates acts like a NamedTuple. this also came up with the 3d domain for soil - returning (x,y,z) coordinates, but wanting similar(coords.z) e.g.

kmdeck:
branch is WIP - rebased to new main, simplified pond model for our example, rebuilt based on changes we've made. Got overzealous and added in 2d spectral element domain for the pond, whcih we can make have a single nodal point and single element for "column" mode. Try to get this to run with pond standalone, then go back to pond_soil_lsm...

src/Domains.jl
kmdeck: fixed by doing
zero_state = map(_ -> zero(FT), coords)
and then do similar(zero_state), instead of similar(coords)

kmdeck:
Once we merge this: https://github.com/CliMA/ClimaLSM.jl/pull/22
this PR will get a lot smaller (no domain changes, coordinate changes)
and then we can discuss whether or not the pond domain should be a plane or something else...

we could also try and add in the pond_soil_lsm model as well...

src/SurfaceWater/Pond.jl
```diff
@@ -0,0 +1,67 @@
...
+
+"""
+    PrescribedRunoff <:  AbstractSurfaceRunoff
+
+The required input for driving the simple pond model: precipitation, as a
+function of time, soil effective saturation at a depth `Δz` below the surface,
+as a function of time, and soil parameters, which affect infiltration.
+"""
+struct PrescribedRunoff{FT} <: AbstractSurfaceRunoff{FT}
+    "Time dependent P⃗ = Pẑ, given in m/s"
```
jb-mackay, line 44: "Time dependent precipitation P = ..."

src/pond_soil_model.jl
```diff
@@ -0,0 +1,117 @@
...
+                Y.surface_water.η,
+                i_c,
+                -land.surface_water.runoff.precip(t),
+            )
+    end
+    return update_aux!
+end
+
+struct PrognosticRunoff{FT} <: Pond.AbstractSurfaceRunoff{FT}
+    precip::Function
```
jb-mackay, line 95: This is called `P` in the PrescrivedRunoff case (unify)

jb-mackay: use `precip` as there are other `p`s

test/pond_soil_lsm.jl
```diff
@@ -0,0 +1,56 @@
...
+        #unsaturated zone only, assumes water table starts at z_∇
+        z_∇ = FT(-3)# matches zmin
+        S = FT((FT(1) + (vg_α * (z - z_∇))^vg_n)^(-vg_m))
+        ϑ_l = S * (ν - θ_r) + θ_r
+        return FT(ϑ_l)
+    end
+    Ysoil.soil.ϑ_l .= hydrostatic_profile.(z, Ref(params))
+end
+init_soil!(Y, coords.soil, land.soil.param_set)
+Y.surface_water.η .= 0.0
```
jb-mackay, line 47: `init_pond`

README.md
```diff
@@ -1,5 +1,38 @@
...
+- prognostic_vars()
+- auxiliary_vars()
+    
+Each model will also have some notion of a domain with coordinates, parameter sets,
+and boundary conditions or other prescribed drivers.
+
+Examples:
+
+Component Models:
+RichardsModel <: AbstractModel [runnable w/o LandModel wrapper as well]
```
jb-mackay, line 27: RichardsModel <: AbstractSoilModel <: AbstractModel

jb-mackay (approved):

test/pond_soil_lsm.jl
```diff
@@ -0,0 +1,56 @@
...
+
+saved_values = SavedValues(FT, ClimaCore.Fields.FieldVector)
+ode! = make_ode_function(land)
+t0 = FT(0);
+tf = FT(200)
+dt = FT(1);
+cb = SavingCallback((u, t, integrator) -> integrator.p, saved_values)
+prob = ODEProblem(ode!, Y, (t0, tf), p);
+sol = solve(prob, Euler(), dt = dt, callback = cb);
```
kmdeck, line 56: this is fine, but we should test some of our infiltration functions too & make sure they work OK in all cases!

test/pond_soil_lsm.jl
kmdeck: added comment indicating this.

PR #11, opened 2022-01-26
added aux variables

kmdeck:
Testing adding aux variables to the soil model.

Works well out of the box! using default initialize_auxiliary, just had to define
`aux_variables(model::RichardsModel)` and `make_update_aux(model::RichardsModel)`.

The only hitch is that `solve!` only returns `integrator.sol`, which has only the prognostic state and time. if we use the `integrator` interface, we do have access to `integrator.p` at each step. otherwise we need to use a callback to get aux variables back - that is what this currently does in the test file. 

Note that `p` is updated in place, so what we pass into `prob` originally, after `solve` finishes, is the final `p`.

https://diffeq.sciml.ai/stable/basics/integrator/#Handing-Integrators
https://github.com/SciML/OrdinaryDiffEq.jl/blob/4bf1d8e8a6b99f104e3abc0f51eaae2fae8ce64b/src/solve.jl#L496

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,15 +1,23 @@
 using Test
+using Statistics
+using DifferentialEquations
+using UnPack
+using NLsolve
+using OrdinaryDiffEq: ODEProblem, solve, Euler
+using ClimaCore
 
+if !("." in LOAD_PATH) # for ease of include
+    push!(LOAD_PATH, ".")
+end
 using ClimaLSM
 using ClimaLSM.Domains: Column, RootDomain
 using ClimaLSM.Soil
 using ClimaLSM.Roots
-using UnPack
-using NLsolve
-using OrdinaryDiffEq: ODEProblem, solve, Euler
 
 
 FT = Float64
+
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -1,16 +1,21 @@
 using Test
+using Statistics
+using DifferentialEquations
```
jb-mackay, line 3: Since `Statistics` and `DifferentialEquations` and some other packages are only being used (I think) in our test suite, we should create a separate test environment (i.e. create a `test/Project.toml`) so that we keep the ClimaLSM dependencies slim.

kmdeck: good idea! Ill try and get this set up.

kmdeck: OK - I have questions about how this works haha, maybe we can hop on a VS code and figure it out at some point.

It is saying:
```- Run `import Pkg; Pkg.add("ClimaLSM")` to install the ClimaLSM package.
```

ClimaLSM is not registered yet. I can add it via URL and it will pull from main, but couldnt we run into issues, where the version it pulls is not the version we need on our branch we are trying to merge (because we have made modifiations)?

PR #13, opened 2022-01-31
Root soil lsm

kmdeck:
This sets up an LSM which includes a root model and a soil model, in a single column. The atmospheric consists of prescribed transpiration from the leaves and prescribed infiltration at the surface of the soil.


kmdeck:
bors try

kmdeck:
@jb-mackay perhaps ready for review? I added doc strings into the files, but have not yet added the docs that make the github pages for it (waiting on that till things are a bit more "final"?).

however, I moved around some files to give it a bit more structure (mimicking future repos with directories) but unfortunately that kind of complicated the files changed comparison. let's aim to get this in ~mid next week, barring major changes?

kmdeck:
I guess once this is merged we should follow up by writing a bunch of unit tests :D

jb-mackay:
> @jb-mackay perhaps ready for review? I added doc strings into the files, but have not yet added the docs that make the github pages for it (waiting on that till things are a bit more "final"?).

Yeah, I think we can wait to set up github pages. I don't think this is worth our energy at the moment.

> however, I moved around some files to give it a bit more structure (mimicking future repos with directories) but unfortunately that kind of complicated the files changed comparison. let's aim to get this in ~mid next week, barring major changes?

Mid next-week is a good goal! Especially since we have already reviewed pieces of this before merging them into this branch.

> I guess once this is merged we should follow up by writing a bunch of unit tests :D

Yes! I sense some pair programming sessions in our future.

src/Vegetation/Roots.jl
```diff
@@ -0,0 +1,432 @@
...
+        b::FT,
+        Kmax::FT,
+    ) where {FT}
+
+Computes the flow of water (moles/sec)  given the height and pressures
+at two points, `do` and `up`. Here, `a`, `b, and `Kmax` are parameters
+which parameterize the hydraulic conductance of the pathway along which
+the flow occurs.
+"""
+function flow(
```
jb-mackay, line 206: What are `do` and `up`? I'm also wondering if the name can be more descriptive - where is this flow occurring? Is it in a root or in any length of a plant stem/root?

kmdeck: I believe "do" and "up" refer to downstream and upstream, but that the model does not require them to be ordered in this way. agreed that it can be clearer! This formula would hold in any length of a plant stem, root, or leaf, assuming steady state flow and constant hydraulic properties & geometry in that segment.

FWIW, this model has to be updated to apply to a bulk plant (or plants per unit ground area) - right now it is sort of for a single plant, but that wont generalize nicely to an LSM with very coarse resolution. but the hookup it demonstrates between soil & roots is still representative.

test/runtests.jl
```diff
@@ -19,6 +19,6 @@ FT = Float64
 
 saved_values = SavedValues(FT, ClimaCore.Fields.FieldVector)
 
-include("./initial_structure_test.jl")
+#include("./initial_structure_test.jl")
```
jb-mackay, line 22: worth holding onto this test?

kmdeck: Hm, this test was just making sure that the variables are stored correctly in Y and p when we have a landmodel, i think. I think it is fine to let this one go as we will have to more carefully write unit tests for all of clima lsm soon, and this structure set up is kind of implicitly tested in the lsm simulation test we do.

test/lsm_test.jl
```diff
@@ -0,0 +1,105 @@
+using Test
```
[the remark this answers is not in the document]
kmdeck: added it, unsure of the right test.

kmdeck:
bors try

PR #18, opened 2022-02-11
Config ideas

kmdeck:
This removes the Configuration concept. Instead of using a single keyword type `Configuration` and dispatching on the type of this for boundary conditions, source terms, boundary flows for the roots, etc, we are making things more explicit, and creating `boundary_conditions`, `source`, `transpiration` attributes that live inside `Soil` and `Roots` model, as makes sense for that model.  Then, for example, on the soil RHS, we could call `compute_source(source)`, rather than `compute_source(config)`. 

Our main motivation for this change is that it was unclear what "Configurations" was buying us, and it had the potential to be confusing, as it lumped a lot of related concepts together. 

Our goal is that the LSM constructor can than make the necessary source, boundary conditions, transpiration, and root extraction structs when it makes the soil model and root model. This could be using info the user passes in as needed.

In a way, this emphasizing a design based on standalone models - the LSM just constructs a version of the standalone model that is appropriate for running in LSM mode, and then runs many of these (snow, rivers, soil, etc) all together (and defines the appropriate methods, etc, to capture interaction exchanges). The alternative - based on Configurations - emphasizes the LSM mode (and a few "configurations" that are relevant for that). We can go back to that, as needed.

### DIFF
--- a/src/ClimaLSM.jl
+++ b/src/ClimaLSM.jl
@@ -1,23 +1,20 @@
 module ClimaLSM
-###What is the role of configurations here?
 using UnPack
 using ClimaCore
 import ClimaCore: Fields
 include("Domains.jl")
 using .Domains
-include("Configurations.jl")
-using .Configurations
 include("models.jl")
 include("Soil.jl")
 using .Soil
+import .Soil: source
 include("Roots.jl")
 using .Roots
+import .Roots: flow_out_roots
 
 export RootSoilModel
 
-
 abstract type AbstractLandModel{FT} <: AbstractModel{FT} end
-
 
 """
     struct RootSoilModel{FT, SM <: AbstractModel{FT}, RM <: AbstractModel{FT}} <: AbstractLandModel{FT}
@@ -30,11 +27,9 @@ No, but should it work ?
 """
 struct RootSoilModel{
     FT,
-    CT <: AbstractConfiguration{FT},
     SM <: AbstractSoilModel{FT},
     VM <: AbstractVegetationModel{FT},
 } <: AbstractLandModel{FT}
-    configuration::CT
     soil::SM
     vegetation::VM
 end
@@ -45,24 +40,31 @@ function RootSoilModel{FT}(;
     vegetation_model_type::Type{VM},
     vegetation_args::NamedTuple = (;),
 ) where {FT, SM <: AbstractSoilModel{FT}, VM <: AbstractVegetationModel{FT}}
-    #configuration may depend on the versions of the soil and root models too
-    configuration = Configurations.RootSoilConfiguration{FT}()
-    soil = soil_model_type(; configuration = configuration, soil_args...)
+    sources = (RootExtraction{FT}(),)
+    boundary_fluxes = FluxBC{FT}(FT(0.0), FT(0.0))
+    transpiration = PrescribedTranspiration{FT}((t::FT) -> FT(0.0))
+    root_extraction = PrognosticSoilPressure{FT}()
+    soil = soil_model_type(;
+        boundary_conditions = boundary_fluxes,
+        sources = sources,
+        soil_args...,
+    )
     vegetation = vegetation_model_type(;
-        configuration = configuration,
+        root_extraction = root_extraction,
+        transpiration = transpiration,
         vegetation_args...,
     )
-    args = (configuration, soil, vegetation)
+    args = (soil, vegetation)
     return RootSoilModel{FT, typeof.(args)...}(args...)
 end
 
 function initialize_interactions(
-    land::RootSoilModel{FT, RootSoilConfiguration{FT}},
-) where {FT}
+    land::RootSoilModel{FT, SM, RM},
+) where {FT, SM <: Soil.RichardsModel{FT}, RM <: Roots.RootsModel{FT}}
+
     soil_coords = land.soil.coordinates
     return (root_extraction = similar(soil_coords),)
 end
-
 
 land_components(land::RootSoilModel) = (:soil, :vegetation)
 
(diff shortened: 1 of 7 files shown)


kmdeck:
@jb-mackay let's talk about this more - if it seems reasonable to you, maybe the first step is to update `main` for the standalone models to incoporate this.

src/ClimaLSM.jl
```diff
@@ -157,7 +158,9 @@ function initialize(model::AbstractModel)
     return Y, p, coords
 end
 
-initialize(model::NotIncluded{FT}) where {FT} = (;model.model_name => FT[]), (;model.model_name => FT[]), (;model.model_name => FT[])
+initialize(model::NotIncluded{FT}) where {FT} = (; model.model_name => FT[]),
```
jb-mackay, line 161: let's change this definition to `function initialize(...)` since it is on multiple lines

src/Soil.jl
```diff
@@ -137,7 +114,11 @@ function make_rhs(model::RichardsModel)
         )
         @. dY.soil.ϑ_l =
             -(divf2c_water(-interpc2f(p.soil.K) * gradc2f_water(p.soil.ψ + z)))
-        dY.soil.ϑ_l .= compute_source(model.configuration, Y, p)
+        map(
+            source -> dY.soil.ϑ_l .+= compute_source(source, Y, p),
```
jb-mackay, line 118: This might be more readable as a `for` loop

src/Soil.jl
```diff
@@ -154,8 +135,47 @@ function make_update_aux(model::RichardsModel)
...
+
+abstract type AbstractSoilSource{FT <: AbstractFloat} end
+
+"""
+    SoilConfiguration{FT, BCT, ST} <: AbstractConfiguration{FT}
+
+The Configuration object for the soil model, holding boundary conditions
+and source terms, which are used for dispatching on within the soil RHS.
+"""
+struct SoilConfiguration{FT, BCT<: AbstractSoilBoundaryConditions{FT},ST} <: AbstractConfiguration{FT}
```
jb-mackay, line 153: As we discussed, I think it doesn't seem like individual models need an internal `Configuration` struct that then stores the source and bcs. We can directly store these in the model instead

src/Roots.jl
```diff
@@ -285,14 +271,48 @@ function make_rhs(model::RootsModel)
         dY.vegetation.rwc[1] = flow_out_roots - flow_out_stem
         dY.vegetation.rwc[2] =
             flow_out_stem -
-            compute_transpiration(model.configuration,t)
+            compute_transpiration(model.configuration.transpiration, t)
```
jb-mackay, line 274: I think having `compute_...` as a prefix is redundant, what do you think?

src/Roots.jl
```diff
@@ -285,14 +271,48 @@ function make_rhs(model::RootsModel)
...
+struct RootsConfiguration{
+    FT,
+    RET <: AbstractRootExtraction{FT},
+    TT <: AbstractTranspiration{FT},
+} <: AbstractConfiguration{FT}
+    root_extraction::RET
+    transpiration::TT
+end
+
+struct PrognosticSoilPressure{FT} <: AbstractRootExtraction{FT} end
```
jb-mackay, line 303: More nitpicking - is calling this `Prognostic` misleading here? It is a prognostic variable for the soil model, but not for the root model.

jb-mackay (commented):
Following our discussion, I think this moves in a more explicit direction. We have a bit more granularity and control on the dispatch in the `rhs` methods. Let's do away with the `Configuration` types for now (we can keep the file but maybe not use them) and add sources and boundary condition fields to the model structs.

Happy to find time to pair when it comes to looking at the LSM constructor.

src/Roots.jl
kmdeck: sure! Ill change them

src/ClimaLSM.jl
kmdeck: the last PR was able to remove NotIncluded, but good point! Will keep an eye out for other multi-line functions defined without `function`.

src/Roots.jl
kmdeck: I appreciate nitpicking!

It's a good point. Perhaps, since we discussed moving these types and methods to the LSM repo eventually, since it is only in the LSM case that they are needed (rather than having them be in the component modules), I will leave it with this name. Then, when we move it, it is consistent that SoilPressure is a prognostic variable, since it is for the LSM....

src/ClimaLSM.jl
```diff
@@ -45,25 +39,35 @@ function RootSoilModel{FT}(;
...
     )
-    args = (configuration, soil, vegetation)
+    args = (soil, vegetation)
     return RootSoilModel{FT, typeof.(args)...}(args...)
 end
 
 function initialize_interactions(
-    land::RootSoilModel{FT, RootSoilConfiguration{FT}},
-) where {FT}
+    land::RootSoilModel{
```
kmdeck, line 61: @jb-mackay this is what I was mentioning on slack....we need to put in placeholders for all the parametric types of Root and Soil models...

not sure why

src/ClimaLSM.jl
jb-mackay: Ah right, good point!

src/Roots.jl
jb-mackay: That's a good point!

src/models.jl
```diff
@@ -8,6 +8,7 @@ export AbstractModel,
     prognostic_vars,
     auxiliary_vars
 
+import .Domains: coordinates
```
kmdeck, line 11: im not sure if this needs to be here?

src/ClimaLSM.jl
```diff
@@ -1,24 +1,21 @@
 module ClimaLSM
-###What is the role of configurations here?
 using UnPack
 using ClimaCore
 import ClimaCore: Fields
 include("Domains.jl")
 using .Domains
-include("Configurations.jl")
-using .Configurations
```
kmdeck, line 9: maybe we should just delete the Configurations file for now, nothing in it really.

jb-mackay: Fine either way

src/models.jl
jb-mackay: This method is extended at the end of the file

src/Roots.jl
```diff
@@ -49,72 +56,41 @@ struct RootsParameters{FT <: AbstractFloat}
...
 
 The RootModel domain must be of type `AbstractPlantDomain`.
 """
-struct RootsModel{FT, PS, D, B} <: AbstractVegetationModel{FT}
+struct RootsModel{FT, PS, D, RE, T} <: AbstractVegetationModel{FT}
     param_set::PS
     domain::D
-    configuration::B
+    root_extraction::RE
+    transpiration::T
```
jb-mackay, line 76: In the soil model, we have `sources` and `boundary_conditions`, while here we have fields for what seem to be specific processes. Does it make sense to group these processes in a `sources` tuple that is stored in the model?

kmdeck: that's a good question! I dont know the answer....

followup questions:
- is a single term on the RHS of an ODE called a source?
- is this the terminology they use in that field, regardless of what mathematicians think?
- the two "sources" appear in different ODEs - would we then need to specify which variable the source terms apply to (in the tuple)? we may need to do this with soil anyways when we add in heat, so not an issue really...
- is it clearer to group by physical concept, rather than function (they are both added in to RHS's)?

kmdeck: made https://github.com/CliMA/ClimaLSM.jl/issues/19

PR #22, opened 2022-03-02
Soil 3d

kmdeck:
This PR adds in a HybridBox domain for the soil model. The code is essentially copied over from Clima Atmos:
https://github.com/CliMA/ClimaAtmos.jl/blob/9816afbb9679a3b4dfea5efe5b83c0dae89d5170/src/Domains/make_function_space.jl#L46

I also took many of the domain unit tests from ClimaAtmos. thank you @akshaysridhar @bischtob @jiahe23 and everyone else who worked on that code. 
________
@jb-mackay  With the addition of non-Column domains, we needed to modify our `coordinates` function. Prior to this, we only had a Column domain, and we would return only `coordinates.z` when we called `coordinates(column::Column)`. In other domains, coordinates has other entries, and we want all of them (not just z).

so, `coordinates` now returns something that behaves like a NamedTuple, with (x,y,z) entries for the HybridBox and only z for the column. This required us to slightly modify how we initialize the state and aux vectors. We can no longer do `similar(coords)`, but rather
```
zero_state = map(_ -> zero(FT), coords)
similar(zero_state)
```
(the zero_state command is something ClimaAtmos does - where we could eventually replace zero(FT) with the type for a specific variable, as needed).

This works regardless of the entries in coords.
____________
I added in a `Plane` domain. Im not 100% sure we need this, but the box domain is made from a 2d domain/mesh/topology/space and a 1d domain/mesh/topology/space, so we can use it to make the Box domain. I also thought this might come in handy for surface models. We can get a single point by given nelements = (1,1) and npolynomial = 0, or a 2d grid of coordinate points.

________
Lastly, we can then run a soil problem in 3d. Thanks to @akshaysridhar  and @jiahe23  for taking a look at the horizontal tendency terms in our RHS.

To test that the horizontal operators are used correctly in the RHS for the 3d soil problem, I created a test with initial conditions in vertical hydrostatic eq, but not in equilibrium in the horizontal. Then I computed the RHS horizontal terms analytically, and compared to the dY coming from our ode function. These agree as well as they do in the case of purely vertical derivative terms contributing. It was tested in both the saturated and unsaturated zone.

### DIFF
--- a/src/ClimaLSM.jl
+++ b/src/ClimaLSM.jl
@@ -31,20 +31,22 @@ and the versions of these component models being used.
 abstract type AbstractLandModel{FT} <: AbstractModel{FT} end
 
 
-function initialize(land::AbstractLandModel)
+function initialize(land::AbstractLandModel{FT}) where {FT}
     components = land_components(land)
     coords_list = map(components) do (component)
         Domains.coordinates(getproperty(land, component))
     end
     Y_state_list = map(zip(components, coords_list)) do (component, coords)
+        zero_state = map(_ -> zero(FT), coords)
         getproperty(
-            initialize_prognostic(getproperty(land, component), coords),
+            initialize_prognostic(getproperty(land, component), zero_state),
             component,
         )
     end
     p_state_list = map(zip(components, coords_list)) do (component, coords)
+        zero_state = map(_ -> zero(FT), coords)
         getproperty(
-            initialize_auxiliary(getproperty(land, component), coords),
+            initialize_auxiliary(getproperty(land, component), zero_state),
             component,
         )
     end
--- a/src/SharedUtilities/models.jl
+++ b/src/SharedUtilities/models.jl
@@ -137,6 +137,8 @@ function initialize_auxiliary(
     end
 end
 
+Domains.coordinates(model::AbstractModel) = Domains.coordinates(model.domain)
+
 """
     initialize(model::AbstractModel)
 
@@ -145,11 +147,10 @@ values; constructs and returns the coordinates for the `model` domain.
 We may need to consider this default more as we add diverse components and 
 `Simulations`.
 """
-function initialize(model::AbstractModel)
-    coords = coordinates(model)
-    Y = initialize_prognostic(model, coords)
-    p = initialize_auxiliary(model, coords)
+function initialize(model::AbstractModel{FT}) where {FT}
+    coords = Domains.coordinates(model)
+    zero_state = map(_ -> zero(FT), coords)
+    Y = initialize_prognostic(model, zero_state)
+    p = initialize_auxiliary(model, zero_state)
     return Y, p, coords
 end
-
-Domains.coordinates(model::AbstractModel) = Domains.coordinates(model.domain)
--- a/src/Soil/Soil.jl
+++ b/src/Soil/Soil.jl
@@ -61,9 +61,9 @@ using ClimaLSM
 using UnPack
 using DocStringExtensions
 using ClimaCore
-import ClimaCore: Fields, Operators, Geometry
+import ClimaCore: Fields, Operators, Geometry, Spaces
 
-import ClimaLSM.Domains: coordinates
+import ClimaLSM.Domains: coordinates, Column, HybridBox
 import ClimaLSM:
     AbstractModel,
     make_update_aux,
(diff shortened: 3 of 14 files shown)


kmdeck:
Im not sure how to test that the RHS is being computed correctly in the 3D case (the horizontal terms). I followed ClimaAtmos examples and asked @jiahe23 and she thought it looked right. but we dont have a great test case for this. I will try and dig one up...though the integration is very slow for the one test case I tried to run (require small timesteps with explicit solver...)

https://github.com/CliMA/ClimaLSM.jl/blob/2f4e5c08e54839272e73c92463fb82016676e8df/src/Soil/Soil.jl#L321

kmdeck:
> Im not sure how to test that the RHS is being computed correctly in the 3D case (the horizontal terms). I followed ClimaAtmos examples and asked @jiahe23 and she thought it looked right. but we dont have a great test case for this. I will try and dig one up...though the integration is very slow for the one test case I tried to run (require small timesteps with explicit solver...)
> 
> https://github.com/CliMA/ClimaLSM.jl/blob/2f4e5c08e54839272e73c92463fb82016676e8df/src/Soil/Soil.jl#L321

edited ordering of dss after speaking with @akshaysridhar, and added in documentation for the functions.
https://github.com/CliMA/ClimaLSM.jl/blob/e70e73e627f3f68054874856cbb9b011d71e433d/src/Soil/Soil.jl#L310

kmdeck:
bors try

kmdeck:
bors try

src/SharedUtilities/models.jl
```diff
@@ -145,11 +147,10 @@ values; constructs and returns the coordinates for the `model` domain.
 We may need to consider this default more as we add diverse components and 
 `Simulations`.
 """
-function initialize(model::AbstractModel)
-    coords = coordinates(model)
-    Y = initialize_prognostic(model, coords)
-    p = initialize_auxiliary(model, coords)
+function initialize(model::AbstractModel{FT}) where {FT}
+    coords = Domains.coordinates(model)
+    zero_state = map(_ -> zero(FT), coords)
```
jb-mackay, line 152: `Fields.zeros`

src/Soil/Soil.jl
```diff
@@ -293,23 +293,81 @@ function make_rhs(model::RichardsModel)
...
+        # Vertical operators - needs boundary conditions
+        # When is the BC a Wvector vs others?
         interpc2f = Operators.InterpolateC2F()
         gradc2f_water = Operators.GradientC2F()
         divf2c_water = Operators.DivergenceF2C(
             top = Operators.SetValue(Geometry.WVector(top_flux_bc)),
             bottom = Operators.SetValue(Geometry.WVector(bot_flux_bc)),
         )
-        @. dY.soil.ϑ_l =
+        @. dY.soil.ϑ_l +=
```
jb-mackay, line 307: `+=` should be `=`

src/Soil/Soil.jl
```diff
@@ -293,23 +293,81 @@ function make_rhs(model::RichardsModel)
...
             -(divf2c_water(-interpc2f(p.soil.K) * gradc2f_water(p.soil.ψ + z)))
```
jb-mackay, line 308: May want a `vertical_components!` method (for spherical domains vs box)

src/ClimaLSM.jl
```diff
@@ -31,20 +31,22 @@ and the versions of these component models being used.
...
 
 
-function initialize(land::AbstractLandModel)
+function initialize(land::AbstractLandModel{FT}) where {FT}
     components = land_components(land)
     coords_list = map(components) do (component)
         Domains.coordinates(getproperty(land, component))
     end
     Y_state_list = map(zip(components, coords_list)) do (component, coords)
+        zero_state = map(_ -> zero(FT), coords)
```
jb-mackay, line 40: Fields.zeros

src/Soil/Soil.jl
```diff
@@ -257,7 +257,7 @@ function pressure_head(
     if S_l_eff <= FT(1.0)
         ψ = matric_potential(α, n, m, S_l_eff)
     else
-        ψ = (ϑ_l - ν_eff) / S_s
+        ψ = (ϑ_l - (ν_eff - θ_r)) / S_s
```
jb-mackay, line 260: Incorrect

src/root_soil_model.jl
```diff
@@ -107,8 +107,8 @@ function initialize_interactions(#Do we want defaults, for land::AbstractLandMod
     land::RootSoilModel{FT, SM, RM},
 ) where {FT, SM <: Soil.RichardsModel{FT}, RM <: Roots.RootsModel{FT}}
 
-    soil_coords = land.soil.coordinates
-    return (root_extraction = similar(soil_coords),)
+    zero_state = map(_ -> zero(FT), land.soil.coordinates)
```
jb-mackay, line 110: Fields.zeros

jb-mackay (approved):

src/Soil/Soil.jl
jb-mackay: (future PR)

src/Soil/Soil.jl
jb-mackay: should be no change

PR #24, opened 2022-03-02
Update `initialize` for models

jb-mackay:
Will update the `initialize` function for `AbstractModels`

Goal: 
Move towards a [ClimaAtmos](https://github.com/CliMA/ClimaAtmos.jl/blob/ec56b8453712bff8e8662030993af3ec89b8b582/src/Models/Nonhydrostatic3DModels/nonhydrostatic_3d_model.jl#L44) implementation that allows variables to have different types and spaces.

Additionally:
Removes the `model_name` nesting in prognostic and aux Field vectors for standalone models


kmdeck:
@jb-mackay added in the variable type stuff. I thought it was working (works with FT type), and it seemed OK with MVectors, but when I actually tried it with the root model (Y.vegetation.rwc is an MVector{k,FT} where k = length(model.compartment_heights), it errors in the test files. so debugging this....

(this error is probably generic, i.e. doesnt matter that we are using the Root model in particular for it)

though, the nice thing is that then we can scrap the RootDomain! I stuck the compartment_heights and root layer depths in the model. Then we can have the domain be a Point, and have the variable be an array at the point (one index for each compartment).

kmdeck:
A couple of things...
``` 
using ClimaLSM
using ClimaLSM.Domains: Point, Column, coordinates
using StaticArrays

FT = Float64

keys = (:a,:b, :c)
types = (FT, MVector{3,FT}, NTuple{2,FT})
coords = coordinates(Point{FT}(0.0))
Y = initialize_vars(keys,types, coords, :foo)
Y.foo.a[1] == 0.0
Y.foo.b[1] == zero(MVector{3,FT})
Y.foo.c[1] == zero(NTuple{2,FT})

col = Column(Float64; zlim = (-1.0, 0.0), nelements = 10)
field_coords = coordinates(col)
Y_col = initialize_vars(keys,types, field_coords, :foo)
## MVector leads to error, but NTuple may work:
keys = (:a,:b)
types = (FT,  NTuple{2,FT})
Y_col = initialize_vars(keys,types, field_coords, :foo)

```

kmdeck:
Here is the carbon example. It is a matrix equation, Ydot = M*Y + B
https://github.com/CliMA/ClimaLSM.jl/blob/bm/initialize/carbon.jl

trying to set it up with Point domain, prognostic type = MVector. much simpler than the root example for debugging.

src/SharedUtilities/models.jl
```diff
@@ -29,13 +32,49 @@ Returns the prognostic variable symbols for the model in the form of a tuple.
...
+- NTuple{k,Float64} for a tuple of
+ length `k` at each coordinate point.
+- Note that Arrays, MVectors are not isbits and cannot be used.
+
+Here, the coordinate points are those returned by coordinates(model).
+"""
+auxiliary_types(m::AbstractModel) = ()
+
+Base.zero(T::DataType = NTuple{k <: Int, FT <: AbstractFloat}) =
+    ntuple(i -> T.types[1](0.0), length(T.types))
```
charleskawczynski, line 76: This is probably not a good idea (overloading a Julia Base method for Julia Base types)

kmdeck: oh geez, for some reason when I tried `zero(NTuple{N,FT})` (with N and FT known) it didnt work, but it clearly works now that I try it again.

thanks for pointing this out.

PR #25, opened 2022-03-02
Make root-soil interaction nonzero

kmdeck:
Adjust plant hdyraulics model to account for area indices, add in root distribution function, add in nonzero soil water extraction.

### DIFF
--- a/src/Vegetation/Roots.jl
+++ b/src/Vegetation/Roots.jl
@@ -75,9 +75,10 @@ import ClimaLSM:
     name
 export RootsModel,
     AbstractVegetationModel,
-    flow,
-    theta_to_p,
-    p_to_theta,
+    ground_area_flux,
+    ground_area_flux_out_roots,
+    θ_to_p,
+    p_to_θ,
     RootsParameters,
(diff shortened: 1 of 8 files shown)


src/Vegetation/Roots.jl
```diff
@@ -243,61 +256,66 @@ function vc_integral_approx(
...
+Computes the volumetric water content given pressure (p).
 """
-function theta_to_p(theta::FT) where {FT}
-    α = FT(5.0) # inverse meters
-    n = FT(2.0)
-    m = FT(0.5)
-    rhog_MPa = FT(0.0098) #MPa/m
-    p = -((theta^(-FT(1) / m) - FT(1)) * α^(-n))^(FT(1) / n) * rhog_MPa
+function θ_to_p(θ::FT) where {FT}
+    θ = min(θ, FT(1.0))
```
gagnelandmanna, line 296: We could use an exponentially growing function P(w) which might be more natural like we had discussed briefly

gagnelandmanna (commented):
I think this is is awesome!

PR #30, opened 2022-03-21
Add in LSM Domain constructors

kmdeck:
We are planning LSM domains of the following type:
 - SingleColumnLSM <: AbstractDomain = Column (with FD space, for soil) + surface Point
 - MultiColumnLSM <: AbstractDomain = HybridBox (soil) + HybridBox with one element (surface 2d space)

We construct them at the same time because we think the 2d spaces need to be the same instance in the MultiColumn case. It is also neat for the user to be able to specify everything they need for an LSM run in a single constructor.

The LSM Models can either take in the `LSM` Domains as arguments, and create the sub-component domains, in order to create the subcomponent models, or the user can make the domain and pass in lsm_domain.surface as the domain for the surface models or lsm_domain.subsurface for the soil model.

### DIFF
--- a/src/SharedUtilities/Domains.jl
+++ b/src/SharedUtilities/Domains.jl
@@ -42,34 +42,6 @@ end
 
 coordinates(domain::Point) = [domain.z_sfc]
 
-### Example of component specific domain
-"""
-    AbstractVegetationDomain{FT} <: AbstractDomain{FT}
-
-An abstract type for vegetation specific domains.
-"""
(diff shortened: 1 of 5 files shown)


jb-mackay:
This PR:
- [x] Add SingleColumnLSM domain
- [ ] Show example of single column pond+soil

Next PR:
Add MultiColumnLSM domain
Show example of multi column pond+soil with no horizontal interactions

kmdeck:
closed via PR 31

PR #31, opened 2022-03-21
Pond soil tutorial

kmdeck:
add a tutorial showing the interface for LSM models. this builds off of the abstractmodel tutorial.


kmdeck:
bors try

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
generated tutorial here: https://clima.github.io/ClimaLSM.jl/previews/PR31/generated/LSM_single_column_tutorial/

kmdeck:
ready for review after PR https://github.com/CliMA/ClimaLSM.jl/pull/30 is in. this is branched off of that

src/SharedUtilities/Domains.jl
```diff
@@ -281,8 +253,65 @@ function coordinates(
...
+    return domain.compartment_heights
+end
+
+
+struct SingleColumnLSM{FT} <: AbstractDomain{FT}
+    subsurface::Column{FT}
+    surface::Point{FT}
+end
+
+function SingleColumnLSM(
```
jb-mackay, line 291: I really like this and the way it is used in the `pond_soil_lsm` test. Should we change the name to make it more clear that this has to do with Domains? maybe `SingleColumnDomain` or `SingleColumnLSMDomain`? Otherwise I think it is confusing, since it is not building the full LSM which I think the name implies.

kmdeck: good idea! agreed that we should change the name to be more clear.

src/SharedUtilities/Domains.jl
```diff
@@ -281,8 +253,65 @@ function coordinates(
...
+    FT::DataType = Float64;
+    zlim::Tuple{G, G},
```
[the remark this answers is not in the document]
kmdeck: hmm yes, that would seem clearer. not sure why I did it this way.

also, is this function type stable (FT is an argument, depending on its value, the type of the output changes), even if we change it to what you suggested? perhaps need to rewrite a bit also?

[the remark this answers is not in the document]
kmdeck: maybe we always need FT to be passed as parametric type, also for the `Column` outer constructor.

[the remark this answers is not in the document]
jb-mackay: Yes, this is type stable since the types of the input determine the output type (and the types of the other internal variables). It doesn't matter that `FT` is ambiguous when writing the code. I think that under the hood Julia would compile a method for each `FT` that is actually used in the driver. But yes, it can compile efficient code here.

src/SharedUtilities/Domains.jl
jb-mackay: I don't know if there is also a way to make it clearer that this is creating the domains for multiple models, a Column + Point domain, e.g.

The "LSM" addition might be enough. I had to remind myself that this was setting up more than a `Column` domain, though.

src/SharedUtilities/Domains.jl
```diff
@@ -281,8 +253,65 @@ function coordinates(
...
+        subsurface = coordinates(domain.subsurface),
+        surface = coordinates(domain.surface),
+    )
+end
+
+
+
 export AbstractDomain, AbstractVegetationDomain
 export Column, Plane, HybridBox, RootDomain, Point
+export SingleColumnLSM
```
jb-mackay, line 314: Long term, I think `Column, Plane, HybridBox, Point` would be in the shared `Models.jl` repo, while `Single/MultiColumnLSM` would live in `ClimaLSM.jl`.

test/domains.jl
```diff
@@ -61,3 +62,24 @@ end
...
+
+@testset "SingleColumnLSM Domain" begin
+    domain = SingleColumnLSM(FT, zlim = zlim, nelements = nelements[3])
+
+    point = domain.surface
+    column = domain.subsurface
+
+    @test typeof(column) == Column{FT}
+    @test typeof(point) == Point{FT}
+    @test point.z_sfc == zlim[2]
```
jb-mackay, line 75: Lines 75, 76 I think are redundant since 74 checks that point is a `Point` domain, and the `Point` domain tests check 75, 76.

jb-mackay: Same for L80-83 for `Column`

docs/tutorials/LSM_single_column_tutorial.jl
```diff
@@ -0,0 +1,386 @@
...
+# some initial condition consistent
+# with Y_soil.soil.ϑ(t=0) before starting a simulation, though initial
+# conditions must be given for `Y`.
+
+# Note also that we have defined methods `make_rhs` and `make_update_aux`, which only
+# take the `model` as argument, and which return the functions `update_aux!`
+# and `rhs!`, [here](https://github.com/CliMA/ClimaLSM.jl/blob/9e2b8df6d8d5b7f878a5b0f02e2bf80fa66aec33/src/Soil/Soil.jl#L292)
+# and [here](https://github.com/CliMA/ClimaLSM.jl/blob/9e2b8df6d8d5b7f878a5b0f02e2bf80fa66aec33/src/Soil/Soil.jl#L400).
+
+# lastly, the coordinates returned by `initialize` contain
```
jb-mackay, line 147:
```suggestion
# Lastly, the coordinates returned by `initialize` contain
```

test/domains.jl
```diff
@@ -61,3 +62,24 @@ end
...
+    @test coordinates(point) == [zlim[2]]
+    @test coordinates(point) == coordinates(domain).surface
+
+    column_coords = coordinates(column)
+    @test column.zlim == FT.(zlim)
+    @test column.nelements == nelements[3]
+    @test eltype(column_coords) == ClimaCore.Geometry.ZPoint{FT}
+    @test typeof(column_coords) <: ClimaCore.Fields.Field
```
jb-mackay, line 83:
```suggestion
```

test/domains.jl
```diff
@@ -61,3 +62,24 @@ end
...
+@testset "SingleColumnLSM Domain" begin
+    domain = SingleColumnLSM(FT, zlim = zlim, nelements = nelements[3])
+
+    point = domain.surface
+    column = domain.subsurface
+
+    @test typeof(column) == Column{FT}
+    @test typeof(point) == Point{FT}
+    @test point.z_sfc == zlim[2]
+    @test coordinates(point) == [zlim[2]]
```
jb-mackay, line 76:
```suggestion
```

docs/tutorials/LSM_single_column_tutorial.jl
```diff
@@ -0,0 +1,386 @@
...
+# - P
+# - I
+# which are akin to boundary conditions. In standalone mode, 
+#  one would need to pass in **prescribed** functions of time and store
+# them inside our pond model, since again, the pond model structure
+# must contain everything needed to make the ode function:
+
+precipitation(t::ft) where {ft} = t < ft(20) ? -ft(1e-5) : ft(0.0) # m/s
+
+infiltration(t::ft) where {ft} = -ft(1e-6) #m/s
```
jb-mackay, line 172:
```suggestion
precipitation(t::FT) where {FT} = t < FT(20) ? -FT(1e-5) : FT(0.0) # m/s

infiltration(t::FT) where {FT} = -FT(1e-6) #m/s
```

kmdeck:
Also to do:
remove FT from arguments, just pass in Tuple{FT,FT}...where {FT}, etc.

Create keyword constructor for the Point, so that all are constructed via keyword

docs/tutorials/LSM_single_column_tutorial.jl
```diff
@@ -0,0 +1,386 @@
...
+# ``
+# -K ∇(ψ+z)|_{z = zmax}  ⋅ ẑ = I(ϑ, η, P)
+# ``
+
+# ``
+# -K ∇(ψ+z)|_{z = zmin}  ⋅ ẑ = 0.0.
+# ``
+
+# These two components
+# interfact via the infiltration term `I`. Infiltration is a boundary
```
jb-mackay, line 230:
```suggestion
# interact via the infiltration term `I`. Infiltration is a boundary
```

docs/tutorials/LSM_single_column_tutorial.jl
```diff
@@ -0,0 +1,386 @@
...
+lsm_domain.surface
+# The subsurface domain is a column from zmin to zmax:
+lsm_domain.subsurface
+
+# Let's now collect the needed arguments for the soil and pond
+# models:
+
+soil_args = (param_set = soil_ps, domain = lsm_domain.subsurface, sources = ());
+surface_water_args = (domain = lsm_domain.surface,);
+# Atmospheric drivers don't "belong" to either component alone:
```
jb-mackay, line 259:
```suggestion
# Atmospheric drivers don't "belong" to either component alone, so we introduce general land model arguments:
```

docs/tutorials/LSM_single_column_tutorial.jl
```diff
@@ -0,0 +1,386 @@
...
+# - the order in which the components are treated in the rhs or in update aux does not matter. What matters is that auxiliary/cache variables are updated first, and within this update, interactions are updated last. We assume that the rhs! function for a component only needs the entire `p` and `Y.component` in making this statement. Similarily, updating the aux variables of a single component does not require interaction variables. Yhis is also the same as saying they can be run in *standalone* mode. 
+# - the code is also modular in terms of swapping out a simple component model for a more complex version.
+
+# Some disadvantages to our interface design:
+# - Even in standalone model, variables are accessed in a nested way: Y.soil, p.soil, etc, which is excessive.
+# - To accomodate the fact that some components involve PDEs, a developer for purely ODE based component does need to at least handle `ClimaCore.Field.FieldVector`s.
+# - standalone models need to play by the rules of `AbstractModel`s, and LSMs need to play by the rules of `ClimaLSM.jl`.
```
jb-mackay, line 386: I'm not sure this is a disadvantage, personally :) but perhaps it will feel that way in the short-term.

jb-mackay (approved):
A few small changes, but largely looks great.

src/SharedUtilities/Domains.jl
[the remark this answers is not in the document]
kmdeck: thanks for the explanation!

src/SharedUtilities/Domains.jl
kmdeck: agreed!

kmdeck:
> Also to do: remove FT from arguments, just pass in Tuple{FT,FT}...where {FT}, etc.
> 
> Create keyword constructor for the Point, so that all are constructed via keyword

this led to updating a number of tests

docs/tutorials/LSM_single_column_tutorial.jl
kmdeck: fair enough! yes, maybe the main disadvantage is that there is more machinery to learn up front for standalone models.

changed to "Possible disadvantages" :)

kmdeck:
bors try

PR #33, opened 2022-04-04
LSM Multi-column domain and pond/soil test

kmdeck:
Goals for this PR:
- define an LSM MultiColumnDomain - Hybrid box + plane or hybrid box + hybrid box with one element
- redefine interaction methods for soil/pond to broadcast correctly between the field of surface variables and the field of the top soil layer, for both single column and multi column domains.
- check that the output for a simulation with no variation in the horizontal leads to the same output in each column, and matches what we expect based on a single column run, or that the RHS eval gives the same result (unit test)
- add unit tests for the new domain type.
- we could even design the RHS for the LSM as looping over columns and computing horizontal terms afterwards; saving this for a later PR is fine!

### DIFF
--- a/src/SharedUtilities/Domains.jl
+++ b/src/SharedUtilities/Domains.jl
@@ -298,8 +298,41 @@ function coordinates(
     return cc
 end
 
+
+struct LSMMultiColumnDomain{FT, SS, SF} <: AbstractDomain{FT}
+    subsurface::SS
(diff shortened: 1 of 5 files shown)


kmdeck:
Closing in lieu of https://github.com/CliMA/ClimaLSM.jl/pull/85

PR #35, opened 2022-04-04
adding in full heat params

kmdeck:
This PR:
- Start by adding in clima parameters to the soil model and adding in the other heat parameterization functions

- Then update the RHS to compute the full heat model

- Test with analytic solution/integration test

Next PR:

- Then update the RHS to compute K properly in the hydraulics model (taking into account T)
- add in other integration tests we had working

Next: Boundary condition types
- add further tests (seperate top and bottom, free drainage, runoff, flux, surface fluxes?)

Next:
- Freezing and thawing

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["Clima Land Team"]
 version = "0.1.0"
 
 [deps]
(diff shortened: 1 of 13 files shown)


kmdeck:
closing in lieu of smaller PRs:
https://github.com/CliMA/ClimaLSM.jl/pull/40

PR #36, opened 2022-04-06
add clima params to models that need them

kmdeck:
Add in the clima parameter set (physical constants and shared clima-wide parameters).

Currently, I am only adding it to the models that need them (Roots, SoilEnergyHydrology), and not to RichardsModel or Pond. though the Root/Richards model does need them, we can access them via Roots.earth_param_set.

### DIFF
--- a/src/Vegetation/Roots.jl
+++ b/src/Vegetation/Roots.jl
@@ -49,6 +49,7 @@ using ClimaCore
 using UnPack
 using DocStringExtensions
 import ClimaCore: Fields
+using CLIMAParameters: AbstractEarthParameterSet
 
 using ClimaLSM.Domains: AbstractVegetationDomain, RootDomain
 import ClimaLSM:
@@ -109,7 +110,7 @@ abstract type AbstractTranspiration{FT <: AbstractFloat} end
 A struct for holding parameters of the Root Model. Eventually to be used with ClimaParameters.
 $(DocStringExtensions.FIELDS)
 """
-struct RootsParameters{FT <: AbstractFloat}
+struct RootsParameters{FT <: AbstractFloat, PSE <: AbstractEarthParameterSet}
     "controls the shape and steepness of conductance vs. pressure curve, for roots: unitless"
     a_root::FT
     "controls the steepness of the relative conductance vs. pressure curve, for roots: inverse MPa"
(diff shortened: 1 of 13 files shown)


src/root_soil_model.jl
```diff
@@ -62,12 +64,18 @@ function RootSoilModel{FT}(;
     sources = (RootExtraction{FT}(),)
     root_extraction = PrognosticSoilPressure{FT}()
 
+    @unpack earth_param_set = land_args
+    if soil_model_type == Soil.SoilEnergyHydrology{FT}
```
kmdeck, line 68: needed as Richards model doesnt have this field.

src/Vegetation/Roots.jl
```diff
@@ -160,9 +161,11 @@ and soil pressure at the root tips, or with a dynamic soil model using `ClimaLSM
 
 $(DocStringExtensions.FIELDS)
 """
-struct RootsModel{FT, PS, D, RE, T} <: AbstractVegetationModel{FT}
+struct RootsModel{FT, PS, PSE, D, RE, T} <: AbstractVegetationModel{FT}
     "Parameters required by the root model"
     param_set::PS
+    "Physical Constants and other clima-wide parameters"
+    earth_param_set::PSE
```
jb-mackay, line 168: I think the intent of CLIMAParameters is that each model will only need a single param set stored. This will be constructed via a mixture of both "clima-wide" parameters and model specific ones. See [these examples](https://clima.github.io/CLIMAParameters.jl/previews/PR63/parameter_structs/).

jb-mackay: Perhaps what you have here is a fine start point, and we can define `RootParameters`, etc, later this month.

src/root_soil_model.jl
kmdeck: removed after talking with @jb-mackay

src/Vegetation/Roots.jl
kmdeck: stuck everything in one `parameters` object as discussed. I did not organize as parameters.soil_params and parameters.earth_params, but rather like:
```
struct Parameters{FT, PS}
a::FT
b::FT
earth_param_set::PS
end
```

kmdeck: it looks *way* cleaner! it's very freeing to not think about the EnergyHydrology model for soil as composed of a "heat model" and a "soil model", (with "soil heat params" and "soil water params" but just as its own thing ("energyhydrology params").

kmdeck:
bors try

jb-mackay (approved):

PR #37, opened 2022-04-06
initialize with types

kmdeck:
This follows ClimaAtmos in adding in prognostic and auxiliary types. We need to support scalars and arrays at each coordinate point.

backup copy from botched rebase of bm/initialize :/


kmdeck:
@jb-mackay @charleskawczynski I seriously messed up a rebase of the bm/initialize branch https://github.com/CliMA/ClimaLSM.jl/pull/24 

but this was a copy I had made just before trying. I can keep trying to salvage that branch or we can continue developing here.

The basic use of variable types works, and the test/carbon_ntuple example shows how it would work for the carbon model RHS (matrix equation, thanks Charlie for your help). How big of a deal is it to use `parent`? I thought this was something to avoid. why?

but still TBD how this will work for the root model. possibly a risk to merge this and then find that it doesnt work well for roots?

jb-mackay:
> @jb-mackay @charleskawczynski I seriously messed up a rebase of the bm/initialize branch #24
> 
> but this was a copy I had made just before trying. I can keep trying to salvage that branch or we can continue developing here.

If I recall, there wasn't a ton on that branch. So it probably isn't worth the time to salvage on your end, especially if you have added new features on this one. That said, I surely have an up to date copy of bm/initialize locally.

> The basic use of variable types works, and the test/carbon_ntuple example shows how it would work for the carbon model RHS (matrix equation, thanks Charlie for your help). How big of a deal is it to use `parent`? I thought this was something to avoid. why?
> 
> but still TBD how this will work for the root model. possibly a risk to merge this and then find that it doesnt work well for roots?

I think we certainly want to avoid `parent` in our `rhs!` functions. My understanding is that `parent` is a bit too low-level and anything that uses it should have a better interface or abstraction that hides `parent` away in ClimaCore. Probably OK to use it in tests, though.

kmdeck:
> I think we certainly want to avoid `parent` in our `rhs!` functions. My understanding is that `parent` is a bit too low-level and anything that uses it should have a better interface or abstraction that hides `parent` away in ClimaCore. Probably OK to use it in tests, though.

ok! maybe this is something we can talk about in a future PP session re: this PR...also, we can try and see (on a different branch) how well this works for the root model - i.e. if we can get rid of the RootDomain object and use Point or Plane and run it many coordinates at the same time...

src/SharedUtilities/models.jl
```diff
@@ -127,16 +158,40 @@ have different dimensions - require defining a new method.
...
+        state,
+        name(model),
+    )
+end
+
+# var_names, var_types, var_spaces a la ClimaAtmos
+function initialize_vars(keys, types, state, model_name) # replace state with type and space info in future
+    FT = eltype(state) # may not be needed
+    # Code commented below was written with the intent of not using model_name, and make Y and p not be
+    # hierarchical for standalone models.
```
jb-mackay, line 175: As discussed, can move to this in future PR.

test/carbon_ntuple.jl
```diff
@@ -0,0 +1,80 @@
...
+end
+
+using ClimaLSM
+using ClimaLSM.Domains
+# Import the functions we are extending for our model:
+import ClimaLSM: name, make_rhs, prognostic_vars, prognostic_types
+
+abstract type AbstractCarbonModel{FT} <: AbstractModel{FT} end
+
+struct BulkNPools{FT} <: AbstractCarbonModel{FT}
```
jb-mackay, line 16: Also as discussed, will move this to `src/Carbon` in future PR.

jb-mackay (approved):

src/SharedUtilities/models.jl
```diff
@@ -96,23 +132,18 @@ have different dimensions - require defining a new method.
 """
 function initialize_prognostic(
     model::AbstractModel{FT},
-    state::Union{ClimaCore.Fields.Field, Vector{FT}, FT},
```
kmdeck, line 99: PS just a note - not sure where to store this mentally:

once Point is a CC domain, we dont need to support "state" being a Vector{FT} as well...

src/SharedUtilities/models.jl
```diff
@@ -127,16 +158,27 @@ have different dimensions - require defining a new method.
 """
 function initialize_auxiliary(
     model::AbstractModel{FT},
-    state::Union{ClimaCore.Fields.Field, Vector{FT}, FT},
+    state::Union{ClimaCore.Fields.Field, Vector{FT}},
```
kmdeck, line 161: is it clearer if we call this "coords"?

kmdeck: rather than `state`

kmdeck (commented):
minor comments to address in subsequent PRs

kmdeck:
(the code cov decrease is because `name(::AbstractLandSurfaceModel) = :land` is never used, but it's fine.)

PR #40, opened 2022-04-11
added in parameterizations for integrated model

kmdeck:
Add in the functions that make the heat parameters a function of water content, and the hydraulic K a function of temperature and ice content. Add unit tests for these functions.

A future PR will add in the freeze thaw source term, and integration tests/examples.


kmdeck:
bors try

PR #43, opened 2022-04-18
added bucket model plan

kmdeck:
A simple land model is useful for two main reasons:
(1) It allows the coupler team to develop machinery to couple the land and the atmos models, including computing surface fluxes, handling different types of surface cover type (snow, soil, etc), etc.
(2) It can serve as a back up land model for the AMIP simulation in the event that we are unable to get implicit time-stepping working for the soil model by then.

I looked at Manabe (1969) for the bucket model implementation. I believe this combines the slab land approach used by the coupler already, with a bucket model for soil moisture and runoff, and adds in snow. I think snow is nice to have because it adds in the surface type complexity for the coupler, but we can also not add it if it seems excessive.

There is a PDF in plans/bucket_model.pdf, the overleaf is here:
https://www.overleaf.com/7694514265pntbczwbhmmb

I am hoping for approval from @LenkaNovak @jb-mackay @tapios @yairchn before implementing the boxed equations in the PDF. Ben, we can discuss how best to implement if this is approved. Yair, I tagged you because the surface fluxes calculation slightly changes depending on coverage type.

Completing this would close this https://github.com/CliMA/ClimaCoupler.jl/issues/41. The Manabe paper is here: https://journals.ametsoc.org/view/journals/mwre/97/11/1520-0493_1969_097_0739_catoc_2_3_co_2.xml?tab_body=pdf


kmdeck:
Thanks to Tapio for the feedback (on the ClimaCoupler PR, addressed in the overleaf) - we will start implementing this next Monday barring any objections! Lenka and Yair I tagged you as reviewers too - but we can discuss in person if that is easier. For @yairchn , the main thing is looking at the soil evaporation expression and discussing how to handle different surface coverage types. @LenkaNovak I will take you up on the pair programming if you are still interested and available :)

LenkaNovak:
@kmdeck thanks for putting this together, and sure thing, I'm down for pair programming! 

Just a couple of comments:
- snow: this will probably require a bit more thinking, but we can probably modify the 0-layer sea ice model in (the appendix of) Semtner, which we now have implemented [here](https://github.com/CliMA/ClimaCoupler.jl/pull/46) (fresh out of the oven, so still in need of reviewing and clean up) 
- radiation fluxes: we will likely need `SurfaceFluxes.jl` to supply radiative fluxes and their derivetives wrt to `T_surface` (and also the derivatives of the turbulent sensible and latent heat fluxes), needed by these simple implicit calcuations. I'm doing this manually in the coupler for the sea ice now, but it should be moved. @yairchn We'll be chatting about this in the Coupler meeting today.
- interface: are we still ok to keep the bucket model in `ClimaLSM`, so that we can be developing `ClimaLSM -> ClimaCoupler` interfaces in parallel? When the bucket is substituted by the full LSM model, the coupler is more than happy to host all the slab-like models. :)

kmdeck:
hey Lenka! Thank you for the comments. I will slack you about pair programming!

a few responses:
Yes, I'm happy to have this in ClimaLSM for the reasons we discussed. It will be good for us to see the coupler interface & build to that.

- Re: the Semtner sea ice model. It is quite similar to what we have in the bucket doc. We also followed Semter, Fig 1 and equation on that page, for the soil. Snow is treated more simply, following Manabe. That said, the models are extremely similar, except:

1.  the soil cannot generate snow like the ocean can generate sea ice, when T falls below the freezing point. so we our "phase change" term only includes melting. However, they are the same form (net flux/latent heat gives volume flux m/s).
2. We also take into account sublimation of snow (E) and precipitation of snow P, so our dSWE/dt is different (=P-E-M). 
3. We track SWE (snow water equivalent - height of water you would get if you melted the snow) instead of snow height, since in the simplest case this allows us to avoid prescribing a snow density.
4. We just assumed that the snow was at T_f (as  in Manabe). We could make it vary from T_sfc to T_f linearly. This is fine to do. Do we need to do it?

The sea temp equation and soil temp equations are also very similar  - Eq 12 in our bucket doc for no snow, and T_sfc_soil = T_f when snow is present. We can also make this better by solving Eqs 11 instead (Semtner Fig 1 but single layer).

The last difference is that we also solve for soil moisture, and this depends on precip, evaporation, and snowmelt.

So, I think we are ready to go? Some code will be reusable between the two, but I dont think we need to require them to be the exact same models.

yairchn:
Hi @kmdeck - great work and great writeup 
Here are some thoughts form the first part of the document (I hope to read it all sometimes tomorrow) :

1. To me the land model should produces values (to temperature and humidity) at the land surface just like the atmosphere model produces corresponding values at the first atmospheric level. The flux is computed by SurfaceFluxes.jl and applied to both.  

in that rational does it make to write and equation of the form of (6) for `q_v_surf`, say:

`E_soil = E₀ /ρₗ`
`q̃vₛₐₜ = qvₛₐₜ(T,p,qt) * f(W,W_f) * H(ϵ-S) + (1-H(ϵ-S))*qt_int`

here  `qvₛₐₜ` is the vapor specific humidity at the surface and `qt_int` is the total specific humidity at the first interior Atmos level. Here since E₀∝(q̃vₛₐₜ-qt_int) when H(ϵ-S) =0 the flux will be zero?

2. You might wanna change the use of Heaviside functions with something smooth (1-exp(-(ϵ-S)/S) otherwise it can make calibration harder.

kmdeck:
Hi Yair! Thanks for the comments. Looking forward to discussing the first two points more.

Re the last one, I had a typo, It should be, with a negative sign in front,:
cdT/dt = -1/d(F_sfc-F_bot). fluxes are positive in the +z direction. so, Fsfc > 0 -> loss of heat to atmos, Fbot>0 -> soil heating from below. 

if F_sfc >0, the soil is losing heat to the atmosphere. If  F_sfc > F_bot >0, we have net cooling (more heat loss to atmos than heating from below). If F_sfc >0 > F_bot, we have net cooling (heat loss to atmos and heat loss to bottom). If F_bot > F_sfc >0, we have more heating from below than heat loss to atmos, so F_sfc-F_bot <0, dT/dt >0.

I think it should work out for all the cases when the negative sign is in place, because then it is a -divergence of fluxes.

kmdeck:
Closing PR as the documentation was approved and is linked on the PR adding in the bucket model, https://github.com/CliMA/ClimaCoupler.jl/issues/50

PR #44, opened 2022-04-19
Add cubed sphere domain

kmdeck:
This adds in the spherical shell domain (thank you ClimaAtmos) and adds support for it in our soil RHS functions. We also add in tests using the spherical shell.

The RHS functions which specify a boundary flux as a WVector (rather than a Covariant3Vector - as our flux is a scalar * gradient of a scalar) seem to be fine. Internally, this must be converted to a Covariant3Vector. the test `test/Soil/soil_bc/jl` shows that you obtain the same outcome regardless of how you pass in the boundary flux vector, provided that you adjust for the difference in scale.

**We have to revisit this once we allow for topography.**


kmdeck:
One other comment is that we should clarify that we are supporting boundary fluxes that are normal! will edit documentation of `FluxBC` to reflect this.

kmdeck:
@simonbyrne we would appreciate a quick look over of the `make_rhs` methods for Soil/rre.jl and Soil/energy_hydrology.jl! want to make sure this works for both cartesian and spherical domains w/o topography. in testing it seemed OK.

src/SharedUtilities/Domains.jl
```diff
@@ -302,13 +302,94 @@ function make_function_space(domain::HybridBox{FT}) where {FT}
...
+
 """
-   coordinates(domain::Union{Column{FT}, Plane{FT}, HybridBox{FT}}) where {FT}
+   coordinates(domain::Union{Column{FT}, Plane{FT}, HybridBox{FT}, SphericalShell{FT}}) where {FT}
 
 Returns the coordinate field for the domain.
 """
 function coordinates(
-    domain::Union{Column{FT}, Plane{FT}, HybridBox{FT}},
+    domain::Union{Column{FT}, Plane{FT}, HybridBox{FT}, SphericalShell{FT}},
```
jb-mackay, line 392: I think this can be `domain::AbstractDomain` which encompasses all of these types

src/Soil/Soil.jl
```diff
@@ -156,12 +156,15 @@ For column domains, no dss is needed.
 function dss!(dY::ClimaCore.Fields.FieldVector, domain::Column) end
 
 """
-   dss!(dY::ClimaCore.Fields.FieldVector,domain::HybridBox)
+   dss!(dY::ClimaCore.Fields.FieldVector,domain::{HybridBox, SphericalShell})
```
jb-mackay, line 159: Missing a `Union`

jb-mackay: We could subtype `AbstractDomain` with something like `AbstractHybridDomain`, perhaps? Depends on what the common feature of these domains and their functionality is.

src/Soil/rre.jl
```diff
@@ -92,14 +92,29 @@ function ClimaLSM.make_rhs(model::RichardsModel)
...
+        # Therefore, we should set boundary conditions in terms of a covariant vector
+        # We set the third component first - supply a Covariant3Vector
+
+        # Without topography only
+        # In Cartesian coordinates, W (z^) = Cov3 (z^)= Contra3 (n^ = z^)
+        # In spherical coordinates, W (r^) = Cov3 (r^) = Contra3 (n^ = r^)
+
+        # It appears that the WVector is converted internally to a Covariant3Vector for the gradient value
+        # at the boundary. Offline tests indicate that you get the same thing if
+        # the bc is WVector(F) or Covariant3Vector(F*Δr) or Contravariant3Vector(F/Δr)
```
jb-mackay, line 110: Feel free to keep or remove these, whichever you feel is more useful at this stage.

test/domains.jl
```diff
@@ -5,15 +5,21 @@ if !("." in LOAD_PATH)
...
+    RootDomain,
+    HybridBox,
+    Plane,
+    Point,
+    LSMSingleColumnDomain,
+    SphericalShell
 using ClimaLSM.Domains: coordinates
 
-FT = Float64
+FT = Float32
```
jb-mackay, line 17: Maybe have TestFloatTypes = (Float32, Float64) and iterate over these in each testset. e.g. [here](https://github.com/CliMA/ClimaCore.jl/blob/23e8c9a7344dd7164bad2fffdd478dcf9f47e0c6/test/DataLayouts/data1d.jl#L8-L11)

test/domains.jl
```diff
@@ -79,3 +85,19 @@ end
...
+
+@testset "Spherical Shell Domain" begin
+    domain = SphericalShell(;
+        radius = FT(100.0),
+        height = FT(30.0),
+        nelements = (6, 20),
+        npolynomial = 3,
+    )
+    @test domain.radius == FT.(100)
+    @test domain.height == FT.(30)
```
jb-mackay, line 97: Don't technically need dot operation here

jb-mackay (approved):

src/SharedUtilities/Domains.jl
kmdeck: good catch!

src/Soil/Soil.jl
kmdeck: Yes, this could be useful! Especially when we rearrange rhs operations to loop over columns, it might come up more often. The common feature is that they are extrudedFD spaces, with a spectral horizontal space, like you said. 

Maybe one other thing to check - what does dss! do if we are on a column? Would it be confusing if we do it all the time, if it doesnt do anything for a column field?

test/domains.jl
kmdeck: ok! I also combined some of the domain tests into a single "testset" (all the CC ones, i.e. not the Point yet, and not the LSMDomain) to clean up the printed output

kmdeck:
bors try

src/Soil/Soil.jl
jb-mackay: Right, we have it written that `dss!` does nothing on the column but actually does dss for the other domains. I don't think this is particularly confusing and the great benefit is that we have a single RHS.

kmdeck: ok cool! once we do the "loop over columns" PR we can simplify this too, or before then if it makes sense

PR #50, opened 2022-04-30
Bucket model

kmdeck:
PR adding in the bucket model detailed in doc
https://www.overleaf.com/project/625c5f7677155b314c98eb75
and in design PR: https://github.com/CliMA/ClimaLSM.jl/pull/43
and the Coupler Issue: https://github.com/CliMA/ClimaCoupler.jl/issues/41

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,6 +9,8 @@ ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+SurfaceFluxes = "49b00bb7-8bd4-4f2b-b78c-51cd0450215f"
+Thermodynamics = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 
 [compat]
--- a/src/Bucket/Bucket.jl
+++ b/src/Bucket/Bucket.jl
@@ -1,0 +1,433 @@
+module Bucket
+# (1) Add snow - is there a better way to handle timesteps where only
+# some of the surface energy is used to melt all of the snow?
+# (2) Make atmos functions a function of space (lat lon? x y? coords?)
+# (3) Look at allocations & performance...
+using UnPack
+using DocStringExtensions
+using CLIMAParameters: AbstractEarthParameterSet
+using CLIMAParameters: Stefan
+using CLIMAParameters.Planet:
+    LH_v0, LH_s0, LH_f0, ρ_cloud_liq, T_freeze, T_0, R_d, cp_d, grav, cp_v
+using SurfaceFluxes
+using SurfaceFluxes.UniversalFunctions
+using Thermodynamics
+using StaticArrays
+
+using ClimaLSM
+import ClimaLSM.Domains: coordinates, Plane
+import ClimaLSM:
+    AbstractModel,
+    make_update_aux,
+    make_rhs,
+    prognostic_vars,
+    auxiliary_vars,
+    name,
+    prognostic_types,
+    auxiliary_types
+export BucketModelParameters,
+    PrescribedAtmosphere,
+    PrescribedRadiation,
+    BucketModel,
+    CoupledRadiation,
+    CoupledAtmosphere,
+    surface_fluxes
+include("./bucket_parameterizations.jl")
+
+abstract type AbstractBucketModel{FT} <: AbstractModel{FT} end
+abstract type AbstractAtmosphericDrivers{FT <: AbstractFloat} end
+abstract type AbstractRadiativeDrivers{FT <: AbstractFloat} end
+
+ClimaLSM.name(::AbstractBucketModel) = :bucket
+
+"""
+    struct BucketModelParameters{
+        FT <: AbstractFloat,
+        PSE <: AbstractEarthParameterSet,
+    }
+
+Container for holding the parameters of the bucket model.
+
+$(DocStringExtensions.FIELDS)
+"""
+struct BucketModelParameters{
+    FT <: AbstractFloat,
+    PSE <: AbstractEarthParameterSet,
+}
+    "Depth of the soil column (m) used in heat equation"
+    d_soil::FT
+    "Temperature at the bottom of the soil column (K); constant"
+    T0::FT
+    "Conductivity of the soil (W/K/m); constant"
+    κ_soil::FT
+    "Volumetric heat capacity of the soil (J/m^3/K); constant"
+    ρc_soil::FT
+    "Shortwave soil albedo; constant"
+    α_soil::FT
+    "Shortwave snow albedo; constant"
+    α_snow::FT
+    "Critical SWE amount (m) where surface transitions from soil to snow"
+    S_c::FT
+    "Capacity of the land bucket (m)"
+    W_f::FT
+    "Roughness length for momentum (m)"
+    z_0m::FT
+    "Roughness length for scalars (m)"
+    z_0b::FT
+    "Earth Parameter set; physical constants, etc"
+    earth_param_set::PSE
+end
+
+BucketModelParameters(
+    d_soil::FT,
+    T0::FT,
+    κ_soil::FT,
+    ρc_soil::FT,
+    α_soil::FT,
+    α_snow::FT,
+    S_c::FT,
+    W_f::FT,
+    z_0m::FT,
+    z_0b::FT,
+    earth_param_set::PSE,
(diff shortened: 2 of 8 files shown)


kmdeck:
in "coupled" mode should we swtich from "Prescribed" to "Coupled" and dispatch surface_fluxes() on type >

Project.toml
```diff
@@ -8,7 +8,10 @@ CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
+RootSolvers = "7181ea78-2dcb-4de3-ab41-2b8ab5a31e74"
```
kmdeck, line 11: can remove this

src/Bucket/bucket_parameterizations.jl
```diff
@@ -0,0 +1,93 @@
...
+Extra inflow when the bucket is at capacity runs off.
+"""
+function infiltration_at_point(
+    W::FT,
+    M::FT,
+    P::FT,
+    E::FT,
+    W_f::FT,
+)::FT where {FT <: AbstractFloat}
+    if (W > W_f) & (P + M > E) # Equation (4b) of text
```
kmdeck, line 49: double check this if E is negative - condensation

src/Bucket/bucket_parameterizations.jl
```diff
@@ -0,0 +1,93 @@
...
+        return FT(E)
+    else
+        return FT(P + M)
+    end
+end
+
+"""
+    β(W::FT, W_f::FT) where {FT}
+
+Returns the coefficient which scales the saturated
```
kmdeck, line 59: eventually this can be a user supplied function...and this can be default?

kmdeck (commented):
add docs here
https://github.com/CliMA/ClimaLSM.jl/blob/c71d2830a3786abf8d86c37fe8874d0c878e652f/src/Bucket/Bucket.jl#L60

kmdeck (commented):
other items that can go in this PR:
- surface density from atmos density and (reference height-surface height) and scale height of atmos
- make surface height general

kmdeck:
@tapios One note/change from design doc is that we had to stick the "beta" factor (decreasing the evaporation from the potential rate) in front of q_sat, so that evaporation is given by
rho*u*C_d*(beta*qsat - q_atmos), rather than beta *rho*u*C_d*(qsat - q_atmos). the reason for this is that the SHF, LHF and E are computed in SF.jl - and SHF and LHF depend on E - and SF.jl does not take into account prefactors (or other resistance terms). so the way we thought to make it work for now is to adjust q_sfc directly. @yairchn 

perhaps a benefit of this is that as W->0, q_sfc ->0, and then we get condensation, whereas with the beta out front, as W->0, water fluxes go to zero, which isnt strictly true - condensation should occur then, we think. @jb-mackay

src/Bucket/Bucket.jl
```diff
@@ -0,0 +1,419 @@
...
+        earth_param_set,
+        sc,
+        SurfaceFluxes.UniversalFunctions.Businger(),
+    )
+
+    α = surface_albedo(α_soil, α_snow, S, S_c)
+    R_n = -((FT(1) - α) * SW_d(t) + LW_d(t) - _σ * T_sfc^FT(4.0))
+    # Land needs a volume flux of water, not mass flux
+    # If this evaporation is positive -> towards atmos
+    E = SurfaceFluxes.evaporation(sc, earth_param_set, conditions.Ch) / _ρ_liq
```
yairchn, line 301: I will add `E` to the output of SF so you won't need to call it here in the future. for now this is fine 
@akshaysridhar what do you think ?

yairchn:
@kmdeck for `beta ->0` this formulation 
 `rho*u*Cd*(beta*qsat - qatmos)`
might not provide zero flux. We prob need 
`beta = qsat * H(beta) + qatmos *(1-H(beta))`
 where H is heavyside function to get that behavior

tapios:
> @tapios One note/change from design doc is that we had to stick the "beta" factor (decreasing the evaporation from the potential rate) in front of q_sat, so that evaporation is given by `rho*u*Cd*(beta*qsat - qatmos)`, rather than `beta *rho*u*Cd*(qsat - qatmos)`. the reason for this is that the SHF, LHF and E are computed in SF.jl - and SHF and LHF depend on E - and SF.jl does not take into account prefactors (or other resistance terms). so the way we thought to make it work for now is to adjust q_sfc directly. @yairchn
> 
> perhaps a benefit of this is that as W->0, q_sfc ->0, and then we get condensation, whereas with the beta out front, as W->0, water fluxes go to zero, which isnt strictly true - condensation should occur then, we think. @jb-mackay

This actually makes physical sense, more so than the standard formulation: The flux should always be proportional to `(q_sfc - q_a)` (downgradient fluxes), with q_sfc limited by saturation. So, yes, this is fine. (Maybe just a question whether the condensation fluxes, when beta is small, should be limited.)

kmdeck:
> @kmdeck for `beta ->0` this formulation `rho*u*Cd*(beta*qsat - qatmos)` might not provide zero flux. We prob need `rho*u*Cd*(beta_H*qsat - qatmos)`
> 
> `beta_H = qsat * H(beta) + qatmos *(1-H(beta))` where H is heavyside function to get that behavior

We'll leave this as is - see discussion below. The behavior of flux ->0 as water content in land ->0 isnt really correct, even if it is what most employ. But youre right that if we want to recapture it, what you wrote would work!

src/Bucket/bucket_parameterizations.jl
kmdeck: This is fine, slightly redefined it in the overleaf to have infiltration I satisfy \vec{F} = -I z^, so infiltration is related to the boundary flux directly.

jb-mackay (approved):

src/Bucket/Bucket.jl
```diff
@@ -0,0 +1,433 @@
...
+
+
+"""
+    PrescribedRadiation{FT, SW, LW} <: AbstractRadiativeDrivers{FT}
+
+Container for the prescribed radiation functions needed to drive the
+bucket model in standalone mode.
+$(DocStringExtensions.FIELDS)
+"""
+struct PrescribedRadiation{FT, SW, LW} <: AbstractRadiativeDrivers{FT}
```
LenkaNovak, line 167: Would `PrescribedRadiationSurfaceFluxes` be a better name or is this obvious to a "land person"? 🤔  Also, for idealized testing in land, would it be useful to have `PrescribedTurbulentSurfaceFluxes` (i.e. bypass `SurfaceFluxes.jl` for LHF and SHF)?

src/Bucket/Bucket.jl
```diff
@@ -0,0 +1,433 @@
...
+    PR <: PrescribedRadiation{FT},
+}
+    @unpack ρ_atmos, T_atmos, u_atmos, q_atmos, h_atmos, ρ_sfc = atmos
+    @unpack α_soil, α_snow, z_0m, z_0b, S_c, earth_param_set = parameters
+    @unpack LW_d, SW_d = radiation
+    _σ = Stefan()
+    _ρ_liq = ρ_cloud_liq(earth_param_set)
+
+    # call surface fluxes for E, SHF, LHF
+    ts_sfc = Thermodynamics.PhaseEquil_ρTq(earth_param_set, ρ_sfc, T_sfc, q_sfc)
```
LenkaNovak, line 275: Should we add a warning that if `ρ_sfc = ρ_atmos`, this will break the flux calculation? Perhaps this better suited for `SurfaceFluxes.jl` though.

test/Bucket/bucket_test.jl
```diff
@@ -0,0 +1,305 @@
...
+    Y.bucket.W .= 0.149
+    Y.bucket.Ws .= 0.0
+    Y.bucket.S .= 0.0
+    ode_function! = make_ode_function(model)
+
+    t0 = 0.0
+    tf = 10000.0
+    prob = ODEProblem(ode_function!, Y, (t0, tf), p)
+    Δt = 100.0
+    # need a callback to get `p`
```
LenkaNovak, line 258: Would using DiffEq's `init` and `step!` work for the time stepping here? `p` (as well as other useful info) could then be accessed via the integrator.

LenkaNovak (approved):

LenkaNovak:
Hi @kmdeck and @jb-mackay I left a few comments but they are mainly stylistic. All checks and interface look great! Good job! 🚀

src/Bucket/Bucket.jl
kmdeck: I changed the name to PrescribedRadiativeFluxes - good suggestion. I think for now we can hold off on adding in a PrescribedTurbulentSurfaceFluxes type (unless there is a use case for it at the moment). Since we have the `AbstractAtmosphericDrivers` and `AbstractRadiativeDrivers` types set up, and two example of each, we are ready to add in when needed.

PR #51, opened 2022-05-06
Update Point Domain to ClimaCore and add interactions support

jb-mackay:
Makes the `Point` domain consistent with the other domains in that is backed by ClimaCore spaces (`ClimaCore.Spaces.PointSpace`). This allows for `PointFields` to be defined for surface column models.

Minimal changes were needed to update the standalone Pond model. The pond-soil LSM model needs debugging around space instances. Roots model will be updated in a following PR.

Additionally, this PR changes the way that `interactions` auxiliary variables are defined, making them more consistent with other model variables through the definition of `interactions_vars`, `interactions_types`, and `interactions_spaces` methods (see comment below).

### DIFF
--- a/src/ClimaLSM.jl
+++ b/src/ClimaLSM.jl
@@ -51,7 +51,8 @@ function initialize(land::AbstractLandModel{FT}) where {FT}
             component,
         )
     end
-    p_interactions = initialize_interactions(land)
+    p_interactions =
+        initialize_interactions(land, NamedTuple{components}(coords_list))
 
     Y = ClimaCore.Fields.FieldVector(;
         NamedTuple(zip(components, Y_state_list))...,
@@ -103,22 +104,92 @@ Returns the component names of the `land` model, by calling
 """
 land_components(land::AbstractLandModel) = propertynames(land)
 
+function prognostic_vars(land::AbstractLandModel)
+    components = land_components(land)
+    prognostic_list = map(components) do model
+        prognostic_vars(getproperty(land, model))
+    end
+    return NamedTuple{components}(prognostic_list)
+end
+
+function prognostic_types(land::AbstractLandModel)
+    components = land_components(land)
+    prognostic_list = map(components) do model
+        prognostic_types(getproperty(land, model))
+    end
+    return NamedTuple{components}(prognostic_list)
+end
+
+function auxiliary_vars(land::AbstractLandModel)
+    components = land_components(land)
+    auxiliary_list = map(components) do model
+        auxiliary_vars(getproperty(land, model))
+    end
+    return NamedTuple{(components..., :interactions)}((
+        auxiliary_list...,
+        interaction_vars(land),
+    ))
+end
+
+function auxiliary_types(land::AbstractLandModel)
+    components = land_components(land)
+    auxiliary_list = map(components) do model
+        auxiliary_types(getproperty(land, model))
+    end
+    return NamedTuple{(components..., :interactions)}((
+        auxiliary_list...,
+        interaction_types(land),
+    ))
+end
+
+"""
+   interaction_vars(m::AbstractModel)
+
+Returns the interaction variable symbols for the model in the form of a tuple.
+"""
+interactios_vars(m::AbstractLandModel) = ()
+
+"""
+   interaction_types(m::AbstractModel)
+
+Returns the shared interaction variable types for the model in the form of a tuple.
+"""
+interaction_types(m::AbstractLandModel) = ()
+
+"""
+   interaction_domains(m::AbstractModel)
+
+Returns the interaction domain symbols in the form of a tuple.
+
+For example, for a boundary interaction term, one would specify a model with a
+domain consisting of the boundary space. Most commonly, these are surface models.
+This is only required for variables shared between land submodels, and only needed
+for multi-component models, not standalone components. Component-specific variables
+should be listed as prognostic or auxiliary variables.
+"""
+interaction_domains(m::AbstractLandModel) = ()
+
 """
     initialize_interactions(land::AbstractLandModel) end
 
 Initializes interaction variables, which are a type of auxiliary
-variable, to empty objects of the correct type
-for the model. 
+variable, to empty objects of the correct type for the model. 
 
-This function should be called during `initialize_auxiliary`,
-and the method will change depending on the type of land model,
-and potentially due to the type of the component models. 
-
-This is a stub which is extended to concrete LSM models.
-
+Interaction variables are specified by `interaction_vars`, their types
+by `interaction_types`, and their spaces by `interaction_spaces`. 
+This function should be called during `initialize_auxiliary` step.
 """
-function initialize_interactions(land::AbstractLandModel) end
+function initialize_interactions(land::AbstractLandModel, land_coords)
+    vars = interaction_vars(land)
+    types = interaction_types(land)
+    domains = interaction_domains(land)
+    interactions = map(zip(types, domains)) do (T, domain)
+        zero_instance = zero(T)
+        map(_ -> zero_instance, getproperty(land_coords, domain))
+    end
 
+    return NamedTuple{vars}(interactions)
+end
 """
     make_interactions_update_aux(land::AbstractLandModel) end
 
--- a/src/SharedUtilities/Domains.jl
+++ b/src/SharedUtilities/Domains.jl
@@ -16,14 +16,17 @@ An abstract type for domains.
 """
 abstract type AbstractDomain{FT <: AbstractFloat} end
 Base.eltype(::AbstractDomain{FT}) where {FT} = FT
+
 """
     coordinates(domain::AbstractDomain)
 
-Method which returns the coordinates appropriate for a given domain.
-
-The coordinates can be Fields or Vectors.
+Returns the coordinate field for the domain.
 """
-function coordinates(domain::AbstractDomain) end
+function coordinates(domain::AbstractDomain)
+    cs, _ = make_function_space(domain)
+    cc = ClimaCore.Fields.coordinate_field(cs)
+    return cc
+end
 
 """
     Point{FT} <: AbstractDomain{FT}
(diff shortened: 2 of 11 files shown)


src/ClimaLSM.jl
```diff
@@ -103,22 +104,89 @@ Returns the component names of the `land` model, by calling
...
+function auxiliary_types(land::AbstractLandModel)
+    components = land_components(land)
+    auxiliary_list = map(components) do model
+        auxiliary_types(getproperty(land, model))
+    end
+    return NamedTuple{(components..., :interactions)}((
+        auxiliary_list...,
+        interactions_types(land),
+    ))
+end
```
jb-mackay, line 143: @kmdeck This extends `prognostic_vars` and `prognostic_types` (and for auxiliary) to `AbstractLandModel`s. It returns a named tuple that nests the prognostic/aux vars for each submodel. In the case of auxiliary vars, the interaction variables are also included (non-nested, just like the state/aux containers).

src/ClimaLSM.jl
```diff
@@ -103,22 +104,89 @@ Returns the component names of the `land` model, by calling
...
-This function should be called during `initialize_auxiliary`,
-and the method will change depending on the type of land model,
-and potentially due to the type of the component models. 
+Returns the symbol of a model on the interaction space in the form of a tuple.
 
-This is a stub which is extended to concrete LSM models.
+For example, for a boundary interaction term, one would specify a model with a
+domain consisting of the boundary space. Most commonly, these are surface models.
+"""
+interactions_spaces(m::AbstractLandModel) = ()
```
jb-mackay, line 167: This is the only piece I'm a little grumpy about. It associates interactions variables with a submodel so that they can be initialized on the correct space. The downside is that these interactions don't really belong to a given submodel. So perhaps there is a better way to specify what boundaries these interactions occur on, but this is fast and easy.

Please suggest any language that would make this function & its docstring more clear.

src/ClimaLSM.jl
```diff
@@ -103,22 +104,89 @@ Returns the component names of the `land` model, by calling
...
+    vars = interactions_vars(land)
+    types = interactions_types(land)
+    domains = interactions_spaces(land)
+    interactions = map(zip(types, domains)) do (T, domain)
+        zero_instance = zero(T)
+        map(_ -> zero_instance, getproperty(land_coords, domain))
+    end
 
+    return NamedTuple{vars}(interactions)
+end
```
jb-mackay, line 189: Generic initialize - we may be able to unify this with `initialize_vars` but that would require changing the output type of that (currently a `FieldVector`). We would probably return a pair of `keys`, `zero_instances` where `keys` are the variable names and the `zero_instances` are the initialized fields. Then we can package that into a `NamedTuple` or `FieldVector`.

src/SharedUtilities/Domains.jl
```diff
@@ -16,14 +16,17 @@ An abstract type for domains.
...
 """
     coordinates(domain::AbstractDomain)
 
-Method which returns the coordinates appropriate for a given domain.
-
-The coordinates can be Fields or Vectors.
+Returns the coordinate field for the domain.
 """
-function coordinates(domain::AbstractDomain) end
+function coordinates(domain::AbstractDomain)
```
jb-mackay, line 25: As we discussed, `coordinates` still calls `make_function_space`. It should be called only once during the initialization step.

src/pond_soil_model.jl
```diff
@@ -242,5 +233,5 @@ function Soil.boundary_fluxes(
     p::ClimaCore.Fields.FieldVector,
     t::FT,
 ) where {FT}
-    return p.soil_infiltration[1], FT(0.0)
+    return p.soil_infiltration[], FT(0.0)
```
jb-mackay, line 236: You can access the values in a `ClimaCore.DataLayouts.DataF` (the backing datatype for the `PointField`) with `[]`. You will get `getindex` bounds errors otherwise.

src/ClimaLSM.jl
kmdeck: Right - this doesnt need to return something can be run in standalone mode. I think I'm fine with it being its own thing & not part of a single `init_vars`  - this is all part of the LSM `initialize` method, which we are specializing for (it's different from the `initialize` of `AbstractModel`, though it make use of some of the generic methods that `init` of abstractmodels employs). We can also change the names to be more clear? `initialize_vars` vs `initialize_vectors`? Not sure if that is necessary but it is an option.

src/SharedUtilities/Domains.jl
kmdeck: sounds good!

src/pond_soil_model.jl
kmdeck: Ah OK. So, the boundary_fluxes method for the soil model returns floats. I guess we can broadcast these methods over fields (and this would still pull out the "float" object correctly) or we can still apply at each column (and it would do the same, since column (planar surface) =Point).

This looks fine for now!

kmdeck (commented):
One other thing to look at before merging:
the drop in code cov seems mainly due to these lines:
https://app.codecov.io/gh/CliMA/ClimaLSM.jl/compare/51/tree/src/ClimaLSM.jl#D1L107

Im wondering why they arent being used?

src/ClimaLSM.jl
kmdeck: Another idea: let's call it `interactions_domain`, b/c that is what we are really talking about, and reserve "spaces" for a future use case where we need to distinguish between centers and faces?

and we can also note that component specific variables do not require this information, because by definition they are defined at each point in the coordinate field of their component's domain. Interaction variables are more flexible, so we had to introduce this.

jb-mackay: > let's call it `interactions_domain`

Sounds good - that makes it more consistent with our use of "domain" in LSM.

> and we can also note that component specific variables do not require this information

I'll add a note to the docstring :)

>Maybe instead of associating them with a component, we can associate them with a part of the LSMDomain? This is split into subsurface and surface domains (domain.subsurface, domain.surface). It would be more in line with them being shared to pass in :surface or :subsurface rather than e.g. :soil?

I like that, need to think about the easiest way to implement that. Mostly, it is then associating those tags with component models.

src/ClimaLSM.jl
jb-mackay: I think it makes sense to unify things, but maybe in a following PR. A benefit to returning the `keys` and `zero_instances` is that we could then have flexibility in how we package these states (FieldVectors, NamedTuples).

src/pond_soil_model.jl
jb-mackay: I need to play with this, but actually you make a good point that we shouldn't need to call into the DataLayout here, we can just return the `PointField`.

jb-mackay:
> One other thing to look at before merging: the drop in code cov seems mainly due to these lines: https://app.codecov.io/gh/CliMA/ClimaLSM.jl/compare/51/tree/src/ClimaLSM.jl#D1L107
> 
> Im wondering why they arent being used?

Right, yes I don't think they are called anywhere in the tests. I will write some unit tests for them.

kmdeck:
> > One other thing to look at before merging: the drop in code cov seems mainly due to these lines: https://app.codecov.io/gh/CliMA/ClimaLSM.jl/compare/51/tree/src/ClimaLSM.jl#D1L107
> > Im wondering why they arent being used?
> 
> Right, yes I don't think they are called anywhere in the tests. I will write some unit tests for them.

I think maybe because we wrote our own `initialize` for land, which calls init_prog for each component model. https://github.com/CliMA/ClimaLSM.jl/blob/03f9afc7ac6e7b3a978b4033388cc65918b2da75/src/ClimaLSM.jl#L35

but we dont actually use `init_prog(land)` anywhere?

jb-mackay:
> > > One other thing to look at before merging: the drop in code cov seems mainly due to these lines: https://app.codecov.io/gh/CliMA/ClimaLSM.jl/compare/51/tree/src/ClimaLSM.jl#D1L107
> > > Im wondering why they arent being used?
> > 
> > 
> > Right, yes I don't think they are called anywhere in the tests. I will write some unit tests for them.
> 
> I think maybe because we wrote our own `initialize` for land, which calls init_prog (with prog_vars, prog_types, etc) for each component model.
> 
> https://github.com/CliMA/ClimaLSM.jl/blob/03f9afc7ac6e7b3a978b4033388cc65918b2da75/src/ClimaLSM.jl#L35
> 
> but we dont actually use `init_prog(land)`, `prog_vars(land)` anywhere.

Yes, the purpose of `prognostic_vars(land)` and `auxiliary_vars(land)` is more for consistency and if you wanted to see what those vars are for a model. Perhaps they could be used in `initialize` too, but not sure if that's necessary or possible.

jb-mackay:
> > > One other thing to look at before merging: the drop in code cov seems mainly due to these lines: https://app.codecov.io/gh/CliMA/ClimaLSM.jl/compare/51/tree/src/ClimaLSM.jl#D1L107
> > > Im wondering why they arent being used?
> > 
> > 
> > Right, yes I don't think they are called anywhere in the tests. I will write some unit tests for them.
> 
> I think maybe because we wrote our own `initialize` for land, which calls init_prog (with prog_vars, prog_types, etc) for each component model.
> 
> https://github.com/CliMA/ClimaLSM.jl/blob/03f9afc7ac6e7b3a978b4033388cc65918b2da75/src/ClimaLSM.jl#L35
> 
> but we dont actually use `init_prog(land)`, `prog_vars(land)` anywhere.

It is actually a good idea to define these for land models! This will make `initialize` much easier to follow. Right now it just means hiding the `do` blocks. Let's do this in a follow on PR as well

kmdeck:
> It is actually a good idea to define these for land models! This will make `initialize` much easier to follow. Right now it just means hiding the `do` blocks. Let's do this in a follow on PR as well

ok! that sounds good then

src/pond_soil_model.jl
jb-mackay: This is resolved - see the changes to `rre.jl`

src/ClimaLSM.jl
jb-mackay: > It would be more in line with them being shared to pass in :surface or :subsurface rather than e.g. :soil?

I think this will change a few other things, so I will add this in a separate PR since this is getting quite long already.

PR #64, opened 2022-05-17
start adding snow

kmdeck:
A first pass at a snow model, based on the design doc linked in the corresponding issue and here https://www.overleaf.com/7694514265pntbczwbhmmb 
SLIM code melting term: https://github.com/marysa/SimpleLand/blob/6e986f9ca7d7183c28b501267721828c3b7110b2/src/main/mml_main.F90#L1064

We set the snow cover fraction to a heaviside function, so we only compute one set of surface fluxes each step. Otherwise we have implemented the equations in the doc as noted.

These conserve energy and water but have not been tested beyond that.

The melt term formulation can be changed as we wish in the future or now.


src/Bucket/Bucket.jl
```diff
@@ -357,37 +363,45 @@ Creates the rhs! function for the bucket model.
...
+        snow_cover_fraction = heaviside.(Y.bucket.σS)
+
+        #Currently, the entire surface is assumed to be snow covered or soil covered,
+        # so there is just one set of fluxes.
+        # Since q itself is modified for soil vs snow, and since the surface temperature is
+        # assumed to be the same for each, the fluxes we get back should be accurate
+        # in either case
+        # The below is NOT CORRECT if we want the snow cover fraction to be intermediate between
+        # 0 and 1.
+        @unpack SHF, LHF, R_n, E, G, F_melt = p.bucket
```
kmdeck, line 379: make sure eqn #s are correct

src/Bucket/bucket_parameterizations.jl
```diff
@@ -12,22 +12,32 @@ function heaviside(x::FT)::FT where {FT}
 end
 
 """
-    surface_albedo(α_soil::FT, α_snow::FT, S::FT, S_c::FT)::FT where {FT <: AbstractFloat}
+"""
```
kmdeck, line 15: doc string

kmdeck:
@LenkaNovak I know you're super busy, but I think this is more or less ready for review!

docs/src/APIs/Bucket.md
```diff
@@ -21,5 +21,7 @@ ClimaLSM.Bucket.BucketModel
 ClimaLSM.Bucket.surface_fluxes
 ClimaLSM.Bucket.surface_air_density
 ClimaLSM.Bucket.liquid_precipitation
+ClimaLSM.Bucket.snow_precipitation
```
kmdeck, line 24: add net_radiation

docs/tutorials/coupled_bucket.jl
```diff
@@ -71,44 +71,52 @@ FT = Float64;
```
kmdeck, line 71: import snow_precipitation

PR #66, opened 2022-05-26
added method, removed coupled types

kmdeck:
Make surface specific humidity a function, and move export. move all coupled items out to put into the coupler.


kmdeck:
> Looks good - we can open a PR in the Coupler that adds a ClimaLSM folder that extends LSM methods with the coupled boundary conditions

thanks Ben! Will merge this.

(Right now the method extensions are in the bucket_init.jl file)

PR #68, opened 2022-06-20
Use new way of handling clima parameters

charleskawczynski:
This PR applies the new way of handling clima parameters:
 - Adds a package module, `Parameters`, where we define a struct specifically for this packages parameters.
 - The struct from this module is then used in all places where we originally called CLIMAParameter methods (e.g., `Planet.grav(param_set)` becomes `Parameters.grav(param_set)`, and I've added some aliases so that this is a bit more terse
 - Removes the use of CLIMAParameters in `src/` completely
 - Adds a folder, `parameters` to the top-level package, where a parameter-creating function was added, which constructs the struct instance that houses all of the necessary parameters are need (from this package, and its dependencies).


kmdeck:
This looks fine to me, Charlie! Thanks for upgrading.

In this, it seems like we'll have LSMParameters holding the physical constants for other modules (surface fluxes, TD) as well as planet params, and then each specific model may introduce its own specific params (soil type, etc) and create a bigger struct like

MyModelParams
- model specific params in flat list
- earth_param_set = LSMParameters()

We can discuss more if this makes sense or if the whole thing should be a flat list?

I am going to go ahead and merge this, because to use the latest ClimaAtmos (with the hot fixes @szy21 and @jiahe23 pushed) in coupled sims, we also need to be using the latest parameters style in the land model.

kmdeck (approved):

PR #101, opened 2022-10-03
included area indexes

### LINKED ISSUE #70: Suggestions of improvements to bring to the roots file
Hi Katherine and Emma! I included a list of items I think we should modify in the roots file, and a link to the most updated file (Mon. June 27) with the roots model. The file also compares the new plant hydraulic model against a finely resolved Richard's equation (truth) for a simulation in which both under and over saturation regime equations are tested, for different simulation times, and computes the absolute error between the model and the truth.

https://github.com/CliMA/ClimaLSM.jl/blob/ALG_roots_Richards/src/Vegetation/root_test_against_Richards.jl

1) To match soil model, and rest of Clima, replace all units with SI units. Change flow equations to flux, and use same units as soil model to facilitate passing variables between both.

2) The model won't currently go above saturation, I think we need to add in this capability.

3) The current flux equation used is a bit more complex (needs two functions) and does not work when the pressure values in adjacent compartments are equal (division by P2-P1) and may lead to numerical errors when the pressures are very similar, so it would be a good idea to include the alternative flux equation described in the docs and in the link included above. 

4) There are several different volume to pressure equations we can use, which are more or less complex. I think adding the linear and Christoffersen volume to pressure functions in, and renaming them to suggested names (below) would be useful. 

5) Like soil model, it would be great to keep parameters in a separate file.

6) It would also be useful to be able to switch between a high resolution single-tree simulation and a grid-scale set up, to suit different needs.

1) Unit changes : flow (in mol s-1) to flux (m s-1). See flux equations in the root_test_against_Richards file in the link above, I made the conversion here. See also bulk roots section. Note : the flux_Yujie equation needs to be revised in the file (working on this), I used the alternative flux equation (entitled flux) for the tests performed in this file.

2) Change state variable to augmented liquid fraction, add in branch in the absolute pressure function. See file.

3) Add in alternative flux equation (see file).

4) Replace the volume to pressure function name with more specific name such as van_genuchten_volume_to_pressure for the one that is already included, and add a function for a linear_pressure_to_volume as this is commonly used (eg. Nathan used this, I did while developing as well), and another for the christoffersen_volume_to_pressure. Please see file for examples of these functions.

5) Move parameters out of file as is done in soil model.

6) Add in equations for grid scale set-up (see Ozark file). Add in a type or functionality to switch between a one-tree set up and a grid scale set-up

**Additional context**
Included a link. See above.

gagnelandmanna:
# PULL REQUEST

## Purpose and Content
The purpose of this PR is to include area indexes (RAI, SAI, LAI) to account for the differences in conducting areas of each tree compartment. The area terms had been previously all set to 1m2 for simplicity.

## Benefits and Risks
The model should more accurately represent water fluxes and water content with this addition. 

## Linked Issues
(Provide references to any link issues. Use closes #issuenum to automatically close an open issue)
- Fixes #70

## PR Checklist
- [ x] This PR has a corresponding issue OR is linked to an SDI.
- [ x] I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- [ x] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [ x] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [ x] Unit tests are included OR N/A.
- [ x] Code used in an integration test OR N/A.
- [ x] All tests ran successfully on my local machine OR N/A.
- [ x] All classes, modules, and function contain appropriate docstrings OR N/A.
- [ x] Documentation has been added/updated OR N/A.

### DIFF
--- a/src/Vegetation/PlantHydraulics.jl
+++ b/src/Vegetation/PlantHydraulics.jl
@@ -112,6 +112,8 @@ A struct for holding parameters of the PlantHydraulics Model.
 $(DocStringExtensions.FIELDS)
 """
 struct PlantHydraulicsParameters{FT <: AbstractFloat, PSE}
+    "RAI, SAI, LAI in [m2 m-2]"
+    area_index::NamedTuple{(:root, :stem, :leaf), Tuple{FT, FT, FT}}
     "water conductivity in the different plant compartments (m/s) when pressure is zero, a maximum"
     K_sat::NamedTuple{(:root, :stem, :leaf), Tuple{FT, FT, FT}}
     "van Genuchten parameter"
@@ -324,7 +326,7 @@ The rhs! function must comply with a rhs function of OrdinaryDiffEq.jl.
 """
 function make_rhs(model::PlantHydraulicsModel)
     function rhs!(dY, Y, p, t)
-        @unpack vg_α, vg_n, vg_m, S_s, ν, K_sat = model.parameters
+        @unpack vg_α, vg_n, vg_m, S_s, ν, K_sat, area_index = model.parameters
 
         n_stem = model.domain.n_stem
         n_leaf = model.domain.n_leaf
@@ -334,6 +336,7 @@ function make_rhs(model::PlantHydraulicsModel)
         for i in 1:(n_stem + n_leaf)
             if i == 1
                 flux_in = flux_out_roots(model.root_extraction, model, Y, p, t)
+                area_index_in = area_index[:root]
             else
                 flux_in = flux(
                     model.domain.compartment_midpoints[i - 1],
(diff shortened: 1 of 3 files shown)


kmdeck (commented):
looks fine to me, aside from comments offline about whether area indices should be computed at the faces, rather than the centers of the compartments.

we also might consider using area_index instead of AI to be clear.

kmdeck (approved):

gagnelandmanna:
> looks fine to me, aside from comments offline about whether area indices should be computed at the faces, rather than the centers of the compartments.
> 
> we also might consider using area_index instead of AI to be clear.

Sounds good, fixed both.

gagnelandmanna:
r+

Issue #108, opened 2022-10-27
Update the ozark test


gagnelandmanna:
**Is your feature request related to a problem? Please describe.**
Need to bring the Ozark test up to date with main, which now includes an updated version of the plant hydraulics. This will allow us to test the coupled plant hydraulics and soil models

**Describe the solution you'd like**
The solution 
- will have SI units consistent with the rest of the model (m3 of H2O instead of moles, fluxes in m/s-1, etc.)
- it will allow for the user to set the number of stem and leaf compartments
- will include forcing data from Fluxnet site instead of from Natan's run
- will include parameters which are currently being used in the PlantHydraulics module (vg consistent wth soil model instead of modified logistic)

**Describe alternatives you've considered**
- using a different site (other than Ozark) which has soil and vegetation data for one tree

**Additional context**
NA

gagnelandmanna:
yes, complete!

PR #109, opened 2022-10-28
Ozark test

gagnelandmanna:
## Purpose 
- The purpose of this PR is to add the Ozark test to main, an experiment which demonstrates coupling the soil and the plant hydraulics models, and to update it so as to make it consistent with the changes brought to the plant hydraulics model this summer. This will link to issue #108. Co-authored with Katherine.

## To-do
- In a future PR, adjust the parameters, especially LAI, RAI, SAI and K_sat to better match site values in order to obtain a better agreement between water potential measurements at the Ozark site and model output. Explore changing the water retention curve to include a Psi50 parameter.

## Content
- The Ozark test was updated. The flux approximation function was simplified, the parameter space was reduced, the test is now in SI units and runs with the soil model, and diagnostics plots were included. All tasks listed in #108 are complete.

I have:

- appropriate docstrings on all functions, structs, and modules, and included relevent documentation.


src/Soil/Soil.jl
```diff
@@ -86,8 +86,9 @@ export RichardsModel,
     AbstractSoilModel,
     AbstractSoilSource,
     FluxBC,
-    FreeDrainage,
```
kmdeck, line 89: keep export?

src/Vegetation/PlantHydraulics.jl
```diff
@@ -113,18 +113,22 @@ $(DocStringExtensions.FIELDS)
...
-    "van Genuchten parameter"
-    vg_m::FT
-    "porosity"
-    ν::FT
-    "storativity"
-    S_s::FT
+    "water conductivity in the above-ground plant compartments (m/s) when pressure is zero, a maximum"
+    above_ground_plant_K_sat::NamedTuple{(:stem, :leaf), Tuple{FT, FT}}
+    "water conductivity in the roots (m/s) when pressure is zero, a maximum"
+    K_sat_root::Function
```
kmdeck, line 119: Why add this?

src/Vegetation/PlantHydraulics.jl
```diff
@@ -113,18 +113,22 @@ $(DocStringExtensions.FIELDS)
...
+    "van Genuchten parameter (1/m)"
+    plant_vg_α::FT
```
kmdeck, line 121: since these are will within the planthydraulics parameter set, may be ok to just call vg_alpha, vg_n, etc?

src/Vegetation/PlantHydraulics.jl
```diff
@@ -195,41 +199,46 @@ in section 6.4 "Bulk plant hydraulics model".
...
+    S_l1 = inverse_water_retention_curve(plant_vg_α, plant_vg_n, plant_vg_m, ψ1, plant_ν, plant_S_s)
+    S_l2 = inverse_water_retention_curve(plant_vg_α, plant_vg_n, plant_vg_m, ψ2, plant_ν, plant_S_s)
 
-    K1_p1 = hydraulic_conductivity(K_sat1, vg_m, S_l1)
-    K2_p2 = hydraulic_conductivity(K_sat2, vg_m, S_l2)
+    K1_ψ1 = hydraulic_conductivity(plant_K_sat1, plant_vg_m, S_l1)
+    K2_ψ2 = hydraulic_conductivity(plant_K_sat2, plant_vg_m, S_l2)
 
-    flux = -(K1_p1 + K2_p2) / 2 * ((p2 - p1) / (z2 - z1) + 1)
+    flux = -(K1_ψ1 + K2_ψ2) / 2 * ((ψ2 - ψ1) / (z2 - z1))
```
kmdeck, line 219: What about the +1? is Psi = matric potential + z, or just matric potential?

src/Vegetation/PlantHydraulics.jl
```diff
@@ -255,65 +264,65 @@ end
...
-        ν::FT,
-        S_s::FT) where {FT}
-Converts absolute pressure (p) to effective saturation (S_l).
+        plant_vg_α::FT,
+        plant_vg_n::FT,
+        plant_vg_m::FT,
+        ψ::FT,
+        plant_ν::FT,
+        plant_S_s::FT) where {FT}
+Converts absolute pressure (ψ) to effective saturation (S_l).
```
kmdeck, line 311: not pressure anymore

src/Vegetation/PlantHydraulics.jl
```diff
@@ -399,7 +407,7 @@ A concrete type used for dispatch when computing the `flux_out_roots`,
 in the case where the soil pressure at each root layer is prescribed.
```
kmdeck, line 407: "soil matric potential"

src/Vegetation/PlantHydraulics.jl
```diff
@@ -431,28 +439,30 @@ function flux_out_roots(
...
-
-    S_l_stem = effective_saturation(ν, Y.vegetation.ϑ_l[1])
-
-    #How to ungeneralize
-
-    p_stem = water_retention_curve.(vg_α, vg_n, vg_m, S_l_stem, ν, S_s)
-    return sum(
+    @unpack plant_vg_α, plant_vg_n, plant_vg_m, plant_ν, plant_S_s, above_ground_plant_K_sat, K_sat_root, root_distribution = model.parameters
+    S_l_stem = effective_saturation(plant_ν, Y.vegetation.ϑ_l[1])
+    # How to ungeneralize
```
kmdeck, line 444: rm comment

test/Vegetation/plant_hydraulics_test.jl
```diff
@@ -1,7 +1,9 @@
 using Test
```
kmdeck, line 1: Does this file need to change at all?

test/Vegetation/multi_layer_root_test.jl
```diff
@@ -0,0 +1,119 @@
...
+    prob = ODEProblem(ode!, Y, (t0, tf), p);
+    cb = SavingCallback(
+        (u, t, integrator) -> copy(integrator.p),
+        sv;
+        saveat = daily,
+    )
+    sol = solve(prob, RK4(); dt = dt, saveat = daily, callback = cb, adaptive=false) 
+    # That the test is running is the main test.
+    # Check that root depth array is monotonic and increasing (this is necessary for coupled model)
+    @test(((root_depths[i+1]-root_depths[i] for i in 1:length(root_depths)-1) .> FT(0)) == Bool.(ones(length(root_depths)-1)))
```
kmdeck, line 116: I think this `@test` should go in the domains test file:
https://github.com/CliMA/ClimaLSM.jl/blob/0a09e0be67b167b2fc3bd8b1a6b33c33dffe4e1d/test/domains.jl#L205

test/Vegetation/multi_layer_root_test.jl
```diff
@@ -0,0 +1,119 @@
...
+    # Check integral of root_distribution(z)dz approx = 1
+    @test(abs(1 - sum(root_distribution.(root_depths) .* (vcat(root_depths, [0.0])[2:end] - vcat(root_depths, [0.0])[1:(end - 1)]))) < 0.2)
```
kmdeck, line 118: This doesnt seem like a very stringent test. if the discrete integral of root distribution(z) dz is not =1, I wonder if we should normalize it to be equal to 1 in the root_extraction computations...

test/Vegetation/multi_layer_root_test.jl
```diff
@@ -0,0 +1,119 @@
...
+
+    sv = SavedValues(FT, ClimaCore.Fields.FieldVector)
+    daily = Array(2:(3600 * 24):(N_days * 3600 * 24))
+    prob = ODEProblem(ode!, Y, (t0, tf), p);
+    cb = SavingCallback(
+        (u, t, integrator) -> copy(integrator.p),
+        sv;
+        saveat = daily,
+    )
+    sol = solve(prob, RK4(); dt = dt, saveat = daily, callback = cb, adaptive=false)
```
kmdeck, line 113: Im not sure this is a very good test. Maybe instead of solving an integration, we can test the rhs function directly? What do you want to show with this test?

kmdeck: can the plant_hydraulics_test incorporate a larger number of layers and supersede this?

test/Vegetation/multi_layer_root_test.jl
kmdeck: I think it's ok, actually!

src/Soil/Soil.jl
gagnelandmanna: i think it appears twice in the list, see above

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: I thought I had too as K_sat_root was of a different type (function) than the above_ground_plant_K_sat values (FT), but maybe not necessary and can include a function in the Tupple? Somthing like Tupple{FT,FT,Function} ?

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: You mean we can write vg_alpha, vg_n as they are inside the PlantHydraulics.jl file, but in the coupled tests driver files, we keep plant_vg_α and soil_vg_α to differentiate them?

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: This was a mistake - forgot to put it back in after having removed it when thinking about the conversion. Psi is matric potential only

src/Vegetation/PlantHydraulics.jl
```diff
@@ -195,41 +199,46 @@ in section 6.4 "Bulk plant hydraulics model".
...
 
-    K1_p1 = hydraulic_conductivity(K_sat1, vg_m, S_l1)
-    K2_p2 = hydraulic_conductivity(K_sat2, vg_m, S_l2)
+    K1_ψ1 = hydraulic_conductivity(plant_K_sat1, plant_vg_m, S_l1)
+    K2_ψ2 = hydraulic_conductivity(plant_K_sat2, plant_vg_m, S_l2)
 
-    flux = -(K1_p1 + K2_p2) / 2 * ((p2 - p1) / (z2 - z1) + 1)
+    flux = -(K1_ψ1 + K2_ψ2) / 2 * ((ψ2 - ψ1) / (z2 - z1))
+    
+    # Future, for science, add this:
```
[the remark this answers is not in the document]
gagnelandmanna: Ok, but was thinking it would be interesting to keep as a comment and include in future,  because the gravity component is supposed to be much smaller than suction component, thought it would be interesting from a science point of view to include a graph in a tutorial (could be in Ozark) in which we highlight this difference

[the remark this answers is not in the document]
gagnelandmanna: what do you think?

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: will change, thanks

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: Ok, will change as well

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: Ok, what was meant by this?

test/Vegetation/multi_layer_root_test.jl
gagnelandmanna: I will include the larger number of layers in plant_hydraulics_test - was just wanting to test the multiple layers

test/Vegetation/multi_layer_root_test.jl
gagnelandmanna: OK, sounds good! Are we ok with this test then? I will include it in plant_hydraulics_test.jl

test/Vegetation/plant_hydraulics_test.jl
gagnelandmanna: I had brought some changes, but will look through it again as well as bring in changes proposed above.

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: fixed this

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: replaced by this comment, following notes in the clima docs :

Converts absolute pressure, represented by the height (ψ)
of the water column that would give rise to this pressure,
to effective saturation (S_l).

What do you think?

test/Vegetation/multi_layer_root_test.jl
gagnelandmanna: did this, now deleting multi layer root test

test/Vegetation/plant_hydraulics_test.jl
gagnelandmanna: done

test/LSM/process_data.jl
```diff
@@ -0,0 +1,81 @@
+using DelimitedFiles
```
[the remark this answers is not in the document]
gagnelandmanna: Hadn't noticed this file :P

Not sure if I need to update this file? How would I use the artifacts?

Could we create an experiments folder and include it there? I feel like it is more than a tutorial

src/Vegetation/PlantHydraulics.jl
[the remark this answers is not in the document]
kmdeck: I think if we want to look at that, we could just compute it offline/after the simulation is done - no need to have the comment in source code?

src/Vegetation/PlantHydraulics.jl
kmdeck: sure!

src/Vegetation/PlantHydraulics.jl
kmdeck: oh, just no need to merge in "#How to ungeneralize" into main :)

Comments do have a place, but good to clean up little notes to yourself that arent relevant anymore before merging

test/Vegetation/multi_layer_root_test.jl
kmdeck: which test? 

- Im not sure it makes sense to include the integral one unless you test it with a larger number of layers and show that it is within the expected error given the discretization of the integral we use. A 20% error tolerance is not very meaningful IMO - does it "prove" the code is doing the right thing?

- the test about the monotonically increasing root depths should be in the domains test.

test/LSM/process_data.jl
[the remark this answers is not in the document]
kmdeck: We could add it to an experiments directory, but we need to make sure the script is tested - we can do that with buildkite.  I think let's not add the "process data" script, just the experiement script. 

You can use artifacts by following the template here:
https://github.com/CliMA/ClimaLSM.jl/blob/34566968763f35cdf40ed001ce0a8b723d7ac5fe/docs/tutorials/Soil/freezing_front.jl#L218

But the files must be added to caltech box & shared to get a link. Did I add these files to caltech box already? I did add a couple for you. or are there additional ones?


 You could add a buildkite job for it as done here:
https://github.com/CliMA/ClimaLSM.jl/blob/34566968763f35cdf40ed001ce0a8b723d7ac5fe/.buildkite/pipeline.yml#L31

(note that you need to have a Project.toml in experiments also, i.e. julia --project=experiments for experiments/ozark_site.jl. This is fine, you might copy the test/Project.toml but strip out all the packages that are not needed).

src/Vegetation/PlantHydraulics.jl
```diff
@@ -206,30 +209,35 @@ function flux(
...
-    flux = -(K1_p1 + K2_p2) / 2 * ((p2 - p1) / (z2 - z1) + 1)
+    flux = -(K1_ψ1 + K2_ψ2) / 2 * ((ψ2 - ψ1) / (z2 - z1) + 1)
+    
+    # Future, for science, add this:
+    # suction_term = -1/(z2 - z1) * (p2 - p1) * (K1_p1 + K2_p2)/2
+    # gravity_term = -(K1_p1 + K2_p2)/2
+    # flux = suction_term + gravity_term
 
-    return flux # units of [m s-1]
+    return flux # (m/s)
```
kmdeck, line 225: this is an example of a helpful comment to that is OK to keep :) although the doc string should also convey this information.

src/Vegetation/PlantHydraulics.jl
```diff
@@ -431,28 +441,29 @@ function flux_out_roots(
...
             vg_α,
             vg_n,
             vg_m,
             ν,
             S_s,
-            K_sat[:root],
-            K_sat[model.domain.compartment_labels[1]],
-        ),
-    )
+            K_sat[:root].(model.domain.root_depths),
```
kmdeck, line 458: My question about K_sat[:root] was why it is a function now, instead of a scalar? Is this something we actually have data to constrain?

test/domains.jl
```diff
@@ -222,6 +222,9 @@ end
         @test testing.compartment_midpoints == comp_points
         @test testing.compartment_surfaces == top_of_compartments
         @test testing.compartment_labels == comp_labels
+        # Check that root depth array is monotonic and increasing 
+        @test(((testing.root_depths[i+1]-testing.root_depths[i] for i in 1:length(testing.root_depths)-1) .> FT(0)) == Bool.(ones(length(testing.root_depths)-1)))
```
kmdeck, line 226: 👍

src/Vegetation/PlantHydraulics.jl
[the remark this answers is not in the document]
gagnelandmanna: ok, sounds good

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: cool :-)

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: which comment though: units of [m s-1] or (m/s)? I shortened it and made it same format everywhere

test/LSM/process_data.jl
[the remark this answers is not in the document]
gagnelandmanna: Sounds good for adding to experiments directory and testing the script. You had already added these files to Caltech box. :-)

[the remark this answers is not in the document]
gagnelandmanna: Sounds good for buildkite

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: Was changed following a convo we had had about converting between conductance and conductivity, but the change isn't necessary in the end as nothing to constrain with indeed.

gagnelandmanna:
r+

PR #111, opened 2022-10-29
Bc functions

### LINKED ISSUE #112: Boundary conditions can be functions of p, t
We'd like to allow for more general BC in order to test our code against experimental data.

https://github.com/CliMA/ClimaLSM.jl/pull/111
Boundary conditions can be State or flux already, but the flux and state values passed in are constants. Here we allow them to be functions of p,t.

There is not a good alternative.

kmdeck:
## Purpose 
Allow boundary conditions (flux or state) to be functions, rather than constrained to constants.

This allows them to be functions of `p` and `t` (which allows for functions of space and time). If we need more arguments, we can add them (e.g. Y,p,t, or pass in the coordinate directly?).

Closes #112

## To-do
- allow BC to be functions
-  update current tests

## Content
Review checklist

I have:

- appropriate docstrings on all functions, structs, and modules, and included relevent documentation.


src/soil_plant_hydrology_model.jl
```diff
@@ -55,8 +55,8 @@ function SoilPlantHydrologyModel{FT}(;
 }
 
     #These may be passed in, or set, depending on use scenario
-    top_flux_bc = FluxBC(FT(0.0))
-    bot_flux_bc = FluxBC(FT(0.0))
+    top_flux_bc = FluxBC{FT}((p, t) -> eltype(t)(0.0))
```
juliasloan25, line 58: should this be `FT(0.0)` instead of `eltype(t)(0.0)`?

juliasloan25 (commented):
looks good!

src/soil_plant_hydrology_model.jl
kmdeck: Script to test:
```module Tmpm
export FluxBC, f

struct FluxBC{T}
    bc::Function
end

function f(bc::FluxBC{T}, t::T) where {T}
    for i in 1:1000
        bc.bc(t)
    end
end

end
```

```
include("Tmpm.jl")
using .Tmpm

using BenchmarkTools
FT = Float32
bc1 = FluxBC{FT}((t) -> FT(0.0))
bc2 = FluxBC{FT}((t) -> eltype(t)(0))
f(bc1, FT(0.0));
f(bc2, FT(0.0));
@btime f(bc1,FT(0))
@btime f(bc2,FT(0))
```

results:
```
julia> include("./tmp.jl")
  89.417 μs (2001 allocations: 31.27 KiB)
  9.375 μs (1001 allocations: 15.64 KiB)
```

so we should be using "eltype(t)" rather than "FT"!


Im not sure we need the `FT` in the BC structs, then. it's never being used.

kmdeck: I removed `FT` from all the BC structs, tests still pass! we can add it in to specific instances as needed.

kmdeck:
@AlexisRenchon this it the BC function PR - once it merges into main, if you like, you could merge in the latest? and then your diff should be much smaller.

PR #115, opened 2022-11-04
bc in shared utilities

kmdeck:
## Purpose 
Add boundary condition and source utilities to SharedUtilities. Then they are able to be used by other models besides soil.

It would be nice to move as much as possible over, but unless we come up with more general way of handling stateBC (e.g. for models with purely diffusive fluxes), those methods will live in individual modules.

## To-do
discuss with Julia :)

## Content

Review checklist

I have:

### DIFF
--- a/src/ClimaLSM.jl
+++ b/src/ClimaLSM.jl
@@ -11,6 +11,8 @@ import .Parameters as LSMP
 include("SharedUtilities/Domains.jl")
 using .Domains
 include("SharedUtilities/models.jl")
+include("SharedUtilities/boundary_conditions.jl")
+include("SharedUtilities/sources.jl")
 include("Bucket/Bucket.jl")
 
 export domain
@@ -228,7 +230,6 @@ using .Pond
 import .Pond: surface_runoff
 include("Soil/Soil.jl")
 using .Soil
-import .Soil: source!, boundary_flux
 include("Vegetation/PlantHydraulics.jl")
 using .PlantHydraulics
 import .PlantHydraulics: flux_out_roots
--- a/src/SharedUtilities/boundary_conditions.jl
+++ b/src/SharedUtilities/boundary_conditions.jl
@@ -1,0 +1,80 @@
+export boundary_flux,
+    AbstractBC,
+    AbstractBoundary,
+    TopBoundary,
+    BottomBoundary,
+    diffusive_flux,
+    get_Δz
+
+"""
+    AbstractBC{FT <: AbstractFloat}
+An abstract type for types of boundary conditions, which will include
+prescribed functions of space and time as Dirichlet conditions or
+Neumann conditions, in addition to other 
+convenient conditions.
+"""
+abstract type AbstractBC{FT <: AbstractFloat} end
+
+"""
+    AbstractBoundary{}
+An abstract type to indicate which boundary we are doing calculations for.
+Currently, we support the top boundary (TopBoundary)
+and bottom boundary (BottomBoundary).
+"""
+abstract type AbstractBoundary{} end
+
+
+"""
+    TopBoundary{} <: AbstractBoundary{}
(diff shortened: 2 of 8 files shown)


src/SharedUtilities/boundary_conditions.jl
```diff
@@ -0,0 +1,75 @@
...
+"""
+struct BottomBoundary <: AbstractBoundary end
+
+"""
+    get_Δz(z)
+A function to return a tuple containing the distance between the top boundary
+and its closest center, and the bottom boundary and its closest center, 
+both as Fields.
+"""
+function get_Δz(z)
```
juliasloan25, line 47: Looking at this again, it might be helpful for future users if we include the type of `z` in the docstring (or specify that it should be the object returned by `Fields.coordinate_fields.z`).

src/Soil/Soil.jl
```diff
@@ -72,15 +71,18 @@ import ClimaLSM:
...
+    AbstractBC,
+    source!,
+    boundary_flux
 export RichardsModel,
     RichardsParameters,
     EnergyHydrology,
     EnergyHydrologyParameters,
-    boundary_flux,
-    AbstractBC,
     FluxBC,
```
juliasloan25, line 83: Should `FluxBC` and `StateBC` exist in SharedUtilities/boundary_conditions.jl? If they're only used by soil models I think it makes sense to keep them here, but if they can be used more generally it might be good to keep all the boundary conditions objects in one place.

src/Soil/Soil.jl
```diff
@@ -181,88 +167,62 @@ function dss!(
...
-function diffusive_flux(K, x_2, x_1, Δz)
-    return @. -K * (x_2 - x_1) / Δz
+function ClimaLSM.boundary_flux(
+    bc::FluxBC{FT},
+    boundary::ClimaLSM.AbstractBoundary,
+    Δz,
+    p,
+    _...,
+) where {FT}
+    return bc.bc .+ ClimaCore.Fields.zeros(axes(Δz))
```
juliasloan25, line 210: Maybe add a comment explaining why the boundary condition value is added to this Field object since it may not be obvious from just reading the code

src/Soil/Soil.jl
kmdeck: I think I am on the fence with this! It would great to have them stored and used generally, but each model will have a different method for how to turn a state into a flux. so we cant have a single StateBC type for all models.
but maybe we can somehow get around this because these will all be of the form: factor_at_center*(BC - state)/dz?

src/ClimaLSM.jl
```diff
@@ -11,6 +11,8 @@ import .Parameters as LSMP
 include("SharedUtilities/Domains.jl")
```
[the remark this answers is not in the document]
kmdeck: sure! I can try and do this

src/Soil/Soil.jl
kmdeck: holding off on this: will address longer term https://github.com/CliMA/ClimaLSM.jl/issues/116

src/ClimaLSM.jl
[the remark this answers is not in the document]
kmdeck: I think I might wait on this...it will require changing a lot of source and test files (that extend e.g. ClimaLSM.function where function is defined in models.jl, or call them). Conceptually it makes sense to me that these would be modules (except maybe source, since it is so short). But I think there was some question of if in the long term, shared utilities would live in a separate repo. This would enable individual components to be developed in their own repos, without all the LSM tests, as long as they used/included the shared utilities repo. and then ClimaLSM would import the shared utilities (LSMs are also models) but also the individual components, and assemble them. Im not sure if this will happen, but im sort of inclined to wait on this until we're sure what the best course is?

PR #124, opened 2022-12-01
update beta factor for evaporation

### LINKED ISSUE #123: Beta factor in evaporation
We want to enable a scaling factor beta of the evaporation, rather than scaling q_sfc.

Please see overleaf: https://www.overleaf.com/read/gwnnjsxqtrsz

Scaling q_sfc leads to unphysical results. Scaling evaporation is the standard bucket model approach.

kmdeck:
## Purpose 
Update how we handle evaporation when the bucket is not saturated: E = beta*E_0(q_sat)

The current code relies on the snow cover fraction being either 0 or 1. Snow and soil covered surfaces have the same surface temperature and roughness lengths, in the model, but the specific humidity at the surface changes depending on the surface coverage. If the snow cover fraction is 0 or 1, we can store a single value q_sfc in `p`, which is computed as the saturation specific humidity over ice or liquid water depending on if SCF is 0 or 1.

Similarly, we have an effective beta = (1-sigma)*beta + sigma*1.

Evaporation is the computed with beta_eff and the single value of q_sfc.

If the snow cover fraction was not 1, we would need to separately store the surface q over both snow and soil, and compute two sets of fluxes, using different values of beta.

Closes #123 

## Review checklist

I have:

----


src/Bucket/Bucket.jl
```diff
@@ -587,21 +592,14 @@ function make_update_aux(model::BucketModel{FT}) where {FT}
...
-                Ref(model.parameters),
-            )
         # This relies on the surface specific humidity being computed
         # entirely over snow or over soil. i.e. the snow cover fraction must be a heaviside
         # here, otherwise we would need two values of q_sfc!
-        p.bucket.q_sfc = @. (
-            (FT(1.0) - snow_cover_fraction) *
-            β(Y.bucket.W, model.parameters.W_f) *
-            q_sat + snow_cover_fraction * q_sat
+        @. p.bucket.q_sfc = saturation_specific_humidity(
```
kmdeck, line 598: just FYI this function "saturation_specific_humidity" returns the value over ice if snow covered and over liq if not.

szy21 (approved):
Thanks Kat! Should we add beta to p.bucket too? Or do you plan to do that in a separate PR?

kmdeck:
> Thanks Kat! Should we add beta to p.bucket too? Or do you plan to do that in a separate PR?

I was thinking of just adding it to the cache in the coupled runs (we do that for a couple of variables). but perhaps it is just easier to do it here! ill look at the coupler code again and let you and Lenka know.

kmdeck:
@LenkaNovak what do you think about computing `beta` in the coupler like it compute albedo, vs having land compute it and store it in `p.bucket`?

E.g. here the coupler computes albedo
https://github.com/CliMA/ClimaCoupler.jl/blob/a80650a7bb16313476e1f9d7fdb3712859eb8066/experiments/AMIP/moist_mpi_earth/push_pull.jl#L141

using 
https://github.com/CliMA/ClimaCoupler.jl/blob/a80650a7bb16313476e1f9d7fdb3712859eb8066/experiments/AMIP/moist_mpi_earth/bucket/bucket_utils.jl#L53

(and `surface_albedo` is a function that is imported from ClimaLSM). We could do the same thing, but calling `beta` instead.

Alternatively, we can store beta in `p.bucket.beta` and have land update the value at each timestep, and the coupler would extract it from there. 

I think in both cases a surface field would be defined in the coupler for beta.

(longer term, when we bring in vegetation, etc, we can have another discussion about where to store things like roughness lengths, resistances, albedos...)

LenkaNovak:
Sure! Presumably, you won't need beta in the stand-alone runs? I think this would make debugging easier too!

kmdeck:
> Sure! Presumably, you won't need beta in the stand-alone runs? I think this would make debugging easier too!

ok! that sounds good to me. we can work on the coupler PR together once this one goes in, then :)

LenkaNovak:
Ok, perfect! Thank you!

kmdeck:
@szy21 and @LenkaNovak 

quick check before I merge :) (no rush - can merge in the new year)

we now have (q, beta) = (q_sat over ice, 1) for snow covered surfaces and (q, beta) = (q_sat over liquid, beta(W)) for bare soil.

so, if q_sat < q_atmos, we still can have "condensation". Not sure if we want to keep something in the coupler that prevents this. I dont see any discussion of treating this differently in SLIM, or the other land model I was following.

szy21:
> @szy21 and @LenkaNovak
> 
> quick check before I merge :) (no rush - can merge in the new year)
> 
> we now have (q, beta) = (q_sat over ice, 1) for snow covered surfaces and (q, beta) = (q_sat over liquid, beta(W)) for bare soil.
> 
> so, if q_sat < q_atmos, we still can have "condensation". Not sure if we want to keep something in the coupler that prevents this. I dont see any discussion of treating this differently in SLIM, or the other land model I was following.

Yep I realized that could happen. I think it's probably fine because that would only happen if T_sfc is much smaller than T_atmos, e.g. at night or in high latitudes. We can come back to it if there is a problem.

PR #125, opened 2022-12-07
Albedo from file

kmdeck:
## For reviewers
This is a big PR - the major files changed are src/Bucket/Bucket.jl, src/Regridder.jl, and src/SharedUtilities/models.jl.

## Purpose 
This PR adds in the capability to read in a NetCDF file of surface albedo from a file, store in the the auxiliary state. Currently this only supports fields that do not vary in time; future PRs will address updating the albedo at each step/ at each X number of steps, and will ensure that the correct entry is being read in and used for the first timestep, based on t0.

This has been tested with the coupler in both MPI and non-MPI runs.

## Content
- Added Regridder utils from ClimaCoupler. Longer term, perhaps we should have one copy of these tools, living in a separate repo (ClimaCoreTempestRemap?), to avoid duplication.
- Created artifacts.jl file in `src/Bucket/artifacts` which can stored the needed artifacts for the global bucket simulation.
- Created a new struct "MapInfo", which stores needed information for reading in netcdf files, regridding and making a field of them. Any parameter which is read from a file will longer term be passed in as type ::MapInfo, rather than e.g. `FT` or a `Function`.
- Created an albedo model type which uses `MapInfo` rather than a prescribed `f::Function` for albedo = f(lat,lon). The two options are now `BulkAlbedoFunction` (where soil albedo is a function) and `BulkAlbedoMap`, where soil albedo is of type `MapInfo`.
- Added soil albedo to auxiliary variables. For both albedo models, we compute the albedo once, up front (see next item), and store in the auxiliary state. If the albedo needs to be updated in time, the `update_aux` function is ready to be used for that.
- Created a `set_initial_aux_state!` function which inits the aux variables (based on initial conditions for the prognostic state) as well as inits any parameters read from a dataset that are also stored in aux
- Added partial testing of new capabilities. The Regridder utils are already tested in ClimaCoupler. Longer term, we should import these functions rather than redefine (and retest) here.
- Updated all bucket tests and tutorials to use the new albedo types.

Review checklist

I have:

----


src/Regridder.jl
```diff
@@ -0,0 +1,388 @@
+# Code from ClimaCoupler Regridder.jl
```
LenkaNovak, line 1: Given that a lot of this is copied from ClimaCoupler, it would be preferable to move this to ClimaCore to avoid duplication. @simonbyrne, what do you think?

docs/tutorials/Bucket/bucket_tutorial.jl
```diff
@@ -157,7 +158,7 @@ earth_param_set = create_lsm_parameters(FT);
 # will support other albedo models.
 α_soil = (coordinate_point) -> FT(0.2);
 α_snow = FT(0.8);
-albedo = BulkAlbedo{FT}(α_snow, α_soil);
+albedo = BulkAlbedoFunction{FT}(α_snow, α_soil);
```
braghiere, line 161: α_soil should be α_surface since this is a combination of soil+vegetation.

braghiere (commented):
Just naming a variable differently to be precise (surface instead of soil), and clarifying the usage of snow vs. soil.

src/Bucket/Bucket.jl
```diff
@@ -238,6 +259,11 @@ function BucketModel(;
     atmosphere::ATM,
     radiation::RAD,
 ) where {FT, PSE, ATM, RAD}
+    if typeof(parameters.albedo) == BulkAlbedoMap{FT}
```
juliasloan25, line 262: This seems like it could maybe be done using multiple dispatch instead, though I'm not sure how we could access `parameters.albedo` in the function signature itself.

src/Bucket/Bucket.jl
```diff
@@ -59,19 +64,35 @@ abstract type AbstractRadiativeDrivers{FT <: AbstractFloat} end
...
+location and across wavelength. Soil albedo is specified via a
+NetCDF file, which can be a function of time, but is treated
+as constant across wavelengths.
+
+Note that this option should only be used with global simulations,
+i.e. with a `ClimaLSM.LSMSphericalShellDomain.`
+"""
+struct BulkAlbedoMap{FT} <: AbstractLandAlbedoModel{FT}
+    α_snow::FT
+    α_soil::MapInfo
```
juliasloan25, line 94: Change to `\alpha_surface` per Renato's comment about the file albedo data.

src/Bucket/Bucket.jl
```diff
@@ -289,9 +315,89 @@ function ClimaLSM.initialize(model::BucketModel{FT}) where {FT}
...
+the surface space of the LSM using ClimaCoreTempestRemap. The resul
+is a ClimaCore.Fields.Field of albedo values.
+"""
+function set_initial_parameter_field!(
+    albedo::BulkAlbedoMap{FT},
+    p,
+    surface_coords,
+) where {FT}
+    α_soil = albedo.α_soil
+    comms_ctx = α_soil.comms
```
juliasloan25, line 386: could unpack these with `(; ...)` syntax to condense this code

src/Regridder.jl
```diff
@@ -0,0 +1,388 @@
...
+        1:length(times),
+    )
+    jldsave(
+        joinpath(REGRID_DIR, hd_outfile_root * "_times.jld2");
+        times = times,
+    )
+end
+
+
+function hdread_regridfile(
```
juliasloan25, line 304: this function is a duplicate of `read_from_hdf5`. `read_from_hdf5` just has a name update to be consistent with `write_to_hdf5`, and adds a default value for `comms_ctx`. It's the more updated version that will be used by the coupler going forward.

src/Regridder.jl
```diff
@@ -0,0 +1,388 @@
...
+            varname,
+            boundary_space;
+            hd_outfile_root = outfile_root,
+            mono = mono,
+        )
+    end
+    ClimaComms.barrier(comms_ctx)
+    file_dates =
+        load(joinpath(REGRID_DIR, outfile_root * "_times.jld2"), "times")
+    field = hdread_regridfile(
```
juliasloan25, line 350: change to `read_from_hdf5` (see above comment)

test/Bucket/albedo_map_bucket.jl
```diff
@@ -0,0 +1,142 @@
...
+        LSMSphericalShellDomain(;
+            radius = FT(100.0),
+            height = FT(3.5),
+            nelements = (2, 10),
+            npolynomial = 2,
+        ),
+    ]
+
+
+    for i in 1:2
```
juliasloan25, line 60: could use `for bucket_domain in bucket_domains:` to improve readability

juliasloan25 (approved):

Project.toml
```diff
@@ -4,10 +4,16 @@ authors = ["Clima Land Team"]
 version = "0.2.1"
 
 [deps]
+ArtifactWrappers = "a14bc488-3040-4b00-9dc1-f6467924858a"
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
+ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
+ClimaCoreTempestRemap = "d934ef94-cdd4-4710-83d6-720549644b70"
```
Yujie-W, line 11: I do not see corresponding items in the `[compat]` session

Yujie-W (changes requested):
1. Add corresponding `[compat]` items in the `Project.toml` file. Otherwise, the registration will be redflagged in the future.
2. I am curious why not using Artifacts.jl and lazy artifacts file? Then you can simply use `artifact"ART_NAME"` in the code? [No changes required]
3. I see there are some lines such as FT(1.0) and FT(0.0). If 1 and 0 are directly used along with FT, such as `1 - ft_number`, you do not need to use `FT(1) - ft_number`. Does this result in any difference in speed? [No changes required]

src/Bucket/Bucket.jl
charleskawczynski: We can use UnionSplitting
```suggestion
    if parameters.albedo isa BulkAlbedoMap
```
which is equivalent to multiple dispatch for small (i.e., a few) unions, ref: https://julialang.org/blog/2018/08/union-splitting/

kmdeck: nice!

src/Regridder.jl
kmdeck: good catch!

kmdeck:
> 1. Add corresponding `[compat]` items in the `Project.toml` file. Otherwise, the registration will be redflagged in the future.
> 
>     3. I see there are some lines such as FT(1.0) and FT(0.0). If 1 and 0 are directly used along with FT, such as `1 - ft_number`, you do not need to use `FT(1) - ft_number`. Does this result in any difference in speed? [No changes required]

good catch Yujie! I fixed the toml and at least some of the unnecessary FT() calls. will keep an eye out for those.

kmdeck:
> 2\. I am curious why not using Artifacts.jl and lazy artifacts file? Then you can simply use `artifact"ART_NAME"` in the code? [No changes required]

With respect to this - Im not sure. ArtifactWrappers uses Artifacts.jl under the hood, with an option for a lazy download (https://clima.github.io/ArtifactWrappers.jl/dev/). I'm not sure we need the extra features the wrapper provides, but I think it might amount to the same core capabilities in the end. What do you think?

test/Bucket/albedo_map_bucket.jl
```diff
@@ -0,0 +1,139 @@
...
+import ClimaLSM
+include(joinpath(pkgdir(ClimaLSM), "parameters", "create_parameters.jl"))
+
+@testset "Initialize aux state with map" begin
+    FT = Float32
+    earth_param_set = create_lsm_parameters(FT)
+
+
+    α_snow = FT(0.8) # snow albedo
+    path = cesm2_land_albedo_dataset_path()
```
juliasloan25, line 33: It might be nice to have an additional `MapInfo` constructor which has default values for all the arguments, so a basic map can be created without having to know about the internals. The coupler has to create an albedo `MapInfo` struct and I think it would be better if it didn't have to use e.g. `cesm2_land_albedo_dataset_path()`. This could be a separate small PR, or maybe part of this one.

kmdeck: ok! I think the default constructor would need to call cesm2_land_albedo_dataset_path() - to get the local path of the file - but that's fine! Ill add this in today.

kmdeck: ok! this is fixed. you can now call it with albedo_model = BulkAlbedoMap{FT}(regrid_dirpath). It also has a default value for alpha snow, and comms context, so you'll have to pass in the comms context from the coupler.
https://github.com/CliMA/ClimaLSM.jl/blob/6d4e3b11a7b407011df0d76570f5c10865f57f98/test/Bucket/albedo_map_bucket.jl#L35
https://github.com/CliMA/ClimaLSM.jl/blob/6d4e3b11a7b407011df0d76570f5c10865f57f98/src/Bucket/Bucket.jl#L100

kmdeck:
@Yujie-W I think you need to approve the review before I can merge (given that you requested changes). I left some comments back. Otherwise I think I can dismiss the review. let me know. thank you!

src/Regridder.jl
kmdeck: @simonbyrne @juliasloan25 perhaps we can merge this PR, and then a later PR would move the duplicate code from the coupler and climalsm into climacore?

Yujie-W (approved):

PR #137, opened 2023-01-18
update ClimaCore, SurfaceFluxes compats

juliasloan25:
## Purpose 
Minor updates to ClimaCore, and SurfaceFluxes versions to ensure compatibility with ClimaCoupler before ClimaLSM release.

----


test/Project.toml
```diff
@@ -2,6 +2,7 @@
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
+ClimaLSM = "7884a58f-fab6-4fd0-82bb-ecfedb2d8430"
```
kmdeck, line 5: We actually do not have ClimaLSM as a project dependency for `test`. Instead, we add the path to ClimaLSM to the path (initial lines in all test files).

It may be that it is better to add it as a dep and use `dev ../`.

What do you think?

Project.toml
```diff
@@ -23,14 +23,14 @@ UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 ArtifactWrappers = "0.2"
 CLIMAParameters = "0.7.0"
 ClimaComms = "0.3"
-ClimaCore = "0.10.18"
+ClimaCore = "0.10.24"
```
kmdeck, line 26: Would you also update the deps in experiment/ and docs/? thank you!

test/Project.toml
juliasloan25: Oops I didn't realize - for some reason I was having trouble precompiling without ClimaLSM as a dependency before, but now it works fine without it.

It might be cleaner/clearer to have it as a dependency rather than adding repeated lines to the top of each test file, but maybe we can address that in a later PR.

kmdeck: sure! That can be a later PR. I have a general "clean up CI" issue that Ill add it to.

Issue #135, opened 2023-01-11
O2.3.6 Implicit solver for full soil model


kmdeck:
## Purpose
We need an implicit solver for the soil model (both for Richards equation alone, and the full soil system), and possibly in the future for other models as well. Currently, the implicit solver for Richards equation is implemented and in use. The full soil implicit solver is still in progress.

## Components
- A way to specify separate explicit and implicit tendencies
  - Richards equation: https://github.com/CliMA/ClimaLand.jl/pull/191
- A way to set tolerances and other convergence requirements (max iters, etc)
  - Included by using ClimaTimeSteppers: https://github.com/CliMA/ClimaLand.jl/pull/218/files
- A way to specify Jacobians, and how often to update them; Function to compute Richards model Jacobian (derivation [here](https://www.overleaf.com/project/63be02f455f84a77642ef485))
  - Richards equation: https://github.com/CliMA/ClimaLand.jl/pull/190

## Inputs
- For Richards equation: https://www.overleaf.com/9252289794ykktrwjxqvcj. We will be implementing modified Picard iterations.

## Results and deliverables/Validation
- [x] For Richards equation: plots of mass conservation and solution error vs timestep for modified Picard iterations, for constant flux and state BC test runs. Comparison to an explicit solver of the same order - obtain max dt the explicit solver runs at. - in our buildkite
- [x] Richards equation: unit test of Jacobian computed by the code with the analytic expression
- [x] Make sure the above validation results are preserved with new CTS interface
-  [x] For full soil model : plots of mass conservation and solution error vs timestep for modified Picard iterations, for constant flux, moisture state BC, and for AtmosDrivenBC. Comparison to an explicit solver of the same order - obtain max dt the explicit solver runs at.
- [x] Unit tests of jacobian
- [x] flexible implementation that allows for LSM models - with many component models - to also have Jacobians specified. Start with soil/canopy model

## Initial task breakdown
A preliminary list of PRs and a preliminary timeline of PRs, milestones, and key results.
- [X] Have reviewers review equations (Tapio)
- [X] Add method computing Jacobian for approximate Newton (modified Picard). Make sure `A` and `b` update every X timesteps, where X is chosen at the top-most level.
- [x]  Simulations with the implicit solver using ClimaTimesteppers. comparison of approx Newton to explicit runs
- [X] State varying boundary conditions treated implicitly
- [x] Implement in general fashion #190 #191 #218
- [x] update to new ClimaTimesteppers interface #528
[...]

PR #141, opened 2023-01-23
Implicit solver for soil moisture (WIP)

juliasloan25:
## Purpose 
This branch is mainly used for experimentation with the implicit solver setup for the soil Richards equation.

Set up the infrastructure necessary to run a basic implicit solver for the Richards equation in the soil model. For now, this includes storing and computing the Jacobian, setting up implicit and explicit tendencies, and solving the ODE problem.

Initial step of SDI #135

This PR includes infrastructure that allows us to use true Picard or modified Picard methods. It uses [Newton's Method](https://clima.github.io/ClimaTimeSteppers.jl/dev/api/newtons_method/) and can use any [ClimaTimeSteppers IMEX algorithm](https://clima.github.io/ClimaTimeSteppers.jl/dev/api/ode_solvers/).

We are currently testing the improvement gained by using these IMEX algorithms as compared to the previous explicit solver implementation. Ideally, for the most performance improvement, we will be able to take much larger timesteps and require very few Newton iterations (i.e. 1 or 2).

Currently, we are able to use timesteps up to ~70s with the explicit solver, up to ~170s using true Picard, and up to ~70s when using modified Picard before the solution becomes unstable. In Bonan's solution using the Crank-Nicolson method, he's able to use a timestep of at least 1800s before the solution becomes unstable. Because we expect to be able to take a much larger timestep with the implicit solver, we believe the modified Picard implementation may have a bug, and we need to debug it.

Steps for this debugging include:
* check our Jacobian calculation
* try using CTS's [Jacobian-free Newton-Krylov method](https://clima.github.io/ClimaTimeSteppers.jl/dev/api/newtons_method/#Jacobian-free-Newton-Krylov-Method) and see if that provides a more stable solution.

Additional minor components that are not necessary for the implementation but should be improved are:
- [x] use `effective_saturation` when calculating `S` in `dψdθ`. This may reduce some instabilities that get introduced when `θ` is slightly smaller than `θ_r`
* add parameters and boundary conditions to `TridiagonalW` struct to eliminate use of global variables
* create field of ones on faces once and store in W to reduce allocations

## Content
- [x] Data structure to store Tridiagonal W matrix for modified Picard
- [x] Data structure to store Identity W matrix for true Picard (and methods to handle this input)
- [x] Function to compute Jacobian
- [x] Thomas algorithm to solve `Ax=b` where `A` is tridiagonal
- [x] Function to setup and use implicit solver

## Testing
- [x] Run explicit, implicit timesteppers with same parameters. Compare for correctness to Bonan's setup detailed in #158 
- [x] Generate theta* by running explicit timestepper with small timestep
- [x] For both explicit and implicit timesteppers: run multiple times while increasing timestep. Plot timestep vs. error and choose reasonable threshold. For error, use some metric like mean or max difference to theta*.
- [x] For explicit, implicit timesteppers: run multiple times while increasing timestep. Plot timestep vs. time to solution while remaining under previously-determined error threshold.

## Results
The following plot was generated from simulatiuons using van Genuchten parameters with a constant flux of -1e-7 m/s for the top boundary condition and free drainage at the bottom. The implicit solver was restricted to use one Newton iteration per timestep. RMSE was calculated with respect to an explicit solver using the same setup and a timestep of 1s, and mass conservation error is the relative error of the mass of water over the entire simulation.
implicit solution

We see similar convergence behavior when running the solver with a convergence condition of maximum error 1e-8 and a maximum of 10 Newton iterations.
implicit solution with convergence condition

We can compare this visually to the explicit solution, which has smaller RMSE but begins to fail at a timestep of around 50s. With the implicit solver, we're able to take much larger timesteps (in the 1000s) before we encounter instability.
explicit solution

We can also test the performance of the implicit solver with a state BC. In this case, we are also using a van Genuchten retention curve, and the soil type is Yolo light clay. The BC is porosity -1e-3. We used a tolerance of 1e-8 and 50 max iterations, with backward Euler. This corresponds to modified Picard iterations.
 state bc

RMSE of the solution at the end scales like dt, which is expected for Backward Euler, and the volume conservation error scales like 1/dt, or like the number of timesteps = total_time/dt. This seems like it is consistent with making a certain amount of roundoff error per step.

----


kmdeck:
These plots are now part of buildkite runs (water_conservation.jl) via PR #227

PR #148, opened 2023-02-08
Refactor climate drivers

kmdeck:
## Background 

In the current version of the repo, the Bucket model makes use of a `PrescribedAtmos` and `PrescribedRadiation` struct for storing the prescribed "driver" functions for running the bucket model in standalone mode. These contain the downwards LW and SW radiation, as functions of time, as well as precipitation, q_atmos, T_atmos, u_atmos, rho_atmos, and the height at which these quantities are measured.

Given this, the bucket model has everything it needs to compute turbulent surface fluxes and a net radiation (as the bucket model provides its own roughness lengths, albedo, and emissivity). These are computed by the functions `surface_fluxes` and `net_radiation`, which are called in `update_aux!` to update the stored values of p.bucket.turbulent_surface_fluxes, p.bucket.evaporation and p.bucket.R_n, which get used in the rhs function as boundary conditions. Note that `surface_fluxes` is a wrapper function - it broadcasts `surface_fluxes_at_a_point` over the input fields. Currently the same is true for `net_radiation`.
E.g: https://github.com/CliMA/ClimaLSM.jl/blob/ff1212db13ad5aff7eff69d7cee6fbebcbfb2f1a/src/Bucket/Bucket.jl#L235, 
https://github.com/CliMA/ClimaLSM.jl/blob/ff1212db13ad5aff7eff69d7cee6fbebcbfb2f1a/src/Bucket/Bucket.jl#L552,
E.g. https://github.com/CliMA/ClimaLSM.jl/blob/ff1212db13ad5aff7eff69d7cee6fbebcbfb2f1a/src/Bucket/Bucket.jl#L531,
https://github.com/CliMA/ClimaLSM.jl/blob/ff1212db13ad5aff7eff69d7cee6fbebcbfb2f1a/src/Bucket/Bucket.jl#L731

If the first argument of the `net_radiation` and `surface_fluxes` is of type `Prescribed`, the land model computes its own fluxes. If the first argument is of type `Coupled`, the method will return the fields in `p` directly under the assumption that the coupler has already filled them in. For example, here is the code where the coupler extends the ClimaLSM method and adds a new `Coupled` type: https://github.com/CliMA/ClimaCoupler.jl/blob/4b17a28ad2e225d031990a5b43c1baa240b8d650/experiments/AMIP/moist_mpi_earth/bucket/bucket_init.jl#L41
https://github.com/CliMA/ClimaCoupler.jl/blob/4b17a28ad2e225d031990a5b43c1baa240b8d650/experiments/AMIP/moist_mpi_earth/bucket/bucket_init.jl#L82

## Purpose 
**Main file to review is src/SharedUtilties/Drivers.jl**

This PR moves all of this machinery from ClimaLSM.Bucket into a module stored in src/SharedUtilities/Drivers.jl, in the anticipation that the Snow model will want to run in standalone mode, as will the soil model, etc, and we dont want to redefine all of these surface flux computations etc for each model. It also anticipates that each model will have stored "p.X.T_sfc, p.X.q_sfc, p.X.\rho_sfc, p.X.alpha, p.X.epsilon", where X is the name of the model. This seems generic enough - a model has to specify these to compute surface fluxes, though it is overkill to store a field of floats if e.g. emissivity is 1 everywhere. But in the long run, for a varied land surface, this makes sense.

We have changed the ordering of the arguments so that all of these methods get `(atmos, p, t)` or `(radiation, p, t)` first. This is a breaking change for the coupled runs. Those methods need to be redefined to take arguments `(atmos, p, t, _...)` and `(radiation,p,t,_...)`.

## Questions 
Should we move the `coupled` methods and struct definitions to this Drivers.jl module? 
After chatting with @LenkaNovak we decided it made sense and worked well to keep them in the coupler repo.

## Review checklist

I have:

----

### DIFF
--- a/src/SharedUtilities/Drivers.jl
+++ b/src/SharedUtilities/Drivers.jl
@@ -1,0 +1,323 @@
+module Drivers
+using Thermodynamics
+using ClimaCore
+using DocStringExtensions
+using UnPack
+using SurfaceFluxes
+using StaticArrays
+import ..Parameters as LSMP
+export AbstractAtmosphericDrivers,
+    AbstractRadiativeDrivers,
+    PrescribedAtmosphere,
+    PrescribedRadiativeFluxes,
+    compute_ρ_sfc,
+    construct_atmos_ts,
+    surface_fluxes_at_a_point,
+    radiative_fluxes_at_a_point,
+    liquid_precipitation,
+    snow_precipitation,
+    surface_air_density
+
+"""
+     AbstractAtmosphericDrivers{FT <: AbstractFloat}
+
+An abstract type of atmospheric drivers of the bucket model.
+"""
+abstract type AbstractAtmosphericDrivers{FT <: AbstractFloat} end
+
+"""
+     AbstractRadiativeDrivers{FT <: AbstractFloat}
+
+An abstract type of radiative drivers of the bucket model.
+"""
+abstract type AbstractRadiativeDrivers{FT <: AbstractFloat} end
+
+"""
+    PrescribedAtmosphere{FT, LP, SP, TA, UA, QA, RA} <: AbstractAtmosphericDrivers{FT}
+
+Container for holding prescribed atmospheric drivers and other
+information needed for computing turbulent surface fluxes when
+driving the bucket model in standalone mode.
+$(DocStringExtensions.FIELDS)
+"""
+struct PrescribedAtmosphere{FT, LP, SP, TA, UA, QA, RA} <:
+       AbstractAtmosphericDrivers{FT}
+    "Precipitation (m/s) function of time: positive by definition"
+    liquid_precip::LP
+    "Snow precipitation (m/s) function of time: positive by definition"
+    snow_precip::SP
+    "Prescribed atmospheric temperature (function of time)  at the reference height (K)"
+    T::TA
+    "Prescribed wind speed (function of time)  at the reference height (m/s)"
+    u::UA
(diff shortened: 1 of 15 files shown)


kmdeck:
@LenkaNovak if it would be easier to discuss in person once you have a look at Drivers.jl, lmk!

src/SharedUtilities/Drivers.jl
```diff
@@ -0,0 +1,333 @@
...
+$(DocStringExtensions.FIELDS)
+"""
+struct PrescribedAtmosphere{FT, LP, SP, TA, UA, QA, RA} <:
+       AbstractAtmosphericDrivers{FT}
+    "Precipitation (m/s) function of time: positive by definition"
+    liquid_precip::LP
+    "Snow precipitation (m/s) function of time: positive by definition"
+    snow_precip::SP
+    "Prescribed atmospheric temperature (function of time)  at the reference height (K)"
+    T_atmos::TA
```
LenkaNovak, line 50: does this need the _atmos subscript?

LenkaNovak:
Hi @kmdeck , sounds good, we can chat tomorrow. The structure looks good, though we might want to add some hooks to turn off the flux calculation.

src/SharedUtilities/Drivers.jl
kmdeck: I guess not! It's kind of repetitive.

PR #154, opened 2023-02-22
Use Ntuple for Plant Hydraulics variable types

kmdeck:
## Purpose 
The plant hydraulics model requires an array of numbers at each coordinate point. Previously, we had kind of hacked this by making the "coordinates" of a PlantHydraulicsDomain an array, and then at each point, we had a scalar variable. But in the long run, we want the coordinates to be the lat/lon of the surface, and we want to have an array-like variable at each point. We also then do not need to introduce a special Domain for plant hydraulics, and can instead use the same domain as the other model components (the surface of the Earth). 

In this PR, we remove the PlantHydraulicsDomain. We make the prognostic and aux variables for the plant hydraulics models Ntuples, instead of scalars. This required changing some of the RHS code in the src/Vegetation/PlantHydraulics.jl file. All other changed files are propagating this into docs, unit tests, and experiments.

## Content
1. Removes the plant hydraulics domain.
2. Sticks the necessary info from that Plant Hydraulics domain struct into the model struct
3. Changes the type of the prognostic variable for PlantHydraulics so that it works with the existing Plane and Point domains (closely linked to ClimaCore domains)
4. Cleans up some repetitive computation in the RHS
5. Propagates changes to unit tests and docs.

Review checklist

I have:

----

### DIFF
--- a/src/Vegetation/PlantHydraulics.jl
+++ b/src/Vegetation/PlantHydraulics.jl
@@ -47,7 +47,7 @@ using UnPack
 using DocStringExtensions
 import ClimaCore: Fields
 
-using ClimaLSM.Domains: AbstractVegetationDomain, PlantHydraulicsDomain
+using ClimaLSM.Domains: AbstractDomain
 import ClimaLSM:
     AbstractModel,
     initialize_prognostic,
@@ -71,10 +71,12 @@ export PlantHydraulicsModel,
     PlantHydraulicsParameters,
     PrescribedSoilPressure,
     PrescribedTranspiration,
-    AbstractRootExtraction
+    AbstractRootExtraction,
+    Layers
 
 """
     AbstractVegetationModel{FT} <: AbstractModel{FT}
+
 An abstract type for vegetation models.
 Concrete types include a plant hydraulics model, but future types will
 include multi-layer canopy models and possibly a big leaf model.
@@ -86,6 +88,7 @@ ClimaLSM.domain(::AbstractVegetationModel) = :surface
 
 """
     AbstractRootExtraction{FT <: AbstractFloat}
+
 An abstract type for types representing different models of
 water exchange between soil and plants.
 Currently, prescribed soil matric potential and prescribed flux models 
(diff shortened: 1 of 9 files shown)


src/Vegetation/PlantHydraulics.jl
```diff
@@ -326,72 +423,83 @@ end
...
+            vg_m,
+            effective_saturation(ν, ϑ_l[1]),
+            ν,
+            S_s,
+        )
+
+        @inbounds for i in 1:(n_stem + n_leaf)
+            AIdz =
+                area_index[model.compartment_labels[i]] * (
+                    model.compartment_surfaces[i + 1] -
```
kmdeck, line 466: would it be more natural to store the height of the compartments?

gagnelandmanna (commented):
This looks great! :)

src/Vegetation/PlantHydraulics.jl
kmdeck: I think we for sure will need canopy height as as parameter. I was also thinking that because we only ever need use the compartment surfaces to get the width of each layer, that maybe it would make sense to just pass the width of the layer (my mistake, I wasnt clear by referring to this as the height).

We can do this in a future PR if we want.

src/Vegetation/PlantHydraulics.jl
```diff
@@ -435,29 +566,36 @@ function flux_out_roots(
...
-            K_sat[:root],
-            K_sat[:stem],
-        ) .* root_distribution.(model.domain.root_depths) .* (
-            vcat(model.domain.root_depths, [0.0])[2:end] -
-            vcat(model.domain.root_depths, [0.0])[1:(end - 1)]
-        ),
-    ) # m/s
+    ψ_stem = p.vegetation.ψ[1]
+    n_root_layers = length(model.root_depths)
+    sum_flux_out_roots = similar(ψ_stem)
```
kmdeck, line 574: We are allocating here, which isnt good, but I think we will come back and address with when we add in the root compartments as prognostic variables.

PR #156, opened 2023-02-22
Compute boundary fluxes at each boundary, rather than by component

kmdeck:
## Purpose 
We are currently computing the boundary fluxes per side and per component (water, heat). This will require duplicate calls to surface fluxes when we drive the soil model with reanalysis data (atmospheric conditions). This is because surface fluxes computes the energy and water fluxes at the same time.

## Content
1. Rearrange BC so they are passed in as `(top = (water = ..., heat = ...), bottom = (water = ..., heat = ....))` instead of `(water = (top = ...., bottom = ....), heat = (top = ...., bottom = ....))`. 
2. Propagate this to Biogeochemistry for consistency, although I dont think it makes a difference at this point.
3. Update tests and tutorials.
4. I also clarified the use of StateBC between water and heat, by introducing `MoistureStateBC` and `TemperatureStateBC`, because before our dispatch relied on # of arguments, and this seemed less reliable.
5. Introduce wrapper function for the soil energy and hydrology model, which calls boundary_flux for each component:

```
function soil_boundary_fluxes(bc::NamedTuple, boundary, Δz, p, t, params)
    return ClimaLSM.boundary_flux(bc.water, boundary, Δz, p, t, params),
    ClimaLSM.boundary_flux(bc.heat, boundary, Δz, p, t, params)
end
```

This is what we can dispatch off of when the bc is of type "AtmosSurfaceFluxes" (or whatever) - when we want to use surface fluxes & prescribed atmos conditions. This is not needed for Richards equation, because there is only a single component.

Review checklist

I have:

----

### DIFF
--- a/src/Soil/boundary_conditions.jl
+++ b/src/Soil/boundary_conditions.jl
@@ -1,0 +1,198 @@
+import ClimaLSM: AbstractBC, boundary_flux
+export TemperatureStateBC, MoistureStateBC, FreeDrainage, FluxBC
+
+"""
+    AbstractSoilBC <: ClimaLSM. AbstractBC
+
+An abstract type for soil-specific types of boundary conditions, like free drainage.
+"""
+abstract type AbstractSoilBC <: ClimaLSM.AbstractBC end
+
+"""
+   MoistureStateBC <: AbstractSoilBC
+
+A simple concrete type of boundary condition, which enforces a
+state boundary condition ϑ_l = f(p,t) at either the top or bottom of the domain.
+"""
+struct MoistureStateBC <: AbstractSoilBC
+    bc::Function
+end
+
+"""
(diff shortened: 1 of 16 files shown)


src/Soil/boundary_conditions.jl
```diff
@@ -0,0 +1,206 @@
...
+
+An abstract type for soil-specific types of boundary conditions, like free drainage.
+"""
+abstract type AbstractSoilBC <: ClimaLSM.AbstractBC end
+
+struct PrescribedSoilDrivers <: AbstractSoilBC
+    radiation::PrescribedRadiativeFluxes
+    atmos::PrescribedAtmosphere
+end
+
```
kmdeck, line 17: Remove

PR #181, opened 2023-04-06
Ozark experiment update

AlexisRenchon:
## Purpose 

This PR is to do an experiment run of ClimaLSM using Ozark FLUXNET data 

## To-do

Run ClimaLSM models using Ozark metadata and meteorological drivers, compare output with measured fluxes

## Content

- [x] Read Ozark data

----


kmdeck:
@AlexisRenchon squash commits and rebased - we can check buildkite once finished and make sure the plots are uploaded for the experiments & then merge!

PR #183, opened 2023-04-06
Including new vulnerability and p to theta curves

gagnelandmanna:
## Content

----


kmdeck:
perhaps better to start directly from the main branch?

Issue #159, opened 2023-02-24
Misc improvements to plant hydraulics


kmdeck:
Some misc improvements that didnt go into an already large PR, but which would be good to carry out:
- use conductance instead of conductivity (if that is what databases have)
- pass layer widths instead of compartment surfaces
- root tips as prognostic variables -  remove allocation in flux_out_roots function
- root_depths is not used in the soil+plant hydraulics version - can we store somewhere else so that we dont need to specify in this case?

kmdeck:
tagging @gagnelandmanna for visibility, but either one of us can work on these or add to this list

gagnelandmanna:
hi Katherine ok - depending on how i can juggle advancing the other pr and my other work, I will try to get to this but it may take a little while.

kmdeck:
oh yeah! not assigning to you, but it's all stuff I'd want your opinion on before implementing.

gagnelandmanna:
okidoki :) always happy to exchange yes

PR #185, opened 2023-04-12
Including new vulnerability and p to theta curves 

gagnelandmanna:
## Purpose 
The purpose of this PR is to change the vulnerability curve (K(P)) and water retention curve (to convert between P and theta) used in plant hydraulics, as well as to add in the infrastructure for adding new models for this/allowing for modularity.

#159 -- this will link to issue 159

Review checklist

I have:

----


src/Vegetation/PlantHydraulics.jl
```diff
@@ -72,18 +77,16 @@ struct PlantHydraulicsParameters{FT <: AbstractFloat}
...
-    "van Genuchten parameter (unitless)"
-    vg_m::FT
     "porosity (m3/m3)"
     ν::FT
     "storativity (m3/m3)"
     S_s::FT
     "Root distribution function P(z)"
     root_distribution::Function
+    "Conductivity model parameters"
+    retention_params::PS
```
kmdeck, line 87: I believe these will need to have different symbols for their types, because they wont be the same type.

src/Vegetation/PlantHydraulics.jl
```diff
@@ -260,34 +258,112 @@ to take into account the change in K_sat halfway
...
+    if ψ <= FT(0) 
+        K = exp(-(ψ / ψ50)^c) 
     else
         K = FT(1)
     end
     return K * K_sat # (m/s)
 end
 
+"""
+    linear{FT} <: AbstractRetentionModel{FT}
```
kmdeck, line 315: maybe linear_retention_curve?

is intercept such that when S = 1, psi = 0?

src/Vegetation/PlantHydraulics.jl
kmdeck: is Ksat now stored inside the conductivity model parameters?

Does that mean each compartment has the same Ksat? I think that is OK because we are talking about conductivity, but if we used conductance this would change. is that correct?

(our previous code let Ksat be different for each)

src/Vegetation/PlantHydraulics.jl
```diff
@@ -260,34 +258,112 @@ to take into account the change in K_sat halfway
...
 
 """
-     hydraulic_conductivity(K_sat::FT, vg_m::FT, S::FT) where {FT}
+    Weibull{FT} <: AbstractConductivityModel{FT}
+The necessary parameters for the Weibull plant hydraulic conductivity model. 
+Default parameters were taken from Natan Holtzman's orginal Ozark test.
+# Fields
+$(DocStringExtensions.FIELDS)
+"""
+struct Weibull{FT, T1, T2} <: AbstractConductivityModel{FT}
```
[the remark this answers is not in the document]
gagnelandmanna: ok, fixed!

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: Hey Katherine, just removed K_sat from above and yes thinking we should start with this as easier. Eventually K_sat_root should be different though because its magnitude is much smaller than that of the stem and leaves.

gagnelandmanna: If we load plant conductance in the future instead of plant conductivity, I think it is correct that the conductance would be different for each compartment because conductivity is the inherent plant property whereas conductance depends on length

We could still load only one plant conductance (=plant conductivity/length of plant from tip of root to top of leaf) in the parameter struct though and multiply by compartment length inside the functions which require units of m/s; or we could pass plant conductance*length compartment directly.

src/Vegetation/PlantHydraulics.jl
gagnelandmanna: Hi Katherine! I fixed this. Btw I found an excellent paper on water potential in wood. The best (most rigorous) I have found so far. (A. Cloutier; Y. Fortin, 1991)

kmdeck (approved):

PR #189, opened 2023-04-27
Canopy+full soil model

kmdeck:
## Purpose 
Run the canopy+full soil model together.

This assumes that the canopy temperature equals the air temperature and that the emissivity of the canopy is 1. 

It computes the radiation at the soil surface taking into account the absorbed radiation in the canopy, but computes the sensible and latent heat fluxes from the soil and the evaporation from the soil as if the canopy was not present otherwise. This is surely incorrect (the canopy must modify MOST) but is an approximation here.

Transpiration is also computed from the canopy, we are not currently storing the total LHF but instead storing T and E separately. We are not currently storing the outgoing radiation from the surface.

## Content

----


src/Soil/boundary_conditions.jl
```diff
@@ -105,18 +105,8 @@ function soil_boundary_fluxes(
...
-    S_l_sfc = ClimaLSM.Domains.top_center_to_surface(
-        effective_saturation.(ν, Y.soil.ϑ_l, θ_r),
-    )
-    τ_a = ClimaLSM.Domains.top_center_to_surface(
-        @. max(eps(FT), (ν - p.soil.θ_l - Y.soil.θ_i)^(FT(5 / 2)) / ν)
-    )
-    S_c::FT = (1 + ((vg_n - 1) / vg_n)^(1 - 2 * vg_n))^(-vg_m)
-    dsl = dry_soil_layer_thickness.(S_l_sfc, S_c, d_ds)
-    r_soil = @. dsl / (_D_vapor * τ_a) # [s\m]
-    r_ae = conditions.r_ae
```
kmdeck, line 119: use this!

src/soil_canopy_model.jl
```diff
@@ -0,0 +1,244 @@
...
+        top = top_bc,
+        bottom = (water = Soil.FreeDrainage(), heat = zero_flux),
+    )
+    soil = soil_model_type(;
+        boundary_conditions = boundary_conditions,
+        sources = sources,
+        soil_args...,
+    )
+
+    if :transpiration in propertynames(land_args)
```
kmdeck, line 93: why?

src/soil_canopy_model.jl
```diff
@@ -0,0 +1,244 @@
...
+            @. max(eps(FT), (ν - p.soil.θ_l - Y.soil.θ_i)^(FT(5 / 2)) / ν)
+        )
+        S_c = hydrology_cm.S_c
+        dsl = dry_soil_layer_thickness.(S_l_sfc, S_c, d_ds)
+        r_soil = @. dsl / (_D_vapor * τ_a) # [s\m]
+        r_ae = soil_conditions.r_ae
+        @. p.soil_evap = soil_conditions.vapor_flux * r_ae / (r_soil + r_ae)
+        @. p.soil_lhf = soil_conditions.lhf * r_ae / (r_soil + r_ae)
+        @. p.soil_shf = soil_conditions.shf
+        p.T_sfc = surface_temperature(land.soil, Y, p, t)
```
kmdeck, line 202: rename to T_soil, etc

experiments/LSM/ozark/ozark.jl
```diff
@@ -0,0 +1,758 @@
...
+    h_stem + h_leaf,
+    z0_m,
+    z0_b,
+    earth_param_set,
+)
+
+canopy_model_args = (; parameters = shared_params, domain = land_domain.surface)
+
+# Integrated plant hydraulics and soil model
+# The default is diagnostics transpiration. In this case, we are testing with prescribed
```
kmdeck, line 162: update

experiments/LSM/ozark/ozark.jl
```diff
@@ -0,0 +1,758 @@
...
+# but will change that soon.
+land_input =
+    (transpiration = transpiration, atmos = atmos, radiation = radiation)
```
kmdeck, line 165: why is this land input and not part of canopy?

experiments/LSM/ozark/ozark.jl
```diff
@@ -0,0 +1,758 @@
...
+    canopy_component_args = canopy_component_args,
+    canopy_model_args = canopy_model_args,
+)
+Y, p, cds = initialize(land)
+exp_tendency! = make_exp_tendency(land)
+
+#Initial conditions
+Y.soil.ϑ_l = FT(0.4)
+Y.soil.θ_i = FT(0.0)
+T_0 = atmos.T(t0)# replace with soil T
```
kmdeck, line 180: let's do this now

experiments/LSM/ozark/ozark.jl
```diff
@@ -0,0 +1,758 @@
...
+for i in 1:2
+    Y.canopy.hydraulics.ϑ_l[i] .=
+        augmented_liquid_fraction.(plant_ν, S_l_ini[i])
+end
+
+
+update_aux! = make_update_aux(land)
+update_aux!(p, Y, 0.0)
+
+# Set up timestepper
```
kmdeck, line 210: move to sim

src/soil_canopy_model.jl
```diff
@@ -0,0 +1,244 @@
...
+    SM <: Soil.EnergyHydrology{FT},
+    VM <: Canopy.CanopyModel{FT},
+} <: AbstractLandModel{FT}
+    "The soil model to be used"
+    soil::SM
+    "The canopy model to be used"
+    canopy::VM
+end
+
+struct CanopyRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT} end
```
kmdeck, line 27: doc string

src/soil_canopy_model.jl
```diff
@@ -0,0 +1,244 @@
...
+    canopy_model_args::NamedTuple = (;),
+) where {FT, SM <: Soil.EnergyHydrology{FT}}
+
+    # These may be passed in, or set, depending on use scenario.
+    (; atmos, radiation) = land_args
+    # These should always be set by the constructor.
+    Δz = minimum(
+        ClimaCore.Fields.Δz_field(ClimaLSM.coordinates(soil_args.domain)),
+    )
+    sources = (RootExtraction{FT}(),)#Soil.PhaseChange{FT}(Δz))
```
kmdeck, line 75: add back in

src/soil_canopy_model.jl
kmdeck: just make it so it is always diagnostic

src/soil_canopy_model.jl
```diff
@@ -0,0 +1,244 @@
...
+        ),
+        atmos = atmos,
+        radiation = radiation,
+        canopy_model_args...,
+    )
+
+    return SoilCanopyModel{FT, typeof(soil), typeof(canopy)}(soil, canopy)
+end
+
+interaction_vars(m::SoilCanopyModel) = (
```
kmdeck, line 123: doc string

src/soil_canopy_model.jl
```diff
@@ -0,0 +1,244 @@
...
+                    p.canopy.hydraulics.ψ[1],
+                ),
+            ) *
+            (land.canopy.hydraulics.parameters.root_distribution(z))
+
+        # Soil boundary fluxes under canopy or for bare soil
+        bc = land.soil.boundary_conditions.top
+        soil_conditions = surface_fluxes(bc.atmos, land.soil, Y, p, t)
+        (; ν, hydrology_cm, θ_r, d_ds, earth_param_set) = land.soil.parameters
+        _D_vapor = FT(LSMP.D_vapor(earth_param_set))
```
kmdeck, line 188: use new functions

src/soil_canopy_model.jl
```diff
@@ -0,0 +1,244 @@
...
+        )
+        S_c = hydrology_cm.S_c
+        dsl = dry_soil_layer_thickness.(S_l_soil, S_c, d_ds)
+        r_soil = @. dsl / (_D_vapor * τ_a) # [s\m]
+        r_ae = soil_conditions.r_ae
+        @. p.soil_evap = soil_conditions.vapor_flux * r_ae / (r_soil + r_ae)
+        @. p.soil_lhf = soil_conditions.lhf * r_ae / (r_soil + r_ae)
+        @. p.soil_shf = soil_conditions.shf
+        p.T_soil = surface_temperature(land.soil, Y, p, t)
+        p.α_soil = surface_albedo(land.soil, Y, p)
```
kmdeck, line 203: why add these?

kmdeck:
I think we could merge this, but I have a feeling that there are still bugs somewhere...for example, the water potential in the leaves is too low and not responsive enough to the drop in soil water content. 

My strong preference would be to improve some of the discrepancies before making things more complex model-wise...or to be sure of what is causing them before trying to fix them by adding model complexity.

braghiere:
Can we schedule a meeting with some leaf water potential experts to have a look at this together? How about changing emissivity to 0.95?

kmdeck:
> Can we schedule a meeting with some leaf water potential experts to have a look at this together? @Yujie-W, could you have a look at leaf water potential part of the code and see if you find any particular reason why this is not responding to soil water content? How about changing emissivity to 0.95?

we cant yet change the canopy emissivity, because it is not a parameter of the model. We could update the equations to reflect this, though

Yujie-W:
> Can we schedule a meeting with some leaf water potential experts to have a look at this together? @Yujie-W, could you have a look at leaf water potential part of the code and see if you find any particular reason why this is not responding to soil water content?

Where can I find the code that computes P_leaf from Psi_soil?

Yujie-W:
> But not assuming steady state, and not modeling the water content in the roots. The flux from the roots into the stem (lower boundary condition) is set based on the root distribution function, soil water content in each layer, potential in the stem, area index of the stem and roots, etc.

Thanks for pointing out the lines of code. Seen from the code you mentioned, it should work. Can you paste the outputs as well, for example, the computed fluxes from prescibed soil psi from -0.001 MPa to -2 MPa. Or any specific outputs that made you think it does not respond to soil moisture.

kmdeck:
> Thanks for pointing out the lines of code. Seen from the code you mentioned, it should work. Can you paste the outputs as well, for example, the computed fluxes from prescibed soil psi from -0.001 MPa to -2 MPa. Or any specific outputs that made you think it does not respond to soil moisture.

I can try and make a plot showing how the fluxes and leaf water potential vary as a function of prescribed soil potential, assuming steady state & using typical atmospheric conditions for the site. 

In the meantime, here is what the outcome often looks like:
plant_hydraulics
soil_water_content
moisture_stress
water_fluxes

Other possibly useful plots:
ET
GPP
soil_temperature

It seems like the moisture stress factor does not decrease enough during the middle of the summer when the soil moisture drops (around 200 days), so we do not get the big drop in leaf water potential, and do not see the dip in GPP and ET that is in the data.


Some caveats:
- LAI is constant throughout the year, but this simulation starts in the late spring so perhaps LAI~fixed is an OK assumption?
- results are highly dependent on soil van Genuchten properties. 
- sensible heat fluxes are really wrong in this simulation, T_soil surface disagrees with data..

kmdeck:
@Yujie-W quick update :) @gagnelandmanna had also been playing around with parameters and realized that our total water storage in the plant was very large. Reducing the porosity we are using helps a lot! Here is another run:
plant_hydraulics
soil_water_content

Yujie-W:
You have a capacity in the system? Then it could be the reason. If the computed beta reasonable now? If not, you will need to look into the value you used for sc and pc as in https://github.com/CliMA/ClimaLSM.jl/blob/ad0b840116df53e0ceb2085366a7cbd2a85f8eb0/src/Vegetation/canopy_parameterizations.jl#L261

AlexisRenchon (approved):
Looks good to merge! 
Just a note on path change if we merge #249 first

braghiere (approved):
Looks good to me.

src/Vegetation/canopy_parameterizations.jl
```diff
@@ -240,7 +240,7 @@ dark respiration (`Rd`), and the moisture stress factor (`β`).
 See Table 11.5 of G. Bonan's textbook, Climate Change and Terrestrial Ecosystem Modeling (2019).
 """
 function net_photosynthesis(Ac::FT, Aj::FT, Rd::FT, β::FT) where {FT}
-    An = max(0, min(Ac * β, Aj) - Rd)
+    An = max(0, min(Ac, Aj) * β - Rd)
```
kmdeck, line 243: This is correct: Rd already has the factor, while Ac and Aj do not. From Bonan's book:
"Other models impose biochemical limitations and indirectly reduce stomatal conduct- ance by reducing An as soil moisture stress increases (multiplying Vcmax and Jmax by betaw)."

src/soil_canopy_model.jl
```diff
@@ -0,0 +1,309 @@
...
+This function is called each ode function evaluation.
+"""
+function make_interactions_update_aux(
+    land::SoilCanopyModel{FT, SM, RM},
+) where {FT, SM <: Soil.EnergyHydrology{FT}, RM <: Canopy.CanopyModel{FT}}
+    function update_aux!(p, Y, t)
+        z = ClimaCore.Fields.coordinate_field(land.soil.domain.space).z
+        (; area_index, conductivity_model) = land.canopy.hydraulics.parameters
+        @. p.root_extraction =
+            (area_index[:root] + area_index[:stem]) / 2 *
```
kmdeck, line 217: shouldnt default to stem, in case we have no stem.

PR #190, opened 2023-05-09
Jacobian update defaults

kmdeck:
## Purpose 
Part of SDI #135 
Adds functions which compute the Jacobian and Jacobian due to the boundary terms in the tendency.
Adds default functions + extends those for Richards equation.

## Content
- adds three functions: `make_update_jacobian`, `make_tendency_jacobian`, and `\partialtendencyBC\partialY`. 
- adds the methods of these for RichardsModel.
- adds unit tests which demonstrate that the Jacobian is computed correctly.

Review checklist

I have:

----


kmdeck:
@juliasloan25 not ready for review yet, but let me know if you think the scope of this PR is appropriate! the next step would be to add unit tests to make sure W has the right values. One thing I wasnt sure about was how specific the TridiagonalW is for the model in question. It seems like it is specific because of the naming of the properties of W (W.\partial \theta_t\partial\theta), and because other models will require more blocks. 

What about the linsolve functions? I think those could go in a different PR...note that they are kind of hardcoded to be for `soil` etc, so we will have to generalize them.

src/SharedUtilities/implicit_tendencies.jl
```diff
@@ -0,0 +1,61 @@
...
+end
+
+"""
+    ∂tendencyBC∂Y(::AbstractModel, ::AbstractBC, ::AbstractBoundary, _...)
+
+A function stub which returns the derivative of the implicit tendency
+term of the `model` arising from the boundary condition,
+with respect to the state Y.
+"""
+function ∂tendencyBC∂Y(
```
juliasloan25, line 52: It seems like we could come up with a better name for this function - maybe `\partial T_BC \partial Y`? That might be worse though because `T_BC` sounds like temperature at the boundary. What do you think?

src/Soil/boundary_conditions.jl
```diff
@@ -270,3 +270,43 @@ function soil_boundary_fluxes(bc::NamedTuple, boundary, model, Y, Δz, p, t)
...
+    p,
+    t,
+)
+    (; ν, vg_α, vg_n, vg_m, S_s, θ_r) = model.parameters
+    K = ClimaLSM.Domains.top_center_to_surface(p.soil.K)
+    ϑ_l = ClimaLSM.Domains.top_center_to_surface(Y.soil.ϑ_l)
+    return ClimaCore.Fields.FieldVector(;
+        :soil => (;
+            :ϑ_l => @. -K / Δz * dψdθ(ϑ_l, ν, θ_r, vg_α, vg_n, vg_m, S_s) /
+               (2 * Δz)
```
juliasloan25, line 290: Just double checking - should K be negative here? And should the second term be divided by `2\Delta z` or just `\Delta z`? From [eqns 20-22 in the overleaf](https://www.overleaf.com/project/63be02f455f84a77642ef485) for `\partialT \partialtheta`, it looks to me like we don't need the negative sign and the second `\Delta z` term doesn't need to be multiplied by 2. I might be misinterpreting though :)

src/Soil/rre.jl
```diff
@@ -214,3 +214,137 @@ function ClimaLSM.make_update_aux(model::RichardsModel)
...
+        W_column_arrays,
+        temp1,
+        temp2,
+        transform,
+    )
+end
+
+
+"""
+    """
```
juliasloan25, line 277: We should add a docstring explaining that `update_jacobian!` should be used as `Wfact!` for CTS/ODE.

juliasloan25 (commented):

src/Soil/rre.jl
kmdeck: will do!

src/SharedUtilities/implicit_tendencies.jl
```diff
@@ -0,0 +1,61 @@
...
+    ::AbstractModel,
+    ::AbstractBC,
+    ::AbstractBoundary,
+    _...,
+)::ClimaCore.Fields.FieldVector end
```
kmdeck, line 57: Does it make sense to declare the type of the output if the stub function doesnt return anything?

src/Soil/boundary_conditions.jl
kmdeck: In the overleaf I wrote the BC for the divergence -d/dz(F), but here we are giving the divergence of the tendency..so the minus sign get absorbed (and we get two factors of dz).

I think the multiply by 2 needs to be there because the distance between the first cell center and the top surface is dz_1/2, where dz_1 is the thickness of the layer (face to face).  This is the distance used to estimate the boundary flux. But then the divergence of the flux is (F_bc{1/2}  - F_{3/2})/dz_1.

src/Soil/boundary_conditions.jl
```diff
@@ -270,3 +270,42 @@ function soil_boundary_fluxes(bc::NamedTuple, boundary, model, Y, Δz, p, t)
     return ClimaLSM.boundary_flux(bc.water, boundary, Δz, p, t, params),
     ClimaLSM.boundary_flux(bc.heat, boundary, Δz, p, t, params)
 end
+
+
```
kmdeck, line 274: add doc string

src/Soil/boundary_conditions.jl
```diff
@@ -270,3 +270,42 @@ function soil_boundary_fluxes(bc::NamedTuple, boundary, model, Y, Δz, p, t)
...
+    ϑ_l = Fields.level(interpc2f_op.(Y.soil.ϑ_l), face_len)
+    return ClimaCore.Fields.FieldVector(;
+        :soil => (;
+            :ϑ_l => @. -K / Δz * dψdϑ(ϑ_l, ν, θ_r, vg_α, vg_n, vg_m, S_s) /
+               (2 * Δz)
+        ),
+    )
+end
+
+function ClimaLSM.∂tendencyBC∂Y(
```
kmdeck, line 301: add doc string

src/Soil/rre.jl
```diff
@@ -214,3 +214,159 @@ function ClimaLSM.make_update_aux(model::RichardsModel)
...
+    N = Spaces.nlevels(center_space)
+
+    tridiag_type = Operators.StencilCoefs{-1, 1, NTuple{3, FT}}
+    ∂ϑₜ∂ϑ = Fields.Field(tridiag_type, center_space)
+
+    ∂ϑₜ∂ϑ.coefs[1] .= FT(0)
+    ∂ϑₜ∂ϑ.coefs[2] .= FT(0)
+    ∂ϑₜ∂ϑ.coefs[3] .= FT(0)
+
+    arr1 = Array{FT}(undef, N - 1)
```
kmdeck, line 273: not used?

src/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -5,7 +5,21 @@ export volumetric_liquid_fraction,
     pressure_head,
     hydraulic_conductivity,
     impedance_factor,
-    viscosity_factor
+    viscosity_factor,
+    dψdϑ
+
+"""
```
kmdeck, line 11: doc string

src/SharedUtilities/implicit_tendencies.jl
kmdeck: changed to Union{Nothing, ClimaCore.Fields.FieldVector}

kmdeck:
@dennisYatunin Big PR, but the area where it would be most helpful for you to take a look is:
Jacobian computation + implementation of boundary contributions
https://github.com/CliMA/ClimaLSM.jl/blob/d68d4fc7d1b7e13b636ddd8635ba9988024aad9c/src/Soil/rre.jl#L353-L366

If you also want to look at the full Jacobian computation and definition, it is the same function/in that same file (src/Soil/rre.jl).

1. We want to be easily adapt to the new stencil interface when you finish the refactoring youre working on,  while at the same time making progress in implementing the working implicit solver into main.  I think having functions that return these boundary values is general and that can be reused, and we'll just have to change how we use the values inside the Jacobian update function to update the boundary matrix entries.  Is that the case, or is there a better way to implement a solution now that will generalize more nicely to the new interface youre working on?

3. Right now we just have a single variable, but in the future this has to generalize to multiple variables. so perhaps TBD on that, but if you have ideas now, that would be helpful.

src/Soil/Soil.jl
```diff
@@ -56,18 +56,20 @@ for interactions between components.
 =#
 
 using ClimaLSM
-using UnPack
 using DocStringExtensions
+using LinearAlgebra
```
juliasloan25, line 60: Why do we need LinearAlgebra in this file?

test/implicit_timestepping/richards_model.jl
```diff
@@ -0,0 +1,189 @@
...
+        dz = FT(0.01)
+        dψdϑ_ic = dψdϑ(FT(0.24), ν, θ_r, vg_α, vg_n, vg_m, S_s)
+
+        @test all(parent(W.temp2) .== FT(0.0))
+        @test all(parent(W.temp2) .== FT(0.0))
+        @test W.transform == false
+        @test typeof(W.W_column_arrays) <:
+              Vector{LinearAlgebra.Tridiagonal{Float64, Vector{Float64}}}
+        @test length(W.W_column_arrays) == 1
+        if typeof(domain) <: Column
```
juliasloan25, line 72: great tests!!

juliasloan25 (approved):
Looks good!

src/Soil/Soil.jl
kmdeck: Hm, we need LinearAlgebra in rre.jl (for making TridiagonalW stuff) but rre.jl is included in the Soil.jl module. We have all the "using" "import" etc statements in Soil.jl. perhaps not so clear?

src/SharedUtilities/implicit_tendencies.jl
kmdeck: I'm going to leave this alone for now, but I'm happy to revisit it as we continue work on implicit solvers! In particular, I think we'll have to change the function a little for cases with more than one variable.

PR #191, opened 2023-05-10
imp, exp tendency functions instead of ode_function/rhs

juliasloan25:
## Purpose 
Part of SDI #135 
Replaces some functions to setup infrastructure for mixed implicit/explicit timestepping schemes:
- `make_ode_function` -> `make_exp_tendency, make_imp_tendency`
- `ode_function` -> `exp_tendency!, imp_tendency!`
- `make_rhs` -> `make_compute_exp_tendency, make_compute_imp_tendency`
- `rhs!` -> `compute_exp_tendency!, compute_imp_tendency!`

For now, the implicit functions aren't used (even for the Richards model). This will be addressed soon in another PR.

This PR also adds two new subtypes of the `AbstractModel` type, `AbstractImExType` and `AbstractExpType`. The former will be used for models which will be treated implicitly, and the former is for models which will be treated explicitly. These types are used for dispatch in functions that behave differently if the model is stepped implicitly or explicitly, such as `make_imp_tendency`.

## Content
- [x] Add `AbstractImExModel`, `AbstractExpModel` types
- [x] Add default functions for AbstractLandModel, AbstractModel, AbstractImExModel, AbstractCanopyComponent types
- [x] Replace all `make_rhs, rhs!` functions and calls with `make_compute_exp_tendency, compute_exp_tendency!`
- [x] Replace all `make_ode_function, ode_function!` functions and calls with `make_exp_tendency, exp_tendency!`
- [x] Update docs and tutorials to explain new interface
- [x] Test that the default functions  for abstract types return `nothing`

----

### DIFF
--- a/docs/tutorials/Usage/domain_tutorial.jl
+++ b/docs/tutorials/Usage/domain_tutorial.jl
@@ -2,8 +2,8 @@
 # Domain Tutorial
 
 ## Goals of the tutorial
-The goal of this is to outline what is currently implemented in ClimaLSM 
-and to serve as a software design document 
+The goal of this is to outline what is currently implemented in ClimaLSM
+and to serve as a software design document
 for future development involving the underlying domains.
 
 ## Background
@@ -11,10 +11,10 @@ for future development involving the underlying domains.
 In both the atmosphere and the ocean, all variables are defined at all locations
 in the region of interest, or domain.  For example, the air density, temperature, pressure,
 and wind speed are defined everywhere in the domain. After choosing a resolution
-and discretizing space, the numerical problem is to advance a 
-system of differential equations, where at each coordinate point a value of 
+and discretizing space, the numerical problem is to advance a
+system of differential equations, where at each coordinate point a value of
 `ρ`, `T`, `P`, and `u⃗` are solved for at each step. The choice of domain is a question "only"
-of geometry: you may be interested in a large eddy simulation (using a box domain), or 
+of geometry: you may be interested in a large eddy simulation (using a box domain), or
 in a global model (where you would need a spherical shell domain
 representing the atmosphere or ocean from some depth to a given height).
 
@@ -42,9 +42,9 @@ A domain represents a region of space. In ClimaLSM, domains are simply
 structs containing parameters that define these regions - for example an
 x-range and y-range that define a plane. In addition, ClimaLSM domains store
 the ClimaCore function space on the
-physical domain. When solving partial differential equations, the spatial 
+physical domain. When solving partial differential equations, the spatial
 discretization is tied to a set of basis functions you wish to use to represent the prognostic
-variable as a function of space. The nodal points - the locations in space where the variable 
+variable as a function of space. The nodal points - the locations in space where the variable
 is solved for - are arranged in `space` in a manner which depends on these basis functions.
 
  Note that this is only needed for domains of variables satisfying PDEs,
@@ -60,7 +60,7 @@ of x and y, and a Point domain only has a coordinate z_sfc.
 
 
 ## Domain types
-All ClimaLSM domains are subtypes of abstract type 
+All ClimaLSM domains are subtypes of abstract type
 [`ClimaLSM.Domains.AbstractDomain`](https://clima.github.io/ClimaLSM.jl/dev/APIs/SharedUtilities/#ClimaLSM.Domains.AbstractDomain).
 A variety of concrete domain types are supported:
 
@@ -69,9 +69,9 @@ A variety of concrete domain types are supported:
 - 2D: [`Domains.Plane`](https://clima.github.io/ClimaLSM.jl/dev/APIs/SharedUtilities/#ClimaLSM.Domains.Plane), [`Domains.SphericalSurface`](https://clima.github.io/ClimaLSM.jl/dev/APIs/SharedUtilities/#ClimaLSM.Domains.SphericalSurface)
 - 3D: [`Domains.HybridBox`](https://clima.github.io/ClimaLSM.jl/dev/APIs/SharedUtilities/#ClimaLSM.Domains.HybridBox), [`Domains.SphericalShell`](https://clima.github.io/ClimaLSM.jl/dev/APIs/SharedUtilities/#ClimaLSM.Domains.SphericalShell).
 
-Single component models (soil, snow, vegetation, canopy airspace...) will be solved on 
+Single component models (soil, snow, vegetation, canopy airspace...) will be solved on
 a single domain. Which domain is appropriate depends on the model equations (PDE vs ODE) and
-on the configuration of interest (single column or 3d). 
+on the configuration of interest (single column or 3d).
 
 Furthermore, these individual model domains are composed when building a multi-component
 Land System Model. In this case, the full LSM domain is split into regions:
@@ -86,9 +86,9 @@ domain being modeled:
 | LSMCartesianBoxDomain    | Plane[^3]       | HybridBox      |
 | LSMSphericalShellDomain    | SphericalSurface[^3]       | SphericalShell      |
 
-It is important to note that the horizontal domain used for the surface and subsurface 
+It is important to note that the horizontal domain used for the surface and subsurface
 domains are identical in LSM simulations. This ensures that we can use the same indexing
-of surface and subsurface domains and variables to correctly compute and 
+of surface and subsurface domains and variables to correctly compute and
 apply boundary fluxes between different component models. Otherwise we would need
 to develop additional infrastructure in order to, for example, select the correct subsurface
 column corresponding to a particular surface location.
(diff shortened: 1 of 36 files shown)


.github/workflows/JuliaFormatter.yml
```diff
@@ -36,8 +36,8 @@ jobs:
...
       run: |
         julia --project=.dev .dev/climaformat.jl .
-          
+
     - name: Check formatting diff
       if: steps.filter.outputs.julia_file_change == 'true'
       run: |
-        git diff --color=always --exit-code
\ No newline at end of file
+        git diff --color=always --exit-code
```
kmdeck, line 43: meant to comment here - changed on purpose?

src/Soil/energy_hydrology.jl
```diff
@@ -239,7 +242,7 @@ function ClimaLSM.make_rhs(model::EnergyHydrology{FT}) where {FT}
         # This has to come last
         dss!(dY, model.domain)
```
kmdeck, line 243: with CTS, was this removed from the tendency function?

.github/workflows/JuliaFormatter.yml
juliasloan25: oops, no. I'll change it back

src/Soil/energy_hydrology.jl
juliasloan25: We can remove it from the tendency if we include it in the `CTS.ClimaODEFunction` when we construct the `ODEProblem` (like [this](https://github.com/CliMA/ClimaLSM.jl/blob/js/implicit-solver/src/Soil/examples/imex_testing_modified_picard.jl#L131-L140)). We're not currently doing that in main though, only on the implicit solver branch.

kmdeck: Does this mean we either define our tendency function with it (and use ODE's infrastructure completely) or else we remove it from the tendency and pass it via  CTS.ClimaODEFunction() but then are required to use ClimaTimesteppers? or can we use e.g. a CTS.ClimaODEFunction but use an ODE timestepper?

kmdeck: (I suspect the latter, but just want to make sure!)

kmdeck:
@AlexisRenchon @braghiere @JianingFang @Yujie-W @cfranken a heads up that we are renaming some core functions in ClimaLSM to account for mixed implicit/explicit timestepping schemes, which are required for the soil model. If a component does not require implicit timestepping, the only change is in naming (make_rhs -> make_compute_exp_tendency), as Julia notes here. If a model requires implicit timestepping, you will also need to define the implicit tendency  in addition to the explicit tendency.

For example, for the soil model, vertical diffusion terms are part of the implicit rhs while horizontal diffusion terms are the explicit rhs.

Please review and ask questions if this affects you!

docs/tutorials/Usage/domain_tutorial.jl
```diff
@@ -118,16 +118,18 @@ for example) will be [θ11, θ12; θ21, θ22]. Your variable always has the same
 ## Future work
 Almost all interactions between variables in land surface models are within column - that is, there is only
 vertical transport and exchanges. The exception to this is the horizontal flow of water on the surface
-and within the soil. The `right hand side` (rhs) functions (the ODE functions) can be split into
-"vertical" and "horizontal" pieces. We envision each step of the land surface model simulation to be solved  
-in two steps: (1) the vertical rhs evaluations are carried out (and can be parallelized), and (2) the horizontal 
-rhs functions are then evaluated (possibly less frequently?) and require communcation between columns.
-In this case, right hand side functions will need to be aware of the domain.
+and within the soil. The tendency (produced by `make_exp_tendency` and `make_imp_tendency`) functions
+(the ODE functions) can be split into "vertical" and "horizontal" pieces. We envision each step of the
```
kmdeck, line 122: "We envision each step of the
land surface model simulation to be solved  in two steps: (1) the vertical tendency evaluations are carried out
(and can be parallelized), and (2) the horizontal tendency functions are then evaluated (possibly less frequently?)
and require communication between columns. 

In general, tendencies reflecting horizontal flow will be treated explicitly and include in the explicit tendency function. Tendencies reflecting vertical flow may be treated explicitly or implicitly depending on the use case. To solve the problem, we then use IMEX (mixed explicit/implicit) methods."

src/SharedUtilities/models.jl
```diff
@@ -100,21 +117,39 @@ function make_update_aux(model::AbstractModel)
...
 """
-    make_ode_function(model::AbstractModel)
+    make_imp_tendency(model::AbstractModel)
+
+Returns an `imp_tendency` that updates auxiliary variables and
+updates the prognostic state of variables that are stepped implicitly.
+
+`compute_imp_tendency!` should be compatible with OrdinaryDiffEq.jl solvers.
+"""
+function make_imp_tendency(model::AbstractModel)
```
kmdeck, line 127: maybe we could discuss these defaults a little more. Just briefly speaking with Dennis, he pointed out that even if the implicit tendency only updates the aux state, we'll still be passing to the timestepper. it would be an unecessary update aux (which we discussed before)  - in fact it would call it every stage, not once per step, which might be less OK with us in terms of unecessary calls.

what to we pass to the ODEFunction if we really want no implicit tendency is `nothing`
```
CTS.ClimaODEFunction(
            T_exp! = explicit_tendency!,
            T_imp! = nothing,
            dss! = ClimaLSM.Soil.dss!,
```

[reminder for future PR to move the dss! function ClimaLSM.Domains?]

src/Vegetation/component_models.jl
```diff
@@ -97,16 +97,31 @@ function initialize_auxiliary(
...
+     ClimaLSM.make_compute_imp_tendency(component::AbstractCanopyComponent, canopy)
+
+Creates the compute_imp_tendency!(dY,Y,p,t) function for the canopy `component`.
+
+Since component models are not standalone models, other information
+may be needed and passed in (via the `canopy` model itself).
+The right hand side for the entire canopy model can make use of
+these functions for the individual components.
+"""
+function ClimaLSM.make_compute_imp_tendency(::AbstractCanopyComponent, canopy)
```
kmdeck, line 124: we may be able to avoid adding this. If we call make_imp_tendency(c::Canopy), and there is no method defined for it (which currently there is not, because it is all explicit), we wont ever call make_imp_tendency for the components.

src/SharedUtilities/models.jl
kmdeck: one idea: we could add a boolean field to `model` like "has_implicit" and if false, imp_tendency is nothing, and if true, it does what we have here

src/Soil/energy_hydrology.jl
juliasloan25: Unfortunately, we can't use a CTS problem setup (e.g. `CTS.ClimaODEFunction`) with an ODE timestepping algorithm. 

If we wanted to use CTS to setup the `ODEProblem` but then an ODE algorithm to step it, we could add the dss step to the end of both the explicit and implicit tendencies, and use a `SplitODEProblem` (instead of an `ODEProblem`) with both an implicit and an explicit tendency (no dss field). However, this approach became numerically unstable after 300 days when used in ClimaAtmos, so it's probably a better approach to remove dss from the tendencies and include it in the problem instead, even though this won't be compatible with ODE algorithms.

src/SharedUtilities/models.jl
juliasloan25: After discussing offline, @kmdeck and I decided to add two subtypes to our `AbstractModel` type: `AbstractImexModel` and `AbstractExpModel`. We can dispatch off of these types to define defaults that better suit the desired behavior - `imp_tendency!` will be nothing for an `AbstractExplicitModel`, but will call `update_aux!` and `compute_imp_tendency!` for an `AbstractImplicitModel`. This will allow us to take advantage of multiple dispatch, and avoid unnecessarily defining an unused `make_compute_implicit_tendency` for `AbstractExplicitModel` types (which will also give a helpful method/type error if `make_compute_implicit_tendency` is called incorrectly).

src/Vegetation/component_models.jl
juliasloan25: Good point! I'll remove it.

src/SharedUtilities/models.jl
```diff
@@ -14,6 +18,7 @@ export AbstractModel,
     name
 
 import .Domains: coordinates
+using ClimaCore
```
kmdeck, line 21: dont need this?

src/SharedUtilities/models.jl
```diff
@@ -22,6 +27,24 @@ An abstract type for all models.
 """
 abstract type AbstractModel{FT <: AbstractFloat} end
 
+"""
+    AbstractImExModel{FT} <: AbstractModel{FT}
+
+An abstract type for models which have both an implicit and an explicit tendency.
```
kmdeck, line 33: "which have tendency terms which must be treated implicitly (and which may also have tendency terms that can be treated explicitly)"

src/SharedUtilities/models.jl
```diff
@@ -22,6 +27,24 @@ An abstract type for all models.
...
+This inherits all the default function definitions from AbstractModel, as well
+as `make_imp_tendency` and `make_compute_imp_tendency` defaults.
+"""
+abstract type AbstractImExModel{FT} <: AbstractModel{FT} end
+
+"""
+    AbstractExpModel{FT} <: AbstractModel{FT}
+
+An abstract type for models which have only an explicit tendency.
```
kmdeck, line 42: similar comment

src/SharedUtilities/models.jl
```diff
@@ -100,21 +137,49 @@ function make_update_aux(model::AbstractModel)
...
+end
+
+"""
+    make_imp_tendency(model::AbstractModel)
+
+Returns an `imp_tendency` that does nothing. This model type is not
+stepped explicity.
+"""
+function make_imp_tendency(model::AbstractModel)
+    function imp_tendency!(dY, Y, p, t) end
```
kmdeck, line 164: Should this be `nothing`, or a function that doesnt do anything?

kmdeck (approved):
looks good! just minor comments/suggestions. looks good to merge otherwise. nice job!!

src/SharedUtilities/models.jl
juliasloan25: Good catch! I'll remove it

src/SharedUtilities/models.jl
```diff
@@ -78,15 +101,29 @@ Here, the coordinate points are those returned by coordinates(model).
...
+Return a `compute_imp_tendency!` function that updates state variables
+that we will be stepped implicitly.
 
-`rhs!` should be compatible with OrdinaryDiffEq.jl solvers.
+`compute_imp_tendency!` should be compatible with OrdinaryDiffEq.jl solvers.
 """
-function make_rhs(model::AbstractModel)
-    function rhs!(dY, Y, p, t) end
-    return rhs!
+function make_compute_imp_tendency(model::AbstractImExModel)
```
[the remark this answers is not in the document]
juliasloan25: Will do!

src/SharedUtilities/models.jl
juliasloan25: Do you mean defining `make_imp_tendency` without the `function imp_tendency!` inside it at all? Or are you asking about `imp_tendency!`?

src/ClimaLSM.jl
```diff
@@ -115,17 +116,32 @@ function initialize_interactions(land::AbstractLandModel, land_coords)
     return NamedTuple{vars}(interactions)
 end
 
-function make_ode_function(land::AbstractLandModel)
+function make_imp_tendency(land::AbstractLandModel)
     components = land_components(land)
```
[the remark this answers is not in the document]
juliasloan25: I'm not sure this would work since `typeof.(components) .<: AbstractImExModel` will return a `BitVector` of 1/0 values for each element in `components`, which can't be used in an if statement. Maybe we could instead have a for loop that goes over each component? The goal with this is to avoid unnecessary `update_aux` calls right?

I also just realized that we can't call `make_compute_imp_tendency` on an `AbstractExpModel` since it's not defined, so we'll need to change how we construct `compute_imp_tendency_list` in case `land` contains an `AbstractExpModel`.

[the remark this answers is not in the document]
kmdeck: let's discuss in person today!

src/SharedUtilities/models.jl
juliasloan25: After discussing offline, Kat and I decided to have `make_imp_tendency` return a function which does nothing. The advantage of this is that all methods of `make_imp_tendency` return a `Function` type, rather than `Union{Function, Nothing}`. A disadvantage of this is an unnecessary function call when `imp_tendency!` is called by the solver, but we figured any performance hit due to this will be minor.

src/ClimaLSM.jl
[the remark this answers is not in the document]
juliasloan25: Kat and I discussed offline - we'll do something like the following
```
if all(c -> typeof(c) .<: AbstractExpModel, components)
    return nothing
else
    do the existing code
end
```

To make sure that we are able to handle the mixed implicit/explicit case, we'll extend the `make_compute_imp_tendency` function for the `AbstractExpModel` argument type, and define this method to return a function which does nothing.

PR #204, opened 2023-05-17
upgrade to julia1.9

### LINKED ISSUE #196: Upgrade to Julia 1.9

kmdeck:
## Purpose 
Upgrade to Julia1.9. Code is compatible with Julia1.8 except for experiments. The user must delete the pulled Manifest.toml if they use Julia1.8 and re-instantiate. This is noted in the README for experiments.
Closes #196 

Review checklist

I have:

----


AlexisRenchon (commented):
We have multiple Project.toml and sometime Manifest.toml (e.g., experiment), can they have different julia compat? 
The Project.toml in experiment should be 1.9 only, but the others could have 1.8 and 1.9
is that good practice?

experiments/Manifest.toml
```diff
@@ -1,20 +1,28 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.5"
+julia_version = "1.9.0"
```
juliasloan25, line 3: Shouldn't we update `experiments/Project.toml` too if we want the Manifest to be updated?

juliasloan25: Actually, looking at Charlie's PR to update ClimaAtmos to Julia v1.9 [here](https://github.com/CliMA/ClimaAtmos.jl/pull/1659/files#diff-9a9d70a0eb6f3494fd934c224bf0867c14b00d46fde8bcc4d65bb2d836898946), it looks like he only updates the Manifests (not Projects) of directories that have the Manifest checked in. In fact, it seems like he doesn't update any of the Projects. Maybe this is because specifying `julia = "1.8"` is the same as `julia = "1.8, 1.9"` (both restrict to `[1.8, 2.0)`)? I'm not sure it matters either way - it could be more clear to add the 1.9 in.

juliasloan25 (commented):
I think we also need to update the YAML files in `.github/workflows`. Also the `.buildkite/pipeline.yml` and maybe `bors.toml`.

Other than that, it looks good!

experiments/Manifest.toml
kmdeck: My thought was that if someone clones the repo and the manifest in experiments was made with 1.9, it will fail if they try and run it with 1.8. so perhaps we should restrict the experiments compat to 1.9? Or else they need to delete it and re-instaniate. Not sure how "out of the box" it needs to run.

But sounds like what you and @AlexisRenchon suggested with respect to the Project.tomls is to leave at just "1.8" because that indicates 1.9 is also supported. I havent tested with 1.7 and Im sure it might fail at some point on an earlier Julia version.. so perhaps leaving the 1.8 there is good since we dont know if we support previous versions?

kmdeck:
> I think we also need to update the YAML files in `.github/workflows`. Also the `.buildkite/pipeline.yml` and maybe `bors.toml`.
> 
> Other than that, it looks good!

good catch!!

experiments/Manifest.toml
kmdeck: (also, our CI will just run with 1.9)

juliasloan25 (approved):
Looks good!

kmdeck:
bors try

kmdeck:
bors try

PR #211, opened 2023-05-22
allow for different hydrology closure models

kmdeck:
## Content

----


src/Soil/soil_heat_parameterizations.jl
```diff
@@ -49,8 +50,9 @@ function phase_change_source(
     T::FT,
     τ::FT,
     params::EnergyHydrologyParameters{FT},
+    cm,
```
kmdeck, line 53: add type?

src/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -41,84 +59,111 @@ function volumetric_liquid_fraction(ϑ_l::FT, ν_eff::FT, θ_r::FT) where {FT}
...
+variably saturated soil, using the van Genuchten matric potential 
+if the soil is not saturated, and
 an approximation of the positive pressure in the soil if the
 soil is saturated.
 """
 function pressure_head(
-    α::FT,
-    n::FT,
-    m::FT,
+    cm::vanGenuchten,
```
juliasloan25, line 118: Should this dispatch on `::vanGenuchten{FT}`? I think it's better to specify the subtype of the struct as well

src/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -128,6 +173,103 @@ function hydraulic_conductivity(K_sat::FT, m::FT, S::FT) where {FT}
 end
 
 
+"""
+     matric_potential(cm::BrooksCorey{FT}, S::FT) where {FT}
```
juliasloan25, line 177: Maybe we should add documentation somewhere explaining which functions we expect to be implemented for each closure model concrete type. Right above `AbstractSoilHydrologyClosure` might be a good place for that.

src/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -128,6 +173,103 @@ function hydraulic_conductivity(K_sat::FT, m::FT, S::FT) where {FT}
...
+    ) where {FT}
+
+A point-wise function returning the pressure head in
+variably saturated soil, using the Brooks and Corey matric potential 
+if the soil is not saturated, and
+an approximation of the positive pressure in the soil if the
+soil is saturated.
+"""
+function pressure_head(
+    cm::BrooksCorey,
```
juliasloan25, line 256: I think the type annotation should be `::BrooksCorey{FT}`

test/LSM/pond_soil_lsm.jl
```diff
@@ -5,9 +5,8 @@ using ClimaLSM.Domains: LSMMultiColumnDomain, LSMSingleColumnDomain
 using ClimaLSM.Soil
 using ClimaLSM.Pond
 
-FT = Float64
 @testset "Pond soil Multi Column LSM integration test" begin
-
+    FT = Float64
```
juliasloan25, line 9: Maybe we should test for Float32 and Float64?

test/Soil/soil_parameterizations.jl
```diff
@@ -48,22 +49,21 @@ include(joinpath(pkgdir(ClimaLSM), "parameters", "create_parameters.jl"))
...
-    (; α, β, Ω, γ, vg_m, γT_ref) = parameters
+    (; α, β, Ω, hydrology_cm, γ, γT_ref) = parameters
     @test α == 0.24
     @test β == 18.3
     @test γ == 2.64e-2
     @test Ω == 7.0
     @test γT_ref == 288.0
-    @test vg_m == 1.0 - 1.0 / vg_n
-    @test typeof.([α, β, γ, Ω, vg_m, γT_ref]) == [FT, FT, FT, FT, FT, FT]
+    @test hydrology_cm.m == 1.0 - 1.0 / vg_n
```
juliasloan25, line 65: Should we also test the other fields of `hydrology_cm`?

test/Soil/soil_test_3d.jl
```diff
@@ -16,7 +16,8 @@ FT = Float64
     S_s = FT(1)#e-3); #inverse meters
     vg_n = FT(2.0)
     vg_α = FT(2.6) # inverse meters
-    vg_m = FT(1) - FT(1) / vg_n
+    vg_m = 1 - 1 / vg_n
```
juliasloan25, line 19: I think `vg_m` is no longer needed here

test/Soil/soiltest.jl
```diff
@@ -185,6 +185,7 @@ end
     vg_n = FT(2.0)
     vg_α = FT(2.6) # inverse meters
     vg_m = FT(1) - FT(1) / vg_n
```
juliasloan25, line 187: I think `vg_m` can be deleted here

test/Soil/soiltest.jl
```diff
@@ -565,6 +562,7 @@ end
...
```
juliasloan25, line 564: `vg_m` can be deleted

test/implicit_timestepping/richards_model.jl
```diff
@@ -19,6 +19,7 @@ include(joinpath(pkgdir(ClimaLSM), "parameters", "create_parameters.jl"))
     vg_n = FT(1.43)
     vg_α = FT(2.6) # inverse meters
     vg_m = FT(1) - FT(1) / vg_n
```
juliasloan25, line 21: `vg_m` can be deleted

src/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -41,84 +59,111 @@ function volumetric_liquid_fraction(ϑ_l::FT, ν_eff::FT, θ_r::FT) where {FT}
...
 end
 
+
+"""
+   dpdϑ(cm::vanGenuchten{FT}, ϑ, ν, θ_r, S_s)
+
+Computes and returns the derivative of the pressure head 
+with respect to ϑ for the van Genuchten formulation.
+"""
+function dpdϑ(cm::vanGenuchten{FT}, ϑ, ν, θ_r, S_s) where {FT}
```
juliasloan25, line 140: Why did you rename this from `d\psi d\theta` to `dpd\theta`? I think if we're still using `\psi` for matric potential it makes sense to use it here too

experiments/LSM/ozark/ozark.jl
```diff
@@ -34,11 +34,9 @@ include(joinpath(climalsm_dir, "experiments/LSM/ozark/ozark_simulation.jl"))
 # Now we set up the model. For the soil model, we pick
 # a model type and model args:
 soil_domain = land_domain.subsurface
-soil_ps = Soil.RichardsParameters{FT}(
+soil_ps = Soil.RichardsParameters{FT, vanGenuchten{FT}}(
     soil_ν,
-    soil_vg_α,
-    soil_vg_n,
-    soil_vg_m,
+    vanGenuchten(; α = soil_vg_α, n = soil_vg_n),
```
juliasloan25, line 39: With this change, we should be able to delete `vg_m` from ozark_parameters.jl

experiments/Standalone/Soil/evaporation.jl
```diff
@@ -26,10 +26,14 @@ K_sat = FT(225.1 / 3600 / 24 / 1000)
 # n and alpha estimated by matching vG curve.
 vg_n = FT(10.0)
 vg_α = FT(6.0)
+vg_m = 1 - 1 / vg_n
```
juliasloan25, line 29: `vg_m` not needed here

docs/src/APIs/Soil.md
```diff
@@ -57,6 +57,8 @@ ClimaLSM.Soil.FreeDrainage
 ClimaLSM.Soil.AtmosDrivenFluxBC
 ClimaLSM.Soil.AbstractSoilSource
 ClimaLSM.Soil.PhaseChange
+ClimaLSM.Soil.vanGenuchten
```
juliasloan25, line 60: I think `AbstractSoilHydrologyClosure` should be added here too

juliasloan25 (commented):
Looks good! I had a few comments but overall I think this is a nice improvement.

I was also just wondering if we test matric potential/K/etc functions with both `S > 1` and `S < 1`. Since the functionality isn't changed here (except for the Brooks Corey extension, which has a test), we shouldn't need to add more tests, but I wasn't sure if we explicitly test behavior around the discontinuity.

src/Soil/soil_heat_parameterizations.jl
kmdeck: ok! I added the abstract type as annotation

src/Soil/soil_heat_parameterizations.jl
```diff
@@ -49,8 +50,9 @@ function phase_change_source(
...
 ) where {FT}
-    @unpack ν, vg_α, vg_n, vg_m, θ_r, earth_param_set = params
+    @unpack ν, θ_r, earth_param_set = params
```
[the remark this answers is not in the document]
kmdeck: oh, this is a better idea. will do!

src/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -6,22 +6,40 @@ export volumetric_liquid_fraction,
...
-"""
-function dψdϑ(ϑ, ν, θ_r, vg_α, vg_n, vg_m, S_s)
-    S = effective_saturation(ν, ϑ, θ_r)
-    if S < 1.0
-        return 1.0 / (vg_α * vg_m * vg_n) / (ν - θ_r) *
-               (S^(-1 / vg_m) - 1)^(1 / vg_n - 1) *
-               S^(-1 / vg_m - 1)
-    else
-        return 1.0 / S_s
+struct vanGenuchten{FT} <: AbstractSoilHydrologyClosure{FT}
```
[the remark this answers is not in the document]
kmdeck: good call!

src/Soil/soil_hydrology_parameterizations.jl
kmdeck: I switched it because I realized that we were actually returning the derivative of the pressure head (I think the matric potential only refers to the suction generated when the soil is unsaturated, while our expression holds when the soil is also saturated. But looking at the design docs, we use psi for pressure head (and psi_m for matric potential). I'll keep it as psi; the doc string is correct and this matches our docs.

docs/src/APIs/Soil.md
kmdeck: added it!

src/Soil/soil_hydrology_parameterizations.jl
kmdeck: good idea!

kmdeck:
> Looks good! I had a few comments but overall I think this is a nice improvement.
> 
> I was also just wondering if we test matric potential/K/etc functions with both `S > 1` and `S < 1`. Since the functionality isn't changed here (except for the Brooks Corey extension, which has a test), we shouldn't need to add more tests, but I wasn't sure if we explicitly test behavior around the discontinuity.

hey! we do have those tests but they are a little hidden (we test the different methods with an array of values for S which lie on each side & at the boundary, but dont note that it is testing both sides). I added a test for dpsitheta for vG though, as we were missing that!

juliasloan25 (approved):
Just a few minor comments left, but other than that this looks great :)

test/Soil/soil_parameterizations.jl
kmdeck: I added a test for S_c, but I think only testing m and S_c (and just S_c) for Brooks Corey is OK. I added these in particular because they are derived parameters in the constructor, not ones that are passed in.

src/Soil/energy_hydrology.jl
```diff
@@ -7,7 +7,7 @@ form, we assume the conductivity and volumetric heat capacity
 of the soil are constant.
 $(DocStringExtensions.FIELDS)
 """
-struct EnergyHydrologyParameters{FT <: AbstractFloat, PSE}
+struct EnergyHydrologyParameters{FT <: AbstractFloat, C, PSE}
```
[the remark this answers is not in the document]
kmdeck: I added this here and in the RichardsParameters (both in the struct def and the constructor where {} statement)

test/LSM/pond_soil_lsm.jl
kmdeck: We could. what about doing it for Float32 and making sure the type of the output is Float32? Then we know there is no promotion. 

since this isnt related to this PR, I think I will leave alone for now. But Ill open an issue to revisit the unit tests and make sure they are testing the different FT capability comprehensively.

test/Soil/soil_parameterizations.jl
charleskawczynski: I think I may have accidentally unresolved this? I was scrolling on my phone, and it happened so fast.

PR #215, opened 2023-05-24
Vcmax optimality model

eviatarbach:
## Content

----

### DIFF
--- a/src/Vegetation/optimality_farquhar.jl
+++ b/src/Vegetation/optimality_farquhar.jl
@@ -1,34 +1,15 @@
-export FarquharParameters, FarquharModel, C3, C4
-
-abstract type AbstractPhotosynthesisMechanism end
-"""
-    C3 <: AbstractPhotosynthesisMechanism
-
-Helper struct for dispatching between C3 and C4 photosynthesis.
-"""
-struct C3 <: AbstractPhotosynthesisMechanism end
-
-"""
-    C4 <: AbstractPhotosynthesisMechanism
-
-Helper struct for dispatching between C3 and C4 photosynthesis.
-"""
-struct C4 <: AbstractPhotosynthesisMechanism end
-
-abstract type AbstractPhotosynthesisModel{FT} <: AbstractCanopyComponent{FT} end
+export OptimalityFarquharParameters, OptimalityFarquharModel
 
 
 """
-    FarquharParameters{FT<:AbstractFloat}
+    OptimalityFarquharParameters{FT<:AbstractFloat}
 
-The required parameters for the Farquhar photosynthesis model.
+The required parameters for the optimality Farquhar photosynthesis model.
 $(DocStringExtensions.FIELDS)
 """
-struct FarquharParameters{FT <: AbstractFloat}
+struct OptimalityFarquharParameters{FT <: AbstractFloat}
     "Photosynthesis mechanism: C3 or C4"
     mechanism::AbstractPhotosynthesisMechanism
-    "Vcmax (mol/m^2/s) at 25 °C"
-    Vcmax25::FT
     "Γstar at 25 °C (mol/mol)"
     Γstar25::FT
     "Michaelis-Menten parameter for CO2 at 25 °C (mol/mol)"
@@ -49,7 +30,7 @@ struct FarquharParameters{FT <: AbstractFloat}
     ΔHRd::FT
     "Reference temperature equal to 25 degrees Celsius (K)"
     To::FT
-    "Intercelluar O2 concentration (mol/mol); taken to be constant"
+    "Intercellular O2 concentration (mol/mol); taken to be constant"
     oi::FT
     "Quantum yield of photosystem II (Bernacchi, 2003; unitless)"
     ϕ::FT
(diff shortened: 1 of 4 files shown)


src/Vegetation/photosynthesis.jl
```diff
@@ -192,3 +221,54 @@ function compute_photosynthesis(
...
+    Kc = MM_Kc(Kc25, ΔHkc, T, To, R)
+    Ko = MM_Ko(Ko25, ΔHko, T, To, R)
+
+    # Light utilization of APAR
+    IPSII = ϕ * APAR / 2
+
+    Jmax_arr = arrhenius_function(T, To, R, ΔHJmax)
+    Vcmax_arr = arrhenius_function(T, To, R, ΔHVcmax)
+
+    if typeof(mechanism) == C3
```
kmdeck, line 265: Im not sure if this would be confusing, but these lines can be reduced to:
```
FT = typeof(jv_ratio)
C = rubisco_assimilation(mechanism, FT(1), ci, Γstar, Kc, Ko, oi) * Vcmax_arr/Jmax_arr/jv_ratio
```

src/Vegetation/photosynthesis.jl
```diff
@@ -192,3 +221,54 @@ function compute_photosynthesis(
...
+        ΔHJmax,
+        ΔHVcmax,
+        ΔHΓstar,
+        f,
+        ΔHRd,
+        To,
+        θj,
+        ϕ,
+        mechanism,
+        sc,
```
kmdeck, line 244: can only unpack the parameters that you need in this function

src/Vegetation/photosynthesis.jl
```diff
@@ -192,3 +221,54 @@ function compute_photosynthesis(
...
+    Jmax_arr = arrhenius_function(T, To, R, ΔHJmax)
+    Vcmax_arr = arrhenius_function(T, To, R, ΔHVcmax)
+
+    if typeof(mechanism) == C3
+        C = Vcmax_arr/Jmax_arr/jv_ratio*(4*(ci + 2*Γstar)/(ci + Kc*(1 + oi/Ko)))
+    elseif typeof(mechanism) == C4
+        C = Vcmax_arr/Jmax_arr/jv_ratio
+    end
+
+    Jmax = (C - 1)*IPSII/(C*(C*θj - 1))
```
kmdeck, line 271: Is this the inverse of https://github.com/CliMA/ClimaLSM.jl/blob/28e5083c1678192357b209cfc78fc832f156c132/src/Vegetation/canopy_parameterizations.jl#L193
?

src/Vegetation/photosynthesis.jl
```diff
@@ -192,3 +221,54 @@ function compute_photosynthesis(
...
+        ψc,
+        oi,
+        Kc25,
+        Ko25,
+        ΔHkc,
+        ΔHko,
+        jv_ratio,
+    ) = parameters
+
+    Γstar = co2_compensation(Γstar25, ΔHΓstar, T, To, R)
```
kmdeck, line 254: Seeing that you also need to compute these here, I am almost leaning towards having a FarquharModel and an OptimalityFarquharModel, with two methods for `compute_photosynthesis`. 

Perhaps we can discuss with Renato what would make the most sense to him, since he has experience with these models. maybe next week?

src/Vegetation/photosynthesis.jl
eviatarbach: Good point! But yes, I think it may be confusing.

src/Vegetation/photosynthesis.jl
eviatarbach: Almost, but not exactly. It is what you get if you set `Ac = Aj`, which for C3 means  `J * (ci - Γstar) / (4 * (ci + 2 * Γstar)) = Vcmax * (ci - Γstar) / (ci + Kc * (1 + oi / Ko))`. So `Jmax` is on both sides of the equation, in `J` on one side and implicitly in `Vcmax` on the other.

kmdeck:
@eviatarbach how is this going? would you like to meet and pair program a bit? As I recall, we were adding a test that Ac = Aj but that one failed.

eviatarbach:
I found the issue: there is no solution for Ac=Aj for certain values of the parameters. What do you think we should do on this case? Maybe have a default fallback for Jmax?

kmdeck:
> I found the issue: there is no solution for Ac=Aj for certain values of the parameters. What do you think we should do on this case? Maybe have a default fallback for Jmax?

is it a finite/small set of edge cases or is it a wide swath of parameter space? 

having a default sounds reasonable.

src/Vegetation/optimality_farquhar.jl
```diff
@@ -0,0 +1,200 @@
...
+    Kc25 = FT(4.049e-4),
+    Ko25 = FT(0.2874),
+    To = FT(298.15),
+    ΔHkc = FT(79430),
+    ΔHko = FT(36380),
+    ΔHVcmax = FT(58520),
+    ΔHΓstar = FT(37830),
+    ΔHJmax = FT(43540),
+    ΔHRd = FT(46390),
+    jv_ratio = FT(1.67),
```
kmdeck, line 90: can remove since this will be calculated
can add their `c = 0.053...` param.
can add params you want to calibrate

src/Vegetation/optimality_farquhar.jl
```diff
@@ -0,0 +1,200 @@
...
+        C = Vcmax_over_Jmax*(4*(ci + 2*Γstar)/(ci + Kc*(1 + oi/Ko)))
+    elseif typeof(mechanism) == C4
+        C = Vcmax_over_Jmax
+    end
+
+    Jmax = (C - 1)*IPSII/(C*(C*θj - 1))
+    J = electron_transport(APAR, Jmax, θj, ϕ)
+    Jmax25 = Jmax/Jmax_arr
+    Vcmax25 = compute_Vcmax25(Jmax25, jv_ratio)
+    Vcmax = Vcmax_over_Jmax*Jmax
```
kmdeck, line 194: this section is the part that could be changed?

eviatarbach:
Pending resolution of https://github.com/SmithEcophysLab/optimal_vcmax_R/issues/3.

kmdeck:
@eviatarbach is there a way to get this to a point where we can merge it? I see that we are waiting on the resolution of an issue in the R code, but Im wondering if we can merge something that matches the R code, and then open an issue identifying what we need to fix at a later date.

PR #218, opened 2023-05-26
Use ClimaTimeSteppers, add implicit helper functions

### LINKED ISSUE #212: Use CTS; update dss usage
Moving forward, we want to use ClimaTimeSteppers timestepping algorithms instead of OrdinaryDiffEq. Beyond changing `ODEProblem` calls to use CTS, we also have to update how we use dss.

When we use ClimaTimeSteppers, we need to move dss out of our tendencies and pass it as an argument to the `CTS.ClimaODEFunction` instead. We also need to update the interface of our dss function to match what is expected by CTS.

Previously, we have performed dss on the tendency (`dY`) within tendency functions (e.g. in [energy_hydrology.jl](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/Soil/energy_hydrology.jl#L247)). ClimaAtmos has experienced that applying dss to the tendency (`dY`) can eventually lead to instabilities, so we should instead apply dss to the state (`Y`) instead. However, with the update to CTS, we just pass them our dss function without having to call it manually, so we don't really need to worry about this.

Additionally, we've been using `DiffEqCallback.SavingCallback` to save the state of our cache variable `p` in some experiments. This callback isn't supported by CTS, so we'll have to implement a similar one ourselves and use that instead.

This issue and #224 replace #205.

Closed by #218.

Part of SDI #135.

## To Do
- [x] Update `ClimaLSM.dss!` function to match CTS interface
  - [x] Add `dss_helper!` function
- [x] Update `ODEProblem` calls to use `CTS.ClimaODEFunction` constructors
  - [x] Make sure all soil models pass `dss!` to the `ClimaODEFunction`
- [x] Remove dss calls from tendency functions
  - [x] RichardsModel, EnergyHydrology
- [x] Implement Callback since `DiffEqCallback.SavingCallback` and `ContinuousCallback` aren't supported by CTS

juliasloan25:
## Purpose 
Use ClimaTimeSteppers algorithms and ODE functions, adapt dss functions, implement new callback function.

Note that since we aren't using adaptive timestepping with CTS (but we were with ODE), we decreased some timesteps to avoid numerical instability.

Closes #212 

Note: This PR increases run time because adaptive timestepping is no longer being used. This should be mitigated soon when we change `RichardsModel` runs to use implicit timestepping. If needed, we could also open an issue to test out larger `dt`s to decrease simulation time.

----


docs/Project.toml
```diff
@@ -2,6 +2,8 @@
 ArtifactWrappers = "a14bc488-3040-4b00-9dc1-f6467924858a"
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
+ClimaLSM = "7884a58f-fab6-4fd0-82bb-ecfedb2d8430"
```
kmdeck, line 5: do we need to also check in the Manifest for docs now? Because we need the dev version of ClimaLSM, not the latest registered version?

src/SharedUtilities/implicit_functions.jl
```diff
@@ -0,0 +1,125 @@
...
+    LinearAlgebra.ldiv!(A.temp2, A, A.temp1)
+    x .= A.temp2
+end
+
+function LinearAlgebra.ldiv!(
+    x::Fields.FieldVector,
+    A::AbstractTridiagonalW,
+    b::Fields.FieldVector,
+)
+    _ldiv!(x, A, b, axes(x.soil.ϑ_l))
```
kmdeck, line 28: Perhaps better to move into Soil since this has hardcoded theta variables?

Im not totally sure how this will generalize to a case with more prognostic variables. I guess it will be handled by ClimaTimesteppers, but as we continue to develop here, prior to this being moved to CTS, Im not sure what is best. For example, if we start testing the full soil implicit problem, or even the richards+canopy problem (canopy prognostic variables present), Im not sure if this will run.

src/SharedUtilities/implicit_functions.jl
```diff
@@ -0,0 +1,125 @@
...
+# related issue that requires the same workaround.
+function LinearAlgebra.ldiv!(x, A::AbstractTridiagonalW, b)
+    A.temp1 .= b
+    LinearAlgebra.ldiv!(A.temp2, A, A.temp1)
+    x .= A.temp2
+end
+
+function LinearAlgebra.ldiv!(
+    x::Fields.FieldVector,
+    A::AbstractTridiagonalW,
```
kmdeck, line 25: maybe this could be RichardsTridiagonalW? not sure :)

src/SharedUtilities/utils.jl
```diff
@@ -17,3 +19,156 @@ end
...
+    field::ClimaCore.Fields.Field,
+    domain::Union{
+        ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace,
+        ClimaCore.Spaces.AbstractSpectralElementSpace,
+    },
+)
+    Spaces.weighted_dss!(field)
+end
+
+"""
```
kmdeck, line 83: should we have unit tests for these different methods?

src/SharedUtilities/utils.jl
```diff
@@ -17,3 +19,156 @@ end
...
+
+Method of `dss_helper!` which unpacks properties of Y when on a
+domain that is 2-dimensional in the horizontal.
+
+The assumption is that Y contains FieldVectors which themselves contain either
+FieldVectors or Fields, and that the final unpacked variable is a Field.
+This method is invoked when the current property itself contains additional
+property(ies).
+"""
+function dss_helper!(field_vec::ClimaCore.Fields.FieldVector, _)
```
kmdeck, line 51: unit test to make sure recursion works?

src/SharedUtilities/utils.jl
```diff
@@ -17,3 +19,156 @@ end
...
+Constructs a DiscreteCallback which saves the time and cache `p` at each time
+specified by `saveat`. The times in `saveat` must be times the simulation is
+evaluated at for this function to work. The first argument must be a named
+tuple containing `t` and `saveval`, each having the same length as `saveat`.
+
+Note that unlike DiffEqCallbacks's SavingCallback, this version does not
+interpolate if a time in saveat is not a multiple of our timestep. This
+function also doesn't work with adaptive timestepping.
+"""
+function NonInterpSavingCallback(saved_values, saveat::Vector{FT}) where {FT}
```
kmdeck, line 163: perhaps we also need a test of this?

Project.toml
```diff
@@ -9,7 +9,9 @@ CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
 ClimaCoreTempestRemap = "d934ef94-cdd4-4710-83d6-720549644b70"
+ClimaTimeSteppers = "595c0a79-7f3d-439a-bc5a-b232dc3bde79"
```
kmdeck, line 12: a mental note that once we move the RK4 tableau to CTS, I think we can remove this dependency.

for now, we just need to add compat entries to keep the Julia registry happy

src/SharedUtilities/implicit_functions.jl
juliasloan25: Hmm this is a good point. Maybe we could recursively unpack the properties of `x` like we do in the new `dss!` implementation. Then we could choose one of the properties and dispatch off of its space for the call to `_ldiv!`.

src/SharedUtilities/implicit_functions.jl
juliasloan25: (related to your comment about hardcoded theta variables in `ldiv!` too)
Maybe we could have `ldiv!` dispatch off of concrete TridiagonalW types (so `RichardsTridiagonalW` here). Then we'll implement one `ldiv!` function for each model (which isn't great), but we would be able to pass the property to access from `x` and `b` (so `soil.\vartheta_l` here) to `_ldiv!/ldiv_serial!`, and keep `_ldiv!`  and `ldiv_serial!` general across all `AbstractTridiagonalW` types.

src/SharedUtilities/utils.jl
juliasloan25: That's probably a good idea. Is there a simple setup that will allow us to tell when dss is performed vs not performed?

kmdeck:
general comment: the plots look good in the docs & in on buildkite except for the model_tutorial.jl. I can help with this one

docs/tutorials/Bucket/bucket_tutorial.jl
```diff
@@ -180,7 +182,10 @@ z_0b = FT(1e-3);
 # Thermal parameters of soil
 κ_soil = FT(0.7);
 ρc_soil = FT(2e6);
-Δt = FT(3600.0);
+# Simulation start time, end time, and timestep
+t0 = FT(0.0);
+tf = FT(7 * 86400);
+Δt = FT(360.0);
```
kmdeck, line 188: is this another case where we need a smaller timestep because we arent using adaptive stepping? or could we keep the timestep the same?

docs/tutorials/Soil/richards_equation.jl
```diff
@@ -148,9 +156,35 @@ t0 = FT(0)
 timeend = FT(60 * 60 * 24 * 36)
 dt = FT(100);
 
+# Now, we choose the timestepping algorithm we want to use.
```
kmdeck, line 159: do we want to leave all the richards examples as using the implicit solver (though currently imp_tendency does nothing)?

might be confusing, but also might be worth it since we'll follow up with the next PR & then this will already be prepped

docs/tutorials/Soil/richards_equation.jl
```diff
@@ -148,9 +156,35 @@ t0 = FT(0)
...
+# We'll use the ARS111 algorithm with a maximum of 1 Newton iteration
+# per timestep.
+stepper = CTS.ARS111();
+ode_algo = CTS.IMEXAlgorithm(
+    stepper,
+    CTS.NewtonsMethod(
+        max_iters = 1,
```
kmdeck, line 166: just a note that when we use the implicit solver, we can use a conv checker and max_iters > 1

src/SharedUtilities/implicit_functions.jl
kmdeck: sure! does it make sense to do this in a different PR? maybe it will be clearer what to do are running the implicit solver?

src/SharedUtilities/utils.jl
kmdeck: hm. I dont know, but we could make a field, and compute dss on it, and save that value. and then compare after running this with the saved value?

kmdeck:
@juliasloan25  I will fix up the model_tutorial.jl today, and try and debug what is going on with the Bonan sand comparison too.

docs/Project.toml
juliasloan25: Hmm, I agree that if ClimaLSM is in the Project, we should check in the Manifest. But I'm confused why ClimaLSM isn't already in this Project before this PR. On main, there are files in docs/ that use ClimaLSM (e.g. [bucket_tutorial](https://github.com/CliMA/ClimaLSM.jl/blob/main/docs/tutorials/Bucket/bucket_tutorial.jl#L140))

kmdeck: I think that it is a mix of using the registered version and things like this:
https://github.com/CliMA/ClimaLSM.jl/blob/c9a04b631b7ad01c73609a533286e30ff787d8d9/docs/tutorials/Soil/richards_equation.jl#L64

the LOADPATH command lets you use the current version, because when we build the docs, we build from ClimaLSM.jl. We used to do this for test and for experiments too. But it was confusing.

kmdeck: /is confusing :)

juliasloan25: Thank you!

src/SharedUtilities/implicit_functions.jl
juliasloan25: Yeah, I think doing this in a different PR is a good idea. I'll open an issue and we can address it soon, maybe once we start using the implicit solver for RRE.

src/SharedUtilities/implicit_functions.jl
juliasloan25: This is a potential solution to #221 - @kmdeck are you okay with addressing this in the PR for that issue?

Project.toml
juliasloan25: CTS dependency removal noted in #219

docs/tutorials/Bucket/bucket_tutorial.jl
juliasloan25: This was using adaptive timestepping before since that's the default for `ODE.solve`, so I do think we need a smaller timestep.

docs/tutorials/Soil/richards_equation.jl
juliasloan25: Discussed offline - While using fewer iterations is better for performance, it also increases the likelihood of larger errors in both our solution values and mass conservation. We'll look into this more later on, but we should test different convergence conditions and find one that gives us reasonable RMSE and mass conservation error. See #222

src/SharedUtilities/implicit_functions.jl
juliasloan25: Discussed offline - This will be addressed in the PR linked to #221

docs/tutorials/Bucket/bucket_tutorial.jl
juliasloan25: Actually, I just tried running this with the same timestep as before (3600s) and it runs locally, so I'll change it back to that.

docs/tutorials/Soil/richards_equation.jl
juliasloan25: I changed the `RichardsModel` examples back to use just the explicit tendency, which I think is more clear

src/SharedUtilities/utils.jl
juliasloan25: I'll add one to this PR

src/SharedUtilities/utils.jl
juliasloan25: Okay, I added a test for dss in this PR

src/SharedUtilities/utils.jl
juliasloan25: I added a test for dss which also checks that the recursion works

kmdeck (approved):
looks great!! Nice work, Julia.

PR #220, opened 2023-06-05
restructure

AlexisRenchon:
## Purpose 

The purpose of this PR is to restructure ClimaLSM, as the codebase becomes more complexe, we need a better organised structure for clarity and simplicity. 

## To-do

What needs to be done is quite simple:
- [ ] create new subfolders
- [ ] update path in code (paths will change with new folders/subfolders structure

## Content

For the first commit, a first attempt of new folder structure in /src 

----


AlexisRenchon:
@kmdeck @braghiere 
This PR is not ready but I figured it'd be a good place to start. 
/src looks like this (standalone and integrated):
image
/src/standalone looks like:
image

We could then think of subfolder at more level... I think this PR should be a long WIP - let's take the time to discuss of the best possible structure! 

Happy to invite more people in the discussion - to add in this PR

kmdeck:
what do you feel like the pros and cons are to the current idea you show? It makes a lot of sense to me...

AlexisRenchon:
For now all I suggest is having `standalone` and `integrated` folders inside /src. (note that this is the naming we have in ClimaLSM README.md)
I think this is definitely needed! /src is quite messy at the moment (6 folders, 6 files), and will become more and more crowded... It will need a restructuration eventually, and it is easier to do it now that later (always easier to do smaller PRs). The new structure has 3 folders, 3 files in /src, and that probably won't change for a while (maybe we will need more for global run?). 

In general, I think structure becomes a personal preference at some point: is it better to have few big files, or many small files, less or more subfolders... 
But good organization makes it easier to introduce/explain, or navigate. 

Having many folders is a bit a similar pro / cons than our nested variable names (like you explain in the docs `"Even in standalone model, variables are accessed in a nested way: Y.soil, p.soil, etc, which is excessive."` but I personally find it clearer.

kmdeck:
> For now all I suggest is having `standalone` and `integrated` folders inside /src. (note that this is the naming we have in ClimaLSM README.md) I think this is definitely needed! /src is quite messy at the moment (6 folders, 6 files), and will become more and more crowded... It will need a restructuration eventually, and it is easier to do it now rather than later (always easier to do smaller PRs). The new structure has 3 folders, 3 files in /src, and that probably won't change for a while (maybe we will need more for global run?).
> 
> In general, I think structure becomes a personal preference at some point: is it better to have few big files, or many small files, less or more subfolders... But good organization makes it easier to introduce/explain, or navigate.
> 
> Having many folders is a bit a similar pro / cons than our nested variable names (like you explain in the docs `"Even in standalone model, variables are accessed in a nested way: Y.soil, p.soil, etc, which is excessive."` but I personally find it clearer.

This sounds good, Alexis. We could start by organizing the src directory, like you wrote (and it already matches the structure in experiments). Let's ask for feedback and incorporate it, then I think we can go ahead and do this.

Then we can restructure docs (we also need to add tutorials and docs for a number of items - so maybe we could have a docs week where we do all of these things)

juliasloan25:
This looks good to me! I think having separate src/Standalone/ and src/Integrated/ folders would be more clear.

On the topic of refactoring/code structure, I noticed that each of the standalone models has its own module, but the integrated ones don't. Is there a reason for this? If not, it might be nice to make them all modules. This would ensure that each of the models has its own namespace, which would prevent conflicts when defining functions/objects with the same name across modules. It also may be a bit faster at runtime since modules can be precompiled.

AlexisRenchon:
I like @juliasloan25 idea of having modules for integrated models!

kmdeck:
closing because we will instead merge #249

PR #238, opened 2023-06-16
Zero lai

kmdeck:
## Purpose 
In the current codebase, setting the area indices of the plant to zero would result in a 0/0, because, for example, we have:
` d theta_leaf/dt = [stem_flux *LAI - canopy_transpiration]/ (LAI x Delta z)`, where canopy_transpiration involves a multiplication by LAI as well. 

Because we cannot have a changing number of prognostic variables across the domain, even when there is no vegetation present, we need to have a canopy model with some prognostic variables, parameters, etc. To turn off the canopy, setting LAI = SAI = RAI = 0 will imply zero fluxes and zero root extraction, and no change in the initial conditions of the plant prognostic variables. 

Therefore, update the codebase so that when LAI=SAI=RAI = 0, `d theta/ dt = 0` and there is no divide by zero.

Once we get closer to global runs, we should do some refactoring as it makes sense to make it so you dont need to define all these parameters where there is no vegetation. See Issue #250.

## Content
1. Change "LAI/(LAI x dz)" terms to "LAI/max(LAI x dz, eps(FT))". When LAI = 0, this gives zero flux and does not divide by zero.
2. Add consistency checks to the PlantHydraulics constructor:
- if LAI == 0, all area indices should be zero (no plant present)
- if LAI is nonzero, RAI must be nonzero (a plant cannot exist without roots)
- if n_stem == 0, the stem area index must be zero for consistency.
3. Adds unit tests to make sure consistency checks pass and that when LAI =RAI = SAI = 0, the root extraction and transpiration are both zero.

Review checklist

I have:

----

### DIFF
--- a/src/Vegetation/PlantHydraulics.jl
+++ b/src/Vegetation/PlantHydraulics.jl
@@ -107,24 +107,74 @@ function PlantHydraulicsParameters(;
     )
 end
 
+"""
+    lai_consistency_check(
+        n_stem::Int64,
+        n_leaf::Int64,
+        area_index::NamedTuple{(:root, :stem, :leaf), Tuple{FT, FT, FT}},
+    ) where {FT}
+
+Carries out consistency checks using the area indices supplied and the number of
+stem elements being modeled.
+
+Note that it is possible to have a plant with no stem compartments
+but with leaf compartments, and that a plant must have leaf compartments
+(even if LAI = 0).
+
+Specifically, this checks that:
+1. n_leaf > 0
+2. if LAI is nonzero or SAI is nonzero, RAI must be nonzero.
+3. if SAI > 0, n_stem must be > 0 for consistency. If SAI == 0, n_stem must
+be zero.
+"""
+function lai_consistency_check(
+    n_stem::Int64,
+    n_leaf::Int64,
+    area_index::NamedTuple{(:root, :stem, :leaf), Tuple{FT, FT, FT}},
+) where {FT}
+    @assert n_leaf > 0
+    if area_index[:leaf] > eps(FT) || area_index[:stem] > eps(FT)
+        @assert area_index[:root] > eps(FT)
+    end
+    # If there SAI is zero, there should be no stem compartment
+    if area_index[:stem] < eps(FT)
+        @assert n_stem == FT(0)
+    else
+        # if SAI is > 0, n_stem should be > 0 for consistency
+        @assert n_stem > 0
+    end
+
+end
 
 """
-    PlantHydraulicsModel{FT, PS, RE, T, B} <: AbstractPlantHydraulicsModel{FT}
+    PlantHydraulicsModel{FT, PS, RE, T} <: AbstractPlantHydraulicsModel{FT}
 
 Defines, and constructs instances of, the PlantHydraulicsModel type, which is used
 for simulation flux of water to/from soil, along roots of different depths,
 along a stem, to a leaf, and ultimately being lost from the system by
 transpiration.
 
-This model can be used in standalone mode by prescribing the transpiration rate
-and soil matric potential at the root tips or flux in the roots, or with a
-dynamic soil model using `ClimaLSM`.
+This model can also be combined with the soil model using ClimaLSM, in which
+case the prognostic soil water content is used to determine root extraction, and
+the transpiration is also computed diagnostically. In  global run with patches
+of bare soil, you can "turn off" the canopy model (to get zero root extraction, zero absorption and
+emission, zero transpiration and sensible heat flux from the canopy), by setting:
+- n_leaf = 1
+- n_stem = 0
+- LAI = SAI = RAI = 0.
+
+A plant model can have leaves but no stem, but not vice versa. If n_stem = 0, SAI must be zero.
+
+Finally, the model can be used in Canopy standalone mode by prescribing
+the soil matric potential at the root tips or flux in the roots. There is also the
+option (intendend only for debugging) to use a prescribed transpiration rate.
+
 $(DocStringExtensions.FIELDS)
 """
 struct PlantHydraulicsModel{FT, PS, RE, T} <: AbstractPlantHydraulicsModel{FT}
-    "The number of stem compartments for the plant"
+    "The number of stem compartments for the plant; can be zero"
     n_stem::Int64
-    "The number of leaf compartments for the plant"
+    "The number of leaf compartments for the plant; must be >=1"
     n_leaf::Int64
     "The depth of the root tips, in meters"
     root_depths::Vector{FT}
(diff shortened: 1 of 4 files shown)


gagnelandmanna (approved):

src/Vegetation/PlantHydraulics.jl
```diff
@@ -418,13 +464,17 @@ function make_compute_exp_tendency(model::PlantHydraulicsModel, _)
         n_leaf = model.n_leaf
         fa = p.canopy.hydraulics.fa
         fa_roots = p.canopy.hydraulics.fa_roots
-
+        FT = eltype(t)
         @inbounds for i in 1:(n_stem + n_leaf)
-            AIdz =
+            # if the area_index is zero, fa is zero.
```
juliasloan25, line 469: Is `fa` only for RAI or for all AIs?

test/Vegetation/plant_hydraulics_test.jl
```diff
@@ -12,6 +12,58 @@ using Insolation
 using Dates
 
 include(joinpath(pkgdir(ClimaLSM), "parameters", "create_parameters.jl"))
+@testset "LAI assertions" begin
```
juliasloan25, line 15: Nice job covering all the cases :)

juliasloan25 (commented):
Looks good! I had a few comments

src/Vegetation/PlantHydraulics.jl
```diff
@@ -108,6 +108,40 @@ function PlantHydraulicsParameters(;
...
+    n_stem::Int64,
+    n_leaf::Int64,
+    area_index::NamedTuple{(:root, :stem, :leaf), Tuple{FT, FT, FT}},
+) where {FT}
+    # If LAI is zero, all area indices should be zero (no plant present)
+    # If LAI is nonzero, there should be a nonzero RAI
+    # If no plant is present, do *not* assert that n_leaf = n_stem = 0, because
+    # we run with the same prognostic variables everywhere: turning the plant "on" vs "off"
+    # is done by setting LAI/RAI/SAI to zero vs not.
+    if area_index[:leaf] < eps(FT)
```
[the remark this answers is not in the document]
kmdeck: it is possible to have a plant with leaves but no stem (I think grass falls in this category, and the simplest big leaf would too). I left a note in the docs now that indicates this.

For the second point, yes, the plant must have leaves. I added the @assert n_leaf > 0 like so:
```
    if area_index[:leaf] < eps(FT)
        @assert all(values(area_index) .== 0)
        @assert n_leaf > 0
    else
        @assert area_index[:root] > eps(FT)
        @assert n_leaf > 0
    end
```
good catch! We also had checked that later on in the PlantHydraulicsModel constructor but moving it hear makes sense to me.

src/Vegetation/PlantHydraulics.jl
```diff
@@ -418,13 +464,17 @@ function make_compute_exp_tendency(model::PlantHydraulicsModel, _)
         n_leaf = model.n_leaf
         fa = p.canopy.hydraulics.fa
         fa_roots = p.canopy.hydraulics.fa_roots
-
+        FT = eltype(t)
         @inbounds for i in 1:(n_stem + n_leaf)
-            AIdz =
+            # if the area_index is zero, fa is zero.
+            # to  prevent dividing by zero, use a small number instead: 0/ϵ = 0
+            AIdz = max(
```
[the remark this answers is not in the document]
kmdeck: ok! I added this to the doc string for the function and as a comment in the code.

src/Vegetation/PlantHydraulics.jl
kmdeck: it is for all fluxes - so the root flux is multiplied by the mean of SAI and RAI, and the flux between the stem and leaves are multiplide by the mean of SAI and LAI, and then transpiration is proportional to LAI as well.

kmdeck: in the terminology of finite difference, the area indicies and fluxes live on faces, while theta is on cell centers..

test/Vegetation/plant_hydraulics_test.jl
kmdeck: :D thanks!

Ill add in the checks for the ones we just added.

kmdeck:
> Looks good! I had a few comments

ok! I addressed them all, I think. If the tests pass now, good to go?

src/Vegetation/PlantHydraulics.jl
```diff
@@ -116,15 +162,27 @@ for simulation flux of water to/from soil, along roots of different depths,
...
+This model can also be combined with the soil model using ClimaLSM, in which
+case the prognostic soil water content is used to determine root extraction, and
+the transpiration is also computed diagnostically. In  global run with patches
+of bare soil, you can "turn off" the canopy model (to get zero root extraction, zero absorption and
+emission, zero transpiration and sensible heat flux from the canopy), by setting:
+- n_leaf = 1
+- n_stem = 0
+- LAI = SAI = RAI = 0.
+
+A plant model can have leaves but no stem, but not vice versa. If n_stem = 0, SAI must be zero.
```
AlexisRenchon, line 174: Did you say above that it is actually possible to have stem but no leaves? (I imagine deciduous trees in winter for example).
 
Sorry I may be a bit confused here.

AlexisRenchon (approved):
Looks good to merge! 
I only had one small comment / question.

src/Vegetation/PlantHydraulics.jl
kmdeck: oh, good question. In the code right now, we require n_leaf > 0 (and then can set LAI to zero as needed if there is no plant). so we would not support n_stem > 0 but n_leaf = 0. 

I think for deciduous trees we would have n_leaf > 0 but then LAI would vary over time and be very low in the winter. 

Is it possible to have zero LAI but nonzero SAI/RAI? What happens then? are there any fluxes (energy, water, carbon) between the plant and the atmosphere? I think no. and then would there be root extraction still? Right now in this PR we assert that if LAI = 0, SAI = RAI = 0 also.

AlexisRenchon: I don't know how it is done in models, but zero LAI with nonzero SAI/RAI definitely happens in winter! 

Photosynthesis is 0, but respiration is not 0 (it is reduced, but not 0). 

Not sure about water - obviously 0 for leaves since no leaves, but maybe reduced in roots (close to 0 but there can be some movement)

About energy, I imagine there is still sensible heat flux, perhaps even latent if trees freeze? (but I imagine that's rare?) 

Not sure what we should do.

kmdeck: ok, well, all those fluxes when they get defined will have to be defined so that they have the right behavior. But I guess that is not in the realm of this PR's goals. For now, are these "rules" OK?

1. n_leaf > 0 [you cant have a plant model with a leaf compartment. 
2.  if n_stem > 0, SAI > 0; if n_stem = 0, SAI =0 [ you can have a plant with only roots and leaves, like grass]
3. if LAI > 0, RAI must be > 0
No check required for the following:
4. you can have LAI zero but SAI and RAI nonzero - e.g. deciduous trees in winter. right now these would have zero transpiration but possibly some water flux between roots/stem and soil.
5. If we want to "turn off" the canopy, we set RAI = SAI = LAI = 0, but at a minimum, the leaf compartment is always there.
Is that ok @juliasloan25 @AlexisRenchon?

kmdeck: [note - I realized that this LAI consistency check function will have to be refactored once we have LAI read in from a file or varying in time, because we wont be passing the values in as fixed float parameters this way....so im sort of on the fence about if we even need to include this function? But perhaps it is good to include it for now to codify these rules and then we can adapt as needed when we refactor]

AlexisRenchon: that's ok with me =)

PR #236, opened 2023-06-15
Add some aqua tests

charleskawczynski:
This PR adds some aqua tests. @juliasloan25, can you follow through on fixing the following?

 - Unbound argument methods
 - `test_stale_deps`
 - `test_undefined_exports`
 - `test_deps_compat`

If you can fix the ambiguities, great, if not, don't worry, they're not always trivial, and sometimes it's just a julia issue.

PR #246, opened 2023-06-22
Fix aqua tests

### LINKED ISSUE #237: Fix aqua tests
#236 introduces some aqua tests for quality assurance. Some of the tests are failing, so we need to go through and figure out how to fix the code to make sure they pass.

**Components**
- Unbound argument methods
- Method ambiguities
- `test_stale_deps`
- `test_undefined_exports`
- `test_deps_compat`

**Additional context**
[ClimaCore.jl aqua tests](https://github.com/CliMA/ClimaCore.jl/pull/1324/files)

juliasloan25:
## Purpose 
Closes #237 
Follow-up for #236 

## To-do
- [x] Unbound argument methods
- [x] Method ambiguities
- [x] test_undefined_exports
- [x] test_stale_deps
- [x] test_deps_compat

----


src/SharedUtilities/models.jl
```diff
@@ -226,7 +226,7 @@ have different dimensions - require defining a new method.
 """
 function initialize_prognostic(
     model::AbstractModel{FT},
-    state::Union{ClimaCore.Fields.Field, Vector{FT}},
+    state::Union{ClimaCore.Fields.Field, ClimaCore.Fields.FieldVector, Vector},
```
charleskawczynski, line 229: This change will result in the `detect_unbound_args_recursively` test to fail-- `FT` is removed from the argument type parameters, but still exists in the `where {FT}` part

juliasloan25: Oops, I didn't mean to remove the `{FT}`. I'll add it back.

src/SharedUtilities/models.jl
```diff
@@ -226,7 +226,7 @@ have different dimensions - require defining a new method.
 """
 function initialize_prognostic(
     model::AbstractModel{FT},
-    state::Union{ClimaCore.Fields.Field, Vector{FT}},
+    state::Union{ClimaCore.Fields.Field, ClimaCore.Fields.FieldVector, Vector{FT}},
```
kmdeck, line 229: just curious - do we use this function with state of type FieldVector? ditto for initialize_auxiliary

src/SharedUtilities/models.jl
juliasloan25: Wait, just to make sure I understand `detect_unbound_args_recursively`, would we expect it to fail in this case even though we have `AbstractModel{FT}` with the `where {FT}`? It looks like all the aqua tests pass for this commit

src/ClimaLSM.jl
```diff
@@ -74,7 +74,10 @@ function Domains.coordinates(model::AbstractLandModel)
     return coords
 end
 
-function initialize_prognostic(model::AbstractLandModel{FT}, coords) where {FT}
+function initialize_prognostic(
+    model::AbstractLandModel{FT},
+    coords::Union{ClimaCore.Fields.Field, ClimaCore.Fields.FieldVector},
```
kmdeck, line 79: ditto here - do we use this with a FieldVector as input?

src/SharedUtilities/models.jl
charleskawczynski: Oh, whoops! I missed the `AbstractModel{FT}`, so `FT` is always present (and the tests should pass)

src/ClimaLSM.jl
juliasloan25: `docs/tutorials/Usage/LSM_single_column_tutorial.jl` calls this function with parameter types `::AbstractLandModel, ::FieldVector`, using a `LandHydrologyModel`.  If I change this line to just `coords::ClimaCore.Fields.Field`, we get a method error when running that file

src/SharedUtilities/models.jl
juliasloan25: I don't think we use this combination of types, but we get a type error if I don't include `FieldVector` here. I think it's because it is a type option for the `AbstractLandModel` version of this function (the aqua test error goes away if I remove it in both places, but then I get the method error described below when running `LSM_single_column_tutorial"), but I'm not totally sure. @charleskawczynski do you have any thoughts?

charleskawczynski: Which aqua test is failing? Ambiguities?

juliasloan25: Yeah, I can push a version showing the exact ambiguity message

juliasloan25: Actually I'll just copy it here. This is with the current version of the code, but with `ClimaCore.Fields.FieldVector` removed from the second parameter of `initialize_prognostic` and `initialize_auxiliary` in this file only.

`julia> include("test/aqua.jl")

[ Info: Precompiling ClimaLSM [7884a58f-fab6-4fd0-82bb-ecfedb2d8430]

method_ambiguity = (initialize_auxiliary(model::ClimaLSM.AbstractLandModel{FT}, coords::Union{ClimaCore.Fields.Field, ClimaCore.Fields.FieldVector}) where FT @ ClimaLSM ~/clima/ClimaLSM.jl/src/ClimaLSM.jl:92, initialize_auxiliary(model::AbstractModel{FT}, state::Union{Vector{FT}, ClimaCore.Fields.Field}) where FT @ ClimaLSM ~/clima/ClimaLSM.jl/src/SharedUtilities/models.jl:253)

method_ambiguity = (initialize_prognostic(model::ClimaLSM.AbstractLandModel{FT}, coords::Union{ClimaCore.Fields.Field, ClimaCore.Fields.FieldVector}) where FT @ ClimaLSM ~/clima/ClimaLSM.jl/src/ClimaLSM.jl:77, initialize_prognostic(model::AbstractModel{FT}, state::Union{Vector{FT}, ClimaCore.Fields.Field}) where FT @ ClimaLSM ~/clima/ClimaLSM.jl/src/SharedUtilities/models.jl:227)

Aqua tests (performance): Test Failed at /Users/jsloan/clima/ClimaLSM.jl/test/aqua.jl:22
  Expression: length(ambs) == 0
   Evaluated: 2 == 0`

charleskawczynski: I think the issue is the union, you could probably match the two signatures and split the union, or just remove the argument types for both methods and let all of dispatch happen based on `model`, that might be simplest.

src/ClimaLSM.jl
kmdeck: oh ok! youre right! all good, sorry about that.

kmdeck (approved):

src/SharedUtilities/models.jl
juliasloan25: I think it makes most sense to just dispatch off of the `model` type, so I'll do that. @kmdeck is this okay with you?

kmdeck: yes! I think the pairing is 
LandModel -> FieldVector
AbstractModel -> Field or Vector?

juliasloan25: If I write the functions with these combinations of type annotations, I get method ambiguities for both `initialize_prognostic` and `initialize_auxiliary` (see above in this thread). Just to confirm, are you okay with the solution of not annotating the `state/coord` parameter and only annotating the `model` type as `AbstractLandModel` or `AbstractModel`?

kmdeck: Is a compromise using 
Abstract Land Model + FieldVector
Abstract Model + no type annotation?

Just wondering, otherwise let's make it clear in the docs that in one case we expect something like a named tuple (FieldVector), and in the other, something like an array (Field or Vector) and then remove the type annotation on the state/coord argument...

juliasloan25: Ahh yes, that's a great compromise.

In the docstrings for the AbstractModel methods, I'll note the types we expect for the second argument

PR #248, opened 2023-06-23
reconcile compats

### LINKED ISSUE #223: Reconcile duplicate compat entries
We currently have `compat` entries in our top level directory Project.toml as well as in docs/, experiments/, and test/ Projects. Package-level compats also apply to all downstream Manifests, so we should only have compats in the top level to reduce duplication and confusion.

**Proposed solution**
We can use [CompatDevTools.jl](https://github.com/charleskawczynski/CompatDevTools.jl) to reconcile duplicated compats and fix this issue.

juliasloan25:
## Purpose 
Closes #223

## To-do
- [x] Test with newest allowed versions
- [x] Test with oldest allowed versions

----


Project.toml
```diff
@@ -23,15 +23,15 @@ UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 [compat]
 ArtifactWrappers = "0.2"
 ClimaComms = "0.3, 0.4, 0.5"
-ClimaCore = "0.10"
-ClimaCoreTempestRemap = "0.3.5, 0.3.6, 0.3.7"
+ClimaCore = "0.10.32"
```
kmdeck, line 26: if this is reducing duplicates in compat entries in downstream folders, do we need to change the compat for the core Project.toml? Can you describe a little more how you got these versions?

docs/Manifest.toml
```diff
@@ -2,7 +2,7 @@
 
 julia_version = "1.9.1"
```
kmdeck, line 3: does this mean that if someone is not using 1.9.1, say, 1.9.0, they need to always update the Manifest to agree with what you'd get using julia 1.9.0?  and so every time someone makes a PR with a different version of Julia the Manifest will change?

this seems kind of frustrating to all developers if they arent using the same Julia version.

Project.toml
juliasloan25: I tried compiling ClimaLSM using ClimaCore@0.10.0 (which would be allowed if we just had `ClimaCore = "0.10"`), but I got an error that ClimaCore.PointSpace wasn't defined [here](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/ClimaLSM.jl#L9). I had a hard time finding exactly in which tag that function was added, but when I ran `add ClimaCore` with the oldest allowed versions of other packages, it added 0.10.32. I agree this isn't a very clear way to arrive at a version to specify, but it seems like we need something more recent than 0.10.0.

docs/Manifest.toml
juliasloan25: Hmm I'm not exactly sure how this works. I just switched to Julia v1.9.0. I was able to precompile everything, but it does give a warning that recommends to resolve the manifest. So I think you're right that in practice this probably means the Manifest Julia version Manifest will change anytime developers with different versions make a PR. I'm not sure what a good way around this is though, since the Manifest has to have some Julia version specified. Maybe we could all try to stay up to date with the latest version?

kmdeck (approved):
offline discussion of comments; looks good! we decided that whenever we update the Manifests for new version of Julia, we will also do a release of ClimaLSM

docs/Manifest.toml
juliasloan25: Discussed offline - Kat and I decided that a good approach may be to create a new release of ClimaLSM using the latest Julia version each time a Julia release is made. We should also make sure that the same julia version is used across Manifests

PR #249, opened 2023-06-26
Ar/restructure2

AlexisRenchon:
## Purpose 

This PR is the same as PR #220 but I recreated a branch from main as I was struggling a bit with git rebase... 
We can delete PR #220 after we merge this PR. 
In this PR, I created `/Standalone` and `/Integrated` folders in `/experiments`, `/src`, and `/test`. 
I also accordingly corrected all paths in `src/ClimaLSM.jl`. 
I believe it should be all set and ready to merge, but it is possible I missed something (a path may be broken, maybe?), we will see if CI passes. 
I also put `src/Regridder.jl` and `src/Parameters.jl` in `src/SharedUtilities`, even if it was already done in PR #245. - I hope that's okay. 
I hope it is ready to merge, let me know! 

## To-do

Just check if CI pass... this is just folder restructure, so it should be straightforward and ready to merge if paths have been corrected properly. 

## Content

Create `/Standalone` and `/Integrated` folder throughout the repository for clarity and organisation. 

----
- [x] I have read and checked the items on the review checklist.

PR #254, opened 2023-06-29
restructure with standalone and integrated folders

AlexisRenchon:
## Purpose 

Restructure repo with standalone and integrated folders (as PR #249 but in one commit)

## To-do

deletes duplicates PR after this is merged

## Content

restructure

----


AlexisRenchon:
Looks like CI passed. (in 1 commit)
Kat mentioned to double check that the docs look right and the buildkite tests produce plots that look correct compared to a successful previous run before merging. 

hopefully we can merge today!

PR #256, opened 2023-06-30
Fix docs broken links

AlexisRenchon:
## Purpose 

Fix broken (outdated) links in docs 

## To-do

Go through the docs, check all links, fix the ones that are broken

## Content

Fixed 1 link to start this PR

----

### DIFF
--- a/docs/src/folderstructure.md
+++ b/docs/src/folderstructure.md
@@ -2,7 +2,7 @@
 
 [ClimaLSM](https://github.com/CliMA/ClimaLSM.jl) home directory has 5 main folders:
 
-- docs: contains files to generate [the documentation website](clima.github.io/ClimaLSM.jl/).  
+- docs: contains files to generate [the documentation website](https://clima.github.io/ClimaLSM.jl/dev/).  
 - experiments: contains simple runs of `ClimaLSM` models. 
 - parameters: contains a file to retrieve constants such as avogadro's number, the speed of light, etc. 
 - src: contains the code of `ClimaLSM` models. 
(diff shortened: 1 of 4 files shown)


AlexisRenchon:
I believe some docs are outdated so it is hard to update links. 
For example, LSM_single_column_tutorial.jl at line [149](https://github.com/CliMA/ClimaLSM.jl/blob/ar/fixdocslinks/docs/tutorials/Usage/LSM_single_column_tutorial.jl#L149) refers to `src/standalone/Soil/Soil.jl` line 292 and 400, but Soil.jl only has 147 lines now.

AlexisRenchon:
Another TO DO to fix the docs is make it track version properly. 
It is currently broken. 
In the bottom left, version gives me option for `dev` or `stable` or `v0.2`
I don't know yet how to fix this... maybe we should ask somewhere?

juliasloan25:
Kat has suggested having a land team docs week to go through the docs and fix these sorts of problems that have come up. Maybe we can fix what we're able to easily now, then note the other issues and address them then. Or, if even what you're working on now is a pain, we could do all of the docs fixes during the docs week later on, and I can make a release of ClimaLSM today without the doc fixes.

juliasloan25 (commented):
This looks good to me! It seems that the `experiments` directory isn't precompiling. I had a similar issue in ClimaCoupler.jl today and just had to `up` all of the packages and re-resolve the Manifest.

kmdeck:
thanks for taking this on!  
- perhaps we can remove links to specific lines of code? Youre right that they are outdated, and it seems almost impossible to maintain this... I guess we could link back to the docs for `make_update_aux`  etc because the method for a SoilModel is listed there too?
- I noticed that these links dont really go anywhere: https://github.com/CliMA/ClimaLSM.jl/blob/e9482924a8f80897f51025b2b269d09b3c44267d/docs/tutorials/Bucket/coupled_bucket.jl#L70 because we moved the driver methods from the Bucket module to the shared utilities Drivers file [ surface_fluxes, net_radiation, liquid_precipitation, snow_precipitation, surface_air_density]. I think we can fix the links just by changing it from referring to Bucket.md to these ones: https://github.com/CliMA/ClimaLSM.jl/blob/main/docs/src/APIs/shared_utilities.md#drivers

PR #259, opened 2023-07-06
Prototype branch: LAI as a function of time

kmdeck:
## Purpose 
- Allow for LAI to be a function of time - and SAI/RAI? -> add LAI to the aux state.
- This PR is meant for site level runs, for example, at the ozark site, where we have LAI data throughout the year. But make this flexible enough to also work in 2d: LAI (lat, long, t).
- remove duplicate definitions of LAI in plant hydraulics parameters and in shared canopy parameters

## To-do
Currently, for met data, we fit spline functions to the data to get e.g. precip(t), radiation(t), and we can also do the same for lai(t). We then store the spline functions in structs and pass to the model. 

Im not sure how well this will work as a function of space (fitting splines as a function of lat/long/t). I think what Julia is working on is the better alternative: we have the parameters that vary in time and space as auxiliary variables, and update them via "update aux" each step when we read from the file and interpolate to the simulation time. 

This is described in https://github.com/CliMA/ClimaLSM.jl/issues/126.

I think we can already plan for Julia's work by making LAI a function of aux and updating it via "update aux", even if for now, in update aux we just call the spline function.

## Content

----

### DIFF
--- a/experiments/integrated/ozark/hydrology_only/ozark.jl
+++ b/experiments/integrated/ozark/hydrology_only/ozark.jl
@@ -102,7 +102,7 @@ photosynthesis_args = (;
     )
 )
 # Set up plant hydraulics
-area_index = (root = RAI, stem = SAI, leaf = LAI)
+area_index = (root = RAI, stem = SAI, leaf = max_LAI) # misleading placeholder - LAI value will be updated each timestep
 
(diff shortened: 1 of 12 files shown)


kmdeck:
closing in lieu of #263 which adds the area indices to the aux state and allows us to update them each step.

PR #260, opened 2023-07-06
Read albedo over time (v1)

### LINKED ISSUE #243: Time-varying albedo
We currently have the capability to use albedo read in from a file. However, the albedo is read only once from the supplied data file, and does not change over time. Here, we want to extend our implementation to read in albedo as it changes over time.

Part of SDI https://github.com/CliMA/ClimaCoupler.jl/issues/276.

Closed by #264, #280

We can follow the methods that ClimaCoupler.jl uses to read in SST and SIC from files that change over time. This will involve copying over ClimaCoupler's `BCReader` code, which should be moved to a shared location (as should its `Regridder` module, which has also been copied to ClimaLSM.jl.

We considered adding this functionality in the coupler instead of ClimaLSM, but ulitmately decided it makes more sense in the land model. This will allow us to use time-varying albedo in uncoupled runs of the land model independent of ClimaCoupler, which we may want to do with, e.g. the canopy model.

**To Do**
- [x] Copy BCReader module from coupler #264 
  - [x] Copy tests
  - [x] Modify source code and tests as needed for ClimaLSM (e.g. don't create CoupledSimulation in tests)
- [x] Add new albedo type in `src/Bucket.jl` - `BulkAlbedoTemporal` #280
  - [x] Use `BCFileInfo` as a backing data structure to store temporal datafile information
  - [x] Rename `BulkAlbedoMap` --> `BulkAlbedoStatic`
- [x] Add a method of `set_initial_parameter_field` for the new albedo type, which calls ~`update_midmonth_data`~ `read_data_fields` and ~`interpolate_midmonth_to_daily`~ `interpolate_data`. #280
  - This is based on the initialization of the sea ice data in the coupler's driver [here](https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/AMIP/modular/coupler_driver_modular.jl#L259-L260).
- [x] Add a `update_albedo` function, which will get called from `update_aux` to update `p.bucket.α_sfc` #280
  - [x] For `BulkAlbedoTemporal`, read in the next data point from the file and interpolate if necessary (see coupler drive equivalent [here](https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/AMIP/modular/coupler_driver_modular.jl#L416-L430))
  - [x] For other albedo types, return the albedo unchanged
- [x] Add tests for new methods #280

juliasloan25:
## Purpose 
Add in functionality to read in data files over time. This is largely copied from ClimaCoupler.jl's BCReader module.

Closes #243 (see for more info)

Broken down into smaller PRs: #264 (done), #280 (done)

----


juliasloan25:
replaced by #264, #280

PR #261, opened 2023-07-10
option to turn off lateral flow for the soil models

kmdeck:
## Purpose 
This adds the option to turn off the flow of heat and water (soil energy/hydrology) or of water (richards equation) even in the case of a 3d domain (hybrid box or spherical shell). 

## Content
1. Adds an additional field to the soil models (EnergyHydrology and RichardsModel)
2. Adds additional argument to `horizontal_components!` function
3. Adds a new method for that function
In summary, we now also use multiple dispatch on the value of a boolean flag in the
`horizontal_components!` function. This function now dispatches off of the type of the domain, value of the lateral_flow flag, and the model type as follows:
 - domain::Column -> no horizontal components, regardless of the lateral_flow flag value
 - domain::Union{HybridBox,SphericalShell}, lateral_flow::Val{false} -> no horizontal components, regardless of soil model type.
 - domain::Union{HybridBox,SphericalShell}, lateral_flow::Val{false}, model::RichardsModel -> adds in horizontal flow of water
  - domain::Union{HybridBox,SphericalShell}, lateral_flow::Val{false}, model::EnergyHydrology -> adds in horizontal flow of water and heat.

Review checklist

I have:

----


src/standalone/Soil/energy_hydrology.jl
```diff
@@ -144,18 +146,21 @@ end
...
 function EnergyHydrology{FT}(;
     parameters::EnergyHydrologyParameters{FT, PSE},
     domain::D,
     boundary_conditions::NamedTuple,
     sources::Tuple,
+    lateral_flow::Bool = true,
 ) where {FT, D, PSE}
     args = (parameters, domain, boundary_conditions, sources)
-    EnergyHydrology{FT, typeof.(args)...}(args...)
+    EnergyHydrology{FT, typeof.(args)...}(args..., lateral_flow)
```
juliasloan25, line 163: Could we add `lateral_flow` as the last item of `args` instead of passing it in separately? That might be a cleaner implementation

src/standalone/Soil/rre.jl
```diff
@@ -61,18 +63,21 @@ end
...
 function RichardsModel{FT}(;
     parameters::RichardsParameters{FT},
     domain::D,
     boundary_conditions::NamedTuple,
     sources::Tuple,
+    lateral_flow::Bool = true,
 ) where {FT, D}
     args = (parameters, domain, boundary_conditions, sources)
-    RichardsModel{FT, typeof.(args)...}(args...)
+    RichardsModel{FT, typeof.(args)...}(args..., lateral_flow)
```
juliasloan25, line 80: Same comment as in energy_hydrology.jl - maybe `lateral_flow` can be in the `args` tuple

test/standalone/Soil/soil_test_3d.jl
```diff
@@ -415,3 +415,49 @@ end
     ]) / ν < 1e-11
 
 end
+
+
+@testset "Lateral flow flag" begin
+    FT = Float32
```
juliasloan25, line 421: Should we test for Float32 and Float64?

test/standalone/Soil/soil_test_3d.jl
```diff
@@ -415,3 +415,49 @@ end
...
+    )
+    column = ClimaLSM.Domains.Column(; zlim = (zmin, zmax), nelements = 10)
+    for d in [boxdomain, shell]
+        f1 = ClimaCore.Fields.zeros(d.space)
+        f2 = ClimaCore.Fields.zeros(d.space)
+        dY = ClimaCore.Fields.FieldVector(; g = f1, h = f2)
+        ClimaLSM.Soil.horizontal_components!(dY, d, Val(false))
+        @test dY.g == f1
+        @test dY.h == f2
+        # If you pass true, you need additional arguments (dispatch off of model type)
```
juliasloan25, line 446: Should we also add a test where the additional arguments are provided and we check that the horizontal components are correctly computed?

juliasloan25 (commented):

src/standalone/Soil/energy_hydrology.jl
kmdeck: hmmm we could but then lateral_flow is not part of the parametric types of the model, so we couldnt do {typeof.(args)...} here then: `EnergyHydrology{FT, typeof.(args)...}`

kmdeck: (I declared lateral_flow as type Bool in the original struct defn)

src/standalone/Soil/rre.jl
kmdeck: same as above. happy to implement something cleaner!

test/standalone/Soil/soil_test_3d.jl
kmdeck: I think here it is OK because we are testing that the function does not change dY when there should be no horizontal components added to dY, and it isnt really related to carrying out computations and making sure the type isnt change by them?

test/standalone/Soil/soil_test_3d.jl
kmdeck: I think this is ok because the other tests in this file were already implicitly testing horizontal_components! in the case where they were nonzero, and those are the default now (with lateral_flow = true).

src/standalone/Soil/energy_hydrology.jl
juliasloan25: Ahhh I figured I was missing something. This looks good then :)

test/standalone/Soil/soil_test_3d.jl
juliasloan25: Good point

juliasloan25 (approved):

PR #263, opened 2023-07-11
Area index in aux

kmdeck:
## Purpose 
- Removes LAI and h_c from SharedCanopyParameters as those are duplicated in the PlantHydraulicsParameters
- Allows LAI to vary in time 
- Builds some infrastructure for future parameters that vary in space and time.

## To-do
- future PR: LAI(t) from Ozark data.

## Content
1. LAI and h_c are removed from SharedCanopyParameters. The height of the canopy is now obtained from the last element of "compartment_surfaces" [plant hydraulics compartment surfaces, in the hydraulics model] and LAI is now obtained from the area_index. This removes duplication of parameter definition.
2. Replace the area_index (the constant named tuple of floats for SAI, RAI, and LAI) in PlantHydraulicsParameters with an area_index "model" (prescribed function for LAI(t), SAI::FT, RAI::FT, or eventually reading from a map).
3. Adds a field of NamedTuples to the auxiliary state: p.canopy.hydraulics.area_index
4. adds a "set initial aux state" method for the canopy which sets globally varying parameters (just area_index for now) with their initial values
5. adds an "update parameter fields" function which can be called in update aux which updates the area index (and eventually any other globally varying fields that change in time).
6. Adds a utils function for indexing fields of NamedTuples by symbol.

## Discussion items
It may be that in the long run, these globally varying fields should be stored in p.canopy directly. we can build that when we need it? distiniguish between prescribed parameters in `p` and diagnostic params that depend on state? [currently not doing this]

One downside is that for some models (bucket, now canopy too and any model involving the canopy) we need to call "set_initial_aux_state!" and for some we dont. perhaps this should be a standard step in tutorials even if it isnt needed for all models? it could be confusing. [ we decided offline that all models will call this, even if not strictly required]

Review checklist

I have:

----


src/standalone/Vegetation/Canopy.jl
```diff
@@ -381,8 +397,8 @@ function ClimaLSM.make_update_aux(
         hydraulics = canopy.hydraulics
         n_stem = hydraulics.n_stem
         n_leaf = hydraulics.n_leaf
-        (; retention_model, conductivity_model, S_s, ν, area_index) =
-            hydraulics.parameters
+        #PlantHydraulics.lai_consistency_check.(n_stem, n_leaf, area_index)
```
kmdeck, line 400: had trouble getting this to run...

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -63,48 +67,22 @@ transpiration (Prescribed or Diagnostic)
...
-        typeof(retention_model),
-    }(
-        area_index,
-        ν,
-        S_s,
-        root_distribution,
-        conductivity_model,
-        retention_model,
-    )
+struct PrescribedSiteAI{FT <: AbstractFloat}
```
kmdeck, line 79: eventually inherits from  "AbstractPrescribedFunction" type? which would encompass any prescribed parameter function?

ties into what we were discussing yesterday @juliasloan25 - but too soon to implement now, I think, until we also have the different examples from the bucket work

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -146,6 +124,50 @@ function lai_consistency_check(
...
+
+"""
+    PlantHydraulicsParameters{FT <: AbstractFloat}
+
+A struct for holding parameters of the PlantHydraulics Model.
+$(DocStringExtensions.FIELDS)
+"""
+struct PlantHydraulicsParameters{FT <: AbstractFloat, CP, RP}
+    "The area index model for LAI, SAI, RAI"
+    area_index::PrescribedSiteAI{FT}
```
kmdeck, line 136: eventually allows also for AI maps from files

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -269,8 +290,46 @@ ClimaLSM.auxiliary_types(model::PlantHydraulicsModel{FT}) where {FT} = (
     NTuple{model.n_stem + model.n_leaf, FT},
     NTuple{model.n_stem + model.n_leaf, FT},
     FT,
+    NamedTuple{(:root, :stem, :leaf), Tuple{FT, FT, FT}},
 )
 
+
+"""
+    set_canopy_parameter_field!(component::PlantHydraulics{FT},
```
kmdeck, line 298: could make more modular once we have more parameters needing updating.....

src/standalone/Vegetation/Canopy.jl
juliasloan25: What was the trouble?

src/standalone/Vegetation/Canopy.jl
```diff
@@ -543,7 +558,7 @@ A helper function which returns the surface height for the canopy
 model, which is stored in the parameter struct.
 """
 function ClimaLSM.surface_height(model::CanopyModel, _...)
-    return model.parameters.h_c
+    return model.hydraulics.compartment_surfaces[end]
```
juliasloan25, line 561: I think we should add a comment in the `PlantHydraulicsModel` struct explaining that the height of the canopy is the last element of the `compartment_surfaces` struct, since that might not be very obvious.

src/standalone/Vegetation/PlantHydraulics.jl
juliasloan25: I agree :)

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -269,8 +290,46 @@ ClimaLSM.auxiliary_types(model::PlantHydraulicsModel{FT}) where {FT} = (
...
+                                t0,
+                                ) where {FT}
+
+"""
+function ClimaLSM.Canopy.set_canopy_parameter_field!(
+    component::PlantHydraulicsModel{FT},
+    p,
+    t0,
+) where {FT}
+    (; LAIfunction, SAI, RAI) = component.parameters.area_index
```
juliasloan25, line 309: Maybe rename `area_index` in `parameters` to `ai_parameterization`, or something else that emphasizes this is the parameterization not the values themselves (like we discussed today)

src/standalone/Vegetation/PlantHydraulics.jl
juliasloan25: I agree, it would be nice for this to be less hardcoded. But we can address maybe adding an abstract type for prescribed parameters once we add more and need to generalize

juliasloan25:
Also, as we discussed earlier today, I agree that all models should call `set_initial_aux_state!` for consistency, even if they don't need it for their functionality. They can maybe just use the default which calls `update_aux`

src/standalone/Vegetation/Canopy.jl
```diff
@@ -510,7 +525,7 @@ function canopy_surface_fluxes(
     canopy_conductance =
         upscale_leaf_conductance.(
             leaf_conductance,
-            model.parameters.LAI,
+            p.canopy.hydraulics.area_index[:leaf],
```
AlexisRenchon, line 528: a bit random place for this comment, but following the discussion from yesterday, I agree that it makes sense to have things (prescribed pool like LAI or prescribed parameter like albedo) varying in time in p. 

Y is for PDE... but p can store those other things we get from functions or files. 

Anyway, just saying I agree with this design choice!

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -63,48 +67,22 @@ transpiration (Prescribed or Diagnostic)
 """
 abstract type AbstractTranspiration{FT <: AbstractFloat} end
 
-
 """
-    PlantHydraulicsParameters{FT <: AbstractFloat}
+   PrescribedSiteAI{FT <:AbstractFloat}
```
AlexisRenchon, line 71: should we spell out AI (for clarity)? PrescribedSiteAreaIndex

kmdeck: 💯

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: good call!

src/shared_utilities/ntuple_utils.jl
```diff
@@ -39,3 +39,18 @@ Base.@propagate_inbounds Base.getindex(
...
+    Base.getindex(field::ClimaCore.Fields.Field,
+                   i::Symbol,
+                  ) 
+
+Returns the element of the ClimaCore.Fields.Field by index.
+"""
+Base.@propagate_inbounds Base.getindex(
+    field::ClimaCore.Fields.Field,
+    s::Symbol,
+) = Base.getproperty(field, s)
```
charleskawczynski, line 56: Can we avoid this? This is type piracy

src/standalone/Vegetation/PlantHydraulics.jl
AlexisRenchon: just a reminder that `parameter` naming feels a bit weird as area index will eventually be a prognostic variable (from leaf carbon pool, resulting from leaf growth (NPP*allocation to leaf) and accounting for leaf mass per area). 

that being said, the name is ok with me for now! just to keep in mind and maybe change later (or not).

AlexisRenchon (approved):
I just had a few food for thoughts comment, but it looks good to me!

src/shared_utilities/ntuple_utils.jl
kmdeck: here too? https://github.com/CliMA/ClimaLSM.jl/blob/26bd6a95fa5fe46be73e5f0c1852d29545e94092/src/shared_utilities/ntuple_utils.jl#L41

Is there a recommended solution that was implemented for EDMF instead? How should we index into field of tuples/named tuples?

kmdeck: oh - maybe the solution is to have ClimaCore define these. then it is not type piracy, right?

charleskawczynski: > Is there a recommended solution that was implemented for EDMF instead?

We overloaded `getindex` for ClimaCore Fields for `Int` to handle working with `NTuple` fields basically to work around https://github.com/CliMA/ClimaCore.jl/issues/77. The eventual way we worked around it was mentioned here: https://github.com/CliMA/ClimaCore.jl/issues/77#issuecomment-1502147802.

charleskawczynski: > maybe the solution is to have ClimaCore define these. then it is not type piracy, right?

Yes, this would not longer be piracy, however, I believe we are trying to avoid defining these methods in ClimaCore because of design considerations.

charleskawczynski: > How should we index into field of tuples/named tuples?

Maybe you can give an example of what you mean?

```julia
using Revise; import ClimaCore
include(joinpath(pkgdir(ClimaCore), "test", "TestUtilities", "TestUtilities.jl"))
import .TestUtilities as TU
FT = Float64
space = TU.PointSpace(FT)
f = fill((;a=(;x=FT(1))), space)
f.a.x
```

kmdeck: Let me try removing the methods we introduced for getindex and use the latest ClimaCore. Ill report back if it doesnt work. thanks!

kmdeck: hm, ok, now other tests fail unrelated to this PR that use getindex (using latest ClimaCore).

kmdeck: Im not sure I can fix all of this in this PR. Would you have time to chat about this? I could either remove the current getindex overwrite we have for Ntuples in a separate PR first, or merge this, and then remove both getindex methods (for Ntuples and NamedTuples)

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: in progress for docs/experiments, but fixed in src and test. will "resolve" this but update those still.

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: I left a note in the docs of CanopyModel

Note that the canopy height is specified as part of the 
PlantHydraulicsModel, along with the area indices of the leaves, roots, and
stems. Eventually, when plant biomass becomes a prognostic variable (by
integrating with a carbon model), some parameters that are specified here will be
treated differently.

kmdeck:
> Also, for discussion item #2, I agree that all models should call `set_initial_aux_state!` for consistency, even if they don't need it for their functionality. They can maybe just use the default which calls `update_aux` (as we discussed earlier today)

ok! Im going to open an issue for this (to add this to our tests/experiments/docs)

kmdeck:
> I just had a few food for thoughts comment, but it looks good to me!

thank you! Ill address your comments. 

Do you think you could take on this issue? https://github.com/CliMA/ClimaLSM.jl/issues/275 related to our discussion last week.

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: I also changed this to be `update_canopy_prescribed_field` and `set_canopy_prescribed_field`

src/standalone/Vegetation/Canopy.jl
kmdeck: it seems to work now? 🤷‍♀️

src/shared_utilities/ntuple_utils.jl
charleskawczynski: That’s unfortunate, but I understand. Can you open an issue to have this removed?

kmdeck: On the plus side, I got it to work for NamedTuples without the extra method based on your example above. so this PR isnt adding something we'll have to remove. I'll open an issue for removing the Ntuple getindex method.

kmdeck: resolving this here; issue opened #277

PR #264, opened 2023-07-12
add FileReader module

juliasloan25:
## Purpose 
Add a FileReader module (based on ClimaCoupler.jl's BCReader module) to read in data files over time

Part of PR #260 and issue #243

## To-do
- [x] Add FileReader module
- [x] Add unit tests for all functions
- [x] Rename/remove any coupler-specific

----


docs/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.9.1"
+julia_version = "1.9.0"
```
kmdeck, line 3: did we decide to stick with 1.9.1?

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+    surface_space::B
+    date_idx::VI
+    date_idx0::VI
+    segment_length::VI
+    regrid_dirpath::S
+    outfile_root::S
+end
+
+FileInfo{FT}(args...) where {FT} = FileInfo{FT, typeof.(args[1:7])...}(args...)
+float_type_fi(::FileInfo{FT}) where {FT} = FT
```
charleskawczynski, line 92: Just FYI: the float type is already stored in `Spaces.undertype(surface_space)`, so this type parameter is in a sense redundant

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+
+"""
+    FileInfo
+
+Stores information specific to each boundary condition from a file and each variable.
+
+# Inputs:
+- file_path::S                    # directory of the data file
+- comms::X                        # communication context used for MPI
+- varname::V                      # name of the variable
```
charleskawczynski, line 67: should `varname` be plural here? (I imagine there's more than one variable per file)

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+    string(string(lpad(Dates.month(datetime), 2, "0"))) *
+    string(lpad(Dates.day(datetime), 2, "0"))
+
+"""
+    FileInfo
+
+Stores information specific to each boundary condition from a file and each variable.
+
+# Inputs:
+- file_path::S                    # directory of the data file
```
charleskawczynski, line 65: should this be `input_file` to compliment `output_file`?

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+    ClimaComms.barrier(comms)
+    data_dates = JLD2.load(
+        joinpath(regrid_dirpath, outfile_root * "_times.jld2"),
+        "times",
+    )
+
+    # init time tracking info
+    current_fields =
+        Fields.zeros(FT, surface_space), Fields.zeros(FT, surface_space)
+    segment_length = [Int(0)]
```
charleskawczynski, line 166:
```suggestion
    segment_length = Int[0]
```

charleskawczynski: This will enforce that the container has the correct eltype

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+different date.
+We could address this by adding a `date_start` field,
+where the date at any time `t` in a simulation is given by `date_start + t`.
+This is relevant since the land model keeps track of time
+as the number of seconds elapsed since the starting time, rather
+than as a standalone DateTime.
+
+# Arguments
+- `FT`: [DataType] Float type.
+- `regrid_dirpath`: [String] directory the data file is stored in.
```
charleskawczynski, line 122:
```suggestion
- `regrid_dirpath`: String[] directory the data file is stored in.
```
I think it's better that we write this in a way that matches julia syntax

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+- `file_path`: [String] NCDataset file containing data to regrid.
+- `varname`: [String] name of the variable to be regridded.
+- `surface_space`: [Spaces.AbstractSpace] the space to which we are mapping.
```
charleskawczynski, line 125: Should we enforce these type restrictions in the type parameters? That way we'll get eager errors if/when users pass in the wrong datatype

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+- `comms`: [ClimaComms.AbstractCommsContext] context used for this operation.
+- `date_idx0`: [Vector{Int}] index of the file data that is closest to date0.
```
charleskawczynski, line 127:
```suggestion
- `date_idx0::Vector{Vector{Int}}`: index of the file data that is closest to date0.
```
I might actually suggest writing the doc string like this

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+function file_info_init(
+    FT,
+    regrid_dirpath,
+    file_path,
+    varname,
+    surface_space,
+    comms;
+    date_idx0 = nothing,
+    mono = true,
+)
```
charleskawczynski, line 143:
```suggestion
) where {FT <: Real}
```

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+- `comms`: [ClimaComms.AbstractCommsContext] context used for this operation.
+- `date_idx0`: [Vector{Int}] index of the file data that is closest to date0.
+- `scaling function`: [Function] scales, offsets or transforms `varname`.
+- `mono`: [Bool] flag for monotone remapping of `file_path`.
+
+# Returns
+- `FileInfo`
+"""
+function file_info_init(
+    FT,
```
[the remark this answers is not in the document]
charleskawczynski: (assuming FT is being kept in here)

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+            Regridder.read_from_hdf5(
+                regrid_dirpath,
+                outfile_root,
+                all_dates[Int(date_idx0)],
+                varname,
+                comms,
+            ),
+            file_info.surface_space,
+        )
+        file_info.data_fields[2] .= deepcopy(file_info.data_fields[1])
```
charleskawczynski, line 236: Why are we calling `deepcopy` here?

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+        file_info.segment_length .= Int(0)
```
charleskawczynski, line 237:
```suggestion
        file_info.segment_length .= 0
```

src/shared_utilities/FileReader.jl
juliasloan25: The type is just String, not a String array. When we were refactoring the coupler we came up with this notation to denote argument types `- 'arg': [Type]`. If you think it's confusing though I can change it

charleskawczynski: Ah, I see. I think we've actually done that in ClimaAtmos, too, now that I think about it. Part of me kind of likes the `::` notation since it follows julia syntax, it may be more familiar to users familiar with julia-- and since we can make arrays with brackets, it avoids the potential confusion with showing an example of types

juliasloan25: Yeah, that makes sense. I'm working on splitting this large struct into smaller structs with a more logical structure, so I can make this change as I go through as well.

src/shared_utilities/FileReader.jl
juliasloan25: Good point, but we're removing this field from the struct anyway since it's only used in the constructor to fill in other parts of the struct.

src/shared_utilities/FileReader.jl
juliasloan25: So far we've only been using files that have one variable we want to read in. We could extend this in the future if we want to use a file that has multiple variables.

src/shared_utilities/FileReader.jl
juliasloan25: In these cases, we want `data_fields[1]` and `data_fields[2]` to store exactly the same information. Would it use less space to just call `copy` instead?

src/shared_utilities/FileReader.jl
juliasloan25: Good point, I'll remove it

src/shared_utilities/FileReader.jl
juliasloan25: The data type should be `Vector{Int}` hopefully once I change the syntax used in the docstrings this will be more clear

src/shared_utilities/FileReader.jl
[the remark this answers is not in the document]
juliasloan25: I removed `FT` as an argument (it's extracted from the space now), so we don't need this change anymore

src/shared_utilities/FileReader.jl
charleskawczynski: `file_info.data_fields[2] .= file_info.data_fields[1]` should be fine. `file_info.data_fields[2] .= deepcopy(file_info.data_fields[1])` first allocates a whole new copy of `file_info.data_fields[1]`, and then values from that array get broadcasted into `file_info.data_fields[2]`

juliasloan25: I've removed the calls to `deepcopy` and am just assigning `data_fields` using `.=`. This is clearer and has fewer allocations

test/shared_utilities/file_reader.jl
```diff
@@ -0,0 +1,396 @@
...
+                missing => 0,
+            )
+            replace!(
+                parent(prescribed_data.file_state.data_fields[2]),
+                NaN => 0,
+            )
+            replace!(
+                parent(prescribed_data.file_state.data_fields[2]),
+                missing => 0,
+            )
```
charleskawczynski, line 266:
```suggestion
            (; data_fields) = prescribed_data.file_state
            replace!(parent(data_fields[1]), NaN => 0)
            replace!(parent(data_fields[1]), missing => 0)
            replace!(parent(data_fields[2]), NaN => 0)
            replace!(parent(data_fields[2]), missing => 0)
```
code golfing here, I might do this

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,413 @@
...
+end
+
+"""
+    FileState
+
+Stores information about the current data being read in from a file.
+
+# Inputs:
+- data_fields::F        # tuple of two fields at consecutive dates, that will be used for interpolation
+- date_idx::I           # index of the data in the file currently being used
```
kmdeck, line 60: if there are two fields stored in `data_fields`, is this a tuple of indices as well?

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,413 @@
...
+- segment_length::I     # length of each date segment (used in the daily interpolation)
```
kmdeck, line 61: maybe "# length of the time interval between the two data field entries; used in temporal interpolation"

kmdeck: (not daily in general)

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,413 @@
...
+        regrid_dirpath,
+        input_file,
+        varname,
+        date_ref,
+        t_start,
+        surface_space;
+        mono = true,
+    )
+
+Regrids from lat-lon grid to cgll grid, saving the output in a new file,
```
kmdeck, line 114: maybe: "Regrids from the input lat-lon grid to the simulation cgll grid, saving the regridded output in a new file found at `regrid_dirpath`, and returns the info required to run the simulation using this prescribed data package into a single struct."
?

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,413 @@
...
+
+
+"""
+    FileInfo
+
+Stores information about the current data being read in from a file.
+
+# Inputs:
+- regrid_dirpath::String     # directory for storing files used in regridding
+- outfile_root::String       # root for files generated when reading data from input file
```
kmdeck, line 40: "root for regridded data files generated when regridding data from input file"

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,413 @@
...
+"""
+function interpolate_data(prescribed_data::PrescribedData, date, space)
+    FT = Spaces.undertype(space)
+    # Interpolate if the time period between dates is nonzero
+    if prescribed_data.file_state.segment_length[1] > FT(0)
+        (; segment_length, date_idx, data_fields) = prescribed_data.file_state
+        (; all_dates) = prescribed_data.file_info
+
+        return interpol.(
+            Regridder.swap_space!(data_fields[1], space),
```
kmdeck, line 353: I think we do not need this here now, though? (swap_space!)

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,413 @@
...
+
+"""
+    FileInfo
+
+Stores information about the current data being read in from a file.
+
+# Inputs:
+- regrid_dirpath::String     # directory for storing files used in regridding
+- outfile_root::String       # root for files generated when reading data from input file
+- varname::String            # name of the variable we're reading from the input file
```
kmdeck, line 41: We should also note somewhere that right now we are assuming this variable is a scalar.

One other thought is to note in the doc strings that by date we mean datetime but leave it as "date" for brevity.

kmdeck (approved):
This looks good to me, Julia! I mostly left minor comments on the docs strings

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+
+# Arguments
+- `date`: object to be converted to DateTime
+"""
+to_datetime(date) = CFTime.reinterpret(DateTime, date)
+
+"""
+    datetime_to_strdate(datetime::DateTime)
+
+Convert from Date to String ("YYYYMMDD") format.
```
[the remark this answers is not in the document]
juliasloan25: I'll just make a small issue for this and we can address it later: #274

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+
+# Inputs:
+- file_path::S                    # directory of the data file
+- comms::X                        # communication context used for MPI
+- varname::V                      # name of the variable
+- all_dates::D                    # vector of all dates contained in the original data file
+- data_fields::C                  # tuple of two fields at consecutive dates, that will be used for interpolation
+- surface_space::B                # the space we're reading data onto
+- date_idx::Vector{Int}           # time index of the data in the file
+- date_idx0::Vector{Int}          # `date_idx` of the file data that is closest to date0
```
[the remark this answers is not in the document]
juliasloan25: I was planning to not store `date0`, and just calculate it when we need it as `date0 = date_ref + Dates.Second(t_start)`. Do you think that's okay?

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+- segment_length::Vector{Int}     # length of each date segment (used in the daily interpolation)
```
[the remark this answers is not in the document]
juliasloan25: `segment_length` actually stores only the length of the current date segment stored in `data_fields`. I updated the docstring to make this more clear. I think you're right that it's needed in case the dates aren't uniformly spaced - otherwise we could just store a single integer for all segments

src/shared_utilities/FileReader.jl
juliasloan25: `date_idx` just gives the time index of the first loaded data field. I'll update the docstring to make this more clear

src/shared_utilities/FileReader.jl
juliasloan25: That sounds good to me :)

src/shared_utilities/FileReader.jl
juliasloan25: Good catch

src/shared_utilities/FileReader.jl
juliasloan25: Which variable are we assuming is a scalar? `varname`?

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+"""
+datetime_to_strdate(datetime::DateTime) =
+    string(lpad(Dates.year(datetime), 4, "0")) *
+    string(string(lpad(Dates.month(datetime), 2, "0"))) *
+    string(lpad(Dates.day(datetime), 2, "0"))
+
+"""
+    FileInfo
+
+Stores information specific to each boundary condition from a file and each variable.
```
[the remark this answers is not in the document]
juliasloan25: This should be addressed in the latest iteration, which divides this large struct into 3 smaller ones with a wrapper, and removes a couple unnecessary fields (e.g. raw file name)

src/shared_utilities/FileReader.jl
```diff
@@ -0,0 +1,371 @@
...
+
+"""
+    datetime_to_strdate(datetime::DateTime)
+
+Convert from Date to String ("YYYYMMDD") format.
+
+# Arguments
+- `datetime`: [DateTime] object to be converted to string
+"""
+datetime_to_strdate(datetime::DateTime) =
```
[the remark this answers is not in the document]
juliasloan25: That's a good point, we can just use `Dates.value(date)` to get an integer. I removed this function and replaced it with that.

src/shared_utilities/FileReader.jl
juliasloan25: I'll add them

src/shared_utilities/FileReader.jl
juliasloan25: I elaborated the docstrings for `varname` and `all_dates` to make them more clear

PR #265, opened 2023-07-12
Two-Stream Radiative Transfer Scheme

Espeer5:
## Purpose 

Adds two-stream radiative transfer model as an alternative to the BeerLambertModel for computing APAR in the CanopyModel, following methods found in [pySellersTwoStream](https://github.com/tquaife/pySellersTwoStream)

## Content

- Two-stream parameters struct and model struct included in radiation.jl
- New plant_absorbed_ppfd method added for the two-stream model which computes APAR using the formulas and energy balance method found [here](https://github.com/tquaife/pySellersTwoStream/blob/master/sellersTwoStream.py#L450). (After testing, the calculations from these two implementations agree, see testing script which compares outputs of both implementations [here](https://github.com/CliMA/ClimaLSM.jl/blob/two-stream/test/standalone/Vegetation/test_two_stream.jl))
- Substituting this model into the radiative transfer for the Ozark experiment with the same parameters yields very little change in the GPP prediction, see plots below, one of which is using two-stream with one layer canopy, the other is using BeerLambert
Two-Stream:
GPP2STR
BeerLambert:
GPP

----


src/standalone/Vegetation/Canopy.jl
```diff
@@ -367,14 +368,15 @@ function ClimaLSM.make_update_aux(
         q::FT = canopy.atmos.q(t)
 
         # update radiative transfer
-        PAR .= compute_PAR(canopy.radiative_transfer, canopy.radiation, t)
         K = extinction_coeff(ld, θs)
+        PAR .= compute_PAR(canopy.radiative_transfer, canopy.radiation, t)
         @. APAR = plant_absorbed_ppfd(
```
kmdeck, line 373: I think it would be better to have multiple methods for plant_absorbed_ppfd based on the type of canopy.radiative_transfer, rather than the # of arguments. the parameters can be unpack within the function, then?

Espeer5: I wanted to do this initially, but there is an issue evaluating that way. Since plant_absorbed_ppfd is called element-wise using "@." to unpack the PAR argument which is stored as a field vector, I couldn't directly pass the radiation model struct or parameters struct without throwing an error from calling @. on the struct. This was the work around I found for this but it is messy so if there is a better way I would love to know.

kmdeck: ah, yes! youre right. There are a couple ways around that...
Take a look at the function `matric_potential` here:
https://github.com/CliMA/ClimaLSM.jl/blob/bccd7217ec7ba45e820c616253af156d177b2ff8/src/standalone/Soil/soil_hydrology_parameterizations.jl#L40

which dispatches off of the type of the first argument. Here is another method for it:
https://github.com/CliMA/ClimaLSM.jl/blob/bccd7217ec7ba45e820c616253af156d177b2ff8/src/standalone/Soil/soil_hydrology_parameterizations.jl#L140

the first option is to do this:
`p.soil.psi .= matric_potential.(Ref(hcm), S) `
where `S` is a field or array and `hcm` is the struct. `Ref` effectively makes it so you can broadcast the struct as well. The downside to this is that we now need two broadcasting calls, one in the `.=` and one in the function itself. doing this:
`@. p.soil.psi = matric_potential(Ref(hcm), S) ` which would reduce it to a single broadcast is more efficient, but it will fail because it will also try and broadcast the function `Ref`.

Second option:
Make your struct broadcastable....
https://github.com/CliMA/ClimaLSM.jl/blob/bccd7217ec7ba45e820c616253af156d177b2ff8/src/standalone/Soil/retention_models.jl#L18

then you can do this:
`@. p.soil.psi = matric_potential(hcm, S)`

I think we will move towards this for most of our structs, see:https://github.com/CliMA/ClimaLSM.jl/issues/209

Espeer5: Got it! Cleaned it up using the second approach you suggested. Thanks

kmdeck: nice! glad it worked

kmdeck:
one idea - for testing, you could save the output of the python code to a file, and I can add it to caltech box account. then we can use it for a comparison in CI.

we do something like that here in non-unit test: https://github.com/CliMA/ClimaLSM.jl/blob/bccd7217ec7ba45e820c616253af156d177b2ff8/experiments/standalone/Soil/richards_comparison.jl#L103

kmdeck:
PS thank you for being so thorough with respect to tests/updating docs, etc :)

braghiere:
This is looking great, @Espeer5! Thanks for doing the test. Thanks for reviewing it, @kmdeck.

braghiere (approved):
Looks good to me.

src/integrated/soil_canopy_model.jl
```diff
@@ -280,7 +280,7 @@ Returns the correct radiative fluxes for bare ground in the case
 where the canopy LAI is zero.
 """
 function net_radiation_at_ground(
-    canopy_radiation::Canopy.BeerLambertModel{FT},
+    canopy_radiation::Canopy.AbstractRadiationModel{FT},
```
kmdeck, line 283: I think we can make the same change in the soil +plant hydrology file

src/standalone/Vegetation/Canopy.jl
```diff
@@ -322,7 +321,7 @@ has a single update_aux! function, given here.
 function ClimaLSM.make_update_aux(
     canopy::CanopyModel{
         FT,
-        <:BeerLambertModel,
+        <:AbstractRadiationModel,
```
kmdeck, line 324: I think we can either keep this as <:AbstractRadiationModel, but then specify in the docs for AbstractRadiationModel that it must have field `parameters` with \Omega, \lambda_gamma, and ld, since these are unpacked from it in this function. 

We could also update the methods for `auxiliary_variables` and `auxiliary_types` in radiation.jl to be for AbstractRadiationModel instead of having the same behavior but defined in two methods for each concrete model type. This would reduce some duplicate code.


otherwise I would suggest replacing this with the type `Union{BeerLambertModel, TwoStreamModel}`

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -23,31 +23,197 @@ export plant_absorbed_ppfd,
...
+    θs::FT,
 ) where {FT}
-    APAR = PAR * (1 - ρ_leaf) * (1 - exp(-K * LAI * Ω))
+    RTP = RT.parameters
+    APAR = PAR * (1 - RTP.ρ_leaf) * (1 - exp(-K * LAI * RTP.Ω))
     return APAR
 end
 
+"""
+    plant_absorbed_ppfd(ld::FT,
```
kmdeck, line 58: likewise here

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -23,31 +23,197 @@ export plant_absorbed_ppfd,
...
+        I_dir_dn_prev = I_dir_dn
+        I_dif_up_prev = I_dif_up
+        I_dif_dn_prev = I_dif_dn
+
+        # Move on to the next layer
+        i += 1
+    end
+
+    # Convert FAPAR into APAR and return
+    return PAR * FAPAR
```
kmdeck, line 214: you might double check that this returns Float64 if the input args are float64 & likewise for float32.

src/standalone/Vegetation/radiation.jl
```diff
@@ -46,22 +47,93 @@ ClimaLSM.name(model::AbstractRadiationModel) = :radiative_transfer
 ClimaLSM.auxiliary_vars(model::BeerLambertModel) = (:apar, :par)
```
kmdeck, line 47: if you wish, you can replace this with either
`ClimaLSM.auxiliary_vars(model::AbstractRadiationModel) = (:apar, :par)`
or `ClimaLSM.auxiliary_vars(model::Union{BeerLambertModel, TwoStreamModel}) = (:apar, :par)`

(see above comment). I think either is fine, and both reduce code duplication.

liekwise for aux types function

src/standalone/Vegetation/radiation.jl
```diff
@@ -46,22 +47,93 @@ ClimaLSM.name(model::AbstractRadiationModel) = :radiative_transfer
...
+"""
+struct TwoStreamParameters{FT <: AbstractFloat}
+    "Leaf angle distribution function (unitless)"
+    ld::FT
+    "PAR canopy reflectance (unitless)"
+    ρ_leaf::FT
+    "leaf element transmittance"
+    τ::FT
+    "Soil albedo"
+    a_soil::FT
```
kmdeck, line 64: we should think about this more. the soil albedo is specified in the soil model when they are run together, and we dont want to specify it twice. in standalone, this would need to be specified. perhaps we can discuss more on a slack call?

src/standalone/Vegetation/radiation.jl
```diff
@@ -46,22 +47,93 @@ ClimaLSM.name(model::AbstractRadiationModel) = :radiative_transfer
...
+    "Clumping index following Braghiere (2021) (unitless)"
+    Ω::FT
+    "Typical wavelength per PAR photon (m)"
+    λ_γ::FT
+    "number of layers to simulate radiative transfer through"
+    n_layers::UInt64
```
kmdeck, line 70: does this need to match n_layers of plant hydraulics scheme, @braghiere ?

src/standalone/Vegetation/radiation.jl
kmdeck: We discussed offline and will follow this up in a separate PR. For now, in the soil_canopy_model constructor, perhaps we can add an @assert that checks that soil.parameters.albedo == canopy.radiative_transfer.soil_albedo?

kmdeck:
@braghiere before merging this  -  does it make sense that the Two stream and Beer Lambert models give the same output? (I checked and the plant_absorbed_ppfd functions do not produce identicaly the same output, but the median difference is 1% of the value...so the plots for GPP etc that @Espeer5 has produced look almost the exact the same)

src/standalone/Vegetation/radiation.jl
kmdeck: any reason why this is a Uint instead of an Int? I just ask because it is a little unintuitive to see the value as
julia> ts_params.n_layers
0x0000000000000014

instead of 20. might throw someone off?

Espeer5: Just because a negative layer count doesn't make any sense using 64 bit Uint gives a larger possible range of valid input values than using a signed 64 bit int, and will automatically throw an error if a user tried to give negative layers.

Espeer5:
> one idea - for testing, you could save the output of the python code to a file, and I can add it to caltech box account. then we can use it for a comparison in CI.
> 
> we do something like that here in non-unit test:
> 
> https://github.com/CliMA/ClimaLSM.jl/blob/bccd7217ec7ba45e820c616253af156d177b2ff8/experiments/standalone/Soil/richards_comparison.jl#L103

Went ahead and did this and added as a unit test, see [here](https://github.com/CliMA/ClimaLSM.jl/blob/two-stream/test/standalone/Vegetation/test_two_stream.jl)

Right now the file is being hosted on my personal Box so let me know how I can have the file added to the CliMA box

src/standalone/Vegetation/radiation.jl
Espeer5: Went ahead and added this assertion and will follow up in a separate PR.

kmdeck:
> 

ok! you should have access to the Clima Land box account now. you can move it there (it wont change the link, so it wont change the PR).

having this in test looks good to me because it is a comparison of function output. the experiments folder has some tests, but they involve carrying out a simulation, so we moved them out of unit tests.

src/standalone/Vegetation/radiation.jl
kmdeck: ok! thanks. I think it's OK for now that the n_layers here do not match the n_layers of the hydraulics model. I'm opening an issue just to keep track of it so we can revisit it later and make sure, but all good for now.

Espeer5:
> @braghiere before merging this - does it make sense that the Two stream and Beer Lambert models give the same output? (I checked and the plant_absorbed_ppfd functions do not produce identicaly the same output, but the median difference is 1% of the value...so the plots for GPP etc that @Espeer5 has produced look almost the exact the same)
> 
> maybe because the default is diff_perc =0?

Here is a plot showing the APAR from each of the Beer-Lambert and the Two-Stream over the course of the Ozark experiment. It appears Two-Stream is always higher due to the impact of backscattering.
APAR_comp

kmdeck (approved):

Issue #266, opened 2023-07-14
Add runoff model


kmdeck:
Add runoff (surface and subsurface). First model supported will be TOPMODEL.
Preliminary PR implementing Option 2: #267 [see below]

Additional preliminary work needed to support this:
- the topmodel parameters as a function of space #318
-  we also need to support non-global regional runs #319 

Followon PR adding TOPMODEL: #511 
**Possible solutions**
Runoff appears as source term and through affecting the surface boundary condition, so there are two options for implementations. It should only appear if the user is using the Richards model with a RichardsAtmosDrivenBC or the EnergyHydrology model with AtmosDrivenBC; that is, it doesnt make sense to use runoff with a prescribed state bc or a prescribed (arbitrary) flux BC.
Additionally, a single runoff model will in most cases specify both the subsurface and surface runoff contributions.
- Option 1: create an AbstractSubsurfaceRunoff and AbstractSurfaceRunoff type. AbstractSubsurfaceRunoff will be an AbstractSource (and when used, included in a tuple of sources in the model).  concrete types of AbstractSurfaceRunoff will be fields in either the RichardsAtmosDrivenBC or AtmosDrivenBC and affect the application of boundary conditions in these cases (boundary conditions are supplied in a different field of the model from sources).  Pro: splits neatly into the two functionalities/two different existing fields of the model.  Cons: we would be defining the same parameters twice, in both the subsurface and the surface runoff model.

- Option 2: Create a single AbstractRunoffModel and have this be a new field in the AtmosDrivenBC structs. When set to NoRunoff, it isnt used, infiltration = precip +evap and no source is added; otherwise we account for runoff in infiltration and we also add the source term (this latter piece is handled in the model constructor). Pros: no duplication of parameters. Cons: specifying a boundary condition type also dictates a source term, which is a blend of concepts.

In both cases, I think we will need a default NoRunoff model.

** Additional Context **
Runoff models will require adding additional auxiliary fields (parameter maps for topography). It seems easier to add these variables in a flexible way based off of the type of runoff, rather than checking to see if runoff is included in the tuple of sources, e.g.
`parameter_aux_fields(runoff::NoRunoff) = () ` and `parameter_aux_fields(runoff::TOPMODELRunoff) = (;\phi)`

PR #267, opened 2023-07-14
Preliminary runoff model

kmdeck:
## Purpose 
Lays the groundwork for including surface runoff (applied as part of the boundary condition for soil) and for subsurface runoff (applied as a sink term for the soil model).

In this implementation, we assume that
- a single runoff model struct will specify both surface and subsurface runoff parameterizations (these are often linked)
and that the use case is  
- modeling runoff (or not) with Richards model or with the full soil model, either in standalone mode or as part of an LSM. 
- only modeling runoff for the special cases of a simulation which prescribes precipitation (Richards model) or precip+atmospheric conditions+radiation conditions (which in turn leads to a computation for evaporation, SHF, LHF, radiative fluxes...; EnergyHydrology model). 
 
I.e. we wont support:
- just adding subsurface runoff, but not surface runoff, or vice versa
- somehow combining runoff with other types of boundary conditions

See Issue #266 

## Contents 
More major changes:
1. introduces RichardsAtmosDrivenFluxBC which serves the same role as AtmosDrivenFluxBC, but for Richards equation. New boundary_flux method for this BC. Not currently used - should we add now? but I think this could be useful to have.
2. methods for soil_boundary_fluxes for bc::AtmosDrivenFluxBC and boundary_flux for bc::RichardsAtmosDrivenFluxBC compute infiltration using a function defined in runoff.jl (soil_surface_runoff())
3.  Both richards and energy hydrology model constructors add the subsurface flow sink term to the equations based on the type of the boundary condition at the top - i.e. it will only do so if the BC is of type RichardsAtmosDrivenFluxBC or AtmosDrivenFluxBC. It has to add something of type AbstractSoilSource - use soil_subsurface_source().

Minor changes
1. Adds docstrings/updates APIs
2. fixes the same bug appearing in two files: docs/tutorials/Usage/LSM_single_column_tutorial.jl and tests/integrated/soil_energy_hydrology_biogeochemistry related to how boundary conditions are specified.
4. additional type signatures for boundary_flux methods, minor changes in arguments to boundary_flux, soil_boundary_fluxes, and ClimaLSM.source! methods.
- soil_boundary_fluxes(bc, boundary, model, dz, Y, p, t)
- boundary_flux(bc, boundary, model, dz, Y, p, t)
- ClimaLSM.source!(dY, src, Y, p, model)
- 
Review checklist

I have:

----


src/standalone/Soil/boundary_conditions.jl
```diff
@@ -118,14 +141,23 @@ function soil_boundary_fluxes(
         ),
     )
     r_ae = conditions.r_ae
-    net_water_flux = @. bc.atmos.liquid_precip(t) +
-       conditions.vapor_flux * r_ae / (r_soil + r_ae)
+    precip = bc.atmos.liquid_precip(t)
```
kmdeck, line 144: update in integrated model!

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -158,6 +166,13 @@ function EnergyHydrology{FT}(;
     sources::Tuple,
     lateral_flow::Bool = true,
 ) where {FT, D, PSE}
+    top_bc = boundary_conditions.top
+    if typeof(top_bc) <: AtmosDrivenFluxBC
+        # If the top BC indicates atmospheric conditions are driving the model
+        # add baseflow as a sink term
+        subsurface_source = subsurface_runoff_source(top_bc.runoff)
+        sources = append_source(subsurface_source, sources)
```
kmdeck, line 174: we will use the same functionality for sublimation, when we add it, since that is a source term that only makes sense for this BC.

src/integrated/soil_plant_hydrology_model.jl
```diff
@@ -59,7 +59,11 @@ function SoilPlantHydrologyModel{FT}(;
     root_extraction = PrognosticSoilPressure{FT}()
 
     boundary_conditions = (;
-        top = (water = FluxBC((p, t) -> eltype(t)(precipitation(t))),),
+        top = (
```
kmdeck, line 62: perhaps note that no runoff is modeled here.

PR #273, opened 2023-07-24
standalone docs, 1st PR

AlexisRenchon:
## Purpose 

This PR is a start of the documentation of standalone models. 
It is a small PR, to start merging #214 piece by piece, with thorough check of docs. 

## To-do

- [ ] is the structure of standalone docs good? (see below) 
- [ ] is the structure of each page good? (see beer's law template below)
- [ ] is the content of beer's law good (any typo? are the equations correct? are the explanations clear? are the units correct? etc.)

## Content

This PR contains: 
- the structure of standalone models (soil, vegetation, snow, standing water), with e.g., within vegetation (radiative transfer, photosynthesis, stomatal conductance, plant hydraulics), and within soil (biogeochemistry, hydrology, energy)
- beer's law model (which serves as a simple template to evaluate, model description with equations, interactive figure, table of variables with full name and acronym, units and typical range)

https://clima.github.io/ClimaLSM.jl/previews/PR273/

Screenshot from 2023-07-24 12-35-17

----


docs/src/standalone/pages/vegetation/radiative_transfer/beer_model.md
```diff
@@ -0,0 +1,59 @@
...
+index, $θ_s$ is the zenith angle, and $Ω$ is the clumping index following Braghiere
+(2021). $K$, $Ω$ and $ρ_{leaf}$ are all unitless. LAI is in m² m⁻².
+In order to compute $K$, we need $θ_s$ in radians and the leaf angle distribution $l_d$
+(unitless). K is then defined as
+
+```math
+K = l_d/\max{(\cos{(\theta_s)}, \epsilon)}
+```
+
+so that at night, when 3π/2 > $θ_s$ > π/2, $K$ is large (lots of extinction) and
```
Espeer5, line 28: Perhaps instead of "lots of extinction" we could say something like very little light penetrates through the canopy? Since it's not that more incident photons "become extinct" but rather that there are just very few incident photons at night.

docs/src/standalone/pages/vegetation/radiative_transfer/beer_model.md
```diff
@@ -0,0 +1,59 @@
...
+
+The portion of PAR that is actually absorbed by the vegetation canopy for photosynthesis is called Absorbed Photosynthetically Active Radiation (APAR). The APAR driving photosynthesis is calculated following the Beer-
+Lambert law:
+
+```math
+APAR(PAR, \theta_s) = (PAR)(1 - \rho_{leaf})(1 - e^{(-K(\theta_s) LAI  \Omega)})
+```
+
+where PAR ≈ SW/2 is the incident moles of photons per meter squared per
+second in the PAR window, approximated as half of the incident shortwave flux.
```
Espeer5, line 16: I think this might be supposed to read "approximated as half of the incident shortwave flux if PAR is not directly available. $p_{leaf}$ is the ...."

I think this change in the punctuation makes it make a little more sense?

PR #276, opened 2023-07-24
Add Harvard and Niwot Ridge experiments

Espeer5:
## Purpose 

Add experiments paralleling the Ozark experiment using data from the FluxNet sites at Harvard Forest and Niwot Ridge.

## To-do

- [ ] Refine experiment parameters as much as possible
- [ ] Update the new experiments to use the soil_driver structure recently merged
- [ ] Add time varying LAI to the new experiments

## Content

 - 2 experiments added in the directories experiments/integrated/harvard and experiments/integrated/niwot
- Currently, the plots of the model and the data together do not agree well with each other, similar to how the GPP plot for Ozark experiment does not agree well. I have been looking for a reason why this is but haven't been able to find anything which resolves the difference.

----

### DIFF
--- a/experiments/integrated/harvard/harvard.jl
+++ b/experiments/integrated/harvard/harvard.jl
@@ -1,3 +1,4 @@
+# External imports
 using DiffEqCallbacks
 import OrdinaryDiffEq as ODE
 import ClimaTimeSteppers as CTS
@@ -8,6 +9,7 @@ using Statistics
 using Dates
 using Insolation
 
+#Import CliMA packages
 using ClimaLSM
 using ClimaLSM.Domains: LSMSingleColumnDomain
 using ClimaLSM.Soil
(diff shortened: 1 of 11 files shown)


braghiere:
That looks good! Thanks. I think we should include the variable LAI capability before merging it. Probably the mismatch model/data has lots to do with it. Another thing is probably related to the parameter tuning. Maybe we could add a comment next to the parameter showing if the value was updated and indicate from which reference it was taken.

Espeer5:
> That looks good! Thanks. I think we should include the variable LAI capability before merging it. Probably the mismatch model/data has lots to do with it. Another thing is probably related to the parameter tuning. Maybe we could add a comment next to the parameter showing if the value was updated and indicate from which reference it was taken.

I've done this for all of the soil hydraulic and composition parameters in both experiments, still need to do this for the thermal properties and for the canopy parameters.

Espeer5:
Closing; These sites will be added and merged by #369

PR #279, opened 2023-07-27
Create prescribed and prognostic soil drivers for the canopy model

### LINKED ISSUE #271: Prognostic vs Prescribed soil conditions for the canopy model
Currently, in the canopy model, we support a `PrescribedSoilPressure` and a `PrognosticSoilPressure` type for computing root extraction. This is set in the [`root_extraction` field](https://github.com/CliMA/ClimaLSM.jl/blob/47bd35d811b97a5d6aa3c52bd058c32f9a44c5a7/src/standalone/Vegetation/PlantHydraulics.jl#L187) of the PlantHydraulicsModel type.  Based off of the type, the `root_flux_per_ground_area!` function returns the root extraction water flux based on different computations; the uses cases are standalone canopy (prescribed soil pressure) and canopy+soil (prognostic soil pressure).

At the same time, in the two stream model, we need to provide the soil albedo. when running the canopy model in standalone mode, this makes sense. But with soil+canopy, the soil should provide the soil albedo. 

The problem is (a) that we define albedo in multiple places and (2) that we have made it a little confusing how to run with prognostic vs prescribed soil (standalone canopy vs not).

Possible solution direction:
- like we supply an `atmos` field of the canopy model, supply `soil` field, which will be of type Prognostic vs. Prescribed. 
- dispatch off of this in root_flux_per_ground_area! instead of a "root extraction" type - get rid of root extraction types?
- make a ground_albedo function which also dispatches off of this type, and either returns soil.parameters.albedo (if prognostic) or canopy.soil.albedo (if prescribed)

one hiccup with this is that we also run the canopy+soil hydrology only (RichardsModel), which does not have a defined soil albedo. 

**Additional context**

Espeer5:
## Purpose 

Adds soil drivers to the canopy model to prevent the requirement to specify soil albedo in multiple places when using a TwoStream radiative transfer model. Will resolve #271

## Content

- Creates the type `AbstractSoilDriver` with 3 different concrete types, and requires a soil driver of this type to be input into the canopy model
- There are now `PrescribedSoil` (fully prescribed by the user), `PrognosticSoil` (for use with SoilEnergyHydrology), and `PrognosticSoilHydrology` (Hydrology only is prognostic and user needs to specify albedo (for SoilPlantHydrologyModel)) which can be the `soil_driver` for the CanopyModel
- Integrated models edited to use these types
- Ozark experiment and canopy tutorials updated to use these types

There are a couple things we may want to think about still before merging
- `plant_absorbed_ppfd` is now a function of the canopy model since it requires the `soil_driver` from the canopy. If it isn't, then we need to supply the soil_driver to the TwoStream model only when soil is prescribed which seems confusing for a user. However, this would mean that `plant_absorbed_ppfd` should be moved out of the parameterizations file and the parameterizations tests need to change.
    - One alternative would be to use the canopy model similar to how we use the integrated CanopySoilModel... provide arguments for the component models to the constructor of the canopy model, then have the CanopyModel instantiate the component models and supply the soil driver to the ones that need it. This way the user would only have to input the soil driver once, but it would be supplied to the hydraulic model and radiative_transfer model where needed.
- Similarly as it stands now, we have to supply the soil_driver to the Hydraulics model only when running standalone or `SoilPlantHydrologyModel`, which seems confusing for a user

----

### DIFF
--- a/docs/tutorials/Canopy/canopy_tutorial.jl
+++ b/docs/tutorials/Canopy/canopy_tutorial.jl
@@ -109,17 +109,31 @@ shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
     earth_param_set,
 );
 
+# For this canopy, we are running in standalone mode, which means we need to 
+# use a prescribed soil driver, defined as follows:
+
+ψ_soil0 = FT(0.0)
+
+soil_driver = PrescribedSoil(
+    root_depths = FT.(-Array(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0),
+    ψ_soil = t -> eltype(t)(ψ_soil0),
+    soil_α = FT(0.2),
+);
+
 # Now, setup the canopy model by component.
 # Provide arguments to each component, beginning with radiative transfer:
 
-rt_params = BeerLambertParameters{FT}(;
-    Ω = FT(0.69),
+rt_params = TwoStreamParameters{FT}(;
     ld = FT(0.5),
     ρ_leaf = FT(0.1),
+    τ = FT(0.1),
+    Ω = FT(0.69),
     λ_γ = FT(5e-7),
+    n_layers = UInt64(20),
+    diff_perc = FT(0.5),
 )
 
-rt_model = BeerLambertModel{FT}(rt_params);
+rt_model = TwoStreamModel{FT}(rt_params);
 
 # Arguments for conductance model:
 
@@ -159,7 +173,7 @@ photosynthesis_model = FarquharModel{FT}(photo_params);
 # Arguments for plant hydraulics model are more complicated. 
 
 # Begin by providing general plant parameters. For the area
-#indices of the canopy, we choose a `PrescribedSiteAreaIndex`,
+# indices of the canopy, we choose a `PrescribedSiteAreaIndex`,
 # which supports LAI as a function of time, with RAI and SAI as constant.
 LAI = 4.2
 LAIfunction = (t) -> eltype(t)(LAI)
@@ -203,19 +217,10 @@ plant_hydraulics_ps = PlantHydraulics.PlantHydraulicsParameters(;
     retention_model = retention_model,
 );
 
-# Define the remaining variables required for the plant hydraulics model. We 
-# will use prescribed soil conditions here, but the model can be coupled 
-# with a soil hydraulics simulation as well.
-
-root_depths = FT.(-Array(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0)
-
-ψ_soil0 = FT(0.0)
-soil_potential = t -> eltype(t)(ψ_soil0)
-root_extraction = PrescribedSoilPressure{FT}(root_depths, soil_potential)
+# Define the remaining variables required for the plant hydraulics model.
 
 plant_hydraulics = PlantHydraulics.PlantHydraulicsModel{FT}(;
     parameters = plant_hydraulics_ps,
-    root_extraction = root_extraction,
     n_stem = n_stem,
     n_leaf = n_leaf,
     compartment_surfaces = compartment_surfaces,
@@ -223,9 +228,10 @@ plant_hydraulics = PlantHydraulics.PlantHydraulicsModel{FT}(;
 );
 
 # Now, instantiate the canopy model, using the atmospheric and radiative 
-# drivers included from the external file. This contains every piece of 
-# information needed to generate the set of ODEs modeling the canopy
-# biophysics, ready to be passed off to a timestepper.
+# drivers included from the external file, as well as the soil driver we 
+# instantiated above. This contains every piece of information needed to 
+# generate the set of ODEs modeling the canopy biophysics, ready to be passed 
+# off to a timestepper.
 
 canopy = ClimaLSM.Canopy.CanopyModel{FT}(;
     parameters = shared_params,
@@ -234,6 +240,7 @@ canopy = ClimaLSM.Canopy.CanopyModel{FT}(;
     photosynthesis = photosynthesis_model,
     conductance = stomatal_model,
     hydraulics = plant_hydraulics,
+    soil_driver = soil_driver,
     atmos = atmos,
     radiation = radiation,
 );
@@ -270,7 +277,7 @@ end;
 t0 = FT(0)
 N_days = 365
 tf = t0 + FT(3600 * 24 * N_days)
-dt = FT(225)
+dt = FT(225);
 
 # Initialize the auxiliary variables for the canopy using the initial 
 # conditions and initial time.
--- a/src/integrated/soil_plant_hydrology_model.jl
+++ b/src/integrated/soil_plant_hydrology_model.jl
@@ -41,6 +41,15 @@ A constructor for the `SoilPlantHydrologyModel`, which takes in the concrete mod
 type and required arguments for each component, constructs those models,
 and constructs the `SoilPlantHydrologyModel` from them.
 
+In this model, the soil_driver, of type PrognosticSoil, must be defined by the 
+user and passed in within the land_args tuple. The soil_driver is used to
+provide the canopy model with the soil albedo, and for the full soil + canopy 
+model, the soil model contains the albedo, and the constructor for the model is 
+able to define and pass the soil_driver, but here for the soil model, the soil 
+albedo is not defined, and therefore the user must manually define the
+soil driver and provide it in land_args.
+
+
 Each component model is constructed with everything it needs to be stepped
 forward in time, including boundary conditions, source terms, and interaction
 terms.
@@ -59,11 +68,10 @@ function SoilPlantHydrologyModel{FT}(;
 ) where {FT, SM <: Soil.AbstractSoilModel{FT}}
 
     # These may be passed in, or set, depending on use scenario.
-    (; atmos, radiation) = land_args
+    (; soil_driver, atmos, radiation) = land_args
     precipitation = atmos.liquid_precip
     # These should always be set by the constructor.
     sources = (RootExtraction{FT}(),)
-    root_extraction = PrognosticSoilPressure{FT}()
 
     boundary_conditions = (;
         top = (water = FluxBC((p, t) -> eltype(t)(precipitation(t))),),
@@ -93,10 +101,10 @@ function SoilPlantHydrologyModel{FT}(;
             canopy_component_args.conductance...,
         ),
         hydraulics = canopy_component_types.hydraulics(;
-            root_extraction = root_extraction,
             transpiration = transpiration,
             canopy_component_args.hydraulics...,
         ),
+        soil_driver = soil_driver,
         atmos = atmos,
         radiation = radiation,
         canopy_model_args...,
@@ -162,27 +170,10 @@ function make_interactions_update_aux(
     return update_aux!
 end
 
-
-## Extending methods of the Plant Hydraulics and Soil Model
-## TBD if these should be linked more explicitly.
-"""
-   PrognosticSoilPressure{FT} <: PlantHydraulics.AbstractRootExtraction{FT}
-
-Concrete type of PlantHydraulics.AbstractRootExtraction, used for dispatch 
-in an LSM with both soil and plant hydraulic components.
-
-This is paired with the source term `Soil.RootExtraction`:both 
-are used at the same time,
-ensuring that the water flux into the roots is extracted correctly
-from the soil.
-"""
-struct PrognosticSoilPressure{FT} <:
-       Canopy.PlantHydraulics.AbstractRootExtraction{FT} end
-
 """
     PlantHydraulics.root_flux_per_ground_area!(
         fa::ClimaCore.Fields.Field,
-        re::PrognosticSoilPressure{FT},
+        s::PrognosticSoil,
         model::Canopy.PlantHydraulics.PlantHydraulicsModel{FT},
         Y::ClimaCore.Fields.FieldVector,
         p::ClimaCore.Fields.FieldVector,
@@ -192,7 +183,7 @@ struct PrognosticSoilPressure{FT} <:
 An extension of the `PlantHydraulics.root_flux_per_ground_area!` function,
  which returns the
 net flux of water between the
-roots and the soil, per unti ground area, 
+roots and the soil, per unit ground area, 
 when both soil and plant
 hydraulics are modeled prognostically. This is for use in an LSM.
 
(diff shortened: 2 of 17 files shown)


kmdeck:
hey! maybe we can discuss a little more in person...

for now, one simplification might be that this function:
https://github.com/CliMA/ClimaLSM.jl/blob/4635205ff88ec6de9e27c2b2ffd8d545987b814f/src/standalone/Vegetation/component_models.jl#L110

is defined such that we can also pass in the canopy model if we need to. right now we dont take advantage of that:
https://github.com/CliMA/ClimaLSM.jl/blob/4635205ff88ec6de9e27c2b2ffd8d545987b814f/src/standalone/Vegetation/PlantHydraulics.jl#L536

but we could, and then we only need to store the `soil_driver` in the Canopy model and not also in plant hydraulics

src/standalone/Vegetation/soil_drivers.jl
```diff
@@ -0,0 +1,47 @@
...
+
+"""
+   PrognosticSoil{FT} <: AbstractSoilDriver{FT}
+
+Concrete type of AbstractSoilDriver used for dispatch in cases where both 
+a canopy model and soil model are run. Conains a pointer to the integrated 
+soil model to allow canopy components to access the soil parameters
+"""
+struct PrognosticSoil{FT} <: AbstractSoilDriver{FT}
+    PrognosticSoilModel::AbstractSoilModel{FT}
```
kmdeck, line 46: just a thought.
we dont necessarily need to store the soil model here, I think. We could perhaps just store the albedo? The instance is made in the soil_canopy constructor so we can extract soil.parameters.albedo there to make it.

Espeer5: Yes you're right we can do that and then we can manipulate that definition to still allow us to use the SoilPlantHydrology model internally as well

Espeer5:
> hey! maybe we can discuss a little more in person...
> 
> for now, one simplification might be that this function:
> 
> https://github.com/CliMA/ClimaLSM.jl/blob/4635205ff88ec6de9e27c2b2ffd8d545987b814f/src/standalone/Vegetation/component_models.jl#L110
> 
> is defined such that we can also pass in the canopy model if we need to. right now we dont take advantage of that:
> 
> https://github.com/CliMA/ClimaLSM.jl/blob/4635205ff88ec6de9e27c2b2ffd8d545987b814f/src/standalone/Vegetation/PlantHydraulics.jl#L536
> 
> but we could, and then we only need to store the `soil_driver` in the Canopy model and not also in plant hydraulics



> 
Went ahead and did this and it definitely helped clean things up a lot

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -66,18 +67,22 @@ of mol photons per m^2 per second (`APAR`).
...
+the canopy model, along with the incident PAR, LAI, extinction coefficient K, 
+and solar zenith angle.
 """
 function plant_absorbed_ppfd(
-    RT::TwoStreamModel,
+    canopy::CanopyModel{FT, TwoStreamModel{FT}},
     PAR::FT,
     LAI::FT,
     K::FT,
     θs::FT,
```
kmdeck, line 79: A couple of other options for these two methods:
- we can dispatch off of parametric types as well. so, I think that if we do pass canopy as an argument, we dont additionally need to pass the RT model as an argument, because we can dispatch off of the parametric type in canopy. A little care needs to be taken to make sure this works properly though. maybe something like...```canopy::CanopyModel{FT, <: TwoStreamModel}```
- however, I think we may also consider passing in albedo (a float) as the last argument and dispatching off of the radiative transfer model, and *not* passing in canopy (unless we think we need the whole model for another reason). The method for plant_absorbed_ppfd() for the BeerLambertModel can then be
```
plant_absorbed_ppfd(
    RT::BeerLambertModel,
    PAR::FT,
    LAI::FT,
    K::FT,
    θs::FT
     _)
```
i.e. it doesnt use the last argument (albedo), even though we always pass it. And, prior to this function being called (in Canopy.jl), we compute albedo to pass in. The reason why I suggest this is because albedo will eventually vary in space, and be stored in the aux state, and not just be a parameter that is set once. 

I am open to leaving this PR as is, though, and addressing this in a future PR when we get there :)

src/standalone/Vegetation/soil_drivers.jl
```diff
@@ -0,0 +1,35 @@
+using ..ClimaLSM.Soil
```
kmdeck, line 1: we can remove this now

test/integrated/lsms.jl
```diff
@@ -61,7 +61,7 @@ using ClimaLSM
         return update_aux!
     end
 
-    function ClimaLSM.make_set_initial_aux_state(m::DummyModel1{FT}) where {FT}
+    function ClimaLSM.make_set_initial_aux_state(m::DummyModel{FT}) where {FT}
```
kmdeck, line 64: typo? should this be changed?

test/standalone/Vegetation/test_two_stream.jl
```diff
@@ -67,7 +66,7 @@ py_FAPAR = FT.(test_set[2:end, column_names .== "FAPAR"])
 
         # Compute the predicted FAPAR using the ClimaLSM TwoStream implementation
         K = extinction_coeff(ld[i], θs[i])
-        FAPAR = plant_absorbed_ppfd(RT, FT(1), LAI[i], K, θs[i])
+        FAPAR = plant_absorbed_ppfd(RT, FT(1), LAI[i], K, θs[i], a_soil[i])
```
kmdeck, line 69: is a_soil an array? (should it be a constant?)

test/integrated/lsms.jl
Espeer5: With the 1 there, there are two definitions of make_set_initial_aux_state for the type `DummyModel1` and none for `DummyModel`, and the tests fail... But I am not sure why the same tests pass on main

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -42,19 +44,20 @@ BeerLambertModel, along with the PAR, LAI, extinction coefficient K, and solar
...
 """
 function plant_absorbed_ppfd(
-    RT::BeerLambertModel,
+    RT::BeerLambertModel{FT},
     PAR::FT,
     LAI::FT,
     K::FT,
-    θs::FT,
+    _,
+    _,
```
kmdeck, line 52: I think just one extra arg

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -23,11 +24,12 @@ export plant_absorbed_ppfd,
...
 
 """
-    plant_absorbed_ppfd(RT::BeerLambertModel,
+    plant_absorbed_ppfd(RT::BeerLambertModel{FT},
                         PAR::FT,
                         LAI::FT,
                         K::FT,
-                        θs::FT,
+                        _,
+                        _,
```
kmdeck, line 32: likewise here

test/standalone/Vegetation/test_two_stream.jl
Espeer5: These tests just run the TwoStream calculation for a large range of input values and compare to the python TwoStream output for the same values, so it reads the soil albedo from the data file for all of the test cases as an array with one entry for each test case. But this is only specific to these tests it is typically just a constant.

src/standalone/Vegetation/canopy_parameterizations.jl
Espeer5: The method for the BeerModel doesn't use the solar zenith angle (it is only used for the computation of `K` which is done outside the function), I actually added it into the function signature when I added the TwoStreamModel. Since it isn't used within the function anywhere should it be an extra arg here like this?

src/standalone/Vegetation/Canopy.jl
```diff
@@ -99,6 +103,8 @@ struct CanopyModel{FT, RM, PM, SM, PHM, A, R, PS, D} <: AbstractExpModel{FT}
     atmos::A
     "Radiative forcing: prescribed or coupled"
     radiation::R
+    "Soil pressure: prescribed or coupled"
```
kmdeck, line 106: "Prescribed or Prognostic" to match naming

src/standalone/Vegetation/Canopy.jl
```diff
@@ -77,6 +78,9 @@ fluxes.
 - The radiative flux conditions, which are either prescribed
 (of type `PrescribedRadiativeFluxes`) or computed via a coupled simulation
 (of type `CoupledRadiativeFluxes`).
+- The soil conditions, which are either prescribed (of type 'PrescribedSoil') or 
+fully or partially computed via a coupled simulation (of type
```
kmdeck, line 82: clarify a little?
"The soil conditions, which are either prescribed (of type `PrecribedSoil`, for running the canopy model in standalone mode), or prognostic (of type `PrognosticSoil`, for running integrated soil+canopy models)"

src/integrated/soil_plant_hydrology_model.jl
```diff
@@ -162,27 +161,10 @@ function make_interactions_update_aux(
...
-from the soil.
-"""
-struct PrognosticSoilPressure{FT} <:
-       Canopy.PlantHydraulics.AbstractRootExtraction{FT} end
-
 """
     PlantHydraulics.root_flux_per_ground_area!(
         fa::ClimaCore.Fields.Field,
-        re::PrognosticSoilPressure{FT},
+        s::Union{PrognosticSoil, PrognosticSoilHydrology},
```
kmdeck, line 167: can remove PrognosticSoilHydrology

src/integrated/soil_plant_hydrology_model.jl
```diff
@@ -59,11 +59,10 @@ function SoilPlantHydrologyModel{FT}(;
 ) where {FT, SM <: Soil.AbstractSoilModel{FT}}
 
     # These may be passed in, or set, depending on use scenario.
-    (; atmos, radiation) = land_args
+    (; soil_driver, atmos, radiation) = land_args
```
kmdeck, line 62: just to be sure I am following, here we still need to pass it in because we need to get the albedo from the user, while in the soil_canopy_model we can set it in the constructor because the albedo is set by the user as part of the soil model?

kmdeck: Could we add a note in the docstring for this model to make that clear?

src/standalone/Vegetation/soil_drivers.jl
```diff
@@ -0,0 +1,35 @@
...
+
+"""
+     PrescribedSoil{} <: AbstractSoilDriver
+
+A container for holding prescribed soil parameters needed by the canopy model
+when running the canopy in standalone mode, including the soil pressure and 
+albedo.
+$(DocStringExtensions.FIELDS)
+"""
+struct PrescribedSoil{FT} <: AbstractSoilDriver{FT}
```
kmdeck, line 17: What do you think about defining an outer constructor for this which takes these as keyword arguments? it might be a little clearer in usage.

src/standalone/Vegetation/soil_drivers.jl
```diff
@@ -0,0 +1,35 @@
...
+end
+
+"""
+   PrognosticSoil{FT} <: AbstractSoilDriver{FT}
+
+Concrete type of AbstractSoilDriver used for dispatch in cases where both 
+a canopy model and soil model are run. Contains the soil albedo to be shared 
+between the canopy and soil models
+"""
+struct PrognosticSoil{FT} <: AbstractSoilDriver{FT}
```
kmdeck, line 33: What do you think about defining an outer constructor for this which takes this as a keyword argument? it might be a little clearer in usage.

I wonder if we could also add to the docstring something like this:

"When running the SoilCanopyModel, the soil model specifies the albedo. In this case, the constructor for the model reads the albedo from the soil model and the user only specifies it once, for the soil model. When running the SoilPlantHydrologyModel - which is intended for internal usage/testing primarily - the user must specify the albedo because the soil model in that case does not have an albedo."

I dont know if that would make it more confusing :P but im just thinking about how to keep all of this clear.

kmdeck (commented):
almost there! one last thing is to add to the APIs, here:
https://github.com/CliMA/ClimaLSM.jl/blob/main/docs/src/APIs/canopy/Canopy.md

(we should add the 3 soil driver types (abstract, prescr, prognostic) to this, then their docstrings will get included in the github docs)

docs/tutorials/Canopy/canopy_tutorial.jl
```diff
@@ -109,17 +109,31 @@ shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
     earth_param_set,
 );
 
+# For this canopy, we are running in standalone mode, which means we need to 
+# use a prescribed soil driver, defined as follows:
+
+ψ_soil0 = FT(0.0)
+
+soil_driver = PrescribedSoil(
+    FT.(-Array(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0),
```
kmdeck, line 118: for example, using kwargs here would be nice to clarify what these are

test/integrated/lsms.jl
kmdeck: Hm, just a definition of make_update_aux and make_set_initial_aux_state for DummyModel1. For DummyModel, we are using the default (defined in src/shared_utilities/model.jl).

test/standalone/Vegetation/test_two_stream.jl
kmdeck: thanks!

src/standalone/Vegetation/canopy_parameterizations.jl
kmdeck: oh, yes! I missed that theta_s was dropped. this makes sense, thank you.

just fyi - you can define methods in Julia with a different # of arguments, but here we always call the function with the same arguments (from canopy.jl). so I think we have to define them with the same # of arguments, otherwise I think we'd run into "No Method Matching" type errors...

test/integrated/lsms.jl
Espeer5: Oh yes you're right it is not a duplicate definition, I can't read.  One of the tests still fails if this is `DummyModel1` though, and I don't know why. I will reset this back to what it was before and see if I can figure out what's going on.

src/standalone/Vegetation/soil_drivers.jl
kmdeck: (note - the default inner constructor will still work, so you dont need to change the tests. maybe just the tutorials and experiments can use the kwargs constructor?)

kmdeck:
I noticed a couple of lines in the two tutorials which print their output and can have a `;` after them. You can take a look here: 
https://clima.github.io/ClimaLSM.jl/previews/PR279/

test/integrated/lsms.jl
Espeer5: Ok well I must have an issue with my local manifest or something because after changing it back it still fails locally on my machine but passes here.

kmdeck (approved):

Espeer5:
Apparently failed the first time due to slow download speed from Box? But passed the same check running outside of the merge so re-running.

PR #282, opened 2023-08-02
Fix `interpol` bug

juliasloan25:
## Purpose 
We have a function `interpol` in the FileReader module that performs interpolation between two values. However, the logic is switched so the interpolation is incorrect. This PR fixes the logic and adds a test to verify correctness.

## Content
- [x] fix interpol logic
- [x] add test

----


kmdeck (approved):
looks great, thank you!

PR #281, opened 2023-08-02
improving comparison to data at Ozark site

kmdeck:
## Purpose 
Test hypotheses for improving the simulation.

## To-do
1. increase amplitude of GPP and T in summer [leaf water potential too low -> stomatal conductance too low?]
2. decrease soil evaporation early in the year [issue with soil temperature or soil parameters? or influence of canopy which is not taken into account?]
3. improve sensible heat flux errors

## Content
- made nicer plots

Review checklist

I have:
- followed the codebase contribution guide: https://clima.github.io/ClimateMachine.jl/latest/Contributing/
- followed the style guide: https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/
- followed the documentation policy: https://github.com/CliMA/policies/wiki/Documentation-Policy
- checked that this PR does not duplicate an open PR.

In the Content, I have included 
- relevant unit tests, and integration tests, 
- appropriate docstrings on all functions, structs, and modules, and included relevant documentation.

----
- [ ] I have read and checked the items on the review checklist.

PR #284, opened 2023-08-04
Nir abs

Espeer5:
## Purpose 
 Compute not only the APAR but also the NIR absorption using the radiative transfer models

## To-do

- [ ] Potentially test out Yujie's suggestion on how to include Beta to see if it improves Ozark GPP?
- [ ] Fix the tutorials which now need the PAR and NIR parameters setup and used

## Content
 In order to calculate the absorbed NIR it was necessary to:
- Add soil albedos for both NIR and PAR to the soil model and canopy soil drivers
    - This required a slight tweak to `surface_albedo` for the soil model since SW albedo = .5 * a_NIR + .5 * a_PAR
- Edit `plant_absorbed_ppfd` so that it computes absorption for any spectral band based on the reflectances and incident radiation passed into it for BeerLambert and TwoStream
    - (i.e if you pass in NIR reflectance and incident NIR, you get NIR absorption)
- absorbed NIR, NIR added as auxiliary variables of the radiative transfer scheme
- Add the absorbed NIR to the calculation in `net_radiation_at_ground`

I also input the NIR and PAR albedos, reflectances, transmittances, etc from values used in CLM from [their docs](https://www2.cesm.ucar.edu/models/cesm2/land/CLM50_Tech_Note.pdf) to the parameters used for the Ozark site.

I also went ahead and put in the changes to the plotting and such from #281 so that this is the fully updated version of the Ozark site.

Here are the updated plots for Ozark including the computed NIR absorption
cumul_p_et
ET
GPP
leaf_water_potential
soil_water_content
soil_temperature

----


src/standalone/Vegetation/radiation.jl
```diff
@@ -14,8 +14,10 @@ $(DocStringExtensions.FIELDS)
 struct BeerLambertParameters{FT <: AbstractFloat}
     "Leaf angle distribution function (unitless)"
     ld::FT
-    "PAR canopy reflectance (unitless)"
+    "PAR leaf reflectance (unitless)"
```
kmdeck, line 17: could we use alpha for both? or rho for both? and is rho the same as alpha PAR?

src/standalone/Vegetation/radiation.jl
```diff
@@ -75,7 +81,9 @@ end
     function TwoStreamParameters{FT}(;
         ld = FT(0.5),
         ρ_leaf = FT(0.3),
-        τ = FT(0.2),
+        τ_leaf = FT(0.05),
```
kmdeck, line 84: 0.2

src/standalone/Vegetation/radiation.jl
Espeer5: Yes sorry we can and should do that, I had changed that originally but then when I reverted to get rid of the SAI I had added I forgot to re-change it whoops

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -24,71 +24,82 @@ export plant_absorbed_ppfd,
...
 
 This applies the two-stream radiative transfer solution which takes into account
 the impacts of scattering within the canopy. The function takes in all 
 parameters from the parameter struct of a TwoStreamModel, along with the 
-incident PAR, LAI, extinction coefficient K, soil albedo from the 
+incident radiation, LAI, extinction coefficient K, soil albedo from the 
 canopy soil_driver, and solar zenith angle.
 """
-function plant_absorbed_ppfd(
+function plant_absorbed_pfd(
```
kmdeck, line 86: ppfd

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -114,8 +125,8 @@ function plant_absorbed_ppfd(
     u₂ = b - c * a_soil
     u₃ = f + c * a_soil
 
-    s₁ = exp(-h * LAI * Ω)
-    s₂ = exp(-K * LAI * Ω)
+    s₁ = exp(-min(h * LAI * Ω, 40))
```
kmdeck, line 128: where does the 40 come from?

src/standalone/Vegetation/Canopy.jl
```diff
@@ -411,15 +423,28 @@ function ClimaLSM.make_update_aux(
...
         # update radiative transfer
+        RT = canopy.radiative_transfer
         K = extinction_coeff(ld, θs)
-        PAR .= compute_PAR(canopy.radiative_transfer, canopy.radiation, t)
-        @. APAR = plant_absorbed_ppfd(
-            canopy.radiative_transfer,
+        RTP = RT.parameters
+        PAR .= compute_PAR(RT, canopy.radiation, t)
+        NIR .= compute_NIR(RT, canopy.radiation, t)
+        @. APAR = plant_absorbed_pfd(RT,
```
kmdeck, line 431: oh ppfd or pfd, whatever makes sense to @braghiere !

kmdeck (approved):
minor comments, looks good!

src/standalone/Vegetation/canopy_parameterizations.jl
Espeer5: Hmm, I changed to pfd (ppfd = photosynthetic flux density, and we are now using this for all flux densities, not just photosynthetic), but if this should be named differently let me know.

kmdeck:
oh -  one other comment. we use 5e-7 as the typical wavelength of a photon in order to convert from energy to moles of photons. @braghiere is it OK to use this value for both PAR and NIR? or should we use a different value for each?

src/standalone/Vegetation/canopy_parameterizations.jl
Espeer5: CLM... https://www2.cesm.ucar.edu/models/cesm2/land/CLM50_Tech_Note.pdf equation 3.40

It may not be necessary, when I had added the SAI I did so according to these equations which is why this got changed. When I took the SAI back out I probably should have put this back.

src/standalone/Vegetation/canopy_parameterizations.jl
kmdeck: sounds good! just a note to change here as well:https://github.com/CliMA/ClimaLSM.jl/blob/main/docs/src/APIs/canopy/RadiativeTransfer.md

braghiere:
> oh - one other comment. we use 5e-7 as the typical wavelength of a photon in order to convert from energy to moles of photons. @braghiere is it OK to use this value for both PAR and NIR? or should we use a different value for each?

Good catch! Using the middle of the Near-Infrared (NIR) range as a reference wavelength is a reasonable approach. The middle of the NIR range would be approximately around 1600 to 1700 nm.

experiments/integrated/ozark/ozark.jl
```diff
@@ -55,7 +55,8 @@ soil_ps = Soil.EnergyHydrologyParameters{FT}(;
     z_0m = z_0m_soil,
     z_0b = z_0b_soil,
     emissivity = soil_ϵ,
-    albedo = soil_α,
+    PAR_albedo = soil_α,
```
braghiere, line 58: Should this be soil_alpha_PAR?

src/integrated/soil_canopy_model.jl
```diff
@@ -309,8 +312,11 @@ function net_radiation_at_ground(
     N_a = FT(LSMP.avogadro_constant(earth_param_set))
     (; λ_γ) = canopy_radiation.parameters
     energy_per_photon = h * c / λ_γ
-    SW_d_beneath_canopy =
-        @. SW_d - p.canopy.radiative_transfer.apar * (energy_per_photon * N_a)
+    SW_d_beneath_canopy = @. SW_d - (
+        (energy_per_photon * N_a) * (
```
braghiere, line 316: In here we should use the NIR average wavelength 1650nm.

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -73,7 +75,8 @@ function EnergyHydrologyParameters{FT}(;
     K_sat::FT,
     S_s::FT,
     θ_r::FT,
-    albedo = FT(0.2),
+    PAR_albedo = FT(0.2),
+    NIR_albedo = FT(0.2),
```
braghiere, line 79: Is this surface albedo or soil albedo?

src/standalone/Vegetation/soil_drivers.jl
```diff
@@ -35,9 +37,10 @@ specify the soil parameters by keyword arguments.
 function PrescribedSoil(;
     root_depths::Vector{FT} = FT.(-Array(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0),
     ψ_soil::Function = t -> eltype(t)(FT(0.0)),
-    soil_α::FT = FT(0.2),
+    soil_α_PAR::FT = FT(0.2),
+    soil_α_NIR::FT = FT(0.2),
```
braghiere, line 41: Soil alpha NIR is usually higher than soil alpha PAR. This should be around 0.4.

src/standalone/Vegetation/soil_drivers.jl
```diff
@@ -65,6 +71,9 @@ end
 An outer constructor for the PrognosticSoil soil driver allowing the user to
 specify the soil albedo by keyword argument.
 """
-function PrognosticSoil(; soil_α::FT = FT(0.2)) where {FT}
-    return PrognosticSoil{FT}(soil_α)
+function PrognosticSoil(;
+    soil_α_PAR::FT = FT(0.2),
+    soil_α_NIR = FT(0.2),
```
braghiere, line 76: Same here. Soil alpha NIR 0.4.

braghiere (changes requested):
Looks good. I added a few comments regarding NIR albedos and representative wavelengths.

src/standalone/Soil/energy_hydrology.jl
kmdeck: soil. It is stored in SoilParameters, so I think the naming is OK

braghiere: So it should be something like: 
SW_albedo = 0.5 * PAR_albedo + 0.5 * NIR_albedo
PAR_albedo = FT(0.2);
NIR_albedo = FT(0.4);

Espeer5:
> hey! I noticed that you are force pushing most of your commits. any reason why?

Just for convenience for squashing. When I build the feature I make seperate commits to have a history, but then when I am just fixing the formatting and such I just git commit --amend and then I have to force push since this makes the commit history diverge. This just gives me less commits to have to squash before merging.

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -155,11 +168,11 @@ function plant_absorbed_ppfd(
     h₉ = (u₂ + μ̄ * h) / (d₂ * s₁)
     h₁₀ = -s₁ * (u₂ - μ̄ * h) / d₂
 
-    # Compute the LAI per layer for this canopy
+    # Compute the (LAI + SAI) per layer for this canopy
```
kmdeck, line 171: minor, could revert this comment back

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -24,71 +24,84 @@ export plant_absorbed_ppfd,
...
+function plant_absorbed_pfd(
     RT::TwoStreamModel{FT},
-    PAR::FT,
+    α_leaf,
+    SW_IN::FT,
     LAI::FT,
     K::FT,
+    τ_leaf,
     θs::FT,
     a_soil::FT,
```
kmdeck, line 96: could call alpha_soil instead of a_soil for consistency

kmdeck: (\alpha)

src/standalone/Vegetation/Canopy.jl
```diff
@@ -411,15 +425,33 @@ function ClimaLSM.make_update_aux(
...
-        @. APAR = plant_absorbed_ppfd(
-            canopy.radiative_transfer,
-            PAR / (energy_per_photon * N_a),
+        RTP = RT.parameters
+        PAR .= compute_PAR(RT, canopy.radiation, t)
+        NIR .= compute_NIR(RT, canopy.radiation, t)
+        # τ is defined only for TwoStream model, not for BeerLambert
+        τ_PAR = typeof(RT) == TwoStreamModel ? RTP.τ_PAR_leaf : FT(0)
+        τ_NIR = typeof(RT) == TwoStreamModel ? RTP.τ_NIR_leaf : FT(0)
+        @. APAR = plant_absorbed_pfd(
```
kmdeck, line 436: Im wondering if it makes more sense to pass RT, PAR/(energy per photon * Na), LAI, K, theta_s, ground_albedo only as args, and then inside the `plant_absorbed_pfd` function, we can unpack what we need from RT.parameters.

then, because we have two methods for this function, we can unpack tau in one case and not the other.

kmdeck: oh, hm, nvm. this wont work because you need to know which \tau to unpack, etc. (NIR or PAR).

kmdeck: Im not sure. perhaps we can leave it alone for now because we'll revisit this in the future at some point if we have an arbitrary number of wavelength bands

kmdeck: but if you have ideas about how to streamline, please feel free!

Espeer5: The only problem is how to tell `plant_absorbed_pfd` which reflectivity and transmittance to use. When we call it for PAR, we use the PAR values, for NIR, the NIR values. We would need to pass another argument which specified which stream to calculate for.

I originally tried to solve this by writing `calculate_absorbances` which dispatched off the type of RT and made the calls to `plant_absorbed_pfd` with the appropriate arguments and then returned a tuple of (APAR, ANIR), but I couldn't get this to work neatly with the broadcasting.

test/standalone/Soil/climate_drivers.jl
```diff
@@ -150,7 +152,8 @@ include(joinpath(pkgdir(ClimaLSM), "parameters", "create_parameters.jl"))
         @test ClimaLSM.surface_emissivity(model, Y, p) == emissivity
         @test ClimaLSM.surface_evaporative_scaling(model, Y, p) == FT(1)
         @test ClimaLSM.surface_height(model, Y, p) == z_sfc
-        @test ClimaLSM.surface_albedo(model, Y, p) == albedo
+        @test ClimaLSM.surface_albedo(model, Y, p) ==
+              FT((0.5 * PAR_albedo) + (0.5 * NIR_albedo))
```
kmdeck, line 156: small note. `a*0.5` will always convert to Float64, but if you write `a/2` it will preserve the type of `a`.

src/standalone/Vegetation/Canopy.jl
kmdeck: nice idea! i think we have an example of something similar here: https://github.com/CliMA/ClimaLSM.jl/blob/386572b94019657ee24b6463ee81c79e3dfa3868/src/standalone/Soil/boundary_conditions.jl#L196
We do the broadcasting inside of this function, and return the fields. 

when we use the function, we dont broadcast: https://github.com/CliMA/ClimaLSM.jl/blob/386572b94019657ee24b6463ee81c79e3dfa3868/src/standalone/Soil/energy_hydrology.jl#L201C13-L201C13

Espeer5: Ah! Thanks this helps a lot.

experiments/integrated/ozark/ozark_parameters.jl
```diff
@@ -25,16 +25,23 @@ soil_vg_α = FT(0.04) # inverse meters
...
-ρ_leaf = FT(0.1)
-λ_γ = FT(5e-7)
+α_PAR_leaf = FT(0.1)
+λ_γ_PAR = FT(5e-7)
+λ_γ_NIR = FT(1.65e-6)
+τ_PAR_leaf = FT(0.05)
+α_NIR_leaf = FT(0.45)
+τ_NIR_leaf = FT(0.25)
+n_layers = UInt64(20)
+diff_perc = FT(0.2)
```
braghiere, line 41: The percentage of diffuse light can have a large impact on energy fluxes. We need to read that from a file or parameterize it. Does Ozark have incident diffuse radiation measurements?

src/standalone/Soil/energy_hydrology.jl
braghiere, line 79: Was this solved? NIR_albedo = 0.4?

braghiere (commented):
I added a few comments on i) NIR_albedo and ii) diffuse radiation percentage. Apart from that, everything looks good.

experiments/integrated/ozark/ozark_parameters.jl
kmdeck: would it be ok to do this in a followup PR? adding the infrastructure to pass in diffuse fraction or fall back on an equation will require some more changes. this PR is already quite large.

braghiere (approved):

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -24,71 +24,164 @@ export plant_absorbed_ppfd,
...
+    α_leaf::FT,
     LAI::FT,
     K::FT,
-    _,
-    _,
 ) where {FT}
     RTP = RT.parameters
-    APAR = PAR * (1 - RTP.ρ_leaf) * (1 - exp(-K * LAI * RTP.Ω))
-    return APAR
+    AR = SW_IN * (1 - α_leaf) * (1 - exp(-K * LAI * RTP.Ω))
```
AlexisRenchon, line 141: Not a big deal but for consistency, we could agree to use either "radiation" (W m-2) or "photon flux density" (umol m-2 s-1) throughout. 

Here we call the function "PFD" but return "AR".

It sounds like we could use "radiation" everywhere, since the term PAR and APAR are more common. We should in theory therefore stick to (W m-2) units, but we could do like everyone in ecology and wrongfully attribute PAR units of (umol m-2 s-1).

AlexisRenchon (approved):

Issue #288, opened 2023-08-07
DSS handling behind the scenes


kmdeck:
**Is your feature request related to a problem? Please describe.**
- We currently allocate extra memory each step when we carry out dss. We need to allocate a buffer and use that in the weighted_dss! call. #292 Note that we need to use dss *whenever* we have a 3D or 2D space, not just when there are horizontal derivatives, and that we need a different buffer for 2d vs 3d fields.
- rename the current `domain(model)` function to `domain_name(model)`. #291

PR #292, opened 2023-08-07
dss buffer

kmdeck:
## Purpose 
See #288

Substantive changes are in shared_utilities: models.jl, utils.jl, ClimaLSM.jl,src/standalone/Vegetation/Canopy.jl, and  src/standalone/Bucket/Bucket.jl.

We had to remove a tutorial but will be rewriting it and adding it back in #296 

## To-do
[X] handle 2d vs 3d buffer
[X] write unit test for new functions, and added tests to make sure the aux state was properly made for standalone models, integrated models, and the bucket model.
[] have all integrations use dss (separate PR)

## Content
1. Adding the dss buffer to the aux state. We have three methods for `initalize_aux`, and one special method for `initialize`, so four files have changes.
- src/shared_utilities/models.jl: since `initialize_vars` no longer returns a field vector, convert the output to a FieldVector within `initialize_prognostic`, but not in `initialize_auxiliary`. Updated docs for `initialize_auxiliary`. In `initialize_auxiliary`, add the dss buffer to `p`. This is the default `initialize_auxiliary` method.
- src/ClimaLSM.jl: in `initialize_auxiliary` method for AbstractLandModels, cycle through the (unique) domains of the component models and add the corresponding buffers (at most, this should add a 3d buffer and a 2d buffer for the surface of that 3d domain)
- src/standalone/Bucket/Bucket.jl: in the method for `initialize` here, add in the call to add_dss_to_aux. The Bucket model has to be made with an LSM domain, which has the surface and subsurface attributes. 
- - src/standalone/Vegetation/Canopy.jl: in the method for `initialize_auxiliary` here, add in the call to add_dss_to_aux. 
@juliasloan25 these would be good for you to review

2. Update the existing shared utilities tests for dss. I also wanted to test the `initialize_auxiliary` was doing the correct thing for: the bucket model, standalone models (default method), and land surface models. i did this somewhat lazily, by adding in additional tests to existing files which already make `p` for these models, rather than by making a new test file with a dummy LSM, dummy standalone model, etc. @juliasloan25 these would be good for you to review

3. Changes ClimaLSM.dss! to use the buffer [shared_utilities/utilities.jl] @akshaysridhar these would be good for you to review

4. creates the add_dss_buffer_to_aux function @akshaysridhar these would be good for you to review

Review checklist

I have:

----


src/shared_utilities/models.jl
```diff
@@ -303,5 +356,11 @@ function initialize(model::AbstractModel{FT}) where {FT}
     coords = Domains.coordinates(model)
     Y = initialize_prognostic(model, coords)
     p = initialize_auxiliary(model, coords)
+    # assumes model.domain exists. We do this already
+    add_dss_buffer_to_aux!(p, model.domain)
     return Y, p, coords
 end
+
+# 1. Make ODEfunction method
+# 2. do we need to account for different field spaces in integrated models in buffer?
```
kmdeck, line 365: yes

src/shared_utilities/models.jl
```diff
@@ -303,5 +356,11 @@ function initialize(model::AbstractModel{FT}) where {FT}
...
+# 3. Bucket model buffer
```
kmdeck, line 366: did so

src/shared_utilities/models.jl
```diff
@@ -303,5 +356,11 @@ function initialize(model::AbstractModel{FT}) where {FT}
     coords = Domains.coordinates(model)
     Y = initialize_prognostic(model, coords)
     p = initialize_auxiliary(model, coords)
+    # assumes model.domain exists. We do this already
+    add_dss_buffer_to_aux!(p, model.domain)
     return Y, p, coords
 end
+
+# 1. Make ODEfunction method
```
kmdeck, line 364: future PR

src/shared_utilities/models.jl
```diff
@@ -291,6 +294,67 @@ end
...
+function add_dss_buffer_to_aux(
+    p::ClimaCore.Fields.FieldVector,
+    domain::Domains.AbstractDomain,
+) end
+
+function add_dss_buffer_to_aux(
+    p::ClimaCore.Fields.FieldVector,
+    domain::Union{Domains.Plane, Domains.SphericalSurface},
+)
+    keys = propertynames(p)
```
kmdeck, line 309: @akshaysridhar do you know of a better way of appending another element to a FieldVector? this doesnt work as is...and is also kind of clunky

akshaysridhar: Does `p` need to be a `FieldVector` here ? Or could it be a general cache container ?

src/ClimaLSM.jl
```diff
@@ -92,10 +92,14 @@ function initialize_auxiliary(
...
-    p = ClimaCore.Fields.FieldVector(;
-        p_interactions...,
-        NamedTuple{components}(p_state_list)...,
-    )
+    p = (; p_interactions..., NamedTuple{components}(p_state_list)...)
+    domains_list = map(components) do (component)
+        submodel = getproperty(model, component)
+        getproperty(submodel, :domain)
+    end
+    for domain in domains_list
```
kmdeck, line 100: unique(domains_list)

docs/make.jl
```diff
@@ -8,7 +8,7 @@ using ClimaLSM
 include("pages_helper.jl")
 tutorials = [
     "For model developers" => [
-        "Using AbstractModel functionality" => "Usage/model_tutorial.jl",
+        #        "Using AbstractModel functionality" => "Usage/model_tutorial.jl",
```
juliasloan25, line 11: I think we can remove the Henon Heiles tutorial in favor of more relevant tutorials, and just delete this instead of commenting it out (as discussed earlier)

src/shared_utilities/models.jl
```diff
@@ -266,26 +270,28 @@ Adjustments to this - for example because different auxiliary variables
...
+    # change that tutorial? 
+    p = add_dss_buffer_to_aux(p, model.domain)
+    return p
 end
 
 function initialize_vars(keys, types, state, model_name)
     FT = eltype(state)
     if length(keys) == 0
-        return ClimaCore.Fields.FieldVector(; model_name => FT[])
+        return (; model_name => FT[])
```
juliasloan25, line 288: we should check the requirements of a NamedTuple that gets converted to a Field Vector. To me it would make more sense to have `(; model_name => (;))` for type consistency with the `else` case, but I'm not sure if we're able to do that.

src/shared_utilities/utils.jl
```diff
@@ -20,21 +20,72 @@ Helper function used in computing tendencies of vertical diffusion terms.
...
+add_dss_buffer_to_aux(p, domain::Domains.AbstractDomain) = p
+
+"""
+    add_dss_buffer_to_aux(
+        p,
+        domain::Union{Domains.Plane, Domains.SphericalSurface},
+    )
+
+Adds a dss buffer corresponding to `domain.space` to `p` with the name `dss_buffer_2d`, 
+appropriate for a 2D domain.
```
juliasloan25, line 38: could you add a comment explaining why we store `dss_buffer_2d` and `dss_buffer_3d` separately and with different names?

src/standalone/Bucket/Bucket.jl
```diff
@@ -206,7 +207,11 @@ end
 
 function BucketModel(;
     parameters::BucketModelParameters{FT, PSE},
-    domain::ClimaLSM.Domains.AbstractLSMDomain,
+    domain::Union{
```
juliasloan25, line 210: why are we restricting the type more here?

test/standalone/Soil/soiltest.jl
```diff
@@ -147,7 +147,8 @@ end
     )
 
     Y, p, coords = initialize(soil)
-
+    # Here we do not add a dss buffer: check that.
```
juliasloan25, line 150: in this comment do you mean we should add a test that `add_dss_buffer_to_aux` doesn't change `p`?

test/standalone/Bucket/soil_bucket_tests.jl
```diff
@@ -98,6 +98,26 @@ for bucket_domain in bucket_domains
         )
         # Initial conditions with no moisture
         Y, p, coords = initialize(model)
+        # test if the correct dss buffers were added to aux
+        if typeof(model.domain) <: Union{
```
juliasloan25, line 102: could you add a comment explaining why we expect buffers for these 3d domains but not LSMSingleColumnDomain please?

test/shared_utilities/utilities.jl
```diff
@@ -126,36 +126,88 @@ end
...
+        )
+
+        domains = (domain1, domain2)
+        for domain in domains
+            space = domain.space
+            field = Fields.zeros(space)
+            subfield1 = Fields.zeros(space)
+            subfield2 = Fields.zeros(space)
+
+            # make fields non-constant
```
juliasloan25, line 162: maybe expand on this - "make fields non-constant to check impact of dss step"

juliasloan25 (commented):
Nice work, this PR is very clear even though it has a lot of changes :) 

I'm okay with keeping all of these changes in one PR, but it may be easier to debug if we break it up. Up to you, but if you're having trouble getting CI to pass we could try splitting it up into the pieces you suggested

src/shared_utilities/models.jl
kmdeck: That makes sense. I think we are not able though:
```
julia> a = (; :foo => Float32[])
(foo = Float32[],)

julia> b = (;)
NamedTuple()

julia> ClimaCore.Fields.FieldVector(;a)
1-blocked 0-element ClimaCore.Fields.FieldVector{Float32, NamedTuple{(:a,), Tuple{ClimaCore.Fields.FieldVector{Float32, NamedTuple{(:foo,), Tuple{Vector{Float32}}}}}}}

julia> ClimaCore.Fields.FieldVector(;b)
ERROR: ArgumentError: Union{} does not have elements
```

src/shared_utilities/utils.jl
kmdeck: "We use a name which specifically denotes that the buffer is on a 3(2)d space
because for some models, we require both a buffer on the 3d space as well as on the surface 
2d space, because some models have prognostic variables that are only defined on the surface 
space."

src/standalone/Bucket/Bucket.jl
kmdeck: I actually realized that this should have been restricted in the first place. The bucket model already relies on having surface and subsurface variables, so it only makes sense to use it with certain domains. So in principle we can use it with the LSM domains, the spherical shell, the hybrid box, or the column. But right now in the `initialize` function we use some features that only exist for LSM domains. 

Ill make a comment.

test/standalone/Soil/soiltest.jl
kmdeck: I meant that that is what this next @test is checking. I'll clarify!

src/shared_utilities/utils.jl
```diff
@@ -44,36 +101,45 @@ FieldVectors or Fields, and that the final unpacked variable is a Field.
...
+end
+
+function dss_helper!(
+    field::ClimaCore.Fields.Field,
+    space::Union{ClimaCore.Spaces.AbstractSpectralElementSpace},
+    p,
 )
-    Spaces.weighted_dss!(field)
+    buffer = p.dss_buffer_2d
+    ClimaCore.Spaces.weighted_dss!(field, buffer)
```
akshaysridhar, line 142: Thanks @kmdeck. This is indeed the funcionality we want!

akshaysridhar (changes requested):
Changes to `dss_buffer`, addition to `p::NamedTuple` and usage in the `dss_helper!` functions are reasonable. 

When I attempt to run the tests within this package locally I see the error in the code block below : Is this some local configuration/ artifact requirement issue? With this addressed I'm happy to approve the PR.   
```
(ClimaLSM) pkg> test 
...
...
Initialize aux state with map: Test Failed at /Users/akshaysridhar/Research/Codes/ClimaLSM.jl/test/standalone/Bucket/albedo_map_bucket.jl:130
  Expression: sprint(showerror, err) == "Using an albedo map requires a global run."
  Test Summary:                 | Pass  Fail  Total   Time
Initialize aux state with map |    2     1      3  31.7s
  ```

**Review by akshaysridhar — APPROVED** [review]
```
Test Summary:                 | Pass  Total   Time
Initialize aux state with map |    3      3  37.2s
```
Passes at commit `da0f6f312`;  LGTM!

juliasloan25 (approved):
Looks good!

PR #301, opened 2023-08-21
Continued Model Improvements

Espeer5:
## Purpose 

This PR contains several different small improvements that we've been experimenting with over the past week.

## To-do
 
- [ ] Verify the formula for ground heat flux and how it should be added to the computation of surface fluxes

## Content

- Clumping Index added to the denominator in compute_GPP
- Empirical formula for fraction of diffuse radiation for two-stream added to capture diurnal cycle
- RMSE and R^2 added to GPP and ET plots to allow more numerical assessment of model agreement
- Beta removed from An, applied to Medlyn term
- Computation of G, ground heat flux added to soil_canopy model as interaction var, and preliminarily applied to calculation of surface fluxes
- New plots added showing average diurnal cycles of GPP and ground heat flux
- LW through the canopy calculated using Boltzmann law

----

### DIFF
--- a/docs/tutorials/Canopy/canopy_tutorial.jl
+++ b/docs/tutorials/Canopy/canopy_tutorial.jl
@@ -134,7 +134,6 @@ rt_params = TwoStreamParameters{FT}(;
     λ_γ_PAR = FT(5e-7),
     λ_γ_NIR = FT(1.65e-6),
     n_layers = UInt64(20),
-    diff_perc = FT(0.5),
 )
 
 rt_model = TwoStreamModel{FT}(rt_params);
--- a/docs/tutorials/Canopy/soil_canopy_tutorial.jl
+++ b/docs/tutorials/Canopy/soil_canopy_tutorial.jl
@@ -197,7 +197,6 @@ radiative_transfer_args = (;
         λ_γ_PAR = FT(5e-7),
         λ_γ_NIR = FT(1.65e-6),
         n_layers = UInt64(20),
-        diff_perc = FT(0.5),
     )
 )
 
--- a/experiments/Manifest.toml
+++ b/experiments/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.9.1"
 manifest_format = "2.0"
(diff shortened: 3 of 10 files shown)


Espeer5:
This is a lot packed into one PR, probably too much, but I'm not sure how to break it up so if there is a way that makes sense, pls let me know.

kmdeck:
@Espeer5 let's break this up into several PRs! the ground heat flux is already included, for example. also, we should add improvements to the model incrementally, and assess what they actually do, before continuing to make the model more complex and harder to debug.
Maybe we could do:
1. improve and add plots, including goodness of fit metrics
2. canopy temperature equation

and then take a step back and see what needs to be added next?

Espeer5:
> @Espeer5 let's break this up into several PRs! the ground heat flux is already included, for example. also, we should add improvements to the model incrementally, and assess what they actually do, before continuing to make the model more complex and harder to debug. Maybe we could do:
> 
> 1. improve and add plots, including goodness of fit metrics
> 2. canopy temperature equation
> 
> and then take a step back and see what needs to be added next?

I know that the CI and Beta changes need to be made as per discussion with Renato last week. I will close this one, and I will create smaller PRs with 
- improve plots and add fit metrics
- Refactor application of beta
- Vary diffuse radiation fraction

We can then decide what needs to be done with the heat fluxes that Renato and I discussed last week after the meeting we have this week.

braghiere:
Good! Thanks. Agreed. Shall we split like this:

PR 1: (more straightforward. easier to merge)
    Clumping Index added to the denominator in compute_GPP
    Empirical formula for fraction of diffuse radiation for two-stream added to capture diurnal cycle
    RMSE and R^2 added to GPP and ET plots to allow more numerical assessment of model agreement
    Beta removed from An, applied to Medlyn term
    New plots added showing average diurnal cycles of GPP and ground heat flux

PR 2: (a bit more complex. a few meetings needed)
    Computation of G, ground heat flux added to soil_canopy model as interaction var, and preliminarily applied to calculation of surface fluxes
    LW through the canopy calculated using Boltzmann law

PR #303, opened 2023-08-21
Varying Diffuse Radiation Fraction

Espeer5:
## Purpose 

Vary the percentage of incoming radiation which is diffuse according to an empirical formula in order to capture the diurnal cycle for the two-stream model

## To-do
- [x] Add reference time to data structs and use to compute DOY for diffuse fraction calculation

## Content

- Clumping index added to the denominator of GPP calculation
- Moisture stress factor application refactored and applied directly to Medlyn term
- Empirical formula used to compute diffuse radiation for two-stream

----


src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -311,6 +318,36 @@ function plant_absorbed_pfd(
...
 
+"""
+    diffuse_fraction(td::FT, T::FT, SW_IN::FT, UR::FT, θs::FT) where {FT}
+
+Computes the fraction of diffuse radiation (`diff_frac`) as a function
+of the solar zenith angle (`θs`), the total surface incident radiation (`SW_IN`),
+the air temperature (`T`), the relative humidity (`UR`), and the day of the year
+(`td`).
+"""
+function diffuse_fraction(td::FT, T::FT, SW_IN::FT, UR::FT, θs::FT) where {FT}
```
braghiere, line 329: This is minor, but could we call it RH, instead of the Portuguese version UR?

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -536,23 +571,21 @@ end
...
     Vcmax25::FT,
-    β::FT,
     f::FT,
     ΔHRd::FT,
     T::FT,
     To::FT,
     R::FT,
 ) where {FT}
-    Rd = f * Vcmax25 * β * arrhenius_function(T, To, R, ΔHRd)
+    Rd = f * Vcmax25 * arrhenius_function(T, To, R, ΔHRd)
```
braghiere, line 588: So, removing β from An and using it in g1 only is ok, but dark respiration still needs to feel the impact of soil moisture. Include β in here again.

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -721,8 +755,9 @@ function medlyn_conductance(
     medlyn_term::FT,
     An::FT,
     ca::FT,
+    β::FT,
 ) where {FT}
-    gs = g0 + Drel * medlyn_term * (An / ca)
+    gs = β * (g0 + Drel * medlyn_term * (An / ca))
```
braghiere, line 760: This should be: 

gs =  (g0 + β * Drel * medlyn_term * (An / ca))

src/standalone/Vegetation/photosynthesis.jl
```diff
@@ -187,6 +187,6 @@ function compute_photosynthesis(
     Kc = MM_Kc(Kc25, ΔHkc, T, To, R)
     Ko = MM_Ko(Ko25, ΔHko, T, To, R)
     Ac = rubisco_assimilation(mechanism, Vcmax, ci, Γstar, Kc, Ko, oi)
-    Rd = dark_respiration(Vcmax25, β, f, ΔHRd, T, To, R)
-    return net_photosynthesis(Ac, Aj, Rd, β)
+    Rd = dark_respiration(Vcmax25, f, ΔHRd, T, To, R)
```
braghiere, line 190: Again, β needs to go back in Rd.

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -311,6 +318,36 @@ function plant_absorbed_pfd(
     return FT(SW_IN * F_abs)
 end
 
+"""
+    diffuse_fraction(td::FT, T::FT, SW_IN::FT, UR::FT, θs::FT) where {FT}
+
+Computes the fraction of diffuse radiation (`diff_frac`) as a function
+of the solar zenith angle (`θs`), the total surface incident radiation (`SW_IN`),
```
braghiere, line 325: the total surface incident shortwave radiation (`SW_IN`),

braghiere (changes requested):

src/standalone/Vegetation/canopy_parameterizations.jl
Espeer5: We should probably look at the plots together with this implemented. With this change, beta doesn't seem to have enough impact on GPP and we overestimate GPP for the entire year.

src/standalone/Vegetation/Canopy.jl
```diff
@@ -427,6 +427,16 @@ function ClimaLSM.make_update_aux(
...
         NIR .= compute_NIR(RT, canopy.radiation, t)
+        e_sat =
+            Thermodynamics.saturation_vapor_pressure.(
+                Ref(thermo_params),
+                T,
+                Ref(Thermodynamics.Liquid()),
+            )
+        e = FT(2.655 * q * P / (q + 1.6455))
+        rel_hum = e / e_sat
+        day = round(t / 24 / 3600)
```
kmdeck, line 438: Im not sure this is correct. `t` is measured from some reference that is set elsewhere. if that reference is not January 1 always, this wouldnt give the right answer.


We implicitly define it as the start date of the data here:
https://github.com/CliMA/ClimaLSM.jl/blob/99ccdaa7a1b68b649827f938898398b92e8b5afd/experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl#L81
since this starts at 0..


and the same time appears here:
https://github.com/CliMA/ClimaLSM.jl/blob/99ccdaa7a1b68b649827f938898398b92e8b5afd/experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl#L118

so, I think we need to store the reference date somewhere? let me ask around and see what the other models do for this.

Espeer5: Hmm good point, I see what you mean

braghiere: There might be a function called julian day or something like that. It gets any date and give back the julian day.

src/standalone/Vegetation/canopy_parameterizations.jl
braghiere: @Yujie-W, could you please double check this equation?

src/standalone/Vegetation/Canopy.jl
kmdeck: When we prescribe albedo (and eventually reanalysis data) from a file (as a function of lat/lon/t), one of the things we need to define is this "SimInfo" object: https://github.com/CliMA/ClimaLSM.jl/blob/ae8985563721f5d44d35b3907e08c077f9c5461d/src/shared_utilities/FileReader.jl#L80

this eventually gets stored in the PrescribedData object which has the info required to drive the simulation with albedo (reanalysis data) as a function of lat/lon/t.

My thought is that what we are doing with the PrescribedAtmosphere and PrescribedRadiativeFluxes is very similar. we use data to drive the system. and we need a reference date.

For now, I think it is OK to assume that when we drive the system with temporal data, we need to define a reference time, and that can get stored in the object that has the prescribed data info. so I'd say let's add a reference date to the prescribed atmos and prescribed RT types.

What do you think @Espeer5 ?

A longer term project for us is to unify the way we prescribe data across climalsm. I think our use cases would be:
- at a site, pass in functions of time or constants
- global runs: use the PrescribedData object that @juliasloan25 has made which will have a static and temporal flavor (or handle both cases neatly under the hood).

Espeer5: This makes sense to me, that way any time we are accessing data at `t` we can see what `t` is in reference to. It is a little bit redundant for now to specify it to both the `PrescribedRadiativeFluxes` and `PrescribedAtmosphere` but it makes sense as a way to standardize it now and try to unify the way it is prescribed down the road as you said.

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -567,7 +601,7 @@ the total net carbon assimilation (`An`), the extinction coefficient (`K`),
 leaf area index (`LAI`) and the clumping index (`Ω`).
 """
 function compute_GPP(An::FT, K::FT, LAI::FT, Ω::FT) where {FT}
-    GPP = An * (1 - exp(-K * LAI * Ω)) / K
+    GPP = An * (1 - exp(-K * LAI * Ω)) / (K * Ω)
```
kmdeck, line 604: @braghiere since beta is now applied to gs directly, and not to An, is GPP affected by beta at all?

just wanted to double check this - i remember we had briefly mentioned this earlier.

braghiere: That is a good point, but no. That's why, in my point of view, it would be better to apply beta on An like Eq (15) in: 
Clark, D. B., et al. 2011: The Joint UK Land Environment Simulator (JULES), model description – Part 2: Carbon fluxes and vegetation dynamics, Geosci. Model Dev., 4, 701–722, https://doi.org/10.5194/gmd-4-701-2011

I think right now GPP is indirectly being impacted by gs only.

Espeer5: RMSE of GPP is lower when applying beta to An, as soon as we remove it we lose the decrease of GPP during the summer. My question is, if beta is applied back to An as in equation 15, should it then not be applied in the dark respiration function and match equation 13 from that paper? If it is applied to both as it was originally, then it gets applied to dark respiration twice

kmdeck: thanks, Edward! I think we are getting at the same thing...

How is GPP affected by gs? We use `gs` to compute the canopy conductance here: https://github.com/CliMA/ClimaLSM.jl/blob/3b2c8bcaadce3d439bba14fd1ab63ff119146550/src/standalone/Vegetation/canopy_parameterizations.jl#L627 and that is used when we compute transpiration. maybe I am missing how `gs` affects GPP? our formula for GPP has An, K, omega, and LAI, none of which depend on gs, right?

Regarding applying it to net photosynthesis: I think we had it OK before?
https://github.com/CliMA/ClimaLSM.jl/blob/ae8985563721f5d44d35b3907e08c077f9c5461d/src/standalone/Vegetation/canopy_parameterizations.jl#L506
but Rd already had the beta. so whatever comes out has just one factor of beta applied to all terms...
but perhaps it would be clearer to remove from Rd formula, and just apply it in net_photosynthesis?

Espeer5: Either way, this is not necessary for this PR. If we want to table this for later discussion (perhaps see the impact of using different functions for beta instead of removing it from An as we thought about a couple weeks ago?) I can remove it from this PR and let this one just focus on the diffuse radiation function.

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -311,6 +318,36 @@ function plant_absorbed_pfd(
     return FT(SW_IN * F_abs)
 end
 
+"""
+    diffuse_fraction(td::FT, T::FT, SW_IN::FT, UR::FT, θs::FT) where {FT}
+
+Computes the fraction of diffuse radiation (`diff_frac`) as a function
+of the solar zenith angle (`θs`), the total surface incident shortwave radiation (`SW_IN`),
+the air temperature (`T`), the relative humidity (`UR`), and the day of the year
+(`td`).
```
kmdeck, line 327: Could we add a reference for where this formula comes from? also, `UR` in doc string still.

braghiere: Gu, L., Baldocchi, D., Verma, S. B., Black, T. A., Vesala, T., Falge, E. M., & Dowty, P. R. (2002). Advantages of diffuse radiation for terrestrial ecosystem productivity. Journal of Geophysical Research: Atmospheres, 107(D6), ACL 2-1. https://doi.org/10.1029/2001JD001242

Braghiere, R. K. (2013). Avaliação dos fluxos turbulentos de CO2, calor sensível e latente em função da profundidade óptica do aerossol sobre o arco do desflorestamento da Amazônia Legal Brasileira. https://doi.org/10.11606/D.14.2013.tde-29082013-165325

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
```diff
@@ -131,7 +133,8 @@ end
 radiation = ClimaLSM.PrescribedRadiativeFluxes(
     FT,
     SW_IN_spline,
-    LW_IN_spline;
+    LW_IN_spline,
+    UTC_DATETIME[1];
```
kmdeck, line 137: we might change this function signature to take in the reference time as an argument, and then we should be able to pass the time in when it is called in update_aux by accessing the reference time from atmos or radiation.
https://github.com/CliMA/ClimaLSM.jl/blob/499bbaefc58eba0b966648ddd92f585604443244/experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl#L112

Keeping the reference time in UTC (rather than local) is a good idea! I'm not sure which we are using right now? in zenith_angle it looks like local time.

src/shared_utilities/drivers.jl
```diff
@@ -60,6 +60,8 @@ struct PrescribedAtmosphere{FT, LP, SP, TA, UA, QA, RA, CA} <:
     P::RA
     "CO2 concentration in atmosphere (mol/mol)"
     c_co2::CA
+    "Reference time - the datetime correpsonding to t=0 for the simulation"
```
kmdeck, line 63: `corresponding`

src/shared_utilities/drivers.jl
```diff
@@ -241,29 +244,32 @@ function surface_fluxes_at_a_point(
...
 $(DocStringExtensions.FIELDS)
 """
-struct PrescribedRadiativeFluxes{FT, SW, LW, T, OD} <:
+struct PrescribedRadiativeFluxes{FT, SW, LW, DT, T, OD} <:
        AbstractRadiativeDrivers{FT}
     "Downward shortwave radiation function of time (W/m^2): positive indicates towards surface"
     SW_d::SW
     "Downward longwave radiation function of time (W/m^2): positive indicates towards surface"
     LW_d::LW
+    "The reference start time for the simulation"
```
kmdeck, line 258: could use the same doc string

src/standalone/Vegetation/Canopy.jl
```diff
@@ -421,12 +422,23 @@ function ClimaLSM.make_update_aux(
...
         K = extinction_coeff(ld, θs)
         PAR .= compute_PAR(RT, canopy.radiation, t)
         NIR .= compute_NIR(RT, canopy.radiation, t)
+        e_sat =
+            Thermodynamics.saturation_vapor_pressure.(
+                Ref(thermo_params),
+                T,
+                Ref(Thermodynamics.Liquid()),
+            )
+        e = FT(2.655 * q * P / (q + 1.6455))
```
kmdeck, line 438: Ill figure this out, but we should use a Thermodynamics.jl function for this: https://github.com/CliMA/Thermodynamics.jl/blob/e7a012f7ddc883fa307f5a2e672304ef3b49285b/src/relations.jl#L122 maybe?

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
Espeer5: We should be using UTC everywhere... in `zenith_angle` it is time in UTC. The simulations starts from 12AM on Jan 1 local, which is UTC + 6 so `DateTime("2005-01-01-06", "yyyy-mm-dd-HH")` gives us UTC.

This is a good idea about changing the function signature and should be easy to implement I will change that.

braghiere (approved):

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
kmdeck: ok! sounds great, thank you!

src/standalone/Vegetation/Canopy.jl
kmdeck: I think this should work:
`e = Thermodynamics.partial_pressure_vapor.(Ref(thermo_params), P, Ref(PhasePartition(q))`

kmdeck (approved):

src/standalone/Vegetation/Canopy.jl
Espeer5: Great! Got it.

PR #300, opened 2023-08-17
BulkAlbedoTemporal - use ClimaCore space

juliasloan25:
## Purpose 
Closes #299 (see for more info)

As of August 21, the coupler is able to use BulkAlbedoTemporal in coupled runs with this change (https://github.com/CliMA/ClimaCoupler.jl/pull/394).

After merging this PR, we need to make a patch release of ClimaLSM.jl (v0.3.1).

## To-do
- [x] change constructor argument from surface (type `ClimaLSM.Domains.AbstractDomain`) to space (type `ClimaCore.Spaces.AbstractSpace`)
- [x] update constructor usage in tests

----
- [x] I have read and checked the items on the review checklist.

PR #304, opened 2023-08-22
bump to version 0.3.1

juliasloan25:
## Purpose 
#300 introduced changes in the BulkAlbedoTemporal constructor interface, so we need to make a new release (v0.3.1).

----

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -212,6 +212,16 @@ git-tree-sha1 = "e30f2f4e20f7f186dc36529910beaedc60cfa644"
 uuid = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 version = "1.16.0"
 
+[[deps.ChangesOfVariables]]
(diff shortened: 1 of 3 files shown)


docs/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.9.1"
+julia_version = "1.8.5"
```
kmdeck, line 3: keep at 1.9.1? I realize this is because the coupler uses 1.8.5 :)

experiments/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.9.1"
+julia_version = "1.8.5"
```
kmdeck, line 3: ditto here

docs/Manifest.toml
juliasloan25: Ahh yes good catch!

PR #308, opened 2023-08-24
type restriction on t

kmdeck:
## Purpose 
See issue https://github.com/CliMA/ClimaCoupler.jl/issues/396

## To-do
Figure out if we need to remove all instances where we assume t is FT, with FT Float32 or Float64, because it will be Float64 when using ClimaTimesteppers.

## Content

----


kmdeck:
Closing in lieu of #353

Issue #277, opened 2023-07-25
type piracy in getindex method


kmdeck:
Remove the extra method defined in src/shared_utilties/ntuple_utils.jl (in fact, remove this file completely)

see discussion here:https://github.com/CliMA/ClimaLSM.jl/pull/263 with @charleskawczynski .

kmdeck:
@charleskawczynski would it be possible to schedule some time to discuss how to remove these additional methods?

PR #310, opened 2023-08-30
Domain stretching

kmdeck:
## Purpose 
Allows for variable domain stretching (finer resolution at the top of the domain compared to the bottom).

Whenever we need dz in source code, we obtain it using the function `get_Δz`, which uses ClimaCore to extract the `dz` values. so we do not make the assumption of uniform spacing in src code. (we do in some tests, but that seems OK if those tests use uniform spacing.)
## To-do
Future PRs: fix piracy and add the aqua test back in:
#277 #311 

## Content

1. Adds domain stretching to all domains with a z coordinate.
2. Renames height to depth for the SphericalShell domain
3. uses the stretched domain in the ozark example
4. adds unit tests.

Note that the user passes the target (dz_bottom, dz_top) desired, as a Tuple, or `nothing`, if no stretching is needed. We then create the ClimaCore.Meshes.GeneralizedExponentialStretching object in the domain constructor appropriately. Instead, the user could pass the stretching object itself, and the default could be `ClimaCore.Meshes.Uniform`. In a way this is nicer (it's not prescribed that they use Uniform or GeneralizedExponentialStretching, gives them all the options that ClimaCore develops in the future, no `if` statement in the constructor), but I felt this might be more confusing than supplying the target layer widths at the top and bottom. Happy to change if anyone feels strongly!

Review checklist

I have:

----

### DIFF
--- a/experiments/integrated/ozark/ozark_domain.jl
+++ b/experiments/integrated/ozark/ozark_domain.jl
@@ -1,10 +1,16 @@
 # Domain setup
 # For soil column
-nelements = 60
+nelements = 10
 zmin = FT(-2)
 zmax = FT(0)
-land_domain =
-    LSMSingleColumnDomain(; zlim = (zmin, zmax), nelements = nelements)
+dz_bottom = FT(0.5)
+dz_top = FT(0.025)
+
+land_domain = LSMSingleColumnDomain(;
+    zlim = (zmin, zmax),
+    nelements = nelements,
+    dz_tuple = (dz_bottom, dz_top),
+)
 
 # Number of stem and leaf compartments. Leaf compartments are stacked on top of stem compartments
 n_stem = Int64(1)
--- a/experiments/integrated/ozark/ozark_simulation.jl
+++ b/experiments/integrated/ozark/ozark_simulation.jl
@@ -1,8 +1,9 @@
 t0 = FT(120 * 3600 * 24)# start mid year
 N_days = 120
 tf = t0 + FT(3600 * 24 * N_days)
-dt = FT(60)
-n = 60
+dt = FT(240)
+# Number of timesteps between saving output
+n = 15
 saveat = Array(t0:(n * dt):tf)
 timestepper = CTS.RK4()
 # Set up timestepper
--- a/src/shared_utilities/Domains.jl
+++ b/src/shared_utilities/Domains.jl
@@ -63,6 +63,8 @@ struct Column{FT, S} <: AbstractDomain{FT}
     zlim::Tuple{FT, FT}
     "Number of elements used to discretize the interval"
     nelements::Tuple{Int}
+    "Tuple for mesh stretching specifying *target* (dz_bottom, dz_top) (m). If nothing, no stretching is applied."
+    dz_tuple::Union{Tuple{FT, FT}, Nothing}
     "Boundary face identifiers"
     boundary_tags::Tuple{Symbol, Symbol}
     "The associated ClimaCore Space"
@@ -72,12 +74,28 @@ end
 """
     Column(;
         zlim::Tuple{FT, FT},
-        nelements::Int) where {FT}
-Outer constructor for the `Column` type.
+        nelements::Int,
+        dz_tuple::Union{Tuple{FT, FT}, Nothing} = nothing) where {FT}
+
+Outer constructor for the `Column` type. 
+
+Using `ClimaCore` tools, the coordinate
+mesh can be stretched such that the top of the domain has finer resolution
+than the bottom of the domain. In order to activate this, a tuple with the
+target dz_bottom and dz_top should be passed via keyword argument. The default is
+uniform spacing. Please note that in order to use this feature, ClimaCore requires that
+the elements of zlim be <=0. Additionally, the dz_tuple you supply may not be compatible
+with the domain boundaries in some cases, in which case you may need to choose
+different values.
+
 The `boundary_tags` field values are used to label the boundary faces 
 at the top and bottom of the domain.
 """
-function Column(; zlim::Tuple{FT, FT}, nelements::Int) where {FT}
+function Column(;
+    zlim::Tuple{FT, FT},
+    nelements::Int,
+    dz_tuple::Union{Tuple{FT, FT}, Nothing} = nothing,
+) where {FT}
     @assert zlim[1] < zlim[2]
     boundary_tags = (:bottom, :top)
     column = ClimaCore.Domains.IntervalDomain(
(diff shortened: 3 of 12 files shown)


src/shared_utilities/Domains.jl
```diff
@@ -72,24 +74,55 @@ end
...
+        nelements::Int,
+        dz_tuple::Union{Tuple{FT, FT}, Nothing} = nothing) where {FT}
+
+Outer constructor for the `Column` type. 
+
+Using `ClimaCore` tools, the coordinate
+mesh can be stretched such that the top of the domain has finer resolution
+than the bottom of the domain. In order to activate this, a tuple with the
+target dz_bottom and dz_top should be passed via keyword argument. The default is
+unform spacing. Please note that in order to use this feature, ClimaCore requires that
```
kmdeck, line 86: uniform

src/shared_utilities/Domains.jl
```diff
@@ -297,31 +362,50 @@ end
...
+        dz_tuple::Union{Tuple{FT, FT}, Nothing} = nothing,
     ) where {FT}
 Outer constructor for the `SphericalShell` domain, using keyword arguments.
+
+Using `ClimaCore` tools, the coordinate
+mesh can be stretched such that the top of the domain has finer resolution
+than the bottom of the domain. In order to activate this, a tuple with the
+target dz_bottom and dz_top should be passed via keyword argument. The default is
+unform spacing. Please note that in order to use this feature, ClimaCore requires that
+the elements of zlim be <=0. Additionally, the dz_tuple you supply may not be compatible
```
kmdeck, line 377: edit, no zlim here

src/shared_utilities/Domains.jl
```diff
@@ -72,24 +74,55 @@ end
...
+    dz_tuple::Union{Tuple{FT, FT}, Nothing} = nothing,
+) where {FT}
     @assert zlim[1] < zlim[2]
     boundary_tags = (:bottom, :top)
     column = ClimaCore.Domains.IntervalDomain(
         ClimaCore.Geometry.ZPoint{FT}(zlim[1]),
         ClimaCore.Geometry.ZPoint{FT}(zlim[2]);
         boundary_tags = boundary_tags,
     )
-    mesh = ClimaCore.Meshes.IntervalMesh(column; nelems = nelements)
```
AlexisRenchon, line 88: I was wondering if you created a new method for `Column`, but I see you have a `if` and no new method. 
I have a simple question / something to consider: would a new method be computationally more efficient? (faster run), or is this function just called once and basically this changes nothing?

experiments/integrated/ozark/ozark_domain.jl
```diff
@@ -1,10 +1,16 @@
 # Domain setup
 # For soil column
-nelements = 60
+nelements = 10
 zmin = FT(-2)
 zmax = FT(0)
-land_domain =
-    LSMSingleColumnDomain(; zlim = (zmin, zmax), nelements = nelements)
+dz_bottom = FT(0.5)
+dz_top = FT(0.025)
```
AlexisRenchon, line 7: Question / something to consider: 
We define dz top and bottom here, and if I understand correctly the other dz in between are exponential. 
Would it be helpful to add and option for other extrapolation, e.g., linear from top to bottom dz? (just to keep in mind, but no need to do this in this PR)

experiments/integrated/ozark/ozark_simulation.jl
```diff
@@ -1,8 +1,8 @@
 t0 = FT(120 * 3600 * 24)# start mid year
 N_days = 120
 tf = t0 + FT(3600 * 24 * N_days)
-dt = FT(60)
-n = 60
+dt = FT(240)
+n = 15
```
AlexisRenchon, line 5: Maybe we can add a comment to say what `n` is, for clarity
```
# save output at every n * dt
```
we see it in the next line, but it's easier to read line by line?

AlexisRenchon (approved):

AlexisRenchon:
FYI, I edited the PR text `get_Δz`

src/shared_utilities/Domains.jl
kmdeck: Good question! I wasnt sure, I tried making a minimum working example and in this simple case the execution time was very similar.
```
module Tmp

struct FlagTrue end
struct FlagFalse end

f(x::FlagTrue) = 1
f(x::FlagFalse) = 2
function g(x)
    if typeof(x) == FlagTrue
        return 1
    else
        return 2
    end
end


end

using .Tmp
using BenchmarkTools
@btime Tmp.f(Tmp.FlagTrue())
@btime Tmp.g(Tmp.FlagTrue())
```

I asked @dennisYatunin and he said that in more recent versions of Julia using an if statement should be the same as writing a new method. You also made a good point about this being only executed once, so it might be okay even if it was slower to use the if statement.

experiments/integrated/ozark/ozark_domain.jl
kmdeck: yes! ok, i had the same question. right now there are 3 grid stretching options in ClimaCore: Uniform, Exponential, and GeneralizedExponential. The way I wrote this PR currently makes it so, as you wrote, we have to choose the GeneralizedExponential. From a user interface point of view I thought it was a little cleaner (the user passes `dz_tuple = (dz_bottom, dz_top)`, instead of `stretch = ClimaCore.Meshes.GeneralizedExponentialStretching(dz_bottom, dz_top)`. But, if they pass the stretch object itself, then we have more flexibility on what they can choose (including, in the future, maybe a linear stretching of the grid). 


and then, actually, we do not need an if statement, because we can make the default value `stretch = ClimaCore.Meshes.Uniform()` and always pass `stretching` into the function that makes the mesh.
`mesh = ClimaCore.Meshes.IntervalMesh(column; nelems = nelements, stretch = stretch)`

Im open to the latter! what do you think?

experiments/integrated/ozark/ozark_simulation.jl
kmdeck: will do, thank you!

experiments/integrated/ozark/ozark_domain.jl
AlexisRenchon: I am also ok with both options... 
I guess the question is, do we need these options
For now I am ok to just keep it simple, as it is now.

@braghiere

kmdeck: ok, let's leave it for now! Right now exponential stretching is the only option, and if we need other options, we'd need to add them in ClimaCore, and that would most likely be a low priority item anyways.

PR #314, opened 2023-08-31
Remove overloaded getindex

charleskawczynski:
This PR removes the overloading of `getindex` on ClimaCore fields.

also closes #317 #312 

We access the individual fields within the Field of Tuples using the following rules:
Previously, regardless of in a loop or not:
```
field[i] .= ....
@. field[i] = g(other field)
```
Now: outside of a loop:
```
field.:i .= ...
```
Now: inside of a loop:

```
field.:($i) .=
@. field.:($$i) =
```

We also need to sometimes access elements that are compute quantities, for example, before,
```
field[n_stem+n_leaf] .= 
```
Now, we need to do something like:
```
i_end =n_stem+n_leaf # define intermediate quantity
field.:($i_end) .= 
```
regardless of if we are in a loop or not.


kmdeck:
We can merge this first https://github.com/CliMA/ClimaLSM.jl/pull/315 and then in this PR, after rebasing, add back in the Aqua piracy test.

src/standalone/Vegetation/Canopy.jl
```diff
@@ -483,9 +483,10 @@ function ClimaLSM.make_update_aux(
                     )
                 ) / 2
         end
-        @. β = moisture_stress(ψ[n_stem + n_leaf] * ρ_l * grav, sc, pc)
-        # We update the fa[n_stem+n_leaf] element once we have computed transpiration, below
+        @. β =
```
kmdeck, line 486: double check why this works/use same style as the other cases

kmdeck:
@charleskawczynski could you review this? I cant tag you since it is your PR :)

src/standalone/Vegetation/Canopy.jl
```diff
@@ -444,36 +444,36 @@ function ClimaLSM.make_update_aux(
...
-            PlantHydraulics.effective_saturation(ν, ϑ_l[1]),
+            PlantHydraulics.effective_saturation(ν, ϑ_l.:1),
             ν,
             S_s,
         )
 
         @inbounds for i in 1:(n_stem + n_leaf - 1)
-
-            @. ψ[i + 1] = PlantHydraulics.water_retention_curve(
+            ip1 = i + 1
```
kmdeck, line 455: @charleskawczynski I wasnt sure how to avoid defining this intermediate quantity

src/standalone/Vegetation/Canopy.jl
```diff
@@ -483,9 +483,10 @@ function ClimaLSM.make_update_aux(
                     )
                 ) / 2
         end
-        @. β = moisture_stress(ψ[n_stem + n_leaf] * ρ_l * grav, sc, pc)
-        # We update the fa[n_stem+n_leaf] element once we have computed transpiration, below
+        i_end = n_stem + n_leaf
```
kmdeck, line 486: likewise here, and Im not totally clear on the rules around broadcasting -> use $$ but this seemed to give what we want!

src/standalone/Vegetation/Canopy.jl
```diff
@@ -506,7 +507,7 @@ function ClimaLSM.make_update_aux(
             canopy_surface_fluxes(canopy.atmos, canopy, Y, p, t)
         transpiration .= canopy_transpiration
         # Transpiration is per unit ground area, not leaf area (mult by LAI)
-        fa[n_stem + n_leaf] .= PlantHydraulics.transpiration_per_ground_area(
+        fa.:($i_end) .= PlantHydraulics.transpiration_per_ground_area(
```
kmdeck, line 510: not totally sure why we needed the $ here, either, since this is not in a loop.

src/standalone/Vegetation/Canopy.jl
charleskawczynski: Maybe `.:($$(i+1))` might work? But this looks fine to me, too.

src/standalone/Vegetation/Canopy.jl
charleskawczynski: Same as above, I think it's fine to add an intermediate if it helps fix this issue.

src/standalone/Vegetation/Canopy.jl
charleskawczynski: We can/should probably document how this is working, but it looks fine to me

charleskawczynski (commented):
Looks good to me!

src/standalone/Vegetation/Canopy.jl
kmdeck: I agree that we should document how the indexing works now, Ill add some comments!

src/standalone/Vegetation/Canopy.jl
```diff
@@ -444,36 +444,36 @@ function ClimaLSM.make_update_aux(
...
+            PlantHydraulics.effective_saturation(ν, ϑ_l.:1),
             ν,
             S_s,
         )
 
         @inbounds for i in 1:(n_stem + n_leaf - 1)
-
-            @. ψ[i + 1] = PlantHydraulics.water_retention_curve(
+            ip1 = i + 1
+            @. ψ.:($$ip1) = PlantHydraulics.water_retention_curve(
```
juliasloan25, line 456: I think it should be okay to use `:($$(i + 1))` here, and it might be more clear. But this way with `ip1` is fine too

src/standalone/Vegetation/Canopy.jl
juliasloan25: Do we need to use two `$$` if we have a broadcast inside of a loop because the broadcast and loop are each an expression, so we need to interpolate the index value twice? If we have a broadcast outside of a loop, do we use just one `$`?

src/standalone/Vegetation/Canopy.jl
juliasloan25: I'm also confused about why we use a `$` here - shouldn't it just be `fa.:i_end`?

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -526,6 +526,7 @@ function make_compute_exp_tendency(model::PlantHydraulicsModel, canopy)
         fa_roots = p.canopy.hydraulics.fa_roots
         FT = eltype(t)
         @inbounds for i in 1:(n_stem + n_leaf)
+            im1 = i - 1
```
juliasloan25, line 529: I think it might be more clear to just use `fa.:($$(i - 1))` here instead of `im1` (like with `ip1` above), but either is fine :)

juliasloan25 (commented):
Looks good! I just had a couple of questions about places that weren't clear to me

src/standalone/Vegetation/Canopy.jl
kmdeck: thanks! this seems to work. I had tried this:
```
julia> @inbounds for i in 1:(n_stem + n_leaf - 1)
           @. ψ.:($$i+1) = FT(0)
       end
ERROR: MethodError: no method matching getproperty(::ClimaCore.Fields.Field{ClimaCore.DataLayouts.DataF{NTuple{11, Float64}, Vector{Float64}}, ClimaCore.Spaces.PointSpace{ClimaCore.DataLayouts.DataF{ClimaCore.Geometry.LocalGeometry{(3,), ClimaCore.Geometry.ZPoint{Float64}, Float64, StaticArraysCore.SMatrix{1, 1, Float64, 1}}, Vector{Float64}}}}, ::Expr)
```

but not with the additional parantheses

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: will change this here too

src/standalone/Vegetation/Canopy.jl
kmdeck: It's weird, actually - that works in one case, but not the other. @juliasloan25 
```
julia> @inbounds for i in 1:(n_stem + n_leaf - 1)
             @. ψ.:($$(i+1)) = FT(0)
       end

julia> @inbounds for i in 1:(n_stem + n_leaf - 1)
             @. ψ.:($$(i+1)) = PlantHydraulics.water_retention_curve(
               retention_model,
               PlantHydraulics.effective_saturation(ν, ϑ_l.:($$(i+1))),
               ν,
               S_s,
           )
       end
ERROR: syntax: invalid syntax "ϑ_l.((i + 1))" around REPL[96]:2
Stacktrace:
 [1] top-level scope
   @ REPL[96]:1
```

src/standalone/Vegetation/Canopy.jl
kmdeck: Hm - this doesnt work after all, 
```
julia> @inbounds for i in 1:(n_stem + n_leaf - 1)
             @. ψ.:($$(i+1)) = FT(0)
       end

julia> @inbounds for i in 1:(n_stem + n_leaf - 1)
             @. ψ.:($$(i+1)) = PlantHydraulics.water_retention_curve(
               retention_model,
               PlantHydraulics.effective_saturation(ν, ϑ_l.:($$(i+1))),
               ν,
               S_s,
           )
       end
ERROR: syntax: invalid syntax "ϑ_l.((i + 1))" around REPL[96]:2
Stacktrace:
 [1] top-level scope
   @ REPL[96]:1
```

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: changing back...

src/standalone/Vegetation/Canopy.jl
kmdeck: Im not sure. the `:i_end` notation says to look for the field associated with the symbol `i_end` if I understand correctly, and so it doesnt work here.
```
julia> i_end = n_stem+n_leaf
11
julia> fa.:i_end
ERROR: Invalid field name i_end for type NTuple{11, Float64}
```

This doesnt work either:
```
julia> fa.:($(n_stem+n_leaf))
ERROR: syntax: invalid syntax "fa.((n_stem + n_leaf))" around REPL[104]:1
Stacktrace:
 [1] top-level scope
   @ REPL[104]:1
```

charleskawczynski: The integer needs `$` to be interpolated otherwise the symbol (used to call `getproperty(fa, ::Symbol)`) is `:i_end`

src/standalone/Vegetation/PlantHydraulics.jl
charleskawczynski: Either way, let's address the confusion by opening an issue for documenting how this interpolated `getproperty` works in a more central place (since it's common to multiple repos)

src/standalone/Vegetation/Canopy.jl
kmdeck: If we broadcast outside of a loop with a known index (like 2), we dont need any dollar signs:
 ```
julia> @. ψ.:2 = PlantHydraulics.water_retention_curve(
               retention_model,
               PlantHydraulics.effective_saturation(ν, ϑ_l.:2),
               ν,
               S_s,
           )
Float64-valued Field:
  [203.938]
```

If we want to index at the result of an expression, it doesnt really work. I think we need to define the intermediate i_end = n_stem+n_leaf.
```

julia> @. ψ.:$(n_stem+n_leaf) = PlantHydraulics.water_retention_curve(
               retention_model,
               PlantHydraulics.effective_saturation(ν, ϑ_l.:$(n_stem+n_leaf)),
               ν,
               S_s,
           )
ERROR: syntax: "(n_stem + n_leaf)" is not a valid function argument name around REPL[118]:1
```

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: ok! That would be great. Ill open an issue in ClimaCore unless there is a better place?

PR #323, opened 2023-09-07
Remove OrdinaryDiffEq.jl dependency

charleskawczynski:
This PR removes the OrdinaryDiffEq.jl dependency in favor of using SciMLBase.

PR #324, opened 2023-09-07
Remove DiffEqCallbacks dependency

juliasloan25:
## Purpose 
SciMLBase.jl has all of the functionality of DiffEqCallbacks.jl that we're using, but is a more lightweight package. If we use SciMLBase instead, we will decrease precompilation time.

Analogous to #323 , but for removing DiffEqCallbacks.jl instead of OrdinaryDiffEq.jl.

Part of #325

## Content
- [x] Remove DiffEqCallbacks.jl code loading from `docs/` and `experiments/`
- [x] Replace DiffEqCallbacks with SciMLBase in `src/`
- [x] Remove DiffEqCallbacks from all projects

----


Project.toml
```diff
@@ -27,11 +27,11 @@ CFTime = "0.1"
 ClimaComms = "0.3, 0.4, 0.5"
 ClimaCore = "0.10.44"
 ClimaCoreTempestRemap = "0.3.5"
-DiffEqCallbacks = "2"
 DocStringExtensions = "0.8, 0.9"
 IntervalSets = "0.5, 0.6, 0.7"
 JLD2 = "0.4"
 NCDatasets = "0.11, 0.12"
+SciMLBase = "<1.51"
```
charleskawczynski, line 34:
```suggestion
SciMLBase = "1"
```
Is the <1.51 for a known incompatibility before that version? If so, then feel free to skip the suggestion! I imagine nobody will resolve to belle that version anyway since it’s relatively old

charleskawczynski (approved):
I left one comment about a compat entry, but the PR generally looks good to me!

Project.toml
juliasloan25: We had this compat in the docs/ and experiments/ Projects before, but I'm not sure why. I'll change it to 1

PR #328, opened 2023-09-13
Add diurnal average plotting

Espeer5:
## Purpose 

Converts the model output plots for the Ozark experiment to show the average diurnal cycle for GPP, SHF, and ET which show strong diurnal cycling behavior, as well as to allow the data to be saved to a csv file using an optional command line argument.

## To-do

- [x] Discuss adding GOF statistics to these plots

## Content

- function diurnal_avg used to compute the avg cycles of passed time series
- Plots of GPP, SHF, ET converted to show average diurnal cycles in Ozark experiment
- Save all model output data to a large csv file by running with the command line argument "save"

----

### DIFF
--- a/experiments/integrated/ozark/ozark.jl
+++ b/experiments/integrated/ozark/ozark.jl
@@ -6,6 +6,7 @@ using Plots
 using Statistics
 using Dates
 using Insolation
+using StatsBase
 
 using ClimaLSM
 using ClimaLSM.Domains: Column
@@ -183,12 +184,11 @@ Y, p, cds = initialize(land)
 exp_tendency! = make_exp_tendency(land)
 
 #Initial conditions
-Y.soil.ϑ_l = SWC[1 + Int(round(t0 / 1800))] # Get soil water content at t0
-# recalling that the data is in intervals of 1800 seconds. Both the data
-# and simulation are reference to 2005-01-01-00 (LOCAL)
+Y.soil.ϑ_l = SWC[1 + Int(round(t0 / DATA_DT))] # Get soil water content at t0
+# Both data and simulation are reference to 2005-01-01-00 (LOCAL)
 # or 2005-01-01-06 (UTC)
 Y.soil.θ_i = FT(0.0)
-T_0 = TS[1 + Int(round(t0 / 1800))] # Get soil temperature at t0
+T_0 = TS[1 + Int(round(t0 / DATA_DT))] # Get soil temperature at t0
 ρc_s =
     volumetric_heat_capacity.(Y.soil.ϑ_l, Y.soil.θ_i, Ref(land.soil.parameters))
 Y.soil.ρe_int =
@@ -225,7 +225,7 @@ sv = (;
 cb = ClimaLSM.NonInterpSavingCallback(sv, saveat)
 
 prob = SciMLBase.ODEProblem(
-    CTS.ClimaODEFunction(T_exp! = exp_tendency!),
+    CTS.ClimaODEFunction((T_exp!) = exp_tendency!),
     Y,
     (t0, tf),
     p,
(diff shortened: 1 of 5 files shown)


Espeer5:
Still not exactly sure what's happening with the tail-off of the GPP starting so soon in the middle of the day. I don't see anything wrong with the time of day in the simulation or anything like that. But it's definitely a trend that's happening in our regular GPP plot not just in these average plots
GPP
GPP

braghiere:
> Still not exactly sure what's happening with the tail-off of the GPP starting so soon in the middle of the day. I don't see anything wrong with the time of day in the simulation or anything like that. But it's definitely a trend that's happening in our regular GPP plot not just in these average plots GPP GPP

Intriguing. Maybe we could plot GPP vs Sun zenith angle instead of time, and then time vs. sun zenith angle too. I guess that's something off with how we convert time to sun zenith angle. Also, in the model GPP is an instantaneous function of the sun zenith angle, while in the flux tower GPP is a function of the previous 30 min. The number of days we're doing that could also be the reason. From day 120 (Apr 30th) to 240 (Aug 28th), we're averaging GPP across multiple seasons, with highly variable sunsets/sunrise times. We could split the year in 4: March-Apr-May (day 60-151), June-July-Aug (day 152-243), Sep-Oct-Nov (day 244-334), and Dec-Jan-Feb(335-365, 1-59) and make 4 plots of GPP.

kmdeck:
Could we add the same plots for LW and SW, and add the plot for E onto the ET plot?

Espeer5:
> Could we add the same plots for LW and SW, and add the plot for E onto the ET plot?

Definitely, will do.

kmdeck:
> Still not exactly sure what's happening with the tail-off of the GPP starting so soon in the middle of the day. I don't see anything wrong with the time of day in the simulation or anything like that. But it's definitely a trend that's happening in our regular GPP plot not just in these average plots GPP GPP

How does it look if we run the simulation for a shorter period of time, maybe just 30 days? In the summer when there is a drought, our model is severely underpredicting GPP (most likely due to water stress factor being uncalibrated?).

experiments/integrated/ozark/ozark.jl
```diff
@@ -229,42 +229,48 @@ sol = SciMLBase.solve(
...
 savedir = joinpath(climalsm_dir, "experiments/integrated/ozark/")
 
+# This function will be used to compute averages over diurnal cycles. Input a
+# data series to average over and the total number of days in the data series, 
+# and it will return a vector of the average value at each timestamp in the 
+# day averaged over every day in the series.
+function diurnal_avg(series, days)
+    daily = floor(Int, length(series) / days)
+    daily_data = [
+        [sample for sample in series[i:1:(i + daily - 1)]] for
```
kmdeck, line 239: would this work instead? 
`[series[i:1:(i + daily - 1)]] for i in 1:daily:(daily * days)]`

kmdeck: Is there any risk that `daily` will correspond to slightly less than a day, and then over time the excerpt of the data we are considering as "a day" will shift in time?

How does one pass in the number of days if it is a fraction? e.g. if we have a non-integer number of days?

Perhaps renaming some of this could be clearer? `daily` -> `n_indices_per_day`, `daily_data` -> `data_by_day`? Im not sure what is best, but the current names dont clearly indicate what they are referring to.

experiments/integrated/ozark/ozark.jl
```diff
@@ -229,42 +229,48 @@ sol = SciMLBase.solve(
...
+end
+
 # GPP
 model_GPP = [
     parent(sv.saveval[k].canopy.photosynthesis.GPP)[1] for
     k in 1:length(sv.saveval)
 ]
 
-plt1 = Plots.plot(size = (1500, 400))
+GPP_daily_avg_model = diurnal_avg(model_GPP, 110)
```
kmdeck, line 252: are these values (110, 120, 48) hardcoded for the dataset or will they change if we change t0, tf, dt?

experiments/integrated/ozark/ozark.jl
```diff
@@ -229,42 +229,48 @@ sol = SciMLBase.solve(
...
     label = "Model",
-    xlim = [minimum(daily), maximum(daily)],
-    title = "GPP [mol/m^2/s]",
+    title = "GPP Diurnal Averages",
 )
 Plots.plot!(
     plt1,
-    seconds ./ 3600 ./ 24,
-    GPP .* 1e6,
+    1:0.5:24.5,
```
kmdeck, line 267: What is this? maybe this could be named?

experiments/integrated/ozark/ozark.jl
Espeer5: They will change if we change the number of days of the simulation, if we change df, or if the datasets have different timesteps. For example in the Ozark dataset the data is in 30 minutes intervals but for Harvard forest they are hourly so the slice you take is different. We can definitely abstract these to variables if we want to

kmdeck: thanks for the explanation! I think Ndays is specified in the ozark_simulation file (but subtract the number of days in spinup from this so we dont include that). And maybe we can define the interval for the data in the met_drivers file? I think 1800 is hardcoded elsewhere. 

(feel free to tackle now or we can take care of in a followup PR)

experiments/integrated/ozark/ozark.jl
Espeer5: I cleaned up the naming conventions and neatened this code but we do still need to think about how we could do this if we had a non-integer number of days. For now it will work automatically without changing any of the args in `ozark.jl` for any integer number of days, but if we have a fractional number of days then the indexing no longer works for this method.

Espeer5:
I added the ability to save the output data to a csv and formatted that csv so that each aux of all of the models or submodels has its own column in the CSV, and so that each row corresponds to one entry/timstamp in sv.saveval. Since some of the auxes are arrays and not single values, I handled this by writing any array representing an aux as a semicolon seperated string (i.e [1, 2, 3, 4] -> 1;2;3;4) (since separated using commas would interfere with the csv formatting).

I think we may want to consider a more sophisticated data output format in the future, possibly JSON instead of csv? With JSON we could have timestamps map to submodels, and then submodels map to their individual auxes in each data entry which would maintain the structure in how we saved it? Or alternatively output csvs for the individual submodels?

Espeer5:
After rebasing on main the issue with the shift of the GPP curve looks a whole lot better:
GPP

Now we are just underestimating as expected.

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
```diff
@@ -77,6 +76,8 @@ replace_missing_with_mean_by_value!(LW_OUT)
 
 LOCAL_DATETIME = DateTime.(string.(driver_data[2:end, 1]), "yyyymmddHHMM")
 UTC_DATETIME = LOCAL_DATETIME .+ Dates.Hour(6)
+DATA_DT = Second(LOCAL_DATETIME[2]-LOCAL_DATETIME[1]).value # seconds
```
kmdeck, line 79: thank you!! this is great

experiments/integrated/ozark/ozark.jl
```diff
@@ -367,198 +356,147 @@ E =
...
-    ylim = [0, 30],
+    data_daily_indices,
+    ET_avg_data,
+    label="Data ET",
+    xlabel="Time (H)",
+    ylabel="ET",
+    margins=10Plots.mm,
+    title="Average Daily ET Cycle",
 )
+Plots.plot!(plt1, 1:1:24, ET_avg_model, label="Model T", ylim=[0, 30])
```
kmdeck, line 373: could label "model ET"

experiments/integrated/ozark/ozark.jl
```diff
@@ -635,130 +574,168 @@ swp = [
...
-Plots.plot!(plt3, legend = :bottomleft)
-Plots.plot(plt3, plt2, layout = (2, 1))
+Plots.plot!(plt3, legend=:bottomleft)
+Plots.plot(plt3, plt2, layout=(2, 1))
 Plots.savefig(joinpath(savedir, "soil_temperature.png"))
 
-
-
-
+# Run script with comand line argument "save" to save model output to CSV
```
kmdeck, line 701: this looks great! how would you feel about a future PR which moves this into src?

@juliasloan25

kmdeck (approved):
looks great, thanks for your hard work on this @Espeer5! The averaged plots are really informative.

PR #331, opened 2023-09-16
Canopy temperature and emissivity

kmdeck:
## Purpose 

We would like to implement an equation solving for T_canopy prognostically. This PR is a precursor to a PR adding in the energy balance equation. 

The purpose of this is to add canopy temperature and emissivity to the equations implemented for the soil/canopy model.  Any time a temperature is used now, it is labeled as T_air or T_canopy for clarity (along with P_air, q_air, etc).  At the same time, we have added emissivity to the canopy RT parameters and that is used now as well.

We have also introduced an Energy component to the canopy. The only supported energy model is one in which the canopy temperature is set to the air temperature.
Therefore, this PR formalizes what we were doing already but has made it more extensible. Future energy component models which compute T_canopy/q_canopy prognostically will need to set that variable correctly via a new method of `canopy_temperature`, etc.  But no code anywhere else would need to change (in terms of flux computations).

https://github.com/CliMA/ProjectManagement/issues/2

## To-do
Review use of T_air/P_air/q_air in update_aux! of Canopy.jl (in computing the Medlyn term, GPP, etc)
Review the equations for net radiative fluxes for the soil, canopy, and outgoing radiation in soil_canopy_model.jl

## Content
- add energy component to the CanopyModel
- sets the default energy component to the PrescribedCanopyTempModel
- creates a function `canopy_temperature` which returns the air temperature when used with a PrescribedCanopyTempModel
- explicitly uses T_canopy, T_air, e_canopy in equations for fluxes instead of implicitly assuming T_canopy=T_air and e_canopy = 1
- Adds e_canopy to the RT parameter struct.

Review checklist

I have:

----

### DIFF
--- a/experiments/integrated/ozark/ozark_simulation.jl
+++ b/experiments/integrated/ozark/ozark_simulation.jl
@@ -1,10 +1,11 @@
 t0 = FT(120 * 3600 * 24)# start mid year
 N_days = 120
+dt = FT(120)
 tf = t0 + FT(3600 * 24 * N_days)
-dt = FT(240)
 # Number of timesteps between saving output
-n = 15
-saveat = Array(t0:(n * dt):tf)
+n = 30
+t_spinup = t0 + FT(10 * 3600 * 24)
+saveat = Array(t_spinup:(n * dt):tf)
 timestepper = CTS.RK4()
 # Set up timestepper
 ode_algo = CTS.ExplicitAlgorithm(timestepper)
--- a/src/integrated/soil_canopy_model.jl
+++ b/src/integrated/soil_canopy_model.jl
@@ -78,8 +78,8 @@ function soil_boundary_fluxes(
         t,
         model.parameters,
     )
-    net_energy_flux = @. p.soil_Rn + p.soil_lhf + p.soil_shf
-    return infiltration, net_energy_flux
+    G = @. -p.soil_LW_n - p.soil_SW_n + p.soil_lhf + p.soil_shf
+    return infiltration, G
 end
 
 """
@@ -168,13 +168,18 @@ included in the integrated Soil-Canopy model.
 """
 interaction_vars(m::SoilCanopyModel) = (
     :root_extraction,
-    :soil_Rn,
+    :soil_LW_n,
+    :soil_SW_n,
     :soil_evap,
     :soil_shf,
     :soil_lhf,
     :T_soil,
     :α_soil,
     :ϵ_soil,
+    :canopy_LW_n,
+    :canopy_SW_n,
+    :LW_out,
+    :SW_out,
 )
 
 """
@@ -184,7 +189,7 @@ The types of the additional auxiliary variables that are
 included in the integrated Soil-Canopy model.
 """
 interaction_types(m::SoilCanopyModel{FT}) where {FT} =
-    (FT, FT, FT, FT, FT, FT, FT, FT)
+    (FT, FT, FT, FT, FT, FT, FT, FT, FT, FT, FT, FT, FT)
 
 """
     interaction_domain_names(m::SoilCanopyModel)
@@ -194,6 +199,11 @@ included in the integrated Soil-Canopy model.
 """
 interaction_domain_names(m::SoilCanopyModel) = (
     :subsurface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
     :surface,
     :surface,
     :surface,
@@ -269,12 +279,12 @@ function make_interactions_update_aux(
         p.T_soil .= surface_temperature(land.soil, Y, p, t)
         p.α_soil .= surface_albedo(land.soil, Y, p)
         p.ϵ_soil .= surface_emissivity(land.soil, Y, p)
-        p.soil_Rn .= net_radiation_at_ground(
+        lsm_radiation_update!(
+            p,
             land.canopy.radiative_transfer,
             land.canopy,
             land.soil,
             Y,
-            p,
             t,
         )
 
@@ -299,12 +309,12 @@ give the canopy radiation model.
 Returns the correct radiative fluxes for bare ground in the case
 where the canopy LAI is zero.
 """
-function net_radiation_at_ground(
+function lsm_radiation_update!(
+    p,
     canopy_radiation::Canopy.AbstractRadiationModel{FT},
     canopy,
     ground_model::Soil.EnergyHydrology,
     Y,
-    p,
     t,
 ) where {FT}
     radiation = canopy.radiation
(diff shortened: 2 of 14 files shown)


src/standalone/Vegetation/Canopy.jl
```diff
@@ -140,6 +148,7 @@ function CanopyModel{FT}(;
     photosynthesis::AbstractPhotosynthesisModel{FT},
     conductance::AbstractStomatalConductanceModel{FT},
     hydraulics::AbstractPlantHydraulicsModel{FT},
+    energy = PrescribedCanopyTempModel{FT}(),
```
kmdeck, line 151: Setting this to be the default so we dont need to change all of other files right now. We can decide later if there should be a default and what it should be.

src/standalone/Vegetation/Canopy.jl
```diff
@@ -510,11 +520,11 @@ function ClimaLSM.make_update_aux(
 
         # We update the fa[n_stem+n_leaf] element once we have computed transpiration, below
         # update photosynthesis and conductance terms
-        medlyn_factor .= medlyn_term.(g1, T, P, q, Ref(thermo_params))
+        medlyn_factor .= medlyn_term.(g1, T_air, P, q, Ref(thermo_params))
```
kmdeck, line 523: should this be T_air? note that we pass P_air (P) and q_air (q). (perhaps we should make that clearer?)

src/standalone/Vegetation/Canopy.jl
```diff
@@ -618,7 +628,7 @@ A helper function which returns the surface temperature for the canopy
 model, which is stored in the aux state.
 """
 function ClimaLSM.surface_temperature(model::CanopyModel, Y, p, t)
-    return model.atmos.T(t)
+    return canopy_temperature(model.energy, model, Y, p, t)
```
kmdeck, line 631: currently, this returns `T_air(t)` because the type of model.energy is `PrescribedCanopyTemp`. When we add the prognostic model, this would return Y`.canopy.energy.T`. This is the temperature used to compute transpiration and sensible heat fluxes.

src/standalone/Vegetation/canopy_energy.jl
```diff
@@ -0,0 +1,13 @@
+export PrescribedCanopyTempModel
+abstract type AbstractCanopyEnergyModel{FT} <: AbstractCanopyComponent{FT} end
+
+ClimaLSM.name(model::AbstractCanopyEnergyModel) = :energy
+
+## PrescribedCanopyTemperature equal to atmos temperature
```
kmdeck, line 6: will add a comment describing this

src/integrated/soil_canopy_model.jl
```diff
@@ -310,20 +320,85 @@ function net_radiation_at_ground(
...
+
+    @. SW_net_canopy =
+        (energy_per_photon_PAR * N_a * APAR) +
+        (energy_per_photon_NIR * N_a * ANIR) # as before
+    SW_d_canopy = @. SW_d - SW_net_canopy # not sure if this is correct
+    @. SW_net_soil = (1 - α_soil) * SW_d_canopy
+    @. SW_out = SW_d - SW_net_canopy - SW_net_soil # SW_out - SW_d = - SW_net_land
 end
 
+"""
```
kmdeck, line 357: everything below here was modified in PR #329

src/integrated/soil_canopy_model.jl
```diff
@@ -78,7 +78,7 @@ function soil_boundary_fluxes(
         t,
         model.parameters,
     )
-    net_energy_flux = @. p.soil_Rn + p.soil_lhf + p.soil_shf
+    net_energy_flux = @. -p.soil_LW_n - p.soil_SW_n + p.soil_lhf + p.soil_shf
```
kmdeck, line 81: currently, LW_n and SW_n are positive if the soil absorbs energy. In the soil PDE, this must be a negative (downwards) flux, hence the minus signs here.

braghiere: net_energy_flux is G in this case?

src/integrated/soil_canopy_model.jl
```diff
@@ -310,18 +320,38 @@ function net_radiation_at_ground(
...
+
+
+    LW_d_canopy = @. (1 - ϵ_canopy) * LW_d + ϵ_canopy * _σ * T_canopy^4 # double checked
+    LW_u_soil = @. ϵ_soil * _σ * T_soil^4 + (1 - ϵ_soil) * LW_d_canopy # double checked
+    @. LW_net_canopy =
+        ϵ_canopy * LW_d - 2 * ϵ_canopy * _σ * T_canopy^4 + ϵ_canopy * LW_u_soil # double checked
+    @. LW_net_soil = ϵ_soil * LW_d_canopy - ϵ_soil * _σ * T_soil^4 # double checked
+    @. LW_out = (1 - ϵ_canopy) * LW_u_soil * ϵ_canopy * _σ * T_canopy^4 # double checked
+
+    @. SW_net_canopy =
```
braghiere, line 349: Sw_net_canopy = absorbed - reflected - transmitted 

these 3 terms should come from the two-stream scheme. 

SW_d_canopy = transmitted

braghiere (approved):
There are a few terms missing in SW balance. I've added a few comments.

src/integrated/soil_canopy_model.jl
kmdeck: yes - Ill change the name in this PR too!

kmdeck:
@braghiere I see that in CLM, they set the emissivity of the canopy to: 1- exp[-(LAI+SAI)/mu] where mu is the average optical depth in the canopy (set to 1m). Equation 4.20:https://www2.cesm.ucar.edu/models/cesm2/land/CLM50_Tech_Note.pdf

is this something we want to do, or do we want to leave it as a free parameter?

src/integrated/soil_canopy_model.jl
kmdeck: Thank you!

src/integrated/soil_canopy_model.jl
```diff
@@ -310,18 +320,38 @@ function net_radiation_at_ground(
...
+        (energy_per_photon_PAR * N_a * APAR) +
+        (energy_per_photon_NIR * N_a * ANIR) # as before
+    SW_d_canopy = @. SW_d - SW_net_canopy # not sure if this is correct
+    @. SW_net_soil = (1 - α_soil) * SW_d_canopy
+    @. SW_out = SW_d - SW_net_canopy - SW_net_soil # SW_out - SW_d = - SW_net_land
```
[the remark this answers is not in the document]
Espeer5: These terms are essentially what I added to the soil_canopy_EB branch in order to compute the surface albedo. I can add them to this branch. It's not hard, It comes from saving the weighted sum of the diffuse and direct upward radiation in the out of the top layer in the two-stream

src/integrated/soil_canopy_model.jl
```diff
@@ -199,6 +204,11 @@ interaction_domain_names(m::SoilCanopyModel) = (
     :surface,
     :surface,
     :surface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
```
charleskawczynski, line 211: What's going here?

kmdeck: In the land model, we have variables that live on different parts of the domain. These methods (along with ones defining variable symbol names and variable types) help us create the prognostic state and auxiliary state in an automated way.

src/integrated/soil_canopy_model.jl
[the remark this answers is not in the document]
kmdeck: thank you! feel free to commit directly to this branch, or you can point me to the lines of code in the other branch if that is easier.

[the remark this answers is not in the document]
kmdeck: I think I found it:
```
   APAR_sl = @. (1 - α_soil_PAR) * (PAR - APAR)
    ANIR_sl = @. (1 - α_soil_NIR) * (NIR - ANIR)
    bulk_reflected_sw = @. PAR + NIR - (APAR + ANIR + APAR_sl + ANIR_sl)
    ```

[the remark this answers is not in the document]
Comment (braghiere) [reply src/integrated/soil_canopy_model.jl:354]: That doesn't look right to me. There's one term missing. PAR - APAR = RPAR + TPAR (1 - alpha_soil). 

APAR + RPAR  + TPAR * (1 - alpha_soil) = 1.0 

A: absorbed
R: reflected 
T: transmitted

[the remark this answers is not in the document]
Comment (Espeer5) [reply src/integrated/soil_canopy_model.jl:354]: Oh I see you're right

In that case we should be able to get it directly from what the two stream computes instead. For layer i==1 the weighted sum of `I_dir_up` and `I_dif_up` should give the total coming out of the top of the canopy. I've tried doing that [here](https://github.com/CliMA/ClimaLSM.jl/blob/add_rpar_rnir/src/standalone/Vegetation/canopy_parameterizations.jl) but I am having a little bit of trouble with returning a tuple of both `RPAR` and `APAR` from `plant_absorbed_pfd` when we call it with broadcasting in `compute_absorbances`.

If we pull it directly from the TwoStream model like this, how should we then compute it for Beer-Lambert?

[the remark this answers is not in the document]
Comment (kmdeck) [reply src/integrated/soil_canopy_model.jl:354]: we should also add RNIR, RPAR, TPAR, and TNIR to the auxiliary state.

Ill be working on this in about an hour if you want to join for pair programming (using VS code LiveShare) but no worries if not!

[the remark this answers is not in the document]
Comment (braghiere) [reply src/integrated/soil_canopy_model.jl]: That's exactly right! RPAR = I_dir_up + I_dif_up at the the top of the canopy (first layer). 

And then TPAR  = (1.0 - APAR - RPAR)/ (1 - alpha_soil)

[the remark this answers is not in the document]
Comment (Espeer5) [reply src/integrated/soil_canopy_model.jl]: I went ahead and added a minimum working example in the script `field_testing.jl`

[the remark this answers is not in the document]
Comment (kmdeck) [reply src/integrated/soil_canopy_model.jl]: thank you!! this was great. I played around with it - and tried three different options. one is keeping the fields y,z as fields of floats, the other is to make a single field of a 2-tuple and the third is a field of NamedTuples (;y = , z=).

Im not sure how to make the two fields of floats non-allocating if our function returns a two things (either as a tuple or named tuple). If our field is tuple-valued or namedtuple-valued, we can broadcast like we want and not allocate.

if you run the test, you can see that the allocating version is ~5x slower for this field size.

so, what about the following? for this PR, we do the allocating version. Then we have a followup version where the aux variables for radiation become :par, :nir and each is NamedTuple valued, with keys = :par, :apar:, :rpar, :tpar, each with float values for example. then if we have a function that returns a named tuple, we should be able to broadcast/assign a field of named tuples to the output of the broadcasted function.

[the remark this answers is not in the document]
Comment (kmdeck) [reply src/integrated/soil_canopy_model.jl]: or possibly, we can make another PR first which fixes up the RT calls, and then we can rebase this on off of that.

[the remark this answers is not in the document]
Comment (kmdeck) [reply src/integrated/soil_canopy_model.jl]: > Oh I see you're right
> 
> In that case we should be able to get it directly from what the two stream computes instead. For layer `i==1` the weighted sum of `I_dir_up` and `I_dif_up` should give the total shortwave coming out of the top of the canopy. I've tried doing that [here](https://github.com/CliMA/ClimaLSM.jl/blob/add_rpar_rnir/src/standalone/Vegetation/canopy_parameterizations.jl) but I am having a little bit of trouble with returning a tuple of both `RPAR` and `APAR` from `plant_absorbed_pfd` when we call it with broadcasting in `compute_absorbances`.
> 
> If we pull it directly from the TwoStream model like this, how should we then compute it for Beer-Lambert?

@braghiere any references or ideas for the transmitted and reflected #s for Beer Lambert?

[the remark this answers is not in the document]
Comment (braghiere) [reply src/integrated/soil_canopy_model.jl]: For Beer-Lambert it should be straightforward:

APAR = (1 - α_leaf) * (1 - exp(-K * LAI * RTP.Ω)*(1 - alpha_soil))
TPAR = exp(-K * LAI * RTP.Ω)
APAR + RPAR + TPAR * (1 - alpha_soil) = 1.0

RPAR = 1. - APAR - TPAR * (1 - alpha_soil)

File: experiments/integrated/ozark/ozark_simulation.jl [hunk 10]
```diff
@@ -1,10 +1,11 @@
 t0 = FT(120 * 3600 * 24)# start mid year
 N_days = 120
+dt = FT(120)
 tf = t0 + FT(3600 * 24 * N_days)
-dt = FT(240)
 # Number of timesteps between saving output
-n = 15
-saveat = Array(t0:(n * dt):tf)
+n = 30
+t_spinup = t0 + FT(10 * 3600 * 24)
```
kmdeck, line 7: Added a spinup of 10 days

kmdeck:
SW out and LW out look pretty good!
SW
LW

PR #342, opened 2023-09-26
replace MapInfo with PrescribedDataStatic

### LINKED ISSUE #295: unify BulkAlbedoStatic, BulkAlbedoTemporal types
As of #280, we have two structs to store information related to reading in data files either at one point in time (static) or repeatedly over time (temporal). These two structs use different structs under the hood but store similar types of information. We should combine the internal structs to reduce the number of types we define and have a cleaner interface.

juliasloan25:
## Purpose 
Currently, the `BulkAlbedoStatic` and `BulkAlbedoTemporal` types make use of very similar data structures, `MapInfo` and `FileInfo` respectively. Instead of keeping both of these around, we should unify them and just keep one. It's easier to make the change to keep `FileInfo`, so I've done that.

Once we've unified these two types, it will be easier to generalize the albedo types we have defined to input data for any variable.

Closes #295

## To Do
- [x] add `infile_path` to `FileInfo`, update constructor calls to include it
- [x] reorder `FileInfo` struct fields so time-related ones are last (makes `PrescribedDataStatic` constructor clearer)
- [x] replace `MapInfo` with `PrescribedDataStatic ` in `BulkAlbedoStatic` constructor
- [x] add `PrescribedDataTemporal` and `PrescribedDataStatic` types
- [x] add `PrescribedDataStatic` constructor which makes a `FileInfo` internally

----

### DIFF
--- a/src/shared_utilities/FileReader.jl
+++ b/src/shared_utilities/FileReader.jl
@@ -19,16 +19,62 @@ using CFTime
 
 using ClimaLSM.Regridder
 
-export to_datetime,
+export AbstractPrescribedData,
+    PrescribedDataTemporal,
+    PrescribedDataStatic,
     FileInfo,
     FileState,
     SimInfo,
-    PrescribedData,
-    prescribed_data_init,
     read_data_fields!,
     next_date_in_file,
-    interpolate_data
+    interpolate_data,
+    to_datetime
 
+
+"""
+    abstract type AbstractPrescribedData
+
+An abstract type for storing prescribed data info. Subtypes
+include temporally-varying prescribed data and static prescribed data.
+"""
+abstract type AbstractPrescribedData end
+
+"""
+    PrescribedDataStatic <: AbstractPrescribedData
+
+Stores information to read in a prescribed variable from a file.
+The data is read in once and stored without changing for the duration of a
+simulation. This type is meant to be used with input data that does not have
+a time dimension.
+Each of the fields of this struct is itself a struct.
+
+# Inputs:
+- file_info::FI     # unchanging info about the input data file
+"""
+struct PrescribedDataStatic{FI} <: AbstractPrescribedData
+    file_info::FI
+end
+
+"""
+    PrescribedDataTemporal <: AbstractPrescribedData
+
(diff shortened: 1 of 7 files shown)


kmdeck (approved):
Hi Julia! This looks good to me in terms of the code. I am just wondering about:
- should we have a PrecribedTemporalData and PrescribedStaticData object? (current PrescribedData -> PrecribedTemporalData) and PrescribedStaticData just has one field, `FileInfo`? Maybe this is just excessive and not needed, but I found it a little conceptually unclear that PrescribedData has three fields, one of which is FileInfo, and then we use FileInfo directly when we prescribed static data. 
- If it's better to leave as is for now until we have more use cases using static info, let's add some documentation explaining that if you prescribed static data, FileInfo is all you need.

juliasloan25:
You make a good point! I agree it's not very clear that one uses PrescribedData and one just uses FileInfo directly... I can make the change in this PR if we want to go in the direction of PrescribedTemporalData and PrescribedStaticData (which I think is a good idea)

kmdeck:
> You make a good point! I agree it's not very clear that one uses PrescribedData and one just uses FileInfo directly... I can make the change in this PR if we want to go in the direction of PrescribedTemporalData and PrescribedStaticData (which I think is a good idea)

ok! let's go for it :)

src/shared_utilities/FileReader.jl
```diff
@@ -19,34 +19,82 @@ using CFTime
...
+    to_datetime
 
 
+"""
+    abstract type AbstractPrescribedData
+
+An abstract type for storing prescribed data info. Subtypes
+include temporally-varying prescribed data and static prescribed data.
+"""
+abstract type AbstractPrescribedData end
```
kmdeck, line 40: looks good! I like the idea of eventually having methods for each concrete type of this for `next_field!` and `set_initial_field!` (similar to what we do with the albedo now, where `next_field!` doenst update anything in one case, but does in another. not sure how that will play out, just an idea for now!

kmdeck (approved):
looks great!

PR #344, opened 2023-10-03
Add heterotrophic respiration to integrated/soil_canopy_model.jl

AlexisRenchon:
## Purpose 

Add heterotrophic respiration (soil CO2 efflux from microbial decomposition of soil organic matter) to our integrated model "soil_canopy_model.jl" - that we are using to do experiments with ozark. 

This model tracks soil CO2 concentration in the soil porous air space prognostically, as a result of co2 production by microbial decomposition of SOM and co2 diffusion through soil layers. co2 production and co2 diffusion are aux variables, varying with soil depth and time, both are function of soil temperature and moisture - and other variables. 

## To-do

- [x] Edit standalone model: prescribed air pressure  
- [x] Add the model in src/integrated/soil_canopy_model.jl
- [x] Update experiment/ozark 
- [x] Update tutorials and docs 
- [x] Update tests
- [x] Edit docstrings
- [x] Get rid of `eltype`

## Content

See the to-do list

----

### DIFF
--- a/docs/tutorials/Canopy/soil_canopy_tutorial.jl
+++ b/docs/tutorials/Canopy/soil_canopy_tutorial.jl
@@ -46,6 +46,7 @@ import ClimaTimeSteppers as CTS
 using ClimaLSM
 using ClimaLSM.Domains: Column, obtain_surface_domain
 using ClimaLSM.Soil
+using ClimaLSM.Soil.Biogeochemistry
 using ClimaLSM.Canopy
 using ClimaLSM.Canopy.PlantHydraulics
 import ClimaLSM
@@ -163,6 +164,65 @@ soil_ps = Soil.EnergyHydrologyParameters{FT}(;
 
 soil_args = (domain = soil_domain, parameters = soil_ps)
 soil_model_type = Soil.EnergyHydrology{FT}
+
+# For the heterotrophic respiration model, see the
+# [documentation](https://clima.github.io/ClimaLSM.jl/previews/PR214/dynamicdocs/pages/soil_biogeochemistry/microbial_respiration/)
+# to understand the parameterisation. 
+# The domain is defined similarly to the soil domain described above.
+ν = soil_ν # defined above
+θ_a100 = FT(0.1816)
+D_ref = FT(1.39e-5)
+b = FT(4.547)
+D_liq = FT(3.17)
+α_sx = FT(194e3)
+Ea_sx = FT(61e3)
+kM_sx = FT(5e-3)
+kM_o2 = FT(0.004)
+O2_a = FT(0.209)
+D_oa = FT(1.67)
+p_sx = FT(0.024)
+
+soilco2_type = Soil.Biogeochemistry.SoilCO2Model{FT}
+
+soilco2_ps = SoilCO2ModelParameters{FT}(;
+    ν = soil_ν,
+    θ_a100 = θ_a100,
+    D_ref = D_ref,
+    b = b,
+    D_liq = D_liq,
+    α_sx = α_sx,
+    Ea_sx = Ea_sx,
+    kM_sx = kM_sx,
+    kM_o2 = kM_o2,
+    O2_a = O2_a,
+    D_oa = D_oa,
+    p_sx = p_sx,
+    earth_param_set = earth_param_set,
+);
+
+# soil microbes args
+Csom = (z, t) -> 5.0; # kg C m⁻³, this is a guess, not measured at the site
+
+soilco2_top_bc = Soil.Biogeochemistry.SoilCO2StateBC((p, t) -> atmos_co2(t));
+soilco2_bot_bc = Soil.Biogeochemistry.SoilCO2StateBC((p, t) -> 0.0);
+soilco2_sources = (MicrobeProduction{FT}(),);
+
+soilco2_boundary_conditions =
+    (; top = (CO2 = soilco2_top_bc,), bottom = (CO2 = soilco2_bot_bc,));
+
+soilco2_drivers = Soil.Biogeochemistry.SoilDrivers(
+    Soil.Biogeochemistry.PrognosticMet(),
+    Soil.Biogeochemistry.PrescribedSOC(Csom),
+    atmos,
+);
+
+soilco2_args = (;
+    boundary_conditions = soilco2_boundary_conditions,
+    sources = soilco2_sources,
+    domain = soil_domain,
+    parameters = soilco2_ps,
+    drivers = soilco2_drivers,
+);
 
 # Next we need to set up the [`CanopyModel`](https://clima.github.io/ClimaLSM.jl/dev/APIs/canopy/Canopy/#Canopy-Model-Structs).
 # For more details on the specifics of this model see the previous tutorial.
@@ -321,6 +381,8 @@ canopy_model_args = (; parameters = shared_params, domain = canopy_domain);
 land_input = (atmos = atmos, radiation = radiation)
 
 land = SoilCanopyModel{FT}(;
+    soilco2_type = soilco2_type,
+    soilco2_args = soilco2_args,
     land_args = land_input,
     soil_model_type = soil_model_type,
     soil_args = soil_args,
@@ -351,6 +413,9 @@ Y.soil.ρe_int =
         T_0,
         Ref(land.soil.parameters),
     )
+
+Y.soilco2.C .= FT(0.000412) # set to atmospheric co2, mol co2 per mol air
+
 ψ_stem_0 = FT(-1e5 / 9800)
 ψ_leaf_0 = FT(-2e5 / 9800)
 
--- a/experiments/integrated/ozark/conservation/ozark_conservation.jl
+++ b/experiments/integrated/ozark/conservation/ozark_conservation.jl
@@ -10,6 +10,7 @@ using Insolation
 using ClimaLSM
 using ClimaLSM.Domains: Column
 using ClimaLSM.Soil
+using ClimaLSM.Soil.Biogeochemistry
 using ClimaLSM.Canopy
 using ClimaLSM.Canopy.PlantHydraulics
 import ClimaLSM
@@ -63,6 +64,50 @@ soil_ps = Soil.EnergyHydrologyParameters{FT}(;
 
 soil_args = (domain = soil_domain, parameters = soil_ps)
 soil_model_type = Soil.EnergyHydrology{FT}
+
+# Soil microbes model
+soilco2_type = Soil.Biogeochemistry.SoilCO2Model{FT}
+
+soilco2_ps = SoilCO2ModelParameters{FT}(;
+    ν = soil_ν,
+    θ_a100 = θ_a100,
+    D_ref = D_ref,
+    b = b,
+    D_liq = D_liq,
+    # DAMM
+    α_sx = α_sx,
+    Ea_sx = Ea_sx,
+    kM_sx = kM_sx,
+    kM_o2 = kM_o2,
+    O2_a = O2_a,
+    D_oa = D_oa,
+    p_sx = p_sx,
+    earth_param_set = earth_param_set,
+);
+
+# soil microbes args
+Csom = (z, t) -> 5.0
+
+soilco2_top_bc = Soil.Biogeochemistry.SoilCO2StateBC((p, t) -> atmos_co2(t))
+soilco2_bot_bc = Soil.Biogeochemistry.SoilCO2StateBC((p, t) -> 0.0)
+soilco2_sources = (MicrobeProduction{FT}(),)
+
+soilco2_boundary_conditions =
+    (; top = (CO2 = soilco2_top_bc,), bottom = (CO2 = soilco2_bot_bc,))
+
+soilco2_drivers = Soil.Biogeochemistry.SoilDrivers(
+    Soil.Biogeochemistry.PrognosticMet(),
+    Soil.Biogeochemistry.PrescribedSOC(Csom),
+    atmos,
+)
+
+soilco2_args = (;
+    boundary_conditions = soilco2_boundary_conditions,
+    sources = soilco2_sources,
+    domain = soil_domain,
+    parameters = soilco2_ps,
+    drivers = soilco2_drivers,
+)
 
 # Now we set up the canopy model, which we set up by component:
 # Component Types
@@ -178,6 +223,8 @@ canopy_model_args = (; parameters = shared_params, domain = canopy_domain)
 # Integrated plant hydraulics and soil model
 land_input = (atmos = atmos, radiation = radiation)
 land = SoilCanopyModel{FT}(;
+    soilco2_type = soilco2_type,
+    soilco2_args = soilco2_args,
     land_args = land_input,
     soil_model_type = soil_model_type,
     soil_args = soil_args,
@@ -204,6 +251,9 @@ Y.soil.ρe_int =
         T_0,
         Ref(land.soil.parameters),
     )
+
+Y.soilco2.C .= FT(0.000412) # set to atmospheric co2, mol co2 per mol air
+
 ψ_stem_0 = FT(-1e5 / 9800)
 ψ_leaf_0 = FT(-2e5 / 9800)
 
--- a/experiments/integrated/ozark/ozark.jl
+++ b/experiments/integrated/ozark/ozark.jl
@@ -11,6 +11,7 @@ using StatsBase
 using ClimaLSM
 using ClimaLSM.Domains: Column
 using ClimaLSM.Soil
+using ClimaLSM.Soil.Biogeochemistry
 using ClimaLSM.Canopy
 using ClimaLSM.Canopy.PlantHydraulics
 import ClimaLSM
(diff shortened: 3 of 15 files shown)


AlexisRenchon:
image

docs/tutorials/Canopy/soil_canopy_tutorial.jl
```diff
@@ -109,6 +110,67 @@ include(
 # Define the parameters for the soil model and provide them to the model 
 # parameters struct:
 
+# Heterotrophic respiration parameters
+P_sfc = FT(101e3)
```
kmdeck, line 114: Im not sure this should be a parameter - should we get it from the PrescribedAtmos struct instead?

docs/tutorials/Canopy/soil_canopy_tutorial.jl
```diff
@@ -109,6 +110,67 @@ include(
...
+ν = FT(0.556)
```
kmdeck, line 115: how do we ensure this is the same as the soil porosity? we should only define 1x

docs/tutorials/Canopy/soil_canopy_tutorial.jl
```diff
@@ -109,6 +110,67 @@ include(
...
+θ_a100 = FT(0.1816)
+D_ref = FT(1.39e-5)
+b = FT(4.547)
+D_liq = FT(3.17)
```
kmdeck, line 119: Let's make a note of which parameters are constant in space and should be added to CLIMAParameters, and open an issue so that we keep track of this

docs/tutorials/Canopy/soil_canopy_tutorial.jl
```diff
@@ -109,6 +110,67 @@ include(
...
+    Ea_sx = Ea_sx,
+    kM_sx = kM_sx,
+    kM_o2 = kM_o2,
+    O2_a = O2_a,
+    D_oa = D_oa,
+    p_sx = p_sx,
+    earth_param_set = earth_param_set,
+);
+
+# soil microbes args
```
kmdeck, line 147: Could you explain a little what you are specifying here? (as this is a tutorial)

also check out the rendered docs. Usually you need a `;` after the last line to prevent it from outputing into the rendered tutorial.

docs/tutorials/Canopy/soil_canopy_tutorial.jl
```diff
@@ -109,6 +110,67 @@ include(
...
+Csom = (z, t) -> eltype(t)(5.0)
+
+C = FT(100)
```
kmdeck, line 150: Could you explain why setting this to 100 is reasonable? What are the units?

docs/tutorials/Canopy/soil_canopy_tutorial.jl
```diff
@@ -109,6 +110,67 @@ include(
...
+
+soilco2_top_bc = Soil.Biogeochemistry.SoilCO2StateBC((p, t) -> eltype(t)(0.0))
```
kmdeck, line 152: Could you explain why zero flux is reasonable? thanks!

kmdeck: note also Julia's message in the current land chat about specifying Float type. `t` is Float64 (and may in the future be a DateTime), so we cant use eltype(t) anymore. instead we convert it internally (her [PR](https://github.com/CliMA/ClimaLSM.jl/pull/353) takes care of this, but we should try and be consistent about how we handle t in tutorials too

experiments/integrated/ozark/conservation/ozark_conservation.jl
```diff
@@ -200,6 +249,9 @@ Y.soil.ρe_int =
         T_0,
         Ref(land.soil.parameters),
     )
+
+Y.soilco2.C .= FT(100)
```
kmdeck, line 253: set equal to C, define above?

experiments/integrated/ozark/ozark.jl
```diff
@@ -62,6 +63,53 @@ soil_ps = Soil.EnergyHydrologyParameters{FT}(;
...
+    O2_a = O2_a,
+    D_oa = D_oa,
+    p_sx = p_sx,
+    earth_param_set = earth_param_set,
+);
+
+# soil microbes args
+Csom = (z, t) -> eltype(t)(5.0)
+
+# C = FT(100)
```
kmdeck, line 90: can remove (across all files where it isnt used)

experiments/integrated/ozark/ozark.jl
```diff
@@ -62,6 +63,53 @@ soil_ps = Soil.EnergyHydrologyParameters{FT}(;
...
+
+soilco2_top_bc =
+    Soil.Biogeochemistry.SoilCO2StateBC((p, t) -> eltype(t)(0.0004)) # top BC set to atmospheric co2 in mol per mol
```
kmdeck, line 93: Should this be set to the atmospheric co2 (your comment)?

experiments/integrated/ozark/ozark_parameters.jl
```diff
@@ -1,3 +1,19 @@
+# Heterotrophic respiration parameters
+P_sfc = FT(101e3)
```
kmdeck, line 2: same comment - should we get this from the atmos struct instead?

src/integrated/soil_canopy_model.jl
```diff
@@ -160,7 +174,22 @@ function SoilCanopyModel{FT}(;
...
+    soilco2 = Soil.Biogeochemistry.SoilCO2Model{FT}(; soilco2_args...)
+
+    # not sure if the if below is needed
+    if typeof(soilco2_args.drivers.met) != PrognosticMet
+        throw(
+            AssertionError(
+                "To run with a prognostic soil energy and hydrology model, the met driver must be of type PrognosticSoil.",
+            ),
+        )
+    end
```
kmdeck, line 186: can also check that soil porosity is the same between soil & soil co2


if we have other "requirements" we want to enforce, here is the place to do them. E.g. setting the BC for the soil co2 to depend on air co2.

kmdeck (changes requested):
a few changes, but almost there! I left several comments, let me know if you have ?s. thanks for adding this!

AlexisRenchon:
> a few changes, but almost there! I left several comments, let me know if you have ?s. thanks for adding this!

Thanks for the great comments! I agree with all and will work on it. 
Once we merge this, I'd like to do add soil O2 as prognostic variable =) (instead of prescribed)

experiments/integrated/ozark/ozark.jl
```diff
@@ -62,6 +63,53 @@ soil_ps = Soil.EnergyHydrologyParameters{FT}(;
...
+soilco2_bot_bc = Soil.Biogeochemistry.SoilCO2FluxBC((p, t) -> eltype(t)(0)) # no flux
+soilco2_sources = (MicrobeProduction{FT}(),)
+
+soilco2_boundary_conditions =
+    (; top = (CO2 = soilco2_top_bc,), bottom = (CO2 = soilco2_bot_bc,))
+
+soilco2_drivers = Soil.Biogeochemistry.SoilDrivers(
+    Soil.Biogeochemistry.PrognosticMet(),
```
AlexisRenchon, line 101: Where is `PrognosticMet()` defined?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -306,6 +307,8 @@ struct SoilDrivers
     met::AbstractSoilDriver
     "Soil SOM driver - Prescribed only"
     soc::AbstractSoilDriver
+    "Prescribed atmospheric pressure"
+    atm_p::PrescribedAtmosphere
```
AlexisRenchon, line 311: @kmdeck I imagine this is where we would have atmospheric pressure (I will remove it from parameters), and I think the type is `::PrescibedAtmosphere` since this exist already, however it now feels weird to put this in `AbstractSoilDriver`. 

I am not sure what the most elegant way to do this is. 
I imagine we could rename `SoilDriver` to `Driver`, or have a struct that hold `SoilDriver` and `AtmosphericDriver`, ...?

docs/tutorials/Canopy/soil_canopy_tutorial.jl
AlexisRenchon: I agree. 
I started working on this - this will modify the biogeochemistry model (deleting a parameter, adding a driver), so there will be a bunch of edits required throughout the repo

docs/tutorials/Canopy/soil_canopy_tutorial.jl
AlexisRenchon: I just did `v = soil_v` - which is what I had in experiment, I forgot to edit it for the tutorial. 
Do you think that's OK?

docs/tutorials/Canopy/soil_canopy_tutorial.jl
AlexisRenchon: Do you mean constants?
I thought we were going to put all parameters in CLIMAParameters (both constants and variable parameters), and then specify what parameter we want to fit when we do data assimilation 
(which seemed weird to me - I like to be explicit about which parameters are constant or not). We can discuss more about this.

docs/tutorials/Canopy/soil_canopy_tutorial.jl
AlexisRenchon: ugh sorry I had updated this in ozark.jl experiment but not the tutorials.

docs/tutorials/Canopy/soil_canopy_tutorial.jl
AlexisRenchon: I am not yet 100% sure what to do at bottom BC. 
I would need to look at actual data of soil co2 concentration at depth - I wonder if it is very low or very high. 
I don't want to have an infinite source of CO2 from bottom BC (which does happen in some volcanic areas though, but that's the exception). 
And if we say it's 0, we'll have co2 diffusing to below BC constantly, which I imagine would not happen at equilibrium. 

I feel like there must be an equilibrium where diffusion to bottom BC is very small? But I am not sure 100%.

experiments/integrated/ozark/conservation/ozark_conservation.jl
AlexisRenchon: Thanks!
I've set it to atmospheric co2, same as ozark.jl and tutorial

experiments/integrated/ozark/ozark.jl
AlexisRenchon: yes - should we grab that from `PrescribedAtmosphere`?

docs/tutorials/Canopy/soil_canopy_tutorial.jl
AlexisRenchon: ah sorry I thought this was bottom bc. I am changing top to atmospheric C

docs/tutorials/Canopy/soil_canopy_tutorial.jl
kmdeck: I think it is OK for now! Once we go to global runs, soil porosity will be stored in the aux state because it will vary everywhere. so we'll have to make sure this model gets the right value for each location.

docs/tutorials/Canopy/soil_canopy_tutorial.jl
kmdeck: yep! I mean any parameters that are the same everywhere in space (whether or not we calibrate them). Not super important right at this stage, but it might be good to open the issue while this is fresh.

docs/tutorials/Canopy/soil_canopy_tutorial.jl
kmdeck: Re: the bottom BC, it's probably OK for now! might be another thing to open an issue on, just to keep track of it?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
kmdeck: yeah! that makes sense to me. and then the SoilDriver can be prescribed or prognostic, and eventually the atmos driver can be prescribed or coupled?

FWIW, I believe both the canopy and the soil model have copies of the prescribed Atmos struct in their models, so having it here too is probably fine. Or, if you want, you can just store the co2 and pressure, and can extract them from the PrescribedAtmos object in the constructor for the integrated soil_canopy model?

[For global runs, we'll have the atmos variables in `p`, and then any model can access them that way]

docs/tutorials/Canopy/soil_canopy_tutorial.jl
AlexisRenchon: I will do that on the prognostic O2 PR - I hope we can merge this one before the sprint Friday

experiments/integrated/ozark/ozark.jl
AlexisRenchon: I tried `atmos_co2` which is our Spline1D created on co2 data, but it bugs. I'll mark this as resolved once we find what works (probably a type problem)

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -371,6 +370,15 @@ function soil_SOM_C(driver::PrescribedSOC, p, Y, t, z)
     return driver.soil_organic_carbon.(z, t)
 end
 
+"""
+    air_pressure(driver::PrescribedAtmosphere, t)
+
+Returns the prescribed air pressure at the top boundary condition at time (t).
+"""
+function air_pressure(driver::PrescribedAtmosphere, p, Y, t) # not sure if/why p and Y are needed?
```
kmdeck, line 378: I think in this case, Y is not needed. we could also remove `p` if you want, but I *think* that when we couple to the atmosphere, the coupler will store the atmos pressure in our aux state somewhere. so when driver -> CoupledAtmosphere(), we may need `p`

test/standalone/Soil/Biogeochemistry/biogeochemistry_module.jl
```diff
@@ -73,7 +101,35 @@ end
     sources = ()
     boundary_conditions = (; top = (CO2 = top_bc,), bottom = (CO2 = bot_bc,))
 
-    soil_drivers = SoilDrivers(PrescribedMet(T_soil, θ_l), PrescribedSOC(Csom))
+    # Make a PrescribedAtmosphere - we only care about atmos_p though
+    precipitation_function = (t) -> FT(1)
```
kmdeck, line 105: this is tricky, but we shouldnt write " (t) -> FT(1)" and pass that into the model. This is because we are asking the model to store this function:
```
function anonymous(t)
    return FT(1)
end
```
when this gets evaluated by the model each step, it will have to look for FT as a global parameter (in global scope), and this is slow. We *used* to get around this by writing `(t) ->eltype(t)(1)`, because `eltype(t)` can be computed without looking up anything in global scope. since we now arent guaranteed that the type of (t) will be the same FT as our simulation, we will just pass `(t) -> 1.0` and convert the output of this function internally to FT (where FT is accessible because it is stored in the model struct itself, or obtained from eltype(Y.soil.co2.C), or something like that.

experiments/standalone/Biogeochemistry/experiment.jl
```diff
@@ -92,10 +93,36 @@ co2_bot_bc = Soil.Biogeochemistry.SoilCO2StateBC((p, t) -> eltype(t)(0.0))
 co2_sources = (MicrobeProduction{FT}(),)
 co2_boundary_conditions =
     (; top = (CO2 = co2_top_bc,), bottom = (CO2 = co2_bot_bc,))
+
+# Make a PrescribedAtmosphere - we only care about atmos_p though
+precipitation_function = (t) -> FT(1)
```
kmdeck, line 98: same comment re: use of FT here.

src/integrated/soil_canopy_model.jl
```diff
@@ -145,7 +166,22 @@ function SoilCanopyModel{FT}(;
...
     end
 
-    return SoilCanopyModel{FT, typeof(soil), typeof(canopy)}(soil, canopy)
+    soilco2 = Soil.Biogeochemistry.SoilCO2Model{FT}(; soilco2_args...)
+
+    # not sure if the if below is needed
+    if typeof(soilco2_args.drivers.met) != PrognosticMet
+        throw(
+            AssertionError(
+                "To run with a prognostic soil energy and hydrology model, the met driver must be of type PrognosticSoil.",
```
kmdeck, line 175: maybe "must be of type PrognosticMet"?

experiments/integrated/ozark/ozark.jl
```diff
@@ -62,6 +63,50 @@ soil_ps = Soil.EnergyHydrologyParameters{FT}(;
...
+    O2_a = O2_a,
+    D_oa = D_oa,
+    p_sx = p_sx,
+    earth_param_set = earth_param_set,
+);
+
+# soil microbes args
+Csom = (z, t) -> 5.0
+
+soilco2_top_bc = Soil.Biogeochemistry.SoilCO2StateBC((p, t) -> atmos_co2(t))
```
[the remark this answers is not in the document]
AlexisRenchon: good point, let's keep this in mind for now!

PR #348, opened 2023-10-05
Surface resistance effects on SHF

kmdeck:
## Purpose 
Previously, we had accounted for the surface resistance to evaporation/vapor loss by calling a wrapper function (defined in ClimaLSM) of SurfaceFluxes functions:
https://github.com/CliMA/ClimaLSM.jl/blob/2b56772e704e4b53f2b46a7d2f7f6805ff8404a1/src/shared_utilities/drivers.jl#L149
https://github.com/CliMA/ClimaLSM.jl/blob/2b56772e704e4b53f2b46a7d2f7f6805ff8404a1/src/shared_utilities/drivers.jl#L195.
and then taking the returned LHF and E and correcting them for the surface resistance. However, we realized E shows up in the SHF as well, and this has to be corrected also.

This PR does two things:
1. Accounts for surface resistance when computing SHF as well as LHF, and E, because E appears in the SHF expression.
2. Moves the r_sfc computation inside of the surface_fluxes wrapper function, which also obtains T_sfc, q_sfc, etc. Then this function only returns the resistance-corrected SHF, LHF, and E, and does not require the correction to happen outside of the surface fluxes wrapper function.

In order to achieve the latter, we defined a new function `surface_resistance(model, ...)` which returns the surface resistance for that land component `model`. The default is zero (no additional resistance).

## Content

Review checklist

I have:

### DIFF
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -158,12 +158,14 @@ function surface_fluxes(
     q_sfc = surface_specific_humidity(model, Y, p, T_sfc, ρ_sfc)
     β_sfc = surface_evaporative_scaling(model, Y, p)
     h_sfc = surface_height(model, Y, p)
+    r_sfc = surface_resistance(model, Y, p, t)
     return surface_fluxes_at_a_point.(
         T_sfc,
         q_sfc,
         ρ_sfc,
         β_sfc,
         h_sfc,
+        r_sfc,
         t,
         Ref(model.parameters),
         Ref(atmos),
@@ -176,6 +178,7 @@ end
                               ρ_sfc::FT,
                               β_sfc::FT,
                               h_sfc::FT,
+                              r_sfc::FT,
                               t::FT,
                               parameters,
                               atmos::PA,
@@ -184,7 +187,9 @@ end
 Computes turbulent surface fluxes at a point on a surface given
 (1) the surface temperature, specific humidity, and air density,
 (2) the time at which the fluxes are needed,
-(3) a factor β_sfc  which scales the evaporation from the potential rate,
+(3) a factor β_sfc  which scales the evaporation from the potential rate
+    (used in bucket models), and/or the surface resistance r_sfc (used
+    in more complex land models),
 (4) the parameter set for the model, which must have fields `earth_param_set`,
 and roughness lengths `z_0m, z_0b`.
 (5) the prescribed atmospheric state, stored in `atmos`.
@@ -198,6 +203,7 @@ function surface_fluxes_at_a_point(
     ρ_sfc::FT,
     β_sfc::FT,
     h_sfc::FT,
+    r_sfc::FT,
     t::FT,
     parameters::P,
     atmos::PA,
@@ -230,17 +236,21 @@ function surface_fluxes_at_a_point(
         sc;
         tol_neutral = SFP.cp_d(surface_flux_params) / 100000,
     )
-    # Land needs a volume flux of water, not mass flux
-    vapor_flux =
-        SurfaceFluxes.evaporation(surface_flux_params, sc, conditions.Ch) /
-        _ρ_liq
-    return (
-        lhf = conditions.lhf,
-        shf = conditions.shf,
-        Ch = conditions.Ch,
-        vapor_flux = vapor_flux,
-        r_ae = 1 / (conditions.Ch * SurfaceFluxes.windspeed(sc)),
-    )
+    grav::FT = LSMP.grav(earth_param_set)
+    cp_d::FT = Thermodynamics.Parameters.cp_d(thermo_params)
+    R_d::FT = Thermodynamics.Parameters.R_d(thermo_params)
+    T_0::FT = LSMP.T_0(earth_param_set)
+    cp_m = Thermodynamics.cp_m(thermo_params, ts_in)
+    T_in = Thermodynamics.air_temperature(thermo_params, ts_in)
+    ΔT = T_in - T_sfc
+    hd_sfc = cp_d * (T_sfc - T_0) + R_d * T_0
+    E0 = SurfaceFluxes.evaporation(surface_flux_params, sc, conditions.Ch)
+    r_ae = 1 / (conditions.Ch * SurfaceFluxes.windspeed(sc))
+    E = E0 * r_ae / (r_sfc + r_ae)
+    Ẽ = E / _ρ_liq
+    H = conditions.shf + hd_sfc * (E0 - E)
+    LH = conditions.lhf * r_ae / (r_sfc + r_ae)
+    return (lhf = LH, shf = H, vapor_flux = Ẽ, r_ae = r_ae)
 end
 
 """
@@ -339,6 +349,25 @@ compute surface fluxes and radiative fluxes at the surface using
 the functions in this file.
 """
 function surface_temperature(model::AbstractModel, Y, p, t) end
+
+"""
+    surface_resistance(model::AbstractModel, Y, p, t)
+
+A helper function which returns the surface resistance for a given
+model, needed because different models compute and store surface resistance in
+different ways and places.
+
+Extending this function for your model is only necessary if you need to
+compute surface fluxes and radiative fluxes at the surface using
+the functions in this file.
+
+The default is 0, which is no additional resistance aside from the usual
+aerodynamic resistance from MOST.
+"""
+function surface_resistance(model::AbstractModel{FT}, Y, p, t) where {FT}
+    return FT(0)
+end
+
 
 """
     surface_air_density(
--- a/docs/src/APIs/shared_utilities.md
+++ b/docs/src/APIs/shared_utilities.md
@@ -70,4 +70,7 @@ ClimaLSM.construct_atmos_ts
 ClimaLSM.surface_air_density
 ClimaLSM.liquid_precipitation
 ClimaLSM.snow_precipitation
+ClimaLSM.surface_temperature
+ClimaLSM.surface_resistance
+ClimaLSM.surface_specific_humidity
 ```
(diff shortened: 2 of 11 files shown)


src/shared_utilities/drivers.jl
```diff
@@ -230,17 +233,21 @@ function surface_fluxes_at_a_point(
...
-        SurfaceFluxes.evaporation(surface_flux_params, sc, conditions.Ch) /
-        _ρ_liq
-    return (
-        lhf = conditions.lhf,
-        shf = conditions.shf,
-        Ch = conditions.Ch,
-        vapor_flux = vapor_flux,
-        r_ae = 1 / (conditions.Ch * SurfaceFluxes.windspeed(sc)),
-    )
+    grav::FT = LSMP.grav(earth_param_set)
```
kmdeck, line 236: This code accounts for the surface resistance in SHF, LHF, and E. Previously, this correction was made outside of this function.

braghiere:
I think this equation is a bit weird (https://github.com/CliMA/SurfaceFluxes.jl/blob/40a11abaf838b5eef00c3b6be47f2eee8ac9951f/src/SurfaceFluxes.jl#L661C7-L661C7):

return -ρ_sfc * Ch * windspeed(sc) * (cp_m * ΔT + ΔΦ) - (hd_sfc) * E

If you look at the temperatures only, the flux would always be from the surface to the atmosphere, because T_canopy >= T_air (always!)

image

But when we look at Sensible heat, there is a weird inflection in the middle of the day:

image

My guess is that we are removing energy from evaporation from sensible heat, so when evaporation goes up in the middle of the day, sensible heat is decreasing. Also, the sign looks inverted. So I would re-write the sensible heat equation like:

return +ρ_sfc * Ch * windspeed(sc) * (cp_m * ΔT + ΔΦ)

kmdeck:
> I think this equation is a bit weird (https://github.com/CliMA/SurfaceFluxes.jl/blob/40a11abaf838b5eef00c3b6be47f2eee8ac9951f/src/SurfaceFluxes.jl#L661C7-L661C7):
> 
> return -ρ_sfc * Ch * windspeed(sc) * (cp_m * ΔT + ΔΦ) - (hd_sfc) * E


Hi Renato - thank you for looking into this more!! I agree that since T_canopy > T_air (T_sfc > T_in), we would expect the SHF to be positive. The sign in the equation is correct, though, because Delta T = T_in - T_sfc would be negative, and then the negative sign out front cancels to make it positive. I think the explanation must be what you indicated, that the `E` contribution is making the SHF negative, because E (Transpiration) >0, and we see a `-E` in the equation for SHF. if this is larger than the Delta T component, the SHF would be negative.

src/shared_utilities/drivers.jl
```diff
@@ -230,17 +236,21 @@ function surface_fluxes_at_a_point(
...
+    T_0::FT = LSMP.T_0(earth_param_set)
+    cp_m = Thermodynamics.cp_m(thermo_params, ts_in)
+    T_in = Thermodynamics.air_temperature(thermo_params, ts_in)
+    ΔT = T_in - T_sfc
+    hd_sfc = cp_d * (T_sfc - T_0) + R_d * T_0
+    E0 = SurfaceFluxes.evaporation(surface_flux_params, sc, conditions.Ch)
+    r_ae = 1 / (conditions.Ch * SurfaceFluxes.windspeed(sc))
+    E = E0 * r_ae / (r_sfc + r_ae)
+    Ẽ = E / _ρ_liq
+    H = conditions.shf + hd_sfc * (E0 - E)
```
akshaysridhar, line 251: Could you point me to the lines where these values in `conditions` are computed?

akshaysridhar: This correction then results in 
$ H = -\rho_{sfc} * C_{h} * |u_{h}| * (c_{pm} * \Delta T + \Delta \Phi) - (h_{d,sfc})  E  + h_{d, sfc} (E_{0} - E)$ - is this right ? 
```-ρ_sfc * Ch * windspeed(sc) * (cp_m * ΔT + ΔΦ) - (hd_sfc) * E + hd_sfc * (E0 - E) ```

Comment (braghiere) [reply src/shared_utilities/drivers.jl:251]: Thanks. Could you please clarify what the rationale is for removing evaporation from sensible heat? In equation 2.5.3 here: https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/tech_note/Fluxes/CLM50_Tech_Note_Fluxes.html, sensible heat is only a function of the temperature gradient between the surface and the atmosphere and the aerodynamic resistance. I don't understand how evaporation is related to sensible heat.

Comment (kmdeck) [reply src/shared_utilities/drivers.jl:251]: Hi Renato - I dont understand this myself, but in the atmosphere design docs, this is where it shows up:
Screenshot 2023-10-10 at 10 14 30

Comment (kmdeck) [reply src/shared_utilities/drivers.jl:251]: > `conditions`

https://github.com/CliMA/ClimaLSM.jl/blob/7fc4fb22f1cddb184ed8f08ef81c81286ce96d8c/src/shared_utilities/drivers.jl#L234

Comment (kmdeck) [reply src/shared_utilities/drivers.jl:251]: > Could you point me to the lines where these values in `conditions` are computed? `boundary_conditions.jl` > `172` it seems.

https://github.com/CliMA/ClimaLSM.jl/blob/7fc4fb22f1cddb184ed8f08ef81c81286ce96d8c/src/shared_utilities/drivers.jl#L234

Comment (kmdeck) [reply src/shared_utilities/drivers.jl:251]: > This correction then results in $$ H = -\rho_{sfc} * C_{h} * |u_{h}| * (c_{pm} * \Delta T + \Delta \Phi) - (h_{d,sfc}) E + h_{d, sfc} (E_{0} - E)$$ - is this right ? `-ρ_sfc * Ch * windspeed(sc) * (cp_m * ΔT + ΔΦ) - (hd_sfc) * E + hd_sfc * (E0 - E) `

sorry @akshaysridhar ! I originally "edited" your comment instead of quoting in my reply.



not quite. E0 should be the evaporation as computed by SurfaceFluxes, without the resistance correction. so this should be 
```
-ρ_sfc * Ch * windspeed(sc) * (cp_m * ΔT + ΔΦ) - (hd_sfc) * E0 + hd_sfc * (E0 - E) 
```
or
```
-ρ_sfc * Ch * windspeed(sc) * (cp_m * ΔT + ΔΦ) - (hd_sfc) * E
```
where `E` has been corrected for the surface resistance

akshaysridhar: Ah - yes I get this now, I misinterpreted the notation `E` when compared to the atmos / sf src.

kmdeck:
Note that in this PR, the SHF looks much better, because the positive contribution from the `E` term gets reduced by the resistance:
image

akshaysridhar:
Small note: Inconsequential to results - but `\tilde` rendering is inconsistent when viewing in `diff` mode vs `view file` mode within the relevant commit. e.g. 
Screenshot 2023-10-10 at 10 33 55 AM
vs 
Screenshot 2023-10-10 at 10 34 40 AM

src/shared_utilities/drivers.jl
akshaysridhar: This is resolved; `E0` is the outcome from SurfaceFluxes.jl therefore the above correction makes sense.

kmdeck:
> Note that in this PR, the SHF looks much better, because the positive contribution from the `E` term gets reduced by the resistance: image
> 
> Slightly better, but still very wrong. We're not just getting the peak wrong, but there's a very weird behavior at sunrise/sunset and the night values are completely inverted. Could we try to build an alternative function for H and G for testing purposes following Eq. 2.5.87 (https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/tech_note/Fluxes/CLM50_Tech_Note_Fluxes.html#equation-5-63) and Eq. 4 (chrome-extension://efaidnbmnnnibpcajpcglclefindmkaj/https://gmd.copernicus.org/articles/4/677/2011/gmd-4-677-2011.pdf).
> 
> It would be great to have the original reference for the evaporation correction in sensible heat. From the surface energy balance perspective, that doesn't make any sense.
> 
> PS: Also the title of this figure is wrong. It should read surface sensible heat instead of soil



@braghiere Agreed that it still looks very wrong. We can definitely try without the E contribution with a different expression for G, but on a different PR. 
With respect to the JULES equations 2-4. Physically this seems to represent a sensible heat flux between the surface with T_star and the atmosphere, and an evaporation (and corresponding latent heat flux) between the surface at T_star and the atmosphere. It seems like soil evaporation is not modeled explicitly, because there is no latent heat term in G. 

or, is r_surface = r_soil+r_canopy? If so, how do we partition this E between E and T? If there is soil evaporation, why doesnt a corresponding latent heat show up in G?

LenkaNovak (approved):
The correction LGTM, @kmdeck . From the conversation above, it seems that the plan is to recompute the fluxes using the traditional formulation (incl. SHF without the contribution of the E term) as part of a different PR. If it's helpful, perhaps this new PR could introduce some granular unit tests that assume bulk fluxes (constant transfer coefficients) with analytical solutions that show the differences between the two formulations.

akshaysridhar:
> The correction LGTM, @kmdeck . From the conversation above, it seems that the plan is to recompute the fluxes using the traditional formulation (incl. SHF without the contribution of the E term) as part of a different PR. If it's helpful, perhaps this new PR could introduce some granular unit tests that assume bulk fluxes (constant transfer coefficients) with analytical solutions that show the differences between the two formulations.

@LenkaNovak further, following this implementation, I think we should provide a generic container within SurfaceFluxes.jl that carries additional resistance parameters, such that the LSM would simply extract values for various surface properties from the `conditions` struct in the current code.

PR #351, opened 2023-10-11
displacement height

kmdeck:
## Purpose 
This adds the displacement height to how we call SurfaceFluxes.jl

## To-do
review

## Content
Adds a displacement_height function, which returns zero by default. This is appropriate for the soil and snow. The displacement height of the canopy is 0.67*h_canopy. This displacement height is then used when we compute surface fluxes.

Review checklist

I have:

----


src/shared_utilities/drivers.jl
```diff
@@ -218,8 +224,10 @@ function surface_fluxes_at_a_point(
     ts_in = construct_atmos_ts(atmos, t, thermo_params)
     ts_sfc = Thermodynamics.PhaseEquil_ρTq(thermo_params, ρ_sfc, T_sfc, q_sfc)
 
-    state_sfc = SurfaceFluxes.SurfaceValues(h_sfc, SVector{2, FT}(0, 0), ts_sfc)
-    state_in = SurfaceFluxes.InteriorValues(h, SVector{2, FT}(u, 0), ts_in)
+    state_sfc =
+        SurfaceFluxes.SurfaceValues(h_sfc - d_sfc, SVector{2, FT}(0, 0), ts_sfc)
```
kmdeck, line 228: @akshaysridhar no rush, but is this how displacement height is passed into SurfaceFluxes? I recall you said it expects "z-d" instead of z.

akshaysridhar: Right - the current interface does not permit a separate parameter `d == displacement height`. So, if a user knows displacement height `d`, then you can recover the appropriate adjusted values by supplying `z-d` e.g. Bonan textbook eq(6.40).

akshaysridhar: In the above instance should that displacement adjustment be associated with `state_in` ?

akshaysridhar: e.g. - recovering profiles shown in Bonan's reference with displacement height d == 19m. https://clima.github.io/SurfaceFluxes.jl/dev/SurfaceFluxes/#Profile-recovery-using-MOST

kmdeck: Hmmm perhaps Bonan's equation 6.41 was a little more helpful for me. When `z_1 = z_0+d`, `theta_1 = theta_sfc`. Then `z_1-d = z_0`. And `z2-d` is `z_atmos-d` still.

So, we should pass in `z_atmos -d` for the height of `state_in`, like you wrote. For the height of `state_in`, do we always pass zero (not `z_0`)? It seems slight more consistent to pass in `z1-d = z0` rather than zero, since we pass in `z2-d = z_atmos-d`.

Looking here: https://github.com/CliMA/SurfaceFluxes.jl/blob/3b51936ee4135a899d0a0294125becab58fbf799/docs/src/plot_profiles.jl#L107-L109

Once we add topography, it will be good to revisit this!

kmdeck: @akshaysridhar I fixed this. I didnt pass `z_0` for the height of `state_in`, though by the above discussion I thought it should be there. let me know what you think!

akshaysridhar: @kmdeck I agree with the comment above (w.r.t Bonan 6.4), the `Atmos` model sees a zero surface height by the grid construction (simply the sfc elevation); within the surface flux calculation we then assume that `u(z) = 0 for 0 ≤ z≤ z0_{m}` and `DSE == constant for 0 ≤ z≤ z0_{h}` for the momentum, tracers respectively. 

https://github.com/CliMA/SurfaceFluxes.jl/blob/ad299300ce44a8b129c71811e19286592aebef9b/src/SurfaceFluxes.jl#L229-L231 This should also clarify the input args to the sf calculation.

kmdeck: I added a clarifying comment to the code. thanks for the review!

akshaysridhar: Thanks @kmdeck. Resolving this thread, LGTM!

akshaysridhar (approved):

PR #352, opened 2023-10-11
upd to most recent dependencies

juliasloan25:
## Purpose 
update to most recent versions of dependencies, and update compat entries

replaces #339
replaces #346

## Problem updating ClimaTimeSteppers
- updating causes error in `solve`, where we map `zero` over a `FieldVector`, so `zero(::Field)` gets called, which isn't defined
- **problem**: this arises when we try to update from ClimaTimeSteppers.jl v0.7.7 to v0.7.8, due to a change in `init_cache` from [this](https://github.com/CliMA/ClimaTimeSteppers.jl/blob/v0.7.7/src/solvers/imex_ark.jl#L33) to [this](https://github.com/CliMA/ClimaTimeSteppers.jl/blob/v0.7.8/src/solvers/imex_ark.jl#L35). This causes `CanopyModel` and `LSM` simulations to fail, since they have a `FieldVector` containing a `Vector{Field}` type, which isn't supported. See #356 for more details.
- **solution**: for now, I'll pin the version of ClimaTimeSteppers to 0.7.7, and we'll fix the root of this problem soon

----


kmdeck (approved):
lgtm! thank you!

Issue #229, opened 2023-06-13
Make sure tests comprehensively cover Float types


kmdeck:
**Is your feature request related to a problem? Please describe.**
we parameterize our types using FT across ClimaLSM. This is to ensure type stability in simulations. Some of our tests do loop over FT, some to check for type promotion, but not all. Revisit tests to make sure they do test different FTs comprehensively.

PR #353, opened 2023-10-12
restore Float32 compatibility

### LINKED ISSUE #327: Cannot run simulations with Float32
## Describe the bug
In ClimaTimesteppers, the float type of the time `t` is Float64. In several places, we require the type of the time variable and the underlying type of the state vectors to be the same. We need to relax this. We also need to change functions where we prescribed driving data as a function of time, so that whatever is returned is converted to the right type.

Close https://github.com/CliMA/ClimaLSM.jl/issues/229 at the same time

See https://github.com/CliMA/ClimaCoupler.jl/issues/396 for discussion with more details

juliasloan25:
## Purpose 
fixes #327, #229

### Notes
- code extending tests to cover Float32 and Float64 is on branch js/float32-tests - will need to cherry pick these changes onto this branch
  - tests won't pass until other changes are made (`t::FT` and `eltype(t)`) - make those changes first
- idea: it may be helpful to add a flag that checks that our t_start/dt/t_end/FT setup won't exceed the FT's precision - if it does, give warning to user

## To-do
- [x] remove `t::FT` type annotations
- [x] remove `eltype(t)(...)` usage
  - ex https://github.com/CliMA/ClimaLSM.jl/blob/main/docs/tutorials/Bucket/bucket_tutorial.jl#L229-L230
  - in tests, this can be replaced by `FT(...)`
  - [x] functions of time that return `eltype(t)(result)` --> instead convert to FT where function called from
  - [x] boundary conditions, PrescribedAtmosphere, PrescribedRadiativeFluxes, canopy prescribed soil moisture, canopy transpiration
- [x] make sure tests cover Float64 and Float32
- [ ] make sure simulations try Float64 and Float32 - this will be a separate PR #384
  - idea: loop over `(FT, t_end)`, run Float32 only for a couple steps. At end of sim, if statement for FT which tests final type for both and makes plots for Float64
- [x] add documentation for how to maintain Float32 and Float64 compatibility (in [wiki](https://github.com/CliMA/ClimaLSM.jl/wiki/Floating-point-precision-issues))

----

### DIFF
--- a/docs/tutorials/Bucket/bucket_tutorial.jl
+++ b/docs/tutorials/Bucket/bucket_tutorial.jl
@@ -226,8 +226,8 @@ ref_time = DateTime(2005);
 # `ρ_a` (kg/m^3) at a reference height `h_a` (m).
 
 # Here we define the model drivers, starting with downward radiation.
-SW_d = (t) -> eltype(t)(300);
-LW_d = (t) -> eltype(t)(300);
+SW_d = (t) -> 300;
+LW_d = (t) -> 300;
 bucket_rad = PrescribedRadiativeFluxes(
     FT,
     SW_d,
@@ -239,15 +239,15 @@ bucket_rad = PrescribedRadiativeFluxes(
 # Prescribed atmospheric variables
 
 # Stochastic precipitation:
-precip = (t) -> eltype(t)(0);
-snow_precip = (t) -> eltype(t)(5e-7 * (t > 3 * 86400) * (t < 4 * 86400));
+precip = (t) -> 0;
+snow_precip = (t) -> 5e-7 * (t > 3 * 86400) * (t < 4 * 86400);
 # Diurnal temperature variations:
-T_atmos = (t) -> eltype(t)(275.0 + 5.0 * sin(2.0 * π * t / 86400 + 7200));
+T_atmos = (t) -> 275.0 + 5.0 * sin(2.0 * π * t / 86400 + 7200);
 # Constant otherwise:
-u_atmos = (t) -> eltype(t)(3.0);
-q_atmos = (t) -> eltype(t)(0.005);
+u_atmos = (t) -> 3.0;
+q_atmos = (t) -> 0.005;
 h_atmos = FT(2);
-P_atmos = (t) -> eltype(t)(101325);
+P_atmos = (t) -> 101325;
 bucket_atmos = PrescribedAtmosphere(
     precip,
     snow_precip,
--- a/docs/tutorials/Canopy/canopy_tutorial.jl
+++ b/docs/tutorials/Canopy/canopy_tutorial.jl
@@ -116,7 +116,7 @@ shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
 
 soil_driver = PrescribedSoil{FT}(
     root_depths = FT.(-Array(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0),
-    ψ = t -> eltype(t)(ψ_soil0),
+    ψ = t -> ψ_soil0,
     α_PAR = FT(0.2),
     α_NIR = FT(0.4),
     T = t -> 298.0,
@@ -185,7 +185,7 @@ AR_model = AutotrophicRespirationModel{FT}(AR_params);
 # indices of the canopy, we choose a `PrescribedSiteAreaIndex`,
 # which supports LAI as a function of time, with RAI and SAI as constant.
 LAI = 4.2
-LAIfunction = (t) -> eltype(t)(LAI)
+LAIfunction = (t) -> LAI
 SAI = FT(0.00242)
 f_root_to_shoot = FT(3.5)
 RAI = (SAI + LAI) * f_root_to_shoot
--- a/docs/tutorials/Usage/LSM_single_column_tutorial.jl
+++ b/docs/tutorials/Usage/LSM_single_column_tutorial.jl
@@ -81,7 +81,7 @@ vg_n = FT(2.0);
 vg_α = FT(2.6); # inverse meters
 hcm = vanGenuchten(; α = vg_α, n = vg_n);
 θ_r = FT(0);
-soil_ps = Soil.RichardsParameters(;
+soil_ps = Soil.RichardsParameters{FT, typeof(hcm)}(;
     ν = ν,
     hydrology_cm = hcm,
     K_sat = K_sat,
@@ -96,8 +96,8 @@ nelems = 20;
 soil_domain = Column(; zlim = (zmin, zmax), nelements = nelems);
 
 # And boundary conditions and source terms (none currently):
-top_flux_bc = FluxBC((p, t) -> eltype(t)(0.0))
-bot_flux_bc = FluxBC((p, t) -> eltype(t)(0.0))
+top_flux_bc = FluxBC((p, t) -> 0.0);
+bot_flux_bc = FluxBC((p, t) -> 0.0);
 sources = ()
 boundary_fluxes =
     (; top = (water = top_flux_bc,), bottom = (water = bot_flux_bc,))
@@ -173,9 +173,9 @@ soil_ode! = make_exp_tendency(soil);
 # them inside our pond model, since again, the pond model structure
 # must contain everything needed to make the tendency function:
 
-precipitation(t::T) where {T} = t < T(20) ? -T(1e-5) : T(0.0) # m/s
+precipitation(t) = t < 20 ? -1e-5 : 0.0 # m/s
 
-infiltration(t::T) where {T} = -T(1e-6) #m/s
+infiltration(t) = -(1e-6) #m/s
 pond_model = Pond.PondModel{FT}(;
     runoff = PrescribedRunoff{FT}(precipitation, infiltration),
 );
--- a/experiments/integrated/ozark/conservation/ozark_conservation.jl
+++ b/experiments/integrated/ozark/conservation/ozark_conservation.jl
@@ -96,8 +96,8 @@ soilco2_boundary_conditions =
     (; top = (CO2 = soilco2_top_bc,), bottom = (CO2 = soilco2_bot_bc,))
 
 soilco2_drivers = Soil.Biogeochemistry.SoilDrivers(
-    Soil.Biogeochemistry.PrognosticMet(),
-    Soil.Biogeochemistry.PrescribedSOC(Csom),
+    Soil.Biogeochemistry.PrognosticMet{FT}(),
+    Soil.Biogeochemistry.PrescribedSOC{FT}(Csom),
     atmos,
 )
 
--- a/experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
+++ b/experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
@@ -106,8 +106,12 @@ atmos_u = Spline1D(seconds, WS[:])
 LW_IN_spline = Spline1D(seconds, LW_IN[:])
 SW_IN_spline = Spline1D(seconds, SW_IN[:])
 atmos_h = FT(32)
-precipitation_function(t::FT) where {FT} = p_spline(t) < 0.0 ? p_spline(t) : 0.0 # m/s
-snow_precip(t) = eltype(t)(0) # this is likely not correct
+
+function precipitation_function(t)
+    spline = p_spline(t)
+    spline < 0 ? spline : 0 # m/s
+end
+snow_precip(t) = 0 # this is likely not correct
 
 
 # Construct the drivers. For the reference time we will use the UTC time at the
@@ -174,4 +178,4 @@ LAI_column_names = LAI_data[1, :]
 LAI_data = LAI_data[2:end, LAI_column_names .== "lai_mean"] #m2.m-2
 # This has the same timestamp as the driver data, so it's ok to use the time column from that file here
 LAIspline = Spline1D(seconds, LAI_data[:])
-LAIfunction = (t) -> eltype(t)(LAIspline(t))
+LAIfunction = (t) -> LAIspline(t)
--- a/src/integrated/pond_soil_model.jl
+++ b/src/integrated/pond_soil_model.jl
@@ -7,7 +7,7 @@ export LandHydrology, infiltration_capacity, infiltration_at_point
         SW <: Pond.AbstractSurfaceWaterModel{FT},
     } <: AbstractLandModel{FT}
 
-A concrete type of land model used for simulating systems with a 
+A concrete type of land model used for simulating systems with a
 soil and surface water component.
 $(DocStringExtensions.FIELDS)"""
 struct LandHydrology{
@@ -97,15 +97,15 @@ I|______________________________________________________________________________
 =#
 
 """
-    infiltration_at_point(η::FT, i_c::FT, P::FT)   
+    infiltration_at_point(η::FT, i_c::FT, P::FT)
 
 Returns the infiltration given pond height η, infiltration capacity,
-and precipitation. 
+and precipitation.
 
 
 This is defined such that positive means into soil.
 """
-infiltration_at_point(η::FT, i_c::FT, P::FT) where {FT} =
+infiltration_at_point(η::FT, i_c::FT, P::FT) where {FT <: AbstractFloat} =
     η > eps(FT) ? i_c : min(i_c, P)
 
 """
@@ -120,6 +120,7 @@ soil characteristics, moisture levels, and pond height.
 Defined such that positive means into soil.
 """
 function infiltration_capacity(Y::ClimaCore.Fields.FieldVector, p::NamedTuple)
+    FT = eltype(Y.soil.ϑ_l)
     face_space = ClimaLSM.Domains.obtain_face_space(axes(Y.soil.ϑ_l))
     N = ClimaCore.Spaces.nlevels(face_space)
     space = axes(Y.surface_water.η)
@@ -146,7 +147,7 @@ function infiltration_capacity(Y::ClimaCore.Fields.FieldVector, p::NamedTuple)
         ),
         space,
     )
-    ψ_face = max.(0.0, Y.surface_water.η)
+    ψ_face = max.(FT(0), Y.surface_water.η)
     return @. (
         K_center * (ψ_face + z_face - (ψ_center + z_center)) /
         (z_face - z_center)
@@ -158,7 +159,7 @@ end
         land::LandHydrology{FT, SM, SW},
     ) where {FT, SM <: Soil.RichardsModel{FT}, SW <: Pond.PondModel{FT}}
 
-A method which makes a function; the returned function 
+A method which makes a function; the returned function
 updates the auxiliary variable `p.soil_infiltration`, which
 is needed for both the boundary condition for the soil model and the source
 term (runoff) for the surface water model.
@@ -174,7 +175,7 @@ function make_interactions_update_aux(
             -infiltration_at_point(
                 Y.surface_water.η,
                 i_c,
-                -land.surface_water.runoff.precip(t),
+                -FT(land.surface_water.runoff.precip(t)),
             )
     end
     return update_aux!
@@ -201,7 +202,7 @@ end
         runoff::PrognosticRunoff{FT},
         Y::ClimaCore.Fields.FieldVector,
         p::NamedTuple,
-        t::FT,
+        t,
     ) where {FT}
 
 Extension of the `Pond.surface_runoff` function, which computes the surface runoff, for use in an LSM when the runoff is determined
@@ -211,9 +212,9 @@ function Pond.surface_runoff(
     runoff::PrognosticRunoff{FT},
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
-    t::FT,
+    t,
 ) where {FT}
-    return @. -(runoff.precip(t) - p.soil_infiltration)
+    return @. FT(p.soil_infiltration - runoff.precip(t))
 end
 
 """
@@ -225,7 +226,7 @@ prognostically. This infiltration is then used to set an upper
 boundary condition for the soil.
 
 
-This is paired with `Pond.PrognosticRunoff`: both are used at the same 
+This is paired with `Pond.PrognosticRunoff`: both are used at the same
 time,
 ensuring that the infiltration used for the boundary condition of soil
 is also used to compute the runoff for the surface water.
@@ -238,9 +239,9 @@ struct RunoffBC <: Soil.AbstractSoilBC end
         ::TopBoundary,
         Δz::FT,
         p::NamedTuple,
-        t::FT,
+        t,
         params,
-    )::ClimaCore.Fields.Field where {FT}
+    )::ClimaCore.Fields.Field
 
 Extension of the `ClimaLSM.boundary_flux` function, which returns the water volume
 boundary flux for the soil.
@@ -252,8 +253,8 @@ function ClimaLSM.boundary_flux(
     ::TopBoundary,
     Δz::ClimaCore.Fields.Field,
     p::NamedTuple,
-    t::FT,
+    t,
     params,
-)::ClimaCore.Fields.Field where {FT}
+)::ClimaCore.Fields.Field
     return p.soil_infiltration
 end
--- a/src/integrated/soil_canopy_model.jl
+++ b/src/integrated/soil_canopy_model.jl
@@ -14,7 +14,7 @@ export SoilCanopyModel
         canopy::VM
     end
 
-A concrete type of land model used for simulating systems with a 
+A concrete type of land model used for simulating systems with a
 canopy and a soil component.
 $(DocStringExtensions.FIELDS)
 """
@@ -36,7 +36,7 @@ end
     CanopyRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT}
 
 A struct used to compute radiative fluxes in land surface models,
-indicating that 
+indicating that
 canopy absorption and emission is taken into account when computing
 radiation at the surface of the soil or snow.
 
@@ -44,7 +44,7 @@ The only other alternative at this stage is
 ClimaLSM.PrescribedRadiativeFluxes, where the prescribed downwelling
 short and longwave radiative fluxes are used directly,
 without accounting for the canopy. There is a different method
-of the function `soil_boundary_fluxes` in this case. 
+of the function `soil_boundary_fluxes` in this case.
 """
 struct CanopyRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT} end
 
@@ -98,7 +98,7 @@ function SoilCanopyModel{FT}(;
     sources = (RootExtraction{FT}(), Soil.PhaseChange{FT}(Δz))
     # add heat BC
     top_bc = ClimaLSM.Soil.AtmosDrivenFluxBC(atmos, CanopyRadiativeFluxes{FT}())
-    zero_flux = FluxBC((p, t) -> eltype(t)(0.0))
+    zero_flux = FluxBC((p, t) -> 0.0)
     boundary_conditions = (;
         top = top_bc,
         bottom = (water = Soil.FreeDrainage(), heat = zero_flux),
@@ -168,7 +168,7 @@ function SoilCanopyModel{FT}(;
 
     soilco2 = Soil.Biogeochemistry.SoilCO2Model{FT}(; soilco2_args...)
 
-    if typeof(soilco2_args.drivers.met) != PrognosticMet
+    if !(soilco2_args.drivers.met isa PrognosticMet)
         throw(AssertionError("Must be of type PrognosticMet."))
     end
 
@@ -182,7 +182,7 @@ end
 """
     interaction_vars(m::SoilCanopyModel)
 
-The names of the additional auxiliary variables that are 
+The names of the additional auxiliary variables that are
 included in the integrated Soil-Canopy model.
 """
 interaction_vars(m::SoilCanopyModel) = (
@@ -203,7 +203,7 @@ interaction_vars(m::SoilCanopyModel) = (
 """
     interaction_types(m::SoilCanopyModel)
 
-The types of the additional auxiliary variables that are 
+The types of the additional auxiliary variables that are
 included in the integrated Soil-Canopy model.
 """
 interaction_types(m::SoilCanopyModel{FT}) where {FT} =
@@ -212,7 +212,7 @@ interaction_types(m::SoilCanopyModel{FT}) where {FT} =
 """
     interaction_domain_names(m::SoilCanopyModel)
 
-The domain names of the additional auxiliary variables that are 
+The domain names of the additional auxiliary variables that are
 included in the integrated Soil-Canopy model.
 """
 interaction_domain_names(m::SoilCanopyModel) = (
@@ -240,7 +240,7 @@ interaction_domain_names(m::SoilCanopyModel) = (
         RM <: Canopy.CanopyModel{FT}
         }
 
-A method which makes a function; the returned function 
+A method which makes a function; the returned function
 updates the auxiliary variable `p.root_extraction`, which
 is needed for a sink term for the soil model and to create the
 lower water boundary condition for the canopy model.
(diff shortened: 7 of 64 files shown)


juliasloan25:
bors cancel

src/integrated/pond_soil_model.jl
```diff
@@ -238,9 +243,9 @@ struct RunoffBC <: Soil.AbstractSoilBC end
         ::TopBoundary,
         Δz::FT,
         p::NamedTuple,
-        t::FT,
+        t,
         params,
-    )::ClimaCore.Fields.Field where {FT}
```
juliasloan25, line 243: this should still have `where {FT}`

docs/tutorials/Usage/LSM_single_column_tutorial.jl
```diff
@@ -173,9 +173,9 @@ soil_ode! = make_exp_tendency(soil);
 # them inside our pond model, since again, the pond model structure
 # must contain everything needed to make the tendency function:
 
-precipitation(t::T) where {T} = t < T(20) ? -T(1e-5) : T(0.0) # m/s
```
juliasloan25, line 176: remove `FT()` from these functions, apply where functions are called instead

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -259,14 +263,15 @@ function ClimaLSM.boundary_flux(
     p::NamedTuple,
     t,
 )::ClimaCore.Fields.Field
+    FT = eltype(Y)
```
juliasloan25, line 266: use eltype(dz) instead

juliasloan25: same on line 303

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
```diff
@@ -121,24 +126,23 @@ lat = FT(38.7441) # degree
...
-    FT(
-        instantaneous_zenith_angle(
-            dt,
-            orbital_data,
-            longitude,
-            latitude,
-            insol_params,
-        )[1],
-    )
+    # TODO do we need to wrap this in FT? can we get FT from a param other than t?
```
juliasloan25, line 138: remove comment, get FT from orbital_data?

src/shared_utilities/drivers.jl
```diff
@@ -81,9 +84,12 @@ struct PrescribedAtmosphere{FT, LP, SP, TA, UA, QA, RA, CA, DT} <:
         args = (liquid_precip, snow_precip, T, u, q, P, c_co2, ref_time)
         return new{typeof(h), typeof.(args)...}(args..., h, gustiness)
     end
-
 end
 
+float_type(
```
juliasloan25, line 89: I added these `float_type` functions but didn't actually end up using them, so I think I'll remove them

docs/tutorials/Bucket/bucket_tutorial.jl
```diff
@@ -390,7 +390,7 @@ plot(
     ylabel = "T_sfc (K)",
     legend = :bottomright,
 )
-plot!(sol.t ./ 86400, T_atmos.(sol.t), label = "Atmospheric Temperature")
+plot!(sol.t ./ 86400, FT.(T_atmos.(sol.t)), label = "Atmospheric Temperature")
```
kmdeck, line 393: do we need to convert to FT to plot?

docs/tutorials/Canopy/canopy_tutorial.jl
```diff
@@ -114,11 +114,11 @@ shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
 
 ψ_soil0 = FT(0.0)
 
-soil_driver = PrescribedSoil(
-    root_depths = FT.(-Array(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0),
-    ψ_soil = t -> eltype(t)(ψ_soil0),
-    soil_α_PAR = FT(0.2),
-    soil_α_NIR = FT(0.4),
+soil_driver = PrescribedSoil{FT}(
+    FT.(-Array(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0), # root_depths
```
kmdeck, line 118: out of curiosity, do we need these FTs in the arguments?

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
```diff
@@ -99,8 +99,13 @@ atmos_u = Spline1D(seconds, WS[:])
 LW_IN_spline = Spline1D(seconds, LW_IN[:])
 SW_IN_spline = Spline1D(seconds, SW_IN[:])
 atmos_h = FT(32)
-precipitation_function(t::FT) where {FT} = p_spline(t) < 0.0 ? p_spline(t) : 0.0 # m/s
-snow_precip(t) = eltype(t)(0) # this is likely not correct
+
+function precipitation_function(t)
```
kmdeck, line 103: Im not sure we need this. the eltype of the spline is Float64, so I dont think this gains us anything over using eltype of t?

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
kmdeck, line 138: can we pass FT as a keyword argument after it is defined (closure)?

it may be stored in the orbital data also

experiments/integrated/ozark/soil_retention_curve.jl
```diff
@@ -16,12 +16,13 @@ swrc_data = readdlm(data, ','; skipstart = 1)
...
 ψ = swrc_data[4:end, 2] .* 1e6 ./ (1e3 * 9.8) # to convert from MPa to m
 
+FT = Float64
 # Assume value for this:
 θ_r = 0.04
 α = 10.0 .^ (log10(5e-3):0.1:1)
 n = Array(1.1:0.05:3.0)
 porosity = [0.45, 0.5, 0.55]
-L2 = zeros(Float64, (length(α), length(n)))
+L2 = zeros(FT, (length(α), length(n)))
```
kmdeck, line 25: I dont think we need these changes really, this is a one off file for fitting parameters but isnt ever used in the simulation.

src/standalone/Vegetation/soil_drivers.jl
```diff
@@ -31,12 +31,12 @@ end
...
      ) where {FT}
 
-An outer constructor for the PrescribedSoil soil driver allowing the user to 
+An outer constructor for the PrescribedSoil soil driver allowing the user to
 specify the soil parameters by keyword arguments.
 """
 function PrescribedSoil(;
     root_depths::Vector{FT} = FT.(-Array(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0),
-    ψ_soil::Function = t -> eltype(t)(FT(0.0)),
+    ψ_soil::Function = t -> 0.0,
```
kmdeck, line 39: might as well return FT(0)?

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -521,12 +521,13 @@ To prevent dividing by zero, we change AI/(AI x dz)" to
 """
 function make_compute_exp_tendency(model::PlantHydraulicsModel, canopy)
     function compute_exp_tendency!(dY, Y, p, t)
+        FT = eltype(p.canopy.hydraulics.area_index.root)
```
kmdeck, line 524: in cases like this, should we get FT from the model (or another struct) instead? it seems a little cleaner to me.

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -591,8 +592,8 @@ function root_flux_per_ground_area!(
     model::PlantHydraulicsModel{FT},
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
-    t::FT,
-) where {FT}
+    t,
+) where {FT <: AbstractFloat}
```
kmdeck, line 596: I dont know that we need to enforce this here, because FT <: AbstractFloat is part of the definition of AbstractCanopyComponent{FT}: https://github.com/CliMA/ClimaLSM.jl/blob/ad31773009c9309d5e45b083603873a8914be286/src/standalone/Vegetation/component_models.jl#L21

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -669,9 +670,9 @@ function transpiration_per_ground_area(
     transpiration::PrescribedTranspiration{FT},
     _,
     _,
-    t::FT,
-)::FT where {FT}
-    return transpiration.T(t) # (m/s)
+    t,
+)::FT where {FT <: AbstractFloat}
+    return FT(transpiration.T(t)) # (m/s)
```
kmdeck, line 675: I think we are also OK here without saying FT<:AbstractFloat (and just "FT"), because of https://github.com/CliMA/ClimaLSM.jl/blob/ad31773009c9309d5e45b083603873a8914be286/src/standalone/Vegetation/PlantHydraulics.jl#L54

src/standalone/Vegetation/radiation.jl
```diff
@@ -174,11 +174,11 @@ for a radiative transfer model.
...
 """
 function compute_NIR(
-    model::AbstractRadiationModel,
+    model::AbstractRadiationModel{FT},
     solar_radiation::ClimaLSM.PrescribedRadiativeFluxes,
     t,
-)
-    return solar_radiation.SW_d(t) / 2
+) where {FT <: AbstractFloat}
+    return FT(solar_radiation.SW_d(t) / 2)
```
kmdeck, line 181: likewise here

src/standalone/Vegetation/Canopy.jl
```diff
@@ -636,8 +636,8 @@ function canopy_surface_fluxes(
     model::CanopyModel,
     Y,
     p,
-    t::FT,
-) where {FT}
+    t,
+) where {FT <: AbstractFloat}
```
kmdeck, line 640: also here: https://github.com/CliMA/ClimaLSM.jl/blob/ad31773009c9309d5e45b083603873a8914be286/src/shared_utilities/drivers.jl#L27, no need for <: AbstractFloat

src/standalone/Soil/rre.jl
```diff
@@ -168,9 +168,9 @@ Construct the tendency computation function for the explicit terms of the RHS,
 which are horizontal components and source/sink terms.
 """
 function ClimaLSM.make_compute_exp_tendency(model::Soil.RichardsModel)
-    function compute_exp_tendency!(dY, Y, p, t::FT) where {FT}
+    function compute_exp_tendency!(dY, Y, p, t)
```
kmdeck, line 171: do we want to get FT from the RichardModel struct?

kmdeck: this is fine the way you have it also though, only needed in one spot.

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -563,14 +563,13 @@ because the surface air density is not a prognostic
...
-    atmos::PrescribedAtmosphere{FT},
+    atmos::PrescribedAtmosphere{FT, CA, DT},
     model::EnergyHydrology{FT},
     Y,
     p,
     t,
     T_sfc,
-) where {FT}
-
+) where {FT <: AbstractFloat, CA, DT}
```
kmdeck, line 572: do we also need to pass in CA, DT?

dont need <: AbstractFloat

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -259,14 +263,15 @@ function ClimaLSM.boundary_flux(
     p::NamedTuple,
     t,
 )::ClimaCore.Fields.Field
+    FT = eltype(Δz)
```
kmdeck, line 266: This is totally fine, but Im wondering if it is easier to get FT from the parametric type of the BC or if that cleaner or if there is no preference.

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -203,7 +206,8 @@ function ClimaLSM.boundary_flux(
     p::NamedTuple,
     t,
 )::ClimaCore.Fields.Field
-    return bc.bc(p, t) .+ ClimaCore.Fields.zeros(axes(Δz))
+    FT = eltype(Δz)
+    return FT.(bc.bc(p, t)) .+ ClimaCore.Fields.zeros(axes(Δz))
```
kmdeck, line 210: do we need this here? or will it get converted appropriately since it is added to a Field with (presumably) a fixed type?

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -155,24 +155,27 @@ presence of a canopy, snow, etc, as in a land surface model,
...
+        <:PrescribedRadiativeFluxes{FT, DT2, T, OD},
+    },
     boundary::ClimaLSM.TopBoundary,
     model::EnergyHydrology{FT},
     Δz,
     Y,
     p,
     t,
-) where {FT}
+) where {FT <: AbstractFloat, CA, DT1, DT2, T, OD}
```
kmdeck, line 173: Let's remove the extra types since they are not used. same comment re <: AbstractFloat

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -431,9 +449,9 @@ function ClimaLSM.boundary_flux(
     Δz::ClimaCore.Fields.Field,
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
-    t::FT,
-)::ClimaCore.Fields.Field where {FT}
-    return bc.bc(p, t) .+ ClimaCore.Fields.zeros(axes(Δz))
+    t,
+)::ClimaCore.Fields.Field
+    return FT.(bc.bc(p, t)) .+ ClimaCore.Fields.zeros(axes(Δz))
```
kmdeck, line 454: same ?: do we need to convert to FT here, since it is added to a Field of (fixed) type? It might be converted automatically.

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -467,12 +485,13 @@ function ClimaLSM.boundary_flux(
     Δz::ClimaCore.Fields.Field,
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
-    t::FT,
-)::ClimaCore.Fields.Field where {FT}
+    t,
+)::ClimaCore.Fields.Field
+    FT = eltype(Δz)
```
kmdeck, line 490: I think my pref is to get FT from the BC type. but we can discuss! otherwise Im not sure the BC's need to have a parametric type at all.

we could also make clear in the BC docs (future PR) that FT is used to convert functions of time into the right type, or something.

src/standalone/Snow/Snow.jl
```diff
@@ -30,13 +30,13 @@ struct SnowParameters{FT <: AbstractFloat, PSE}
     "Thermal conductivity of ice (W/m/K)"
     κ_ice::FT
     "Timestep of the model"
-    Δt::FT
+    Δt::Any
```
kmdeck, line 33: do we need to even pass in a type if it is Any? What does it gain us?

src/standalone/Snow/Snow.jl
```diff
@@ -63,7 +63,7 @@ function SnowParameters{FT}(
     fS_c = FT(0.2),
     κ_ice = FT(2.21),
     earth_param_set::PSE,
-) where {FT, PSE}
+) where {FT <: AbstractFloat, PSE}
```
kmdeck, line 66: not sure we need the FT <: AbstractFloat here? https://github.com/CliMA/ClimaLSM.jl/blob/ad31773009c9309d5e45b083603873a8914be286/src/standalone/Snow/Snow.jl#L13

src/standalone/Bucket/bucket_parameterizations.jl
```diff
@@ -55,21 +55,21 @@ a helper function which computes and returns the surface air density for the buc
...
 function ClimaLSM.surface_air_density(
-    atmos::PrescribedAtmosphere,
+    atmos::PrescribedAtmosphere{FT, CA, DT},
     model::BucketModel,
     Y,
     p,
     t,
     T_sfc,
-)
+) where {FT <: AbstractFloat, CA, DT}
```
kmdeck, line 64: what does {CA, DT} gain us?

src/standalone/Bucket/Bucket.jl
```diff
@@ -156,11 +156,11 @@ should be used.
 function BulkAlbedoTemporal{FT}(
     regrid_dirpath::String,
     date_ref::Union{DateTime, DateTimeNoLeap},
-    t_start::FT,
+    t_start,
     space::ClimaCore.Spaces.AbstractSpace;
     infile_path = Bucket.cesm2_albedo_dataset_path(),
     varname = "sw_alb",
-) where {FT}
+) where {FT <: AbstractFloat}
```
kmdeck, line 163: do we need this? or is FT already defined to be a subtype of AbstractFloat somewhere else (defn of BulkAlbedoTemporal?)

src/shared_utilities/utils.jl
```diff
@@ -85,16 +85,12 @@ end
...
 
 Computes the weighted direct stiffness summation and updates `Y` in place.
 In the case of a column domain, no dss operations are performed.
 """
-function dss!(
-    Y::ClimaCore.Fields.FieldVector{FT},
-    p::NamedTuple,
-    t::FT,
-) where {FT}
+function dss!(Y::ClimaCore.Fields.FieldVector{FT}, p::NamedTuple, t) where {FT}
```
kmdeck, line 93: the root of it all 😅

src/shared_utilities/drivers.jl
```diff
@@ -324,17 +333,25 @@ end
...
 
 """
     snow_precipitation(atmos::PrescribedAtmosphere, p, t)
 
 Returns the precipitation in snow (m of liquid water/s) at the surface.
 """
-function snow_precipitation(atmos::PrescribedAtmosphere, p, t)
-    return atmos.snow_precip(t)
+function snow_precipitation(
+    atmos::PrescribedAtmosphere{FT, CA, DT},
```
kmdeck, line 350: same comment here as down below

src/shared_utilities/drivers.jl
```diff
@@ -303,7 +312,7 @@ function net_radiation(
     model::AbstractModel{FT},
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
-    t::FT,
+    t,
 ) where {FT <: AbstractFloat}
```
kmdeck, line 316: same comment here as down below

src/shared_utilities/drivers.jl
```diff
@@ -147,12 +149,12 @@ It solves for these given atmospheric conditions, stored in `atmos`,
...
 function surface_fluxes(
-    atmos::PrescribedAtmosphere{FT},
+    atmos::PrescribedAtmosphere{FT, CA, DT},
     model::AbstractModel{FT},
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
-    t::FT,
-) where {FT <: AbstractFloat}
+    t,
+) where {FT <: AbstractFloat, CA, DT}
```
kmdeck, line 157: same comment here as down below

docs/tutorials/Bucket/bucket_tutorial.jl
juliasloan25: probably not! I can remove these

docs/tutorials/Canopy/canopy_tutorial.jl
juliasloan25: I wanted to enforce that they'll all be FT, but I think it's sufficient to pass FT as a type argument to `PrescribedSoil`, so I can remove it for the individual arguments

experiments/integrated/ozark/conservation/ozark_conservation.jl
```diff
@@ -264,7 +264,7 @@ root_sink = [sum(-1 .* sv.saveval[k].root_extraction) for k in 2:length(sol.t)]
 # Free Drainage BC, F_bot = -K_bot
 soil_bottom_flux = [-parent(sv.saveval[k].soil.K)[1] for k in 2:length(sol.t)]
 # Precip is not stored in the aux state, evaluated using sol.t
-precip = [atmos.liquid_precip.(sol.t[k]) for k in 1:(length(sol.t) - 1)]
+precip = [FT.(atmos.liquid_precip.(sol.t[k])) for k in 1:(length(sol.t) - 1)]
```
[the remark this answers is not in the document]
juliasloan25: good catch! I'll remove it

experiments/integrated/ozark/soil_retention_curve.jl
juliasloan25: okay, I'll remove them

src/standalone/Snow/Snow.jl
juliasloan25: We don't, but for some reason the formatter keeps adding this... I'll try to remove it again

docs/tutorials/Canopy/canopy_tutorial.jl
juliasloan25: btw, there are some places in main where we do this currently (like [here](https://github.com/CliMA/ClimaLSM.jl/blob/main/docs/tutorials/Canopy/canopy_tutorial.jl#L127-L137)), but I think it's okay to have a mix of approaches and leave it out here

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
juliasloan25: I don't think I understand - how would we use a closure to get FT?

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
juliasloan25: You're right, we can remove it since we wrap the value in FT after it's returned

src/standalone/Vegetation/PlantHydraulics.jl
juliasloan25: sure! I'll change this to get FT from the model and try to find other instances like this

src/standalone/Soil/boundary_conditions.jl
juliasloan25: I don't think we're storing FT in the BC (code [here](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/standalone/Soil/boundary_conditions.jl#L14)). We could add it as a type parameter for BCs, but I think using dz is easier

src/standalone/Soil/boundary_conditions.jl
juliasloan25: good point! I think it will get converted

src/standalone/Snow/Snow.jl
juliasloan25: Yeah, the formatter just adds it back if I try to remove it :/

src/standalone/Bucket/bucket_parameterizations.jl
juliasloan25: I don't think we need them, I'll remove them

src/standalone/Vegetation/PlantHydraulics.jl
juliasloan25: Ahh good point, I'll remove all the `<: AbstractFloat`s

src/shared_utilities/FileReader.jl
```diff
@@ -154,15 +154,15 @@ end
...
+    PrescribedDataTemporal{FT}(
         regrid_dirpath,
         infile_path,
         varname,
         date_ref,
         t_start,
         surface_space;
         mono = true,
-    )
+    ) where {FT <: AbstractFloat}
```
juliasloan25, line 165: I realized that I'm passing `FT` to these external `PrescribedDataTemporal` constructors, but `PrescribedDataTemporal` itself doesn't actually have `FT` as a parametric type, and so doesn't constrain `FT <: AbstractFloat`. Maybe I should add `FT as a type parameter for it? @kmdeck

docs/tutorials/Canopy/canopy_tutorial.jl
kmdeck: ok! and good to keep in mind. as I come across those, Ill remove them as long as they dont clutter PR diffs!

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
kmdeck: hm, I guess I thought this would be a closure
```
FT = Float64
function foo(a, b; FT = FT)
 # use FT
end
```
and then pass foo into the model. But, maybe that isnt a closure. 

I think we can get FT from the insol params too :)

src/standalone/Soil/boundary_conditions.jl
kmdeck: good point!

src/standalone/Snow/Snow.jl
kmdeck: hahah, the formatter is being creative

src/integrated/soil_canopy_model.jl
```diff
@@ -395,18 +394,18 @@ function PlantHydraulics.root_flux_per_ground_area!(
     model::Canopy.PlantHydraulics.PlantHydraulicsModel{FT},
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
-    t::FT,
-) where {FT}
+    t,
+) where {(FT)}
```
kmdeck, line 398: parentheses?

kmdeck: do you think it is good practice to not use the `where {FT}`/include parameteric types if we arent dispatching off them/dont need the FT in the function?

not to change in this PR. more a question for the future

src/shared_utilities/FileReader.jl
kmdeck: hmmm that makes sense to me! if FT is needed in setting up the data from it or anything, having it handy in the type would be useful. and when we set up the simulation we pick the float type and can set it consistently. we can also possibly enforce that this FT is the same as the model FT...possibly in the model constructor? not sure

src/shared_utilities/drivers.jl
```diff
@@ -240,24 +242,27 @@ function surface_fluxes_at_a_point(
...
 
 """
-    PrescribedRadiativeFluxes{FT, SW, LW, DT, T, OD} <: AbstractRadiativeDrivers{FT}
+    PrescribedRadiativeFluxes{FT, DT, T, OD} <: AbstractRadiativeDrivers{FT}
 
 Container for the prescribed radiation functions needed to drive land models in standalone mode.
 $(DocStringExtensions.FIELDS)
+
+The arguments labeled with "function of time" are typically either of type
+`Function` or type `Spline1D` from the `Dierckx.jl` package.
```
kmdeck, line 265: thanks for clarifying this!

src/shared_utilities/utils.jl
kmdeck, line 93: this should be fine! but seeing this reminded me of this https://github.com/CliMA/ClimaLSM.jl/issues/356 (which we need to fix anyways, but which made me thing that right now our FieldVectors are not always with parametric type {FT}.

FT can be anything here, so it doesnt affect things, but also FT isnt used in the function. maybe here we can just get rid of it?

src/standalone/Soil/boundary_conditions.jl
kmdeck: oh - whatever you think is best is fine - but I just noticed that we have FT available in the model also, and that is an argument.

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -142,8 +142,8 @@ function PlantHydraulicsParameters(;
 ) where {FT}
     return PlantHydraulicsParameters{
         FT,
-        typeof(conductivity_model),
-        typeof(retention_model),
+        Base.typeof(conductivity_model),
```
kmdeck, line 145: do we need Base. here?

test/shared_utilities/variable_types.jl
```diff
@@ -17,97 +17,99 @@ using ClimaLSM.Domains: HybridBox, Column, Point
...
-
-    dm_compute_imp_tendency! = make_compute_imp_tendency(dm)
-    @test dm_compute_imp_tendency!(x[1], x[2], x[3], x[4]) == nothing
-    @test x == [0, 1, 2, 3]
-
-    dm_update_aux! = make_update_aux(dm)
-    @test dm_update_aux!(x[1], x[2], x[3]) == nothing
-    @test x == [0, 1, 2, 3]
-end
+for FT in (Float32, Float64)
```
kmdeck, line 20: Im not sure we need the different values of FT for this test? (I know it doesnt really make a difference and is good practice. but not sure if we want to be a little more judicious about where we test multiple types, since doing everywhere will double CI time (I think?))

kmdeck (approved):
Julia - thanks much for taking on this mammoth task! I left a couple other comments that are small, and had a question about if we needed the FT loop for all of our tests. Otherwise I think it is good to go once CI passes :)

kmdeck:
@Espeer5 @AlexisRenchon it might be good to take a look at this PR and see the broad type of change @juliasloan25  is making, as well as take a look at the docs she has written https://github.com/CliMA/ClimaLSM.jl/wiki/Floating-point-precision-issues. To make her life easier with this PR, let's try and make any new PRs (vaira experiment, other experiments, heterotrophic respiration), respect the changes made here. The overarching idea is that we cannot rely on `t` being the same type as the FT we use for the state, parameters, model, etc. It will always be Float64.

e.g. avoid using `t::FT`, `(p,t) -> eltype(t)(value)`

AlexisRenchon:
is this PR far from being merged? 
if it will be merged soon we could wait for it before we merge ours. (and when we do, we'll avoid `eltype` etc.)

kmdeck:
> is this PR far from being merged? if it will be merged soon we could wait for it before we merge ours. (and when we do, we'll avoid `eltype` etc.)

I think Julia will merge this pretty quickly once she is back, but I dont think that is until next week.

AlexisRenchon:
I'm ok to wait, this is a big PR

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
kmdeck: This works:
```
function zenith_angle(
    t,
    orbital_data,
    ref_time;
    latitude = lat,
    longitude = long,
    insol_params::Insolation.Parameters.InsolationParameters{FT} = earth_param_set.insol_params,
) where {FT}
    # This should be time in UTC
    dt = ref_time + Dates.Second(round(t))
    # Orbital Data uses Float64, so we need to convert to our sim
    # FT.
    FT(
        instantaneous_zenith_angle(
            dt,
            orbital_data,
            longitude,
            latitude,
            insol_params,
        )[1],
    )
end

```

Ill change to this for the Ozark and Vaira site runs (PR adding the Vaira site), we can change next week when youre back if you prefer a different approach!

test/shared_utilities/variable_types.jl
juliasloan25: I agree, I don't think we need both FTs for this test. On one hand, it might be good practice to cover both FTs everywhere, and I don't think it will quite double CI time since precompilation still just happens once. I'm okay with either covering both FTs in all tests, or doing it for just the ones we think require it (maybe in the latter case we could add a note justifying why we don't need to cover both FTs)

experiments/integrated/ozark/ozark_met_drivers_FLUXNET.jl
juliasloan25: Looks good, thank you!

src/shared_utilities/FileReader.jl
juliasloan25: I feel conflicted... FT would only be used in interpolation, I think ([here](https://github.com/CliMA/ClimaLSM.jl/blob/10e457ad9e6930b90181924f8bddd6b33ae446c5/src/shared_utilities/FileReader.jl#L390-L412)). I think it's fine to continue getting FT from the space as we do now, and maybe simpler

src/integrated/soil_canopy_model.jl
juliasloan25: I'm not sure - I don't think we get any quantitative benefit from specifying `where {FT}` in these cases, but maybe it's a bit more readable/clear to the user? I'm fine with either way

test/shared_utilities/variable_types.jl
kmdeck: I think my preference would be to only do where required but it's your call! youre right that it wont double CI.

src/integrated/soil_canopy_model.jl
kmdeck: ok! maybe this is something we can clean up as we merge PRs that affect these functions if we decide to do this. I can see both sides :) but might lean towards removing (might be confusing to newer users of Julia?).

PR #355, opened 2023-10-13
fix soil/en/hydr/biogeo test

### LINKED ISSUE #354: test not called in runtests
## Describe the bug
test/integrated/soil_energy_hydrology_biogeochemistry.jl is not actually called in runtests.jl
see here: https://github.com/CliMA/ClimaLSM.jl/blob/main/test/runtests.jl#L15-L17
this should call `test/integrated/soil_energy_hydrology_biogeochemistry.jl` instead of duplicating `test/integrated/pond_soil_lsm.jl`

juliasloan25:
## Purpose 
closes #354 


kmdeck (approved):
good catch! thank you!

PR #361, opened 2023-10-24
albedo artifacts race condition fixed

### LINKED ISSUE #359: race condition in BulkAlbedoStatic
## Describe the bug
When we create a BulkAlbedoStatic, we call the artifact function bareground_albedo... from all processes, but this needs to be called by the root so that the artifact is downloaded only once, by the root, before any processes try to use the artifact. There seems to be a race condition if multiple processes try to download the artifact at the same time

found in https://github.com/CliMA/ClimaCoupler.jl/issues/470

## To Do
- [x] add move bareground_albedo call inside of function, add iamroot check [here](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/standalone/Bucket/Bucket.jl#L115-L123)
- [x] makesimilar change for BulkAlbedoTemporal

## Alternate solution
- [ ] if we check in the `src/standalone/Bucket/artifacts/Artifacts.toml` file, we'll avoid this issue because it won't need to recreate it during each run

juliasloan25:
## Purpose 
closes #359

----


juliasloan25:
@kmdeck what do you think of these changes? I tried to fix the race condition we've seen in the coupler with the fewest changes to ClimaLSM.

One thing to note is that now for PrescribedDataStatic and PrescribedDataTemporal, we have to pass in a data file as a function that downloads the file, whereas before we just passed a string of the filename for the data. I think this is okay bc this is our current use case, and if we want to be able to pass strings in the future we can add a method for that. Simon has suggested rethinking how we handle artifacts for CliMA as a whole, so this may all be redone anyway.

src/shared_utilities/FileReader.jl
```diff
@@ -144,10 +146,17 @@ simulation grid. Date-related args (last 3 to FileInfo) are unused for static
...
+    get_infile_path::Function,
     regrid_dirpath::String,
     varname::String,
+    comms_ctx::ClimaComms.AbstractCommsContext,
 )
+    # Download `infile_path` artifact on root process first to avoid race condition
+    if ClimaComms.iamroot(comms_ctx)
+        infile_path = get_infile_path()
+    end
+    infile_path = get_infile_path()
```
kmdeck, line 158: Im a little confused here. if not ClimaComms.iamroot(comms_ctx), does anything change? we still get the file path

src/shared_utilities/FileReader.jl
```diff
@@ -205,8 +216,18 @@ function PrescribedDataTemporal{FT}(
...
+        NCDataset(infile_path, "r") do ds
+            if !("time" in keys(ds))
+                error(
+                    "Using a temporal albedo map requires data with time dimension.",
+                )
+            end
+        end
     end
     ClimaComms.barrier(comms_ctx)
+    infile_path = get_infile_path()
```
kmdeck, line 229: similar question here

src/standalone/Bucket/Bucket.jl
```diff
@@ -143,13 +150,13 @@ end
...
         Space::ClimaCore.Spaces.AbstractSpace;
-        infile_path = Bucket.cesm2_albedo_dataset_path(),
+        get_infile_path = Bucket.cesm2_albedo_dataset_path,
         varname = "sw_alb"
     ) where {FT}
 
 Constructor for the BulkAlbedoTemporal struct.
 The `varname` must correspond to the name of the variable in the NetCDF
-file specified by `infile_path`.
+file specified by `get_infile_path`.
```
kmdeck, line 159: any more clarification needed that this is now a function that returns the path, and downloads the data?

src/standalone/Bucket/Bucket.jl
```diff
@@ -101,24 +101,31 @@ end
...
         varname = "sw_alb",
-        infile_path = Bucket.bareground_albedo_dataset_path(),
+        get_infile_path::Function = Bucket.bareground_albedo_dataset_path,
     ) where {FT}
 
 Constructor for the BulkAlbedoStatic that implements a default albedo map, `comms` context, and value for `α_snow`.
 The `varname` must correspond to the name of the variable in the NetCDF file specified by `path`.
 
-The `bareground_albedo_dataset_path` artifact can be used as a default with this type.
+The `bareground_albedo_dataset_path` artifact will be used as a default with this type.
```
kmdeck, line 114: maybe some clarity needed here re: path vs artifact? (is the artifact the data itself, while the path points to the data?)

kmdeck (approved):
This looks fine to me. Maybe just some clarity in docs about paths vs functions that get paths (and download the data if called the first time?).

kmdeck:
> @kmdeck what do you think of these changes? I tried to fix the race condition we've seen in the coupler with the fewest changes to ClimaLSM.
> 
> One thing to note is that now for PrescribedDataStatic and PrescribedDataTemporal, we have to pass in a data file as a function that downloads the file, whereas before we just passed a string of the filename for the data. I think this is okay bc this is our current use case, and if we want to be able to pass strings in the future we can add a method for that. Simon has suggested rethinking how we handle artifacts for CliMA as a whole, so this may all be redone anyway.

I think this is fine! it may be that paths can be handled internally as artifacts get downloaded somewhere into the .julia folder, and usually the person running it doesnt really pick where. 

is it clear how to set up a function that downloads the file, or would we provide a small wrapper function (or something) that does that?

src/shared_utilities/FileReader.jl
juliasloan25: The goal is for the root to download the file first, then other files can try to download it after (since it's already downloaded, they'll just access it instead of downloading). If they all call `get_infile_path()` at the same time (like we had before), some threads try to read the file while another thread is still downloading it, so we get a `No such file or directory` error like [here](https://buildkite.com/clima/climacoupler-longruns/builds/251#018b3fee-be6a-4792-8df6-63daaf906e49)

Also, later on in the function, `infile_path` is used by all processes (not just root), so it has to be defined as a variable on all processes

Now that I think about it some more, I should have a barrier after the `iamroot` if statement, to enforce that all the threads have to wait for the root process to finish downloading. I'll add that right now

src/shared_utilities/FileReader.jl
juliasloan25: hopefully the comment above explains both!

src/standalone/Bucket/Bucket.jl
juliasloan25: I updated the docstring, hopefully it's more clear now

src/shared_utilities/FileReader.jl
kmdeck: yep! this makes sense to me, I see the `barrier` here 👍

src/shared_utilities/FileReader.jl
kmdeck: makes sense! Thanks!

src/standalone/Bucket/Bucket.jl
juliasloan25: I added some info to the docstrings, and also renamed `get_infile_path` --> `get_infile` to hopefully emphasize that it downloads the data rather than only returning the filepath

juliasloan25:
> > @kmdeck what do you think of these changes? I tried to fix the race condition we've seen in the coupler with the fewest changes to ClimaLSM.
> > One thing to note is that now for PrescribedDataStatic and PrescribedDataTemporal, we have to pass in a data file as a function that downloads the file, whereas before we just passed a string of the filename for the data. I think this is okay bc this is our current use case, and if we want to be able to pass strings in the future we can add a method for that. Simon has suggested rethinking how we handle artifacts for CliMA as a whole, so this may all be redone anyway.
> 
> I think this is fine! it may be that paths can be handled internally as artifacts get downloaded somewhere into the .julia folder, and usually the person running it doesnt really pick where.
> 
> is it clear how to set up a function that downloads the file, or would we provide a small wrapper function (or something) that does that?

If we wanted to download the data ourselves, I think we could do something similar to what ArtifactWrappers.jl does in their [src code](https://github.com/CliMA/ArtifactWrappers.jl/blob/main/src/ArtifactWrappers.jl), but I think it makes sense to just keep using that package for now

PR #362, opened 2023-10-25
Update to SF v0.8.0

akshaysridhar:
Use SFv0.8.0 in ClimaLSM. 	
Update compat:   Project.toml
Update interface:   src/shared_utilities/drivers.jl


akshaysridhar (commented):
@kmdeck Do we also want to bump the `julia` compat to 1.9 here?

akshaysridhar:
It seems the experiments in buildkite are currently failing because we are limited to `ClimaLSMv0.3.3` which pins the code to `SurfaceFluxesv0.7.2`.
```(experiments) pkg> dev ..``` should make this work.

File: .buildkite/pipeline.yml [hunk 1]
```diff
@@ -24,6 +24,7 @@ steps:
       - "julia --project -e 'using Pkg; Pkg.precompile()'"
       - "julia --project -e 'using Pkg; Pkg.status()'"
 
+      - "julia --project=experiments -e 'using Pkg; Pkg.develop(path = \".\")'"
```
kmdeck, line 27: Hm, we have a manifest checked in for experiments and that has the path specified: https://github.com/CliMA/ClimaLSM.jl/blob/ad31773009c9309d5e45b083603873a8914be286/experiments/Manifest.toml#L215

so I *think* we should be ok without this? not a problem to add in if needed, but I wonder why

docs/Manifest.toml
kmdeck, line 1: we check this in :) and the experiment Manifest.toml

experiments/standalone/Soil/evaporation.jl
```diff
@@ -220,25 +220,24 @@ for i in 1:length(sol.t)
...
-        z0m = z_0m,
-        z0b = z_0b,
+        z_0m,
+        z_0b,
         beta = 1.0,
     )
     potential_conditions = SurfaceFluxes.surface_conditions(
         surface_flux_params,
         sc;
-        tol_neutral = SFP.cp_d(surface_flux_params) / 100000,
```
kmdeck, line 241: I think we may need this for this experiment but I dont recall why. Ill double check the plots once they are uploaded to buildkite and see if it makes a difference!

kmdeck (commented):
looks good! Ill just double check the plots once buildkite runs (let me know if you have issues and I can help too).

.buildkite/pipeline.yml
akshaysridhar: Ah - noted. I'll check the Manifest files in.

docs/Manifest.toml
akshaysridhar: Noted thanks - will add this back in.

kmdeck (approved):

PR #366, opened 2023-10-30
Implement series rescaling bug fix

Espeer5:
## Purpose 
 This is a very simple PR to fix an issue with data scaling for calculating the GOF statistics in the experiment code. Before, running an experiment with a different data rate than the Ozark site didn't work.

## To-do

- [ ] Verify the code doesn't need additional changes following the canopy energy equation PR which is soon to be merged

## Content

- New function, match_indices implemented within experiment to rescale each data series with respect to each other

----

### DIFF
--- a/experiments/integrated/ozark/ozark.jl
+++ b/experiments/integrated/ozark/ozark.jl
@@ -315,13 +315,11 @@ GPP_daily_avg_model = diurnal_avg(model_GPP)
 GPP_daily_avg_data =
     diurnal_avg(GPP[Int64(t_spinup ÷ DATA_DT):Int64(tf ÷ DATA_DT)])
 RMSD =
-    StatsBase.rmsd(
-        match_indices(GPP_daily_avg_model, GPP_daily_avg_data)...,
-    ) * 1e6
+    StatsBase.rmsd(match_indices(GPP_daily_avg_model, GPP_daily_avg_data)...,) *
+    1e6
 R² =
     Statistics.cor(
-        match_indices(GPP_daily_avg_model, GPP_daily_avg_data)...,
-    )^2
+        match_indices(GPP_daily_avg_model, GPP_daily_avg_data)...)^2
 
 plt1 = Plots.plot(size = (800, 400))
 Plots.plot!(
--- a/experiments/integrated/ozark/ozark.jl
+++ b/experiments/integrated/ozark/ozark.jl
@@ -268,6 +268,20 @@ function diurnal_avg(series)
         [mean([daily_data[i][j] for i in 1:num_days]) for j in 1:daily_points]
     return daily_avgs
 end
+
+# This function scales 2 data series to the same length and matches their
+# indices such that they may be used to compute GOF stats
+function match_indices(model, data)
+    dpm = dt * n / DATA_DT
+    model_per_data = 1
+    if dpm < 1
+        model_per_data = Int64(1 / dpm)
+        dpm = 1
+    end
+    data_per_model = Int64(dpm)
+    return model[1:model_per_data:end], data[1:data_per_model:end]
+end 
+
 # Autotrophic Respiration
 AR = [
     parent(sv.saveval[k].canopy.autotrophic_respiration.Ra)[1] for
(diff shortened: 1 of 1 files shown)


experiments/integrated/ozark/ozark.jl
```diff
@@ -268,6 +268,20 @@ function diurnal_avg(series)
...
+    dpm = dt * n / DATA_DT
+    model_per_data = 1
+    if dpm < 1
+        model_per_data = Int64(1 / dpm)
+        dpm = 1
+    end
+    data_per_model = Int64(dpm)
+    return model[1:model_per_data:end], data[1:data_per_model:end]
+end 
+
```
kmdeck, line 284: This works when dpm is an integer, but otherwise there is an error:
```
DATA_DT = 1800
dt = 45
n = 1
data= 0:DATA_DT:(24*3600-DATA_DT)
model = 0:n*dt:(24*3600-n*dt)

m, d = match_indices(model, data);
# looks good!
m == d


DATA_DT = 1800
dt = 400
n = 1
data= 0:DATA_DT:(24*3600-DATA_DT)
model = 0:n*dt:(24*3600-n*dt)
m, d = match_indices(model, data);
# errors
```

Im not sure how general we need to be. we could add an @assert statement that n*dt mod DATA_DT is zero or vice versa?

I think the other way to do this would be with linear interpolation - for example, if the data is sparser than the model output, we would use the model output nearest to the data at a given time in the data and interpolate the model output to that time in the data. We'd leave the data untouched. Linear interpolation itself is straightforward, but Im not sure how fast algorithm are at finding the bracketing times on either side.

anyways, just a thought. if you want to leave this as is and given an @assert plus some comments that's fine too.

experiments/integrated/ozark/ozark.jl
```diff
@@ -268,6 +268,20 @@ function diurnal_avg(series)
         [mean([daily_data[i][j] for i in 1:num_days]) for j in 1:daily_points]
     return daily_avgs
 end
+
+# This function scales 2 data series to the same length and matches their
+# indices such that they may be used to compute GOF stats
+function match_indices(model, data)
```
kmdeck, line 274: could we add this function to a new julia file in experiments/integrated? maybe "statistical_tools.jl" or something. then we can use this in all scripts without duplicating. not sure if your current work on making a single script would obviate the need for this already, though!

experiments/integrated/ozark/ozark.jl
kmdeck: (we can also use Interpolations.jl directly if that sounds easiest, instead of implementing linear interpolation ourselves, if you want to go that route)

experiments/integrated/ozark/ozark.jl
Espeer5: I think my current work making a single script even exacerbates this need! I am going to add some data checking tools and things like this that I am thinking of moving to a separate file.

experiments/integrated/ozark/ozark.jl
Espeer5: Ah! You're right, I meant to be performing integer division to avoid this error which is what I did in my previous code.

I think you're also right that interpolation is a better solution. I was thinking of this PR of a quick and dirty way of fixing the bug so that the main branch code could work for sites with slower data (i.e Harvard Forest) and then including a better solution within the data tools we will need for the PR with the general experiment script.

kmdeck: ok! sounds fine to me.

kmdeck:
@Espeer5 so to clarify youll do the data tools etc in a future pr? this one as is? i can approve it then

Espeer5:
> @Espeer5 so to clarify youll do the data tools etc in a future pr? this one as is? i can approve it then

Yes. I do need to fix the issue with this one first but then in a follow-up I will add some better data tools this one is just a patch.

kmdeck (approved):
Looks good to merge once commits squashed. thanks for these improvements!

kmdeck:
@Espeer5 this will close when we merge your overhaul PR?

Espeer5:
> @Espeer5 this will close when we merge your overhaul PR?

Yes. Sorry I mean to close this and mention that I decided it was irrelevant to merge since it never made a difference in the Ozark and experiment and that was all that was on main. The fix applies to the Harvard site and it in the overhaul PR.

Espeer5:
Closing; will be resolved by merge of #369

PR #374, opened 2023-11-06
Run CI at every PR

Sbozzolo:
Bors is gone.

### DIFF
--- a/.github/workflows/CodeCov.yml
+++ b/.github/workflows/CodeCov.yml
@@ -2,10 +2,6 @@ name: CodeCov
 
 on:
   push:
-    branches:
-      - main
-      - trying
-      - staging
     tags: '*'
   pull_request:
 
--- a/.github/workflows/JuliaFormatter.yml
+++ b/.github/workflows/JuliaFormatter.yml
@@ -2,10 +2,6 @@ name: JuliaFormatter
 
 on:
   push:
-    branches:
-      - main
-      - trying
-      - staging
     tags: '*'
   pull_request:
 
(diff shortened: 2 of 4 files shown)


Sbozzolo:
@kmdeck, when this has merged, can you try merging a new PR that removes `bors.toml` to check that everything works as you would expect?

To merge, you can just click the "Enable auto-merge" button on your PR.

juliasloan25:
Just to make sure I understand - this change will run the whole CI pipeline on each commit pushed to Github, right? So in terms of changes to workflow, we'll now need to do the following:
- rebase on main when a PR is ready
- enable auto-merge when we want to merge a rebased PR (instead of `bors r+` as before)

We should also continue using `[skip ci]` frequently to avoid running all these checks unnecessarily (though this isn't new)

Sbozzolo:
> Just to make sure I understand - this change will run the whole CI pipeline on each commit pushed to Github, right? So in terms of changes to workflow, we'll now need to do the following:
> 
> * rebase on main when a PR is ready
> * enable auto-merge when we want to merge a rebased PR (instead of `bors r+` as before)
> 
> We should also continue using `[skip ci]` frequently to avoid running all these checks unnecessarily (though this isn't new)

Yes, correct. We can enable more granularity if needed. For example, we can exclude draft PRs from buildkite.

juliasloan25:
Great, thank you for updating this!

kmdeck (approved):

kmdeck:
@juliasloan25 @Sbozzolo I updated the permission to require a review before merging, and then approved this. (this was previously a req, Im not sure why it was not enforced here)

Julia, maybe we can talk more about how to skip CI in certain cases.

Gabriele, will try removing bors.toml before the end of the day

PR #378, opened 2023-11-08
update buildkite

simonbyrne:
## Content

----


juliasloan25:
GHA are failing because we're still using NCDatasets' old syntax on this branch. #380 updates it - once that's merged you can rebase to fix the index errors

.buildkite/pipeline.yml
```diff
@@ -1,32 +1,30 @@
...
       - "julia --project -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-      - "julia --project -e 'using Pkg; Pkg.precompile()'"
       - "julia --project -e 'using Pkg; Pkg.status()'"
 
+      - echo "--- Instantiate test"
       - "julia --project=experiments -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-      - "julia --project=experiments -e 'using Pkg; Pkg.precompile()'"
       - "julia --project=experiments -e 'using Pkg; Pkg.status()'"
+    agents:
+      slurm_cpus_per_task: 8
```
simonbyrne, line 25:
```suggestion
      slurm_cpus_per_task: 8
      slurm_mem_per_cpu: 8G
```

.buildkite/pipeline.yml
```diff
@@ -58,3 +57,6 @@ steps:
       - label: "Water conservation"
         command: "julia --color=yes --project=experiments experiments/standalone/Soil/water_conservation.jl"
         artifact_paths: "experiments/standalone/Soil/water_conservation*png"
+  - wait
+
+  - command: "compact_depot"
```
simonbyrne, line 62:
```suggestion
  - command: "compact_depot"

```

.buildkite/pipeline.yml
```diff
@@ -1,32 +1,31 @@
...
-      - "julia --project -e 'using Pkg; Pkg.precompile()'"
       - "julia --project -e 'using Pkg; Pkg.status()'"
 
+      - echo "--- Instantiate test"
       - "julia --project=experiments -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-      - "julia --project=experiments -e 'using Pkg; Pkg.precompile()'"
       - "julia --project=experiments -e 'using Pkg; Pkg.status()'"
+    agents:
+      slurm_cpus_per_task: 8
+      slurm_mem_per_cpu: 8G
```
juliasloan25, line 26: why did you increase the memory here?

.buildkite/pipeline.yml
juliasloan25, line 62: does this cache the artifacts from this job on the cluster? do we need a DepotClear for ClimaLSM in buildkite now?

juliasloan25 (approved):

.buildkite/pipeline.yml
simonbyrne: It seems to speed up the compilation a bit.

.buildkite/pipeline.yml
simonbyrne: it's different than the shared depot we use in ClimaAtmos: basically in this job we have 2 depots: the first 1 is "clean", and the 2nd one just contains artifacts. We install everything into the 1st one, but at the end of each build, we move all the artifacts to the 2nd one

juliasloan25: so then do these 2 depots persist across jobs so they can be reused?

juliasloan25: also, will we ever need to clear these depots or do any other manual management of them?

simonbyrne: the 2nd one persists, the 1st one does not

simonbyrne: > also, will we ever need to clear these depots or do any other manual management of them?

I hope not, but I can't say for sure.

PR #385, opened 2023-11-13
experiments use Float32 and Float64 [wip]

### LINKED ISSUE #384: test Float32 and Float64 in exps
#353 restores compatibility with Float32

We should test out this behavior in all of our experiments before making a release so these changes can be propogated to the coupler

## To Do
- for each exp, loop over (FT, t_end) - run Float32 only for a couple steps, Float64 for entire simulation
  - at end of sim, check final type for both, make plots for Float64 

juliasloan25:
## Purpose 
closes #384

this is a test branch


juliasloan25:
superseded by #389

PR #389, opened 2023-11-15
float32 in experiments [replacement PR]

### LINKED ISSUE #384: test Float32 and Float64 in exps

juliasloan25:
## Purpose 
replaces #385 (this PR rebased on main)
closes #384

We need to test out our Float32 compatibility in the experiments (which act as integration tests).

### To do
- [x] for each exp, loop over (FT, t_end) - run Float32 only for a couple steps, Float64 for entire simulation
at end of sim, check final type for both, make plots for Float64

### QA
- This increases CI time slightly, but only by 1 minute max per run (and typically more like 20 seconds)


experiments/integrated/ozark/conservation/ozark_conservation.jl
```diff
@@ -15,442 +15,485 @@ using ClimaLSM.Canopy
...
-        f2 = f2,
+include(joinpath(climalsm_dir, "parameters", "create_parameters.jl"))
+
+# Define simulation times
+t0 = Float64(120 * 3600 * 24)# start mid year
+N_days = 10
+tf = t0 + Float64(3600 * 24 * N_days)
+dt = Float64(150)
+n = 1
+saveat = Array(t0:(n * dt):tf)
```
juliasloan25, line 27: I moved these here so that they're defined before we create `FT` and `tf`, since `tf` depends on them. They were in ozark_simulation.jl before, but that defines the convergence condition which needs to use `FT`, so I wanted that to be able to be called after `FT` is defined. I don't think this is the best setup, but I'm not sure what would be better

test/standalone/Bucket/albedo_types.jl
```diff
@@ -328,7 +328,7 @@ end
             ref_time,
             h_atmos,
         )
-        Δt = FT(1.0)
+        τc = FT(1.0)
```
juliasloan25, line 331: I renamed these bc at first it was confusing to me why `\Delta t` was being defined with `FT`

experiments/integrated/ozark/conservation/ozark_conservation.jl
```diff
@@ -15,442 +15,485 @@ using ClimaLSM.Canopy
...
+
+for (float_type, tf) in ((Float32, t0 + 2 * dt), (Float64, tf))
+    # Make these global so we can use them in other ozark files
+    global FT = float_type
+    global earth_param_set = create_lsm_parameters(FT)
```
juliasloan25, line 32: I also don't think this setup is super clean, but since these are defined in the for loop they need to be global so the other included files can access them

experiments/integrated/ozark/conservation/ozark_conservation.jl
kmdeck: What if we include ozark_simulation inside the loop, like the other files, and then have an if statement setting the final time (introduce a new variable = tf in one case and = t0+3*dt in another)?

BTW only the conservation test uses a timestepper with a convergence criterion - the other ones use RK4. I think because the only way to use forward Euler is with ARS111, which is an ImEx algorithm, whereas RK4 is an explicit algorithm. in each case, though, we only have an explicit tendency for now.

(that said, these will eventually use an implicit solver).

experiments/integrated/ozark/conservation/ozark_conservation.jl
kmdeck: it's ok! but if we move ozark_sim inside the loop, then everything is inside the loop I think?

experiments/standalone/Soil/water_conservation.jl
kmdeck, line 1: Do you think for this one it is worth having plots for both Float types? Im curious if it still works well with Float32 + our convergence criteria update.

kmdeck (approved):
Looks good, my only hesitation is that we might only want the Float64/Float32 checks on one of the integrated tests. For example, maybe we could run the ozark conservation script at each Float type? but not run the ozark + vaira sim script. Not sure, but they are all running the same model, just with different input/params. 

Also, it might be nice to keep a simple copy of running a sim with one float type + plots, as an example of how to run (and as what Alexis and Eddie are building off of to run at multiple sites, splitting plotting functionality into it's own script). 

One other possible idea is to pass the FT as a command line argument, since that would remove the loop in the code/simplify the include statements.

experiments/integrated/ozark/conservation/ozark_conservation.jl
juliasloan25: even when ozark_sim is called from inside the loop, `FT` and `earth_param_set` have to be specified as global to be accessed from other files. I don't exactly understand how the scope is defined, but I get an error that `FT` is undefined from within ozark_sim if I don't make it global :/

experiments/standalone/Soil/water_conservation.jl
juliasloan25: Sure! But I think if we want to plot it, we should run it for the full simulation length with Float32 so the results make sense. I'll change it to run the full sim for both and we can see how long it takes (I think it'll still be much shorter than the other tests)

PR #397, opened 2023-11-17
use Float32 in docs

### LINKED ISSUE #387: update FT use in docs
the last part of Float32 compatibility (I hope) :)

Go through docs and make sure FT is used as intended.

specifically:
- [x] don't set t, t0, tf, dt as FT (e.g. [here](https://github.com/CliMA/ClimaLSM.jl/blob/main/docs/tutorials/Soil/soil_energy_hydrology.jl#L247))
- [x] change docs to use Float32 since Float64 compatibility is easier to achieve

closed by #397

juliasloan25:
## Purpose 
- [x] update FT usage according to new guidelines
- [x] use Float32 instead of Float64 in docs/ to ensure compatibility

closes #387


test/shared_utilities/file_reader.jl
```diff
@@ -27,23 +27,23 @@ isdir(regrid_dir_temporal) ? nothing : mkpath(regrid_dir_temporal)
 FT = Float32
 @testset "test interpol, FT = $FT" begin
```
juliasloan25, line 28: I only replaced `t` with `x` in this test bc it kept coming up when I searched for `t = FT...`

kmdeck: :D

kmdeck (approved):
thank you, Julia! looks good. just had one question as noted.

docs/tutorials/standalone/Bucket/bucket_tutorial.jl
```diff
@@ -195,7 +197,7 @@ bucket_parameters = BucketModelParameters(
     W_f,
     z_0m,
     z_0b,
-    Δt,
+    τc,
```
kmdeck, line 200: Not sure if this changes the behavior? but if dt = 3600 was passed before, we are now passing tc=100. Not sure if tc was even used before? Perhaps for consistency, can keep behavior as is (which may mean making tc = 3600?)

juliasloan25: Sure! I had changed it because other tests were using smaller values (1, 10, 100). But you're right that we don't want to change the behavior in this PR :)

PR #400, opened 2023-11-22
tag v0.4.0, use julia v1.9.4

### LINKED ISSUE #388: update to julia v1.9.4
use latest julia version
(maybe we can do this the next time we update dependencies and make a release)

closed by #400

juliasloan25:
## Purpose 
- [x] tag new release v0.4.0
- [x] use latest julia v1.9.4

closes #388


kmdeck (approved):
Looks good! Though, in our CI .github workflow, we only specify julia "1.9". Do you think that means it should work will all julia 1.9 versions? if so, we can update the README to say that it is not tested with versions < 1.9.0?

juliasloan25:
ClimaAtmos and ClimaCore also just specify 1.9 in their CI... I would assume it'll use the most recent julia version available but not totally sure. Our README has a line saying the package is meant to be used with julia v1.9.4 and earlier versions aren't tested, which I think is sufficient since our buildkite tests will run only 1.9.4 now

PR #404, opened 2023-11-22
More steps toward GPU compatibility

Sbozzolo:
- Fix type instability in `precipitation` in test
- Remove one abstract type from one function
- Start extracting the floating point type from the state
- Ensure that time is always Float64 even in the savet array

### DIFF
--- a/src/integrated/pond_soil_model.jl
+++ b/src/integrated/pond_soil_model.jl
@@ -59,7 +59,7 @@ function LandHydrology{FT}(;
     (; precip) = land_args
 
     sources = ()
-    surface_runoff = PrognosticRunoff{FT}(precip)
+    surface_runoff = PrognosticRunoff(precip)
     boundary_conditions =
         (; top = (water = RunoffBC(),), bottom = (water = Soil.FreeDrainage(),))
 
@@ -182,7 +182,7 @@ function make_interactions_update_aux(
 end
 
 """
-    PrognosticRunoff{FT} <: Pond.AbstractSurfaceRunoff{FT}
+    PrognosticRunoff <: Pond.AbstractSurfaceRunoff
 
 Concrete type of `Pond.AbstractSurfaceRunoff` for use in LSM models,
 where precipitation is passed in, but infiltration is computed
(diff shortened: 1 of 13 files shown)


src/integrated/pond_soil_model.jl
```diff
@@ -193,27 +193,29 @@ ensuring the infiltration used for the boundary condition of soil
...
 """
 function Pond.surface_runoff(
-    runoff::PrognosticRunoff{FT},
+    runoff::PrognosticRunoff,
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
     t,
-) where {FT}
+)
+    FT = FTfromY(Y)
```
juliasloan25, line 218: why is it preferred to get `FT` like this?

src/shared_utilities/utils.jl
```diff
@@ -273,3 +274,8 @@ function NonInterpSavingCallback(saved_values, saveat::Vector{FT}) where {FT}
         save_positions = (false, false),
     )
 end
+
+function FTfromY(Y::ClimaCore.Fields.FieldVector)
```
juliasloan25, line 278: can we call this `FT_from_Y`? I think that better matches the convention we use for function naming

juliasloan25 (commented):
looks good! I just had a question and a comment

src/integrated/pond_soil_model.jl
Sbozzolo: Before, we were getting `FT` from the type. However, the type does no longer encode `FT`. So, we have to get it from somewhere else. The main source of truth for that is the `space` (`Field`s have to share the same floating point type as the `space), so ideally we would get it with `Spaces.undertype(Y)`. Now, this does not work because `Y` is a `FieldVector`, so for the moment I am getting it as with `eltype`.

src/shared_utilities/utils.jl
Sbozzolo: I went with `FTfromY` because the `Julia` conventions recommend not using underscores when the name is short and clearly readable (as in this case, which is helped by the capital letters). 

I can change to `FT_from_Y` if you prefer.

juliasloan25: Okay, that's fine!

juliasloan25 (approved):
Looks good! Feel free to merge (ideally after squashing the commits down to 1)

Sbozzolo:
> Looks good! Feel free to merge (ideally after squashing the commits down to 1)

I'd like to keep the commits split because they are logically distinct and each one of them implements a different (and working) change

PR #406, opened 2023-11-22
rm UnPack

### LINKED ISSUE #403: remove UnPack
remove UnPack dependency
we can unpack natively in julia now so we don't need this package

closed by #406

juliasloan25:
## Purpose 
remove dependency on UnPack

note: UnPack still appears in Manifests because it's required by other dependencies

closes #403


Sbozzolo (approved):
Thank you!

PR #410, opened 2023-11-24
More GPU

Sbozzolo:
This PR:
- Removes most abstract types from most structs (149e4cc0)
- Removes `Symbol`s from broadcasted expressions (4750c500, ca2cf0e1)
- Fixes some types and mark GPU compatibility for some tests (ecf9c205, 00a56e80, 9f4c04b0, c7950a92)

This PR requires ClimaCore 0.11.5 (not released yet at the time of writing). With this PR, **27/28** tests are now passing!

With this PR, the Ozark experiment also runs if one uses CUDA 5 and substitutes Spline1D objects with something that is GPU compatibile


src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -616,6 +616,8 @@ function root_water_flux_per_ground_area!(
     ψ_soil::FT = s.ψ(t)
     fa .= FT(0.0)
     @inbounds for i in 1:n_root_layers
+        local_area_index = getproperty(area_index, model.compartment_labels[1])
```
Sbozzolo, line 619: @kmdeck can you please comment on a good variable name for this?

(This occurs in a couple of other places)

kmdeck: ah. this is the area index of the first compartment (above ground), which can be stem (e.g. trees) or leaf (e.g. grass). it can be outside of the loop, I guess. 

maybe...`above_ground_area_index`? im a little stuck. im fine with a wordy name if it's clear

Sbozzolo: I went with `above_ground_area_index` :)

docs/tutorials/integrated/soil_canopy_tutorial.jl
```diff
@@ -309,7 +309,8 @@ maxLAI = FT(4.2)
 K_sat_plant = FT(1.8e-8)
 RAI = (SAI + maxLAI) * f_root_to_shoot;
 # Note: LAIfunction was determined from data in the script we included above.
-ai_parameterization = PrescribedSiteAreaIndex{FT}(LAIfunction, SAI, RAI)
+ai_parameterization =
+    PrescribedSiteAreaIndex{FT, typeof(LAIfunction)}(LAIfunction, SAI, RAI)
```
kmdeck, line 313: I forget if this was allowed: a constructor that takes in these args, but then gets the typeof(LAIfunction) and does the full parametric typing internally to hide it from the user. if so, that seems like a nicer interface for the user. what do you think? I can open an issue so we keep track of it

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -181,15 +189,21 @@ option (intendend only for debugging) to use a prescribed transpiration rate.
 
 $(DocStringExtensions.FIELDS)
 """
-struct PlantHydraulicsModel{FT, PS, T} <: AbstractPlantHydraulicsModel{FT}
+struct PlantHydraulicsModel{
+    FT,
+    PS,
+    T,
+    AA1 <: AbstractArray{FT},
```
kmdeck, line 196: Im curious if AA1 and AA2 are both needed, if they are the same type (same length, element type)?

src/standalone/Vegetation/optimality_farquhar.jl
```diff
@@ -112,14 +112,25 @@ function OptimalityFarquharParameters{FT}(;
 end
 
 """
-    OptimalityFarquharModel{FT} <: AbstractPhotosynthesisModel{FT}
+    OptimalityFarquharModel{FT,
```
kmdeck, line 115: Does this one need something like the following? I saw you added it to some of the other canopy parameterizations.

`Base.eltype(::BigLeafEnergyParameters{FT}) where {FT} = FT`

test/shared_utilities/file_reader.jl
```diff
@@ -23,6 +23,16 @@ regrid_dir_temporal =
 isdir(regrid_dir_static) ? nothing : mkpath(regrid_dir_static)
 isdir(regrid_dir_temporal) ? nothing : mkpath(regrid_dir_temporal)
 
+"""
+Set NaN and Missing to zero (GPU compatbile)
+"""
+function replace_nan_missing!(field)
```
kmdeck, line 29: not a big deal, but should we note that the field needs to be a ClimaCore.Fields.Field?

test/shared_utilities/file_reader.jl
```diff
@@ -278,10 +288,8 @@ if !Sys.iswindows()
         # Test that both data fields store the same data
         # Remove NaNs and missings before comparison
         (; data_fields) = prescribed_data.file_state
-        replace!(parent(data_fields[1]), NaN => 0)
-        replace!(parent(data_fields[1]), missing => 0)
-        replace!(parent(data_fields[2]), NaN => 0)
-        replace!(parent(data_fields[2]), missing => 0)
+
+        foreach(replace_nan_missing!, data_fields)
```
kmdeck, line 292: nice!

test/shared_utilities/utilities.jl
```diff
@@ -156,11 +157,20 @@ end
 
         # make fields non-constant in order to check the
         # impact of the dss step
-        for i in eachindex(parent(field))
-            parent(field)[i] = sin(i)
-            parent(subfield1)[i] = cos(i)
-            parent(subfield2)[i] = sin(i) * cos(i)
-        end
+        ArrayType = ClimaComms.array_type(ClimaComms.device())
```
kmdeck, line 160: It might be nice to have a short comment explaining why these steps are needed (compared with the previous)

kmdeck (commented):
I went through it - thanks for the really clear PR description.  I wonder if there are some doc strings that need updating now, with the different parametric type signatures.

I was also wondering about this type of method defintion: `Base.eltype(::BigLeafEnergyParameters{FT}) where {FT} = FT` and when it was ok to define new Base methods.

docs/tutorials/integrated/soil_canopy_tutorial.jl
kmdeck: sounds good! but we can for sure do in a follow-on PR, and it doesnt have to be you that does it either.

src/standalone/Vegetation/PlantHydraulics.jl
Sbozzolo: No, if we can assume that they are exactly same length and type, they don't have to be distinct. I consolidated to one.

src/standalone/Vegetation/optimality_farquhar.jl
Sbozzolo: If I didn't add it, it means that it is not used anywhere at the moment. We can add it if needed.

test/shared_utilities/file_reader.jl
Sbozzolo: It doesn't really have to be a `ClimaCore.Fields.Field`, the function would work for any struct that has a `parent` (e.g., a `view` of an array).

However, in practice, it would be unwise to use this function for things that are not fields, so I added the annoitation.

test/shared_utilities/file_reader.jl
Sbozzolo: ^_^

Sbozzolo:
> I went through it - thanks for the really clear PR description. I wonder if there are some doc strings that need updating now, with the different parametric type signatures.

Yes, I should double check those. 


> I was also wondering about this type of method defintion: `Base.eltype(::BigLeafEnergyParameters{FT}) where {FT} = FT` and when it was ok to define new Base methods.

I haven't defined that specific method because it is not needed/used. We can add it for completion if you prefer (I won't oppose to it).

We defined new methods for Base functions all the times. For example, when we want to provide our `show` or `summary` functions. We are defining a method for a struct that is not in Base, so this is not type piracy. In fact, we are expected to extended `Base` so that users find that our types behave similarly to primitive ones.

docs/tutorials/integrated/soil_canopy_tutorial.jl
Sbozzolo: I did it in this PR :)

kmdeck:
> In fact, we are expected to extended `Base` so that users find that our types behave similarly to primitive ones.

thanks!

src/standalone/Bucket/Bucket.jl
```diff
@@ -75,13 +75,20 @@ is specified as a function
...
 wavelength; surface is this context refers to soil and vegetation.
 """
-struct BulkAlbedoFunction{FT} <: AbstractLandAlbedoModel{FT}
+struct BulkAlbedoFunction{FT, F <: Function} <: AbstractLandAlbedoModel{FT}
     α_snow::FT
-    α_sfc::Function
+    α_sfc::F
+end
+
+function BulkAlbedoFunction{FT}(
```
kmdeck, line 83: There must be a default constructor that says `BulkAlbedoFunction(α_snow, α_sfc) == BulkAlbedoFunction{typeof(α_snow), typeof(α_sfc)}(α_snow, α_sfc)`? In many instances in the code (in this PR), we have the former. So we might not need this particular constructor too? Im not sure though, it's nice to have the same format for everything `Struct{FT}(args...)`. up to you!

kmdeck (approved):
this looks good to me - thanks very much!

src/standalone/Bucket/Bucket.jl
Sbozzolo: `BulkAlbedoFunction(α_snow, α_sfc) == BulkAlbedoFunction{typeof(α_snow), typeof(α_sfc)}(α_snow, α_sfc)` this is precisely the default constructor for every struct!

I agree that it is nice to have the format with `{FT}`, and I added a constructor to have that. However, I also think that it is a good idea to let julia infer the types when possible. 

We can add `{FT}` around to have symmetry and consistency, but they are not really needed for this specific struct. Let me know what you prefer.

kmdeck: :D 

ok! let's keep it as is.

PR #413, opened 2023-11-28
Cancel previous Github Action runs with new pushes

Sbozzolo:
As seen on https://stackoverflow.com/a/72224714

With this change, when a new commit is pushed and some old github actions are running for the same PR/branch/tag, the old ones are killed.


juliasloan25 (approved):
thank you!

PR #414, opened 2023-11-30
Kd/unify bc computation

kmdeck:
## Purpose 
-Renames the functions in the canopy and soil model which compute the actual fluxes used in the tendency.
-Allocates space for these in `aux`
-updates the values stored in these pre-allocated spaces with `update_boundary_fluxes!`, which must be called in tendencies and jacobian computations

## To-do
Future PR (more nebulous)
- move outside of the tendency function, but still call after update_aux! and before the tendency evaluation
- define a method of `make_update_boundary_fluxes` for SoilCanopyModel (and other integrated models, or is there a default?)
- does this take the place of interactions update aux?  yes?
- add an explanation of what is in aux and when it is updated (tutorial) and docs

## Content

Review checklist

I have:

----

### DIFF
--- a/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
+++ b/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
@@ -10,6 +10,7 @@ import ClimaLSM:
     AbstractExpModel,
     make_update_aux,
     make_compute_exp_tendency,
+    make_update_boundary_fluxes,
     prognostic_vars,
     auxiliary_vars,
     name,
@@ -182,10 +183,35 @@ ClimaLSM.prognostic_vars(::SoilCO2Model) = (:C,)
 ClimaLSM.prognostic_types(::SoilCO2Model{FT}) where {FT} = (FT,)
 ClimaLSM.prognostic_domain_names(::SoilCO2Model) = (:subsurface,)
 
-ClimaLSM.auxiliary_vars(::SoilCO2Model) = (:D, :Sm)
-ClimaLSM.auxiliary_types(::SoilCO2Model{FT}) where {FT} = (FT, FT)
-ClimaLSM.auxiliary_domain_names(::SoilCO2Model) = (:subsurface, :subsurface)
+ClimaLSM.auxiliary_vars(::SoilCO2Model) = (:D, :Sm, :top_bc, :bottom_bc)
+ClimaLSM.auxiliary_types(::SoilCO2Model{FT}) where {FT} = (FT, FT, FT, FT)
+ClimaLSM.auxiliary_domain_names(::SoilCO2Model) =
+    (:subsurface, :subsurface, :surface, :surface)
 
+function make_update_boundary_fluxes(model::SoilCO2Model)
+    function update_boundary_fluxes!(p, Y, t)
+        z = ClimaCore.Fields.coordinate_field(model.domain.space.subsurface).z
+        Δz_top, Δz_bottom = get_Δz(z)
+        p.soilco2.top_bc .= boundary_flux(
+            model.boundary_conditions.top.CO2,
+            TopBoundary(),
+            Δz_top,
+            Y,
+            p,
+            t,
+        )
+        p.soilco2.bottom_bc .= boundary_flux(
+            model.boundary_conditions.bottom.CO2,
+            BottomBoundary(),
+            Δz_bottom,
+            Y,
+            p,
+            t,
+        )
+
+    end
+    return update_boundary_fluxes!
+end
 
 """
     make_compute_exp_tendency(model::SoilCO2Model)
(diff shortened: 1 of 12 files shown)


src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -483,8 +494,14 @@ function ClimaLSM.boundary_flux(
 )::ClimaCore.Fields.Field
     FT = eltype(Δz)
     p_len = Spaces.nlevels(axes(p.soilco2.D))
-    D_c = Fields.level(p.soilco2.D, p_len)
-    C_c = Fields.level(Y.soilco2.C, p_len)
+    D_c = Fields.Field(
+        Fields.field_values(Fields.level(p.soilco2.D, p_len)),
+        axes(Δz),
+    )
```
juliasloan25, line 500: I'm confused why we have the `Field` and `field_values` calls now - shouldn't `Fields.level(p.soilco2.D, p_len)` return a Field? Unless we need to remap onto the space of `\Delta z`? If we need the more complicated chain of calls could you add a comment please :)

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -127,6 +127,11 @@ function AtmosDrivenFluxBC(atmos, radiation; runoff = NoRunoff())
     return AtmosDrivenFluxBC{typeof.(args)...}(args...)
 end
 
+function create_soil_bc_named_tuple(water_flux, heat_flux)
```
juliasloan25, line 130: could you add a docstring for this please?

src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -140,6 +140,12 @@ function ClimaLSM.surface_air_density(
     return compute_ρ_sfc.(Ref(thermo_params), Ref(ts_in), T_canopy)
 end
 
+function make_update_boundary_vars(canopy::CanopyModel)
+    function update_boundary_vars!(p, Y, t)
+        canopy_boundary_fluxes!(p, canopy, canopy.radiation, canopy.atmos, Y, t)
```
juliasloan25, line 145: do we need a separate `canopy_boundary_fluxes!` function? I think we can just move that code into here and remove that function

src/standalone/Soil/rre.jl
```diff
@@ -168,6 +177,7 @@ Construct the tendency computation function for the explicit terms of the RHS,
 which are horizontal components and source/sink terms.
 """
 function ClimaLSM.make_compute_exp_tendency(model::Soil.RichardsModel)
+    # No update boundars required for this model's explicit tendency currently.
```
juliasloan25, line 180: typo: boundars -> boundaries

juliasloan25: also, could you clarify that this is because we only have BCs in the vertical and those are stepped implicitly? (at first I thought this was suggesting this model doesn't have any BCs)

src/standalone/Soil/rre.jl
```diff
@@ -400,6 +411,7 @@ function ClimaLSM.make_update_jacobian(model::RichardsModel{FT}) where {FT}
             ),
         )
         # Hardcoded for single column: FIX!
+        # The derivative of the tendency should live in boundary vars and be updated there
```
juliasloan25, line 414: I'm confused - aren't we calculating the derivative of the tendency here (in ∂T_bc∂YN )? are you saying that should get stored in something like `p.bc.dT`?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
kmdeck: yeah, for some reason after upgrading to CC  0.11 I had to change this, but now I am wondering why since we are using CC 0.11 in main. Let me double check!

src/standalone/Vegetation/canopy_boundary_fluxes.jl
kmdeck: I think this may be nice to have still, because in the SoilCanopyModel, which will have it's own `update_boundary_vars!` function, we can call `canopy_boundary_fluxes!` instead of copying that code over?

src/standalone/Soil/rre.jl
kmdeck: yeah...I was thinking eventually when we have the implicit solver working for energy/hydrology and know how itll work for integrated models, we might want to allocate space for this and update it as part of `update_boundary_vars!`...but Im not sure. more of a note for the future!

src/standalone/Soil/rre.jl
kmdeck: I updated! note that the follow up PR moving `update_boundary_vars!` outside of the tendency will remove this comment also.

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
kmdeck: yeah, idk! without it I get this error: `  Broacasted spaces are not the same.`

src/standalone/Vegetation/canopy_boundary_fluxes.jl
juliasloan25: Good point! That sounds good

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
juliasloan25: Hmm interesting. It's weird that it works fine with main using CC 0.11... I can't find the change in CC that would have caused this, so I guess we can just keep the chain of calls that works 🤷‍♀️

juliasloan25:
> @juliasloan25 what do you think about renaming from `update_boundary_vars` to `update_boundary_fluxes`? it might be more clear?

I think either one is fine! I don't feel strongly either way

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
kmdeck: agreed. I can keep looking into this, itd be nice to keep it simple.

PR #415, opened 2023-11-30
Optimality model for Vcmax

kmdeck:
## Purpose 
Closes #215 
Adds Farquhar optimality model of Smith et al. 2019. This model removes Vcmax25 as a parameter and instead computes it each timestep.

## To-do
[X] Test with ozark site and make sure output looks reasonable (offline) - do we want to have this running?
[X] add more unit tests
Future PR. - add intercellular concentration optimality model - see #419 

## Content
Adds Vcmax25 to aux
Adds OptimalityFarquharModel and Parameters
Adds update_photosynthesis! function which dispatches off of the photosynthesis model type and updates An, Rd, and Vcmax25 in place.

----


AlexisRenchon (approved):
Looks good! I don't have changes to suggest

PR #424, opened 2023-12-08
run all tests windows

### LINKED ISSUE #422: test TempestRemap on windows
TR should work using the latest package versions (CCTR, HDF5, etc). We should test this out by running all our tests on windows (remove if !windows checks)

## Update
ClimaCoreTempestRemap casts a `long *` to a `size_t *` in [this line](https://github.com/ClimateGlobalChange/tempestremap/blob/master/src/netcdf.cpp#L1125C85-L1125C90), which point to different sizes of memory on windows. So we won't be able to use TempestRemap on Windows after all.

juliasloan25:
## Purpose 
Run all tests (even those using TempestRemap) on windows

closes #422

## Update
ClimaCoreTempestRemap casts a `long *` to a `size_t *` in [this line](https://github.com/ClimateGlobalChange/tempestremap/blob/master/src/netcdf.cpp#L1125C85-L1125C90), which are different on windows. So we won't be able to use TempestRemap on Windows after all.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -10,97 +10,97 @@ gpu_broken = ClimaComms.device() isa ClimaComms.CUDADevice
 end
 
 # Shared ClimaLSM utilities tests
-@safetestset "Richards model implicit timestepping tests" begin
-    include("shared_utilities/implicit_timestepping/richards_model.jl")
-end
-@safetestset "Domains module tests" begin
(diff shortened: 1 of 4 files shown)


juliasloan25:
issue opened here: https://github.com/ClimateGlobalChange/tempestremap/issues/121

PR #426, opened 2023-12-15
ClimaLandSimulations package

AlexisRenchon:
## Purpose 

This PR adds a package `ClimaLandSimulations.jl` inside ClimaLSM, this is mostly a refactor of what we had in experiments/integrated/fluxnet. The goal is to offer simple methods for users (for example, fluxnet PIs) to run ClimaLSM at single site (and later global runs).

This PR also adds many new functionality, such as methods to generate dataframes (using DataFrames.jl) for fluxnet inputs and climalsm outputs, in SI units and in commonly used units. It also transition from Plots to Makie, to make plotting methods that generate publication quality plots comparing fluxnet observations and climalsm model.  

## To-do

- [x] Create a new repository, `ClimaLandSite` at the root of ClimaLSM repo, which is the typical format of a julia package (src folder, test folder, etc.)
- [x] Fluxnet inputs dataframe, SI units and commonly used units
- [x] ClimaLSM outputs dataframe
- [ ] Makie (CairoMakie, vector graphics plots) plotting functions -> WIP
    - [x] time series (WIP)
    - [x] fingerprint plot (WIP)
    - [x] diurnals (WIP)
    - [ ] seasonal
    - [ ] response curve (e.g., GPP vs. PAR, RECO vs SWC and T)
    - [ ] energy conservation (L + H = Rn - G)
    - [ ] water budget (cumulative precip and ET)
    - [ ] data quality (e.g., night turbulent CO2 and storage vs. u*)
    - [ ] wavelet analysis (coherence at all time scales + time lags)
- [ ] Call those functions for all fluxnet sites
- [x] Move old plotting (Plots.jl) from run_fluxnet.jl to old_plots.jl
- [ ] run_fluxnet as a function --> return sv, sv is empty?

## Content

See to do above

----


AlexisRenchon:
@kmdeck @Espeer5 Can you guys have a look at the [buildkite error](https://buildkite.com/clima/climalsm-ci/builds/2749#018c6ed1-9022-45e0-8cf0-dd15579dc8ae) and let me know

Espeer5:
> @kmdeck @Espeer5 Can you guys have a look at the [buildkite error](https://buildkite.com/clima/climalsm-ci/builds/2749#018c6ed1-9022-45e0-8cf0-dd15579dc8ae) and let me know what you think is happening? I am not sure...

It's a race condition between the "Makie" test and the normal Ozark test. They are running concurrently and both use the Artifact file for the Ozark data, but the normal Ozark script does `rm()` on the file when it finishes so that the one of the tests ends up trying to access it when it's already been deleted

AlexisRenchon:
> > @kmdeck @Espeer5 Can you guys have a look at the [buildkite error](https://buildkite.com/clima/climalsm-ci/builds/2749#018c6ed1-9022-45e0-8cf0-dd15579dc8ae) and let me know what you think is happening? I am not sure...
> 
> It's a race condition between the "Makie" test and the normal Ozark test. They are running concurrently and both use the Artifact file for the Ozark data, but the normal Ozark script does `rm()` on the file when it finishes so that the one of the tests ends up trying to access it when it's already been deleted (Or rather here it ended up trying to delete it when the other thread had already deleted it)
> 
> Ozark_conservation/Ozark don't experience this problem because the Ozark_conservation test doesn't `rm` the Artifact file. Not sure what to do about it here.

ugh. Any idea @kmdeck ?

kmdeck:
I think we could check if the file exists and if so, delete it, if not, `nothing`?

AlexisRenchon:
hmm I tried but got the same error
```
if isfile(joinpath(climalsm_dir, "experiments/integrated/fluxnet/$site_ID/Artifacts.toml"))
    rm(joinpath(climalsm_dir, "experiments/integrated/fluxnet/$site_ID/Artifacts.toml"))
else
    nothing
end
```
https://github.com/CliMA/ClimaLSM.jl/blob/ar/makie_plots/experiments/integrated/fluxnet/run_fluxnet.jl#L327

Espeer5:
> hmm I tried but got the same error
> 
> ```
> if isfile(joinpath(climalsm_dir, "experiments/integrated/fluxnet/$site_ID/Artifacts.toml"))
>     rm(joinpath(climalsm_dir, "experiments/integrated/fluxnet/$site_ID/Artifacts.toml"))
> else
>     nothing
> end
> ```
> 
> https://github.com/CliMA/ClimaLSM.jl/blob/ar/makie_plots/experiments/integrated/fluxnet/run_fluxnet.jl#L327

Different error this time actually... Before IOError at line 2 of buildkite_test, now in line 5. Instead of looking for `Artifacts.toml` now it can't find `timeseries.svg`. Fairly certain this is happening because there is no directory called `figures`.

Before when plotting we handled this by calling `mkdir` only if the desired output directory didn't already exist, see [here](https://github.com/CliMA/ClimaLSM.jl/pull/426/files#:~:text=if%20!isdir,end)

experiments/integrated/fluxnet/buildkite_test.jl
```diff
@@ -0,0 +1,13 @@
+ARGS = ["US-MOz"]
+include("run_fluxnet.jl")
+include("inputs_dataframe.jl")
+include("climalsm_output_dataframe.jl") 
+
+if isdir("figures")
```
Espeer5, line 6: I think the test runs from home directory of the repo. I would suggest `mkdir` of the whole path, using `joinpath`, to combine the path to the repo (gotten using `pkgdir`) and then combining with the full path from the repo to the figures directory, I think this would be `experiments/integrated/fluxnet/$(site_ID)/figures`

An example of this is [here](https://github.com/CliMA/ClimaLSM.jl/blob/c730511de8f04d99eed9add6d352193783a745a3/experiments/integrated/fluxnet/run_fluxnet.jl#L34)

AlexisRenchon: thanks! I'll try that

AlexisRenchon:
Ok it works now but those files .svg are a bit heavy (2 or 3 MB per file), I wonder if that's okay with our buildkite (do we care about file size?)

PR #369, opened 2023-10-30
Overhaul Experiments Script

Espeer5:
## Purpose 
 Uses a single script to run any FLUXNET site experiment and simplifies the process of adding a new FLUXNET site to run on. For each new site, we now just need to add a parameters file with all site specific variable values and a link to download the data from somewhere as a CSV.

## To-do

- [x] Implement the bug fix from #366
- [x] Include some data validation code, i.e check the necessary columns are included in the given data file
- [x] Fix bug in Harvard site?
- [x] For data used only for initial conditions/plotting, use more general values if data absent
- [x] For comparison plots - only plot comparison if data quality flag not "missing"
- [x] Add Vaira Ranch Site (with this PR, just add parameters!)

## Content

- Removes the old Ozark experiment
- Creates one script, `run_fluxnet_jl` which can be run on any FLUXNET site
- For each site which we want to run the model on, creates a directory named with the site ID, and a parameters file for each site ID needed
- Abstracts the plotting utility functions and data utility functions to new files
- Does all data comparison based on interpolation to fix previous bugs

----
- [x] I have read and checked the items on the review checklist.

PR #427, opened 2023-12-15
Restore Experiments Plots

Espeer5:
## Purpose 

Following #369, there are a couple issues left to be resolved by this PR, including restoring the ground heat flux plots to what they were before, fixing the labels for Harvard plots (plots when the data is missing), and fixing the start and end dates of MODIS LAI data fetching.

##TODO:

- [ ] View Buildkite plots for all sites to ensure that the plotting corrections are correct.

## Content

- Harvard plots relabelled to report model is being plotted
- MODIS LAI fetched based on the data date column (before was always 2005)
- Ground heat flux plot restored to what it was prior to being reverted by #369

----


AlexisRenchon (approved):
Thanks! If it passes CI and has 1 commit go ahead with merge

PR #428, opened 2023-12-18
streamline artifact downloading

### LINKED ISSUE #423: add download_artifacts function
We should download all data at beginning of simulation drivers. This will hopefully address the bucket albedo race condition when run with MPI, which has resurfaced in ClimaCoupler https://github.com/CliMA/ClimaCoupler.jl/issues/538

closed by #428

juliasloan25:
## Purpose 
We currently download artifacts in the code as we need them. This is causing race conditions in ClimaCoupler when we run the bucket with MPI.

I think a better approach would be to download all the artifacts at the beginning of a run during initialization, and then just access them as needed while running.

### Some potential problems
- We have a lot of artifacts, not all of which are needed for every run
  - Can we avoid downloading everything for every run and still have a general downloading workflow?
- We have two types of artifacts:
  1. data we compare results to (used for experiments, tutorials, tests)
  2. data we use in the model itself (in src, e.g. albedo data for the bucket)
  - Maybe these two need to be handled differently?
  - Idea: Add (i) to `artifacts/ `and download on buildkite. Add (ii) to `src/bucket/artifacts/` and call `download_bucket_artifacts` at the start of each bucket driver

### Questions I have
- We can download artifacts on buildkite like ClimaAtmos does [here](https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/pipeline.yml#L57), but then when/where do we download artifacts for local runs?

closes #423

## Content

----


Sbozzolo:
Can we close this in favor of using ClimaArtifacts?

juliasloan25:
> Can we close this in favor of using ClimaArtifacts?

yes!

juliasloan25:
Will be implemented using ClimaArtifacts instead - see docs [here](https://clima.github.io/ClimaUtilities.jl/dev/climaartifacts/)

PR #430, opened 2023-12-19
data tools

kmdeck:
## Purpose 
Changes to "datatools.jl" for readability and to ensure all driver fields are processed.
## Content
- Data with -9999 was not being removed; this is now
- Split between "absent", "incomplete", or "complete" now corresponds to branches in the `filter_column` function

Review checklist

I have:

----

### DIFF
--- a/experiments/integrated/fluxnet/data_tools.jl
+++ b/experiments/integrated/fluxnet/data_tools.jl
@@ -11,32 +11,54 @@ using StatsBase
 # Define the valid data statuses
 @enum DataStatus complete = 1 absent = 2 incomplete = 3
 
-function replace_missing_with_mean!(field, flag)
-    """Replace values indicated to be missing by a QC flag with the mean value
-    in the column"""
+"""
+    replace_poor_quality_with_mean!(field, flag)
+
+Replace values indicated to be poor quality by a fluxnet 
+QC `flag` (Array) with the mean value
+in the `field` (Array).
+
+This uses the Fluxnet convention of 0 = measured, 1 = good quality
+gap fill, 2 = medium quality, and 3 = poor quality, and replaces
+data with QC flag = 2,3. 
+"""
+function replace_poor_quality_with_mean!(field, flag)
     good_indices = (flag .== 0) .|| (flag .== 1)
     fill_value = mean(field[good_indices])
     field[.~good_indices] .= fill_value
     return field
 end
 
+"""
+    replace_replace_missing_with_mean_by_value!(field)
+
+Replace values indicated to be missing
+ with the mean value in the `field` (Array).
+
+This uses the Fluxnet convention of a value of -9999 indicating
+missing data.
+"""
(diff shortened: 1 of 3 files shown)


experiments/integrated/fluxnet/data_tools.jl
```diff
@@ -54,81 +76,84 @@ function check_column(file_matrix::Matrix, column_name::String)
...
-            @info "Warning: Data for $column_name $(
-            round(perc_missing, sigdigits=3))% missing. Filling with mean value"
-            replace_missing_with_mean!(col_dat, QC_col)
+    elseif status == complete # all data is present but some may be poor quality
+        # Check for a QC flag column and use it to clean data if possible
+        column_QC = column_name * "_QC"
+        QC_col, QCstatus = column_status(driver_data, column_QC)
+        # QC flag = [0,1] -> good data, [2,3] -> poorer data
+        num_poor = count(x -> x >= 2, QC_col)
+        if QCstatus != absent && num_poor > 0
```
Espeer5, line 120: For some of the sites (including the Harvard site which is why the buildkite fails) there are QC columns showing that the entire column is poor quality. Then when we try to fill poor quality values there aren't any good indices to use. We just need a case to catch that case and mark the column as absent.

Espeer5 (changes requested):

experiments/integrated/fluxnet/data_tools.jl
kmdeck: If LW_IN is all poor quality and we mark as absent, how do we run the sim?

Espeer5:
I see that all of the LW_IN data have QC flag 2. Not sure what we want the behavior to be in this case. We have a required driver column which does have data, but all of it is flagged as poor quality. 

For a required column would we rather have this warn the user that the column may be unreliable and continue the experiment? This way we can at least still complete the run.

experiments/integrated/fluxnet/data_tools.jl
Espeer5: Ah, before what I did was to mark the QC column as absent if it was all low quality instead of marking the data column as absent if the QC was all low quality to avoid this issue.

This just kind of ignores that all the data is low quality but I don't see too many options if the column is a required driver.

kmdeck:
> I see that all of the LW_IN data have QC flag 2. Not sure what we want the behavior to be in this case. We have a required driver column which does have data, but all of it is flagged as poor quality.
> 
> For a required column would we rather have this warn the user that the column may be unreliable and continue the experiment? This way we can at least still complete the run.

good idea! Ill try and make that change today

juliasloan25:
@kmdeck will this be merged soon? It would be nice to have the NCDatasets update in main for our next release of lsm so I can propagate it to the coupler :)

Espeer5 (approved):
All looks great to me! Looked at the plots on buildkite and they are looking as I would have expected.

PR #431, opened 2023-12-19
fix state FieldVec bug

### LINKED ISSUE #356: empty model components create FieldVector of Vector of Fields
## Describe the bug
In some of our model setups, we end up with a state that is a `FieldVector` containing other `FieldVectors` (this is okay), and also containing `Vectors` of `Fields` :exploding_head: (this is not okay).

Specifically, simulations using `CanopyModel` (`canopy_tutorial.jl, soil_canopy_tutorial.jl, ozark.jl`) fail when trying to update ClimaTimeSteppers from v0.7.7 to v0.7.8 (or any more recent version). We believe this is because `initialize_vars` sets empty model components to `FT[]` [here](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/shared_utilities/models.jl#L302). This is preventing us from using the latest ClimaTimeSteppers.

The change in CTS that causes this setup to break is that [v0.7.7](https://github.com/CliMA/ClimaTimeSteppers.jl/blob/v0.7.7/src/solvers/imex_ark.jl#L33) calls `similar(state)`, and [v0.7.8](https://github.com/CliMA/ClimaTimeSteppers.jl/blob/v0.7.8/src/solvers/imex_ark.jl#L35) calls `zero(state)`, which fails for this `FieldVector{Vector{Field}}` structure.

## Possible fix
Instead of returning `FT[]` from initialize_vars, return `nothing`. Then, where `initialize_vars` is called from (`initialize_prognostic`, `initialize_aux`), we can filter out these empty components when constructing the actual state or cache object.

## To Reproduce
```julia
using ClimaCore: Geometry, Meshes, Domains, Topologies, Spaces, Fields
using ClimaComms
using ClimaLSM.Domains: Point


FT = Float64
domain = Point(; z_sfc = FT(0.0))
space = domain.space.surface

field1 = Fields.ones(space)
field2 = Fields.zeros(space)
fieldvec = [field1, field2]
fv1 = Fields.FieldVector(; field1, field2)
fv2 = Fields.FieldVector(; fv1, fieldvec)

zero(fv1)
zero(fv2)
```

## To Do
- implement fix
- add test for undesired behavior

cc @kmdeck

juliasloan25:
## Purpose 
closes #356

Change how Y and p are created to make sure they always have the correct FieldVector structure.

This will allow us to use the most recent ClimaTimeSteppers :)

## To Do
- [ ] open issue re: allocations introduced by `filter_nt`

## Content
- [x] change `initialize_vars` to return `nothing` if component doesn't have prognostic/aux vars
- [x] add `filter_nt` function to remove key/value pairs containing `nothing` from a NamedTuple
- [x] call `filter_nt` from where `initialize_vars` is called for canopy models
- [x] add test to verify that `zero(Y)` can be called without error
- [x] remove ClimaTimeSteppers pin and use v0.7.15

## Notes/questions
- Is it okay to just use `filter_nt` in canopy case? `initialize_vars` change is made for all models so could require filtering for other model types too
- Canopy tests are kind of hard-coded to only check hydraulics - is this okay?
- We probably don't need to use `filter_nt` for `initialize_aux`, but it might be nice to be consistent between Y and p

----

### DIFF
--- a/src/shared_utilities/models.jl
+++ b/src/shared_utilities/models.jl
@@ -61,6 +61,10 @@ name(model::AbstractModel) =
    prognostic_vars(m::AbstractModel)
 
 Returns the prognostic variable symbols for the model in the form of a tuple.
+
+Note that this default suggests that a model has no prognostic variables,
+which is an invalid model setup. This function is meant to be extended for
+all models.
 """
 prognostic_vars(m::AbstractModel) = ()
 
@@ -70,6 +74,10 @@ prognostic_vars(m::AbstractModel) = ()
 Returns the domain names for the prognostic variables in the form of a tuple.
 
 Examples: (:surface, :surface, :subsurface).
+
+Note that this default suggests that a model has no prognostic variables,
+which is an invalid model setup. This function is meant to be extended for
+all models.
 """
 prognostic_domain_names(m::AbstractModel) = ()
 
@@ -87,15 +95,19 @@ coordinate point)
  length `k` at each coordinate point.
 
 Here, the coordinate points are those returned by coordinates(model).
+
+Note that this default suggests that a model has no prognostic variables,
+which is an invalid model setup. This function is meant to be extended for
+all models.
 """
 prognostic_types(m::AbstractModel) = ()
+
 """
    auxiliary_vars(m::AbstractModel)
 
 Returns the auxiliary variable symbols for the model in the form of a tuple.
 """
 auxiliary_vars(m::AbstractModel) = ()
-
 
 """
    auxiliary_domain_names(m::AbstractModel)
@@ -105,7 +117,6 @@ Returns the domain names for the auxiliary variables in the form of a tuple.
 Examples: (:surface, :surface, :subsurface).
 """
 auxiliary_domain_names(m::AbstractModel) = ()
-
 
 """
    auxiliary_types(m::AbstractModel{FT}) where {FT}
@@ -306,6 +317,11 @@ Adjustments to this - for example because different prognostic variables
 have different dimensions - require defining a new method.
 """
 function initialize_prognostic(model::AbstractModel{FT}, state) where {FT}
+    if length(prognostic_vars(model)) == 0
+        throw(
+            AssertionError("Model must have at least one prognostic variable."),
+        )
+    end
     state_nt = initialize_vars(
         prognostic_vars(model),
         prognostic_types(model),
@@ -355,7 +371,7 @@ end
 function initialize_vars(keys, types, domain_names, state, model_name)
     FT = eltype(state)
     if length(keys) == 0
-        return (; model_name => FT[])
+        return (; model_name => nothing)
     else
         zero_states = map(zip(types, domain_names)) do (T, D)
             zero_instance = ClimaCore.RecursiveApply.rzero(T)
--- a/src/standalone/Vegetation/Canopy.jl
+++ b/src/standalone/Vegetation/Canopy.jl
@@ -274,6 +274,32 @@ function auxiliary_types(canopy::CanopyModel)
 end
 
 """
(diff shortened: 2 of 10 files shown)


src/shared_utilities/models.jl
```diff
@@ -317,7 +317,7 @@ end
 function initialize_vars(keys, types, domain_names, state, model_name)
     FT = eltype(state)
     if length(keys) == 0
-        return (; model_name => FT[])
+        return (; model_name => nothing)
```
juliasloan25, line 320: With these changes, `nothing` is only filtered out for canopy models, but any model could potentially have `nothing` in `Y` or `p` since they all use this function. Is this okay, or should we call `filter_nt` in every case? It might be nice to avoid as long as we can, since it will add more allocations

src/standalone/Vegetation/Canopy.jl
```diff
@@ -321,7 +333,9 @@ function initialize_auxiliary(model::CanopyModel{FT}, coords) where {FT}
         submodel = getproperty(model, component)
         getproperty(initialize_auxiliary(submodel, coords), component)
     end
-    p = (; name(model) => NamedTuple{components}(p_state_list))
+    # `p_state_list` contains `nothing` for components with no auxiliary
+    #  variables, which we need to filter out before constructing `p`
+    p = (; name(model) => filter_nt(NamedTuple{components}(p_state_list)))
```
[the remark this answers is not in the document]
kmdeck: I agree that it would be nice to have consistency, even if we dont need the filtering in `p`

src/shared_utilities/models.jl
kmdeck: darn! and I guess we cant easily change `filter_nt` so that if there are no `nothing`s, it doesnt re-allocate? I kind of think we want to call it in every case to cover our bases...

kmdeck: Not sure if this is too much, but we could write a function which recursively gets the values of all keys in the NT. If none of them are `nothing` we can do nothing, but if some are, we `filter_nt`?

kmdeck (commented):
Given our current default for `initialize_vars`, which can return nothing as the value, I think we would need to add `filter_nt` into the default, even if only the canopy needs it right now. And for the canopy, if we filter Y, we should filter `p` for consistency.

That said, I think it would be great if we could only do all of this only for the Canopy. Perhaps we could change the default `initialize_prognostic` to error if a model has no prognostic variables? This could make sense because you cant run a standalone model with no prognostic variables. But, having no auxiliary parameters is fine (from a modeling point of view and from the bug point of view)

And then canopy specific methods could do the filtering for Y and p?

src/standalone/Vegetation/Canopy.jl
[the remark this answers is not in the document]
kmdeck: at least for the canopy!

src/standalone/Vegetation/Canopy.jl
```diff
@@ -273,6 +273,16 @@ function auxiliary_types(canopy::CanopyModel)
     return NamedTuple{components}(auxiliary_list)
 end
 
+function filter_nt(nt::NamedTuple)
```
kmdeck, line 276: maybe a short doc string explaining?

src/shared_utilities/models.jl
```diff
@@ -60,6 +60,10 @@ name(model::AbstractModel) =
    prognostic_vars(m::AbstractModel)
 
 Returns the prognostic variable symbols for the model in the form of a tuple.
+
+Note that this default suggests that a model has no prognostic variables,
+which is an invalid mode setup. This function is meant to be extended for
```
kmdeck, line 65: mode -> model

kmdeck (approved):
yay! LGTM, thank you Julia!

PR #436, opened 2023-12-21
add driver variables to cache

kmdeck:
## Purpose 
Add driver variables to the cache

## Content
1. Creates a function which specifies the driver variables based on driver type (`driver_cache`)
2. Creates a function which returns the drivers for a model (`get_drivers`)
3. Employs these functions in `add_drivers_to_cache` which is called in `initialize`

----

### DIFF
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -516,3 +516,73 @@ function vapor_pressure_deficit(T_air, P_air, q_air, thermo_params)
     )
     return es - ea
 end
+
+"""
+    initialize_drivers(r::PrescribedAtmosphere{FT}, coords) where {FT}
+
+Creates and returns a NamedTuple for the `PrescribedAtmosphere` driver, 
+with variables `P_liq`, `P_snow`, and air temperature `T`, pressure `P`,
+horizontal wind speed `u`, specific humidity `q`, and CO2 concentration
+`c_co2`.
+"""
+function initialize_drivers(a::PrescribedAtmosphere{FT}, coords) where {FT}
+    keys = (:P_liq, :P_snow, :T, :P, :u, :q, :c_co2)
+    types = ([FT for k in keys]...,)
+    domain_names = ([:surface for k in keys]...,)
+    model_name = :drivers
+    # intialize_vars packages the variables as a named tuple,
+    # as part of a named tuple with `model_name` as the key.
+    # Here we just want the variable named tuple itself
+    vars =
+        ClimaLSM.initialize_vars(keys, types, domain_names, coords, model_name)
+    return vars.drivers
+end
+
+"""
+    initialize_drivers(r::PrescribedRadiativeFluxes{FT}, coords) where {FT}
+
+Creates and returns a NamedTuple for the `PrescribedRadiativeFluxes` driver,
+ with variables `SW_d`, `LW_d`, and zenith angle `θ_s`.
+"""
+function initialize_drivers(r::PrescribedRadiativeFluxes{FT}, coords) where {FT}
+    keys = (:SW_d, :LW_d, :θs)
+    types = ([FT for k in keys]...,)
+    domain_names = ([:surface for k in keys]...,)
+    model_name = :drivers
+    # intialize_vars packages the variables as a named tuple,
+    # as part of a named tuple with `model_name` as the key.
+    # Here we just want the variable named tuple itself
+    vars =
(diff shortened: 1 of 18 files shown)


src/shared_utilities/models.jl
```diff
@@ -383,6 +385,26 @@ end
...
 Domains.coordinates(model::AbstractModel) = Domains.coordinates(model.domain)
 
+"""
+    add_drivers_to_cache(p::NamedTuple, model::AbstractModel, coords)
+
+"""
+function add_drivers_to_cache(p::NamedTuple, model::AbstractModel, coords)
+    (atmos, radiation) = get_drivers(model)
+    vars = initialize_drivers(atmos, radiation, coords)
+    return merge(p, vars)
```
juliasloan25, line 395: might be better to use `merge!` here and update `p` in place?

src/shared_utilities/drivers.jl
```diff
@@ -516,3 +516,51 @@ function vapor_pressure_deficit(T_air, P_air, q_air, thermo_params)
...
+    r::PrescribedRadiativeFluxes{FT},
+    coords,
+) where {FT}
+    keys = (:P_liq, :P_snow, :T, :P, :u, :q, :c_co2, :SW_d, :LW_d, :θs)
+    types = ([FT for k in keys]...,)
+    domain_names = ([:surface for k in keys]...,)
+    model_name = :drivers
+    vars =
+        ClimaLSM.initialize_vars(keys, types, domain_names, coords, model_name)
+    return vars
```
juliasloan25, line 536: This code is shared between both methods of this function. If we expect we'll add more methods, maybe we could move this to a different function (e.g. `get_driver_vars(keys, coords)`) that they all call, but if we just have two methods I think duplicating is okay

src/standalone/Soil/Soil.jl
```diff
@@ -88,7 +88,8 @@ import ClimaLSM:
     surface_emissivity,
     surface_air_density,
     surface_height,
-    surface_resistance
+    surface_resistance,
+    get_drivers
```
juliasloan25, line 92: do we need to extend `get_drivers` for the Soil model?

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -696,3 +696,14 @@ function ClimaLSM.surface_height(model::EnergyHydrology{FT}, Y, p) where {FT}
     )
     return z_sfc
 end
+
+function ClimaLSM.get_drivers(model::EnergyHydrology)
+    bc = model.boundary_conditions.top
+    if typeof(bc) <: AtmosDrivenFluxBC{
```
juliasloan25, line 702: maybe this should have an else case too in case we want to run e.g. without a `PrescribedAtmos`

juliasloan25 (approved):

src/shared_utilities/drivers.jl
kmdeck: I kept this part the same, but only because it's possible we might have different behavior in the future - e.g., if we ever want to run with more wavelength bands, we might want the type of `SW_d` to be a Tuple, with a value for each wavelength portion (instead of all FT). 

If on the other hand more methods are added in the future that call the same code, we can extract this into a function like you suggested then.

kmdeck: BTW: I tried to make it a bit more modular by instead defining `initialize_drivers(atmos, coords)` and `initialize_drivers(radiation, coords)`. Then `initalize_drivers(atmos, radiation, coords)` composes them by calling each on its own.

src/standalone/Soil/Soil.jl
kmdeck: I think since each model can store the drivers in a different spot (e.g. bucket.atmos, bucket.radiation, vs the EnergyHydrology model storing them in the boundary condition) we do need a different method for most models. but e.g. RichardsEquation does not need a new method since it never uses the drivers.

src/standalone/Soil/energy_hydrology.jl
kmdeck: good catch!!

src/shared_utilities/models.jl
kmdeck: Im not totally sure, but `merge!` is not defined for NamedTuples: https://docs.julialang.org/en/v1/base/collections/#Base.merge!
but,  dicts and named tuples seem closely related - collections of key-value pairs? - but unless we start working with Dicts instead of named tuples in `initialize` and only return the NT at the very end, i am not sure if we can update in place.

PR #439, opened 2024-01-02
add driver update callback and functions, tests

kmdeck:
## Purpose 
Adds the capability to update the drivers (atmospheric and radiative) with a callback. In the future, we will use this instead of updating each timestep as part of the tendency (i.e. instead of within`update_aux!`).

## To-do
- [x] double check logic of when we update the drivers and whether the times should be multiples of the timestep
- [x] Tutorial explaining drivers
- future PR: move all updating of drivers outside of update aux and use the callback instead.

## Content
1. Adds a callback which updates the p.driver fields at the times passed to the callback (`updateat`), which do not need to be a multiple of the timestep. Note that nothing restricts the user from wanting to update the drivers multiple times per step (even though this wouldnt make sense to do - the point of this is to update the drivers less frequently than the number of timesteps). 
2. Creates `update_driver!` functions for our different supported drivers: PrescribedAtmosphere, PrescribedRadiativeFluxes.

Review checklist

I have:

----

### DIFF
--- a/experiments/integrated/ozark/conservation/ozark_conservation.jl
+++ b/experiments/integrated/ozark/conservation/ozark_conservation.jl
@@ -313,8 +313,11 @@ for float_type in (Float32, Float64)
         t = Array{Float64}(undef, length(saveat)),
         saveval = Array{NamedTuple}(undef, length(saveat)),
     )
-    cb = ClimaLSM.NonInterpSavingCallback(sv, saveat)
+    saving_cb = ClimaLSM.NonInterpSavingCallback(sv, saveat)
 
+    updateat = deepcopy(saveat)
+    updatefunc = ClimaLSM.make_update_drivers(atmos, radiation)
+    driver_cb = ClimaLSM.DriverUpdateCallback(updateat, updatefunc)
     prob = SciMLBase.ODEProblem(
         CTS.ClimaODEFunction(
             T_exp! = exp_tendency!,
@@ -325,6 +328,7 @@ for float_type in (Float32, Float64)
         (t0, tf),
         p,
     )
+    cb = SciMLBase.CallbackSet(driver_cb, saving_cb)
     sol = SciMLBase.solve(
         prob,
         ode_algo;
@@ -341,6 +345,17 @@ for float_type in (Float32, Float64)
 
     # Plotting for Float64 simulation
     if FT == Float64
+        # Check that the driver variables stored in `p` are
+        # updated correctly - just check one from radiation,
+        # and one from atmos.
+        cache_θs = [parent(sv.saveval[k].drivers.θs)[1] for k in 1:length(sv.t)]
+        cache_Tair =
+            [parent(sv.saveval[k].drivers.T)[1] for k in 1:length(sv.t)]
+        @assert mean(
+            abs.(radiation.θs.(sv.t, radiation.ref_time) .- cache_θs),
+        ) < eps(FT)
+        @assert mean(abs.(atmos.T.(sv.t) .- cache_Tair)) < eps(FT)
+
         daily = sol.t[2:end] ./ 3600 ./ 24
         savedir =
             joinpath(climalsm_dir, "experiments/integrated/ozark/conservation")
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -17,7 +17,8 @@ export AbstractAtmosphericDrivers,
     liquid_precipitation,
     snow_precipitation,
     vapor_pressure_deficit,
-    displacement_height
+    displacement_height,
+    make_update_drivers
 
 """
      AbstractAtmosphericDrivers{FT <: AbstractFloat}
@@ -79,7 +80,7 @@ struct PrescribedAtmosphere{FT, LP, SP, TA, UA, QA, RA, CA, DT} <:
         ref_time,
         h::FT;
         gustiness = FT(1),
-        c_co2 = nothing,
+        c_co2 = (t) -> 4.2e-4,
     ) where {FT}
         args = (liquid_precip, snow_precip, T, u, q, P, c_co2, ref_time)
         return new{typeof(h), typeof.(args)...}(args..., h, gustiness)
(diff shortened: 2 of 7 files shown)


src/shared_utilities/drivers.jl
```diff
@@ -79,7 +80,7 @@ struct PrescribedAtmosphere{FT, LP, SP, TA, UA, QA, RA, CA, DT} <:
         ref_time,
         h::FT;
         gustiness = FT(1),
-        c_co2 = nothing,
+        c_co2 = (t) -> 4.2e-4,
```
kmdeck, line 83: instead of the default being `nothing`, I set it to return the current co2 concentration.

src/shared_utilities/drivers.jl
```diff
@@ -586,3 +587,70 @@ function initialize_drivers(
...
+
+"""
+    make_update_drivers(a::PrescribedAtmosphere{FT}) where {FT}
+
+Creates and returns a function which updates the driver variables
+in the case of a PrescribedAtmosphere at a point.
+"""
+function make_update_drivers(a::PrescribedAtmosphere{FT}) where {FT}
+    function update_drivers!(drivers, t)
+        drivers.P_liq .= FT(a.liquid_precip(t))
```
kmdeck, line 628: @Sbozzolo - what do you think about these update functions? drivers.field is a field, in the site level case it is a point field and being set to a single value. for global runs, we would have to modify this (maybe with another type than `PrescribedAtmosphere`?) to set the field in a different way. 

does this look flexible enough given how we will move away from calling a function at a point for computing the forcing at a particular time?

(these functions would be computed in a callback)

Sbozzolo: I think that we would have something like `set_field!(drivers.T, a, t)` and internally do what needs to be done. 

We will probably need to move away from calling fields directly with `(t)`.

src/shared_utilities/drivers.jl
```diff
@@ -586,3 +587,70 @@ function initialize_drivers(
...
+    make_update_drivers(r::PrescribedRadiativeFluxes{FT}) where {FT}
+
+Creates and returns a function which updates the driver variables
+in the case of a PrescribedRadiativeFluxes at a point.
+"""
+function make_update_drivers(r::PrescribedRadiativeFluxes{FT}) where {FT}
+    function update_drivers!(drivers, t)
+        drivers.SW_d .= FT(r.SW_d(t))
+        drivers.LW_d .= FT(r.LW_d(t))
+        if ~(r.θs isa Nothing)
```
juliasloan25, line 649: I think it's recommended to use `isnothing(var)` to check if something is nothing, but this is fine too

src/shared_utilities/utils.jl
```diff
@@ -194,6 +194,77 @@ function dss_helper!(
...
+
+"""
+    (affect!::DriverAffect)(integrator)
+
+This function is used by `DriverUpdateCallback` to perform the updating.
+"""
+function (affect!::DriverAffect)(integrator)
+    # If there are still update times in the queue and
+    # they are less than the current simulation time,
+    # cycle through until you reach the most `updateat` value
```
juliasloan25, line 218: typo - remove "most" ?

src/shared_utilities/utils.jl
```diff
@@ -194,6 +194,77 @@ function dss_helper!(
...
+        cond = curr_t <= integrator.t && curr_t > (integrator.t - integrator.dt)
+        if cond
+            # update all drivers to curr_t
+            affect!.updatefunc(integrator.p.drivers, curr_t)
+        end
+    end
+end
+
+"""
+    DriverUpdateCallback(updateat::Vector{FT})
```
juliasloan25, line 233: docstring doesn't match function

experiments/integrated/ozark/conservation/ozark_conservation.jl
```diff
@@ -313,8 +313,11 @@ for float_type in (Float32, Float64)
         t = Array{Float64}(undef, length(saveat)),
         saveval = Array{NamedTuple}(undef, length(saveat)),
     )
-    cb = ClimaLSM.NonInterpSavingCallback(sv, saveat)
+    saving_cb = ClimaLSM.NonInterpSavingCallback(sv, saveat)
 
+    updateat = deepcopy(saveat)
```
juliasloan25, line 318: maybe we could have a more general term for both `updateat` and `saveat` (`callbackat`?), so that the similarity between these two is emphasized and we maybe don't have to create a copy

test/shared_utilities/drivers.jl
```diff
@@ -27,3 +27,86 @@ FT = Float32
...
+    f = (t) -> 10.0
+    pa = ClimaLSM.PrescribedAtmosphere(f, f, f, f, f, f, f, FT(1);)
+    pr = ClimaLSM.PrescribedRadiativeFluxes(FT, f, f, f)
+    coords = (; surface = [1])
+    drivers = ClimaLSM.initialize_drivers(nothing, nothing, coords)
+    nothing_update! = ClimaLSM.make_update_drivers(nothing, nothing)
+    nothing_update!(drivers, 0.0)
+    @test drivers == (;)
+    drivers = ClimaLSM.initialize_drivers(pa, nothing, coords)
+    atmos_only_update! = ClimaLSM.make_update_drivers(pa, nothing)
```
juliasloan25, line 89: maybe we should also have a `rad_only_update!` where we pass in `ClimaLSM.make_update_drivers(nothing, pr)`

juliasloan25 (approved):
I left a few suggestions but this looks really good!

src/shared_utilities/drivers.jl
kmdeck: I changed this!

src/shared_utilities/utils.jl
kmdeck: good catch!

experiments/integrated/ozark/conservation/ozark_conservation.jl
kmdeck: I think we might need both, because in the callback affect! we popfirst! the vector of times and modify it. I think if we used the same vector, we would end up only executing one callback.

more generally, we might save more or less frequently than we update the forcing?

test/shared_utilities/drivers.jl
kmdeck: good call, im adding this!

experiments/integrated/ozark/conservation/ozark_conservation.jl
juliasloan25: good point! I agree

PR #443, opened 2024-01-09
add global prescribed atmos and radiation

### LINKED ISSUE #442: add global atmos and radiation types
We have a PrescribedAtmosphere type ([src/shared_utilities/drivers.jl](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/shared_utilities/drivers.jl#L50)) which can be used at a point. This holds things like atmospheric CO2, wind speed, pressure, etc. We need a new PrescribedAtmosphere type for running global simulations.

Same for radiation!

closed by #443 
part of #398

juliasloan25:
## Purpose 
Add types for global atmos and radiation. Write new methods to initialize and update them.

I'll implement this for atmos first and then duplicate for radiation.

closes #442

## To Do
- [x] PrescribedAtmosphere becomes PrescribedAtmosSite, PrescribedRadiation becomes PrescribedRadSite
- [ ] add new types PrescribedAtmosGlobal and PrescribedRadGlobal
  - these will be somewhat analogous to BulkAlbedoFunction and BulkAlbedoTemporal, where one stores FT-valued parameters, and the other stored Field-valued parameters over the entire globe
- [ ] extend `PrescribedDataTemporal/FileInfo/hdwrite_regridfile_rll_to_cgll` to work for files with multiple `varname`s
- [ ] implement methods of [initialize_drivers](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/shared_utilities/drivers.jl#L529-L540) to initialize global atmos and radiation types
- [ ] implement methods of [make_update_drivers](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/shared_utilities/drivers.jl#L626-L637) to update global atmos and radiation types
  - analogous to `next_albedo` (maybe rename to `update_albedo` for consistency)
    - uses read_data_fields!(::PrescribedDataTemporal, date, space)

## Notes
- `PrescribedAtmosGlobal` is largely following `BulkAlbedoTemporal`, but needs an array of `varname`s since the input file contains many variables we want to read in
  - extend `PrescribedDataTemporal` constructor to dispatch on `varname` type?
  - `varname` goes into `FileInfo` --> loop over weights in hdwrite_regridfile_rll_to_cgll (avoids having to recreate regridding objects)
  - can we assume all variables in same file will be on same grid, or do we need to create new weight matrix for each? --> assume same grid since output from same simulation
- should probably regrid only the dates we want to use in our simulation, rather than everything in file (as we've been doing)

----

### DIFF
--- a/docs/src/APIs/shared_utilities.md
+++ b/docs/src/APIs/shared_utilities.md
@@ -63,7 +63,7 @@ ClimaLSM.get_drivers
 
 ## Drivers
 ```@docs
-ClimaLSM.PrescribedAtmosphere
+ClimaLSM.PrescribedAtmosSite
 ClimaLSM.PrescribedRadiativeFluxes
 ClimaLSM.AbstractAtmosphericDrivers
 ClimaLSM.AbstractRadiativeDrivers
--- a/docs/tutorials/shared_utilities/driver_tutorial.jl
+++ b/docs/tutorials/shared_utilities/driver_tutorial.jl
@@ -9,11 +9,11 @@
 # # Types of forcing data
 
 # We currently support site-level simulations and have two site-level
-# driver types, `PrescribedAtmosphere` and `PrescribedRadiativeFluxes`.
+# driver types, `PrescribedAtmosSite` and `PrescribedRadiativeFluxes`.
 
-# The atmosphere driver stores the atmospheric state data as a 
+# The atmosphere driver stores the atmospheric state data as a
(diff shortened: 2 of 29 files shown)


Sbozzolo:
I think that there's some overlap with #451.

I've changed `PrescribedAtmosphere` and `PrescribedRadiation` to use the new `Interpolator`s. My assumption was that they would be used for global run as well.

juliasloan25:
will not be implemented - doing #451 instead

PR #447, opened 2024-01-12
update TagBot release notes

juliasloan25:
## Purpose 
Update TagBot.yml so the automatically generated release notes are more useful.

With these changes, developers should tag issues and/or PRs with the following labels so that their work appears in the release notes:
- `API` or `breaking change` for API/breaking changes
- `enhancement` for new features
- `documentation` for doc updates
- `bug` or `bugfix` for bug fixes

This will be really helpful for the coupler team to update for each new release :)

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -13,3 +13,61 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           ssh: ${{ secrets.DOCUMENTER_KEY }}
+          changelog: |
+            ## {{ package }} {{ version }}
+
+            {% if previous_release %}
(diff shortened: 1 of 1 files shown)


kmdeck (approved):
this looks very useful, thanks for implementing it! I guess we can quite verify that it runs in this PR, but we can adjust if needed next time we make a release

juliasloan25:
> this looks very useful, thanks for implementing it! I guess we can quite verify that it runs in this PR, but we can adjust if needed next time we make a release

Yeah it's a bit hard to test without actually making a release but at least it's a start. This is directly copied from Atmos's TagBot so I'm pretty confident in what it will produce, and we can update if we want something different :)

Issue #392, opened 2023-11-16
Add GPU compatibility in ClimaLSM


Sbozzolo:
# GPU compatibility 

As of 15 Jan 2024, the key missing elements to ensure that ClimaLSM is GPU compatible are file processing and interpolations. This issue is now turned into a Software Design Issue to describe how these will be implemented.

## The problem

`ClimaLSM` reads in and evaluates various data for boundary conditions, forcings, parameters, and so on. As of now, the data is mostly used to prescribe atmosphere, radiation, and snow/precipitation. For a single site, files are read, data is preprocessed (e.g., to deal with missing elements), and 1D splines are produced. Then, `ClimaLSM` is run assuming that we can simply call `atmos_T(t)` to obtain the temperature of the atmosphere at time `t`. 
This pattern works fine for single-site CPU runs, but has to be generalized for global/GPU runs. The main reasons for this are:
- Splines are not transpently GPU-compatible,
- The amount of data that has to be read for global runs can vastly exceed the system limit, and it is not possible to read everything before a run starts.

We want to support (with the same user/developer-interface for CPUs and GPUs)
1. analytic functions (of arbitrary dimensions),
2. 1D temporal data read from files (e.g., for FLUXNET data),
3. 2 + 1D spatiotemporal data read from files (e.g., ERA5 data).

We also want to support arbitrary interpolation schemes. In this way, we can start for the more difficult cases with the simple nearest neighbor interpolation (0th order), and later implement a linear interpolation (1st order), while leaving a pathway to implement higher order interpolation schemes in the future (if needed). This might also be the place where we implement more complex behaviors, e.g., using the same yearly data for a multi-year run.

## The proposed solution

The proposed solution consists in defining a single way to deal with all the aformentioned cases with multiple backends. We propose to define a new abstract class `AbstractInterpolator` and a new function to evaluate the interpolator. The function might look like
```julia
"""
    interpolate(interpolator::AbstractInterpolator, point)

Evaluate `interpolator` on the given `point`. 
"""
function interpolate(interpolator::AbstractInterpolator, point) end
``` 
Or (more likely)
```julia
"""
    interpolate!(destination, interpolator::AbstractInterpolator, point)

Evaluate `interpolator` on the given `point` and write the output to `destination`. 
"""
function interpolate!(destination, interpolator::AbstractInterpolator, point) end
``` 
[...]

PR #451, opened 2024-01-15
Remove splines, add `TimeVaryingInput`s

Sbozzolo:
This PR removes all the splines (which are not GPU compatibile) and substitutes them with `TimeVaryingInput`s (which are GPU compatible). This PR mostly follows what described in #392, but does not implement the 2/3D case.

This PR also introduces a new boundary condition for `Soil`, `AtmosFixedCO2FluxBC`, such that the `co2` is read from the atmos.

With this PR, experiments are GPU compatible﹡. At the moment, they are very slow.

﹡ = `root_distribution` doesn't compile on GPU because of inference problems with `FT` and `rooting_depth`. This can be worked around by making `rooting_depth` constant.


docs/src/Contributing.md
```diff
@@ -28,9 +28,3 @@ Currently a number of checks are run per commit for a given PR.
   are consistent and generate valid output.
 - `Tests` runs the file `test/runtests.jl`,  using `Pkg.test()`. These are a mix of
   unit tests and fast integration tests.
-
-We use [`bors`](https://bors.tech/) to manage merging PR's in the the `ClimaLSM` repo.
```
juliasloan25, line 32: could you also delete `docs/src/folder_structure.md` line 21? then we should finally be rid of bors

src/shared_utilities/interpolating_time_varying_input1d.jl
```diff
@@ -0,0 +1,138 @@
...
+    range::R
+end
+
+# GPU compatibility
+function Adapt.adapt_structure(to, itp::InterpolatingTimeVaryingInput1D)
+    times = Adapt.adapt_structure(to, itp.times)
+    vals = Adapt.adapt_structure(to, itp.vals)
+    method = Adapt.adapt_structure(to, itp.method)
+    range = Adapt.adapt_structure(to, itp.range)
+    # On a GPU, we have a "ClimaCore.DeviceSideContext"
```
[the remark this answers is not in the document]
Sbozzolo: This is called when you run a GPU kernel. It is needed to "adapt" a struct so that it can be moved to the GPU (e.g., possibly removing isbits objcets). In this case it is adapting the CuArrays.

src/shared_utilities/interpolating_time_varying_input1d.jl
```diff
@@ -0,0 +1,138 @@
...
+        method,
+        DeviceSideContext(),
+        range,
+    )
+end
+
+function evaluate!(destination, itp::InterpolatingTimeVaryingInput1D, time)
+    time in itp || error("TimeVaryingInput does not cover time $time")
+    if ClimaComms.device(itp.context) isa ClimaComms.CUDADevice
+        CUDA.@cuda evaluate!(parent(destination), itp, time, itp.method)
```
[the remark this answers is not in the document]
Sbozzolo: Yes. I wrote `evalulate!` in such a way that it could be run both on CPU and GPU. This switch controls where it is run.

src/shared_utilities/interpolating_time_varying_input1d.jl
```diff
@@ -0,0 +1,138 @@
...
+"""
+    in(time, itp::InterpolatingTimeVaryingInput1D)
+
+Check if the given `time` is in the range of definition for `itp`.
+"""
+function Base.in(time, itp::InterpolatingTimeVaryingInput1D)
+    return itp.range[1] <= time <= itp.range[2]
+end
+
+function evaluate!(
```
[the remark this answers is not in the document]
Sbozzolo: I added a few lines. I also realized that I don't need one of the edge cases conditions that I added because I check that time is in the range of validity.

src/shared_utilities/utils.jl
```diff
@@ -1,7 +1,7 @@
 import ClimaCore
 import SciMLBase
 
-export FTfromY
+export FTfromY, searchsortednearest
```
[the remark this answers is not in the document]
Sbozzolo: Right.

docs/tutorials/standalone/Canopy/canopy_tutorial.jl
```diff
@@ -314,7 +315,7 @@ end;
 # solution required, and the timestepping algorithm you are using.
 
 t0 = 0.0
-N_days = 365
+N_days = 364
```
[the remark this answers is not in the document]
Sbozzolo: I forgot to mention that. The original spline was being evaluate outside of the range of the input data, which is no longer allowed with the new interpolators.

src/shared_utilities/interpolating_time_varying_input1d.jl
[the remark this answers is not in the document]
Sbozzolo: Actually, I did miss an edge case.

lib/ClimaLandSimulations/src/utilities/met_drivers_FLUXNET.jl
```diff
@@ -198,9 +198,8 @@ function setup_drivers(site_ID)
     LAI_dt = Second(MODIS_LAI[2, 1] - MODIS_LAI[1, 1]).value
     LAI_seconds = FT.(0:LAI_dt:((length(MODIS_LAI[:, 1]) - 1) * LAI_dt))
 
-    # LAI spline for radiative transfer
-    LAIspline = Spline1D(LAI_seconds, MODIS_LAI[:, 2])
-    LAIfunction = (t) -> eltype(t)(LAIspline(t))
+    # LAI function for radiative transfer
+    LAIfunction = TimeVaryingInput(LAI_seconds, FT.(MODIS_LAI[:, 2]); context)
```
kmdeck, line 202: sorry that you had to do this twice! we are keeping the experiments folder around/as is while this is being built up...

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -455,6 +456,37 @@ struct SoilCO2FluxBC{F <: Function} <: AbstractSoilCO2BC
     bc::F
 end
 
+"""
+    AtmosFixedCO2FluxBC <: AbstractSoilCO2BC
+
+Set the CO2 concentration to the atmospheric one.
+"""
+struct AtmosFixedCO2FluxBC <: AbstractSoilCO2BC end
```
kmdeck, line 464: This is replacing a state boundary condition, so I think we might rename this to `AtmosCO2StateBC`?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -455,6 +456,37 @@ struct SoilCO2FluxBC{F <: Function} <: AbstractSoilCO2BC
...
+"""
+function ClimaLSM.boundary_flux(
+    bc::AtmosFixedCO2FluxBC,
+    boundary::ClimaLSM.AbstractBoundary,
+    Δz::ClimaCore.Fields.Field,
+    Y::ClimaCore.Fields.FieldVector,
+    p::NamedTuple,
+    t,
+)::ClimaCore.Fields.Field
+    return p.drivers.c_co2
```
kmdeck, line 486: This should return a flux, but currently it returns the CO2 concentration (state). I believe we need to convert it into a flux as we do for `SoilCO2StateBC` type, but getting the bc value from p.drivers.co2
https://github.com/CliMA/ClimaLSM.jl/blob/b10b56d3a36af014f73625b524fb506be3e1aafc/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl#L548

kmdeck: (as background, we support a number of different boundary condition types, but internally they are all converted into a flux before being applied to the divergence operator.

src/shared_utilities/general_utils.jl
```diff
@@ -0,0 +1,18 @@
+export searchsortednearest
```
kmdeck, line 1: not sure, should this just go into src/shared_utilities/utils.jl instead of making a new file? Perhaps a future PR can be to split utils.jl into more descriptive pieces. I can open an issue if we want to do this.

src/shared_utilities/general_utils.jl
```diff
@@ -0,0 +1,18 @@
...
+"""
+    searchsortednearest(a, x)
+
+Find the index corresponding to the nearest value to `x` in `a`.
+"""
+function searchsortednearest(a, x)
```
kmdeck, line 7: are we assuming that the data timeseries will be in order? That is a great assumption and requirement! not sure if it is obvious enough that we need to document it somewhere in a docstring?

src/shared_utilities/drivers.jl
```diff
@@ -41,15 +41,21 @@ Container for holding prescribed atmospheric drivers and other
 information needed for computing turbulent surface fluxes when
 driving land models in standalone mode.
 
-The arguments labeled with "function of time" are typically either of type
-`Function` or type `Spline1D` from the `Dierckx.jl` package.
-
 Since not all models require co2 concentration, the default for that
 is `nothing`.
```
kmdeck, line 45: this was something I forgot to update when I changed the default. we now just use the current atmos co2 concentration as the default. 

"The default CO2 concentration is a constant as a function of time, equal to 4.2e-4 mol/mol."

src/shared_utilities/drivers.jl
```diff
@@ -41,15 +41,21 @@ Container for holding prescribed atmospheric drivers and other
...
 $(DocStringExtensions.FIELDS)
 """
-struct PrescribedAtmosphere{FT, LP, SP, TA, UA, QA, RA, CA, DT} <:
-       AbstractAtmosphericDrivers{FT}
+struct PrescribedAtmosphere{
+    FT,
+    LP <: Union{Nothing, AbstractTimeVaryingInput},
```
kmdeck, line 50: When does the Nothing case arise?

src/shared_utilities/interpolating_time_varying_input1d.jl
```diff
@@ -0,0 +1,165 @@
...
+Perform linear interpolation between the two neighboring points.
+"""
+struct LinearInterpolation <: AbstractInterpolationMethod end
+
+"""
+    InterpolatingTimeVaryingInput1D
+
+The constructor for InterpolatingTimeVaryingInput1D is not supposed to be used directly, unless you
+know what you are doing. The constructor does not perform any check and does not take care of
+GPU compatibility. It is reasonability of Integrator() to do so.
```
kmdeck, line 22: could you clarify this? " It is reasonability of Integrator() to do so" What is Integrator()?

src/shared_utilities/interpolating_time_varying_input1d.jl
```diff
@@ -0,0 +1,165 @@
...
+
+"""
+struct InterpolatingTimeVaryingInput1D{
+    AA1 <: AbstractArray,
+    AA2 <: AbstractArray,
+    M <: AbstractInterpolationMethod,
+    CC <: ClimaComms.AbstractCommsContext,
+    R <: Tuple,
+} <: AbstractTimeVaryingInput
+    # AA1 and AA2 could be different because of different FTs
```
kmdeck, line 32: Could we change this to "AA1 and AA2 may have different float types, but they must be the same length, and we assume that they have been sorted to be monotonically increasing in time, without repeated values for the same timestamp." (or we can remove the last clause if we dont make that assumption."

I also wonder if this might go in the docstring instead of a code comment.

src/shared_utilities/interpolating_time_varying_input1d.jl
```diff
@@ -0,0 +1,165 @@
...
+    return nothing
+end
+
+function TimeVaryingInput(
+    times::AbstractArray,
+    vals::AbstractArray;
+    method = LinearInterpolation(),
+    context = ClimaComms.context(),
+)
+    issorted(times) || error("Can only interpolate with sorted times")
```
kmdeck, line 83: nice!

src/shared_utilities/TimeVaryingInputs.jl
```diff
@@ -0,0 +1,82 @@
...
+# Constructors will automatically identify which of the three implementations to use based
+# on the input data, and the existence of three concrete structs should be considered an
+# implementation detail.
+#
+# The three TimeVaryingInputs are:
+# - AnalyticTimeVaryingInput,
+# - InterpolatingTimeVaryingInput1D,
+# - InterpolatingTimeVaryingInput3D.
+#
+# Along side these TimeVaryingInputs, we also define InterpolationMethods that implement
```
kmdeck, line 24: very clear!

kmdeck: one more thing to note should be that the time variable needs to be in "simulation time", i.e. seconds since a reference datetime. This is because we evaluate! using `t` from the simulation.

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
Sbozzolo: Thanks! I thought I had picked the right one between the two.

test/shared_utilities/time_varying_inputs.jl
```diff
@@ -0,0 +1,104 @@
...
+            context,
+            method = TimeVaryingInputs.NearestNeighbor(),
+        )
+
+        # Test in
+        @test FT(3.0) in input
+        @test !(FT(-3.0) in input)
+
+        # Test with different types of spaces
+        for dest in (point_field, column_field)
```
kmdeck, line 72: in the column case, we are setting the same value at each level?

Im wondering if actually the TimeVaryingInput1D should be 0D, since it is handling a timeseries of a scalar. would 1D be a timeseries of a vector, and 2D a timeseries of an image?

src/shared_utilities/general_utils.jl
Sbozzolo: The `utils.jl` file has of ClimaLSM-specific functions, where this one is very general (hence `general_utils`). I don't want to import `utils.jl` in the `TimeVaryingInput` module  because it brings a lot of stuff I don't need.

kmdeck (commented):
Looks great and unifies a lot for us in addition to the GPU support, thanks very much for your hard work on this. 

A couple questions/commments left in the PR. Ill check buildkite next.

src/shared_utilities/drivers.jl
Sbozzolo: I found it here:
https://github.com/CliMA/ClimaLSM.jl/blob/058a66b39652cf72477fbcd174c5fbb4066d8703/test/shared_utilities/drivers.jl#L8-L17

But if we don't need it, I will remove support for `Nothing`

src/shared_utilities/analytic_time_varying_input.jl
```diff
@@ -0,0 +1,14 @@
+struct AnalyticTimeVaryingInput{F <: Function} <: AbstractTimeVaryingInput
```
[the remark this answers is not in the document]
Sbozzolo: Yes!

docs/tutorials/standalone/Canopy/canopy_tutorial.jl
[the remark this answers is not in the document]
Sbozzolo: @kmdeck is there any problems with this?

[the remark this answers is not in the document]
kmdeck: I dont think so. I would guess that maybe the data subset we saved to the CSV on box is short 1 day by accident, and we never caught that before now? but it's not a big deal to go for 364 days.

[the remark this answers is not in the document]
kmdeck: and only allowing simulations to be run where we have data is good!

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
Sbozzolo: Before this was set to return `(p, t) -> atmos_co2(t)` (in `Field` form) and we have that `evaluate!(p.drivers.c_co2, a.c_co2, t)` so isn't it the same, (with `a.c_co2(t) = atmos_co2(t)`)?

src/shared_utilities/general_utils.jl
Sbozzolo: Added

src/shared_utilities/drivers.jl
Sbozzolo: Added

src/shared_utilities/interpolating_time_varying_input1d.jl
Sbozzolo: Right, `Integrator` has nothing to do with this, it was a typo.

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
kmdeck: internally, the boundary_flux method for a StateBC turns a state into a flux (https://github.com/CliMA/ClimaLSM.jl/blob/b10b56d3a36af014f73625b524fb506be3e1aafc/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl#L549-L561) , but this method is just returning the state. so I think they are doing different things

src/shared_utilities/interpolating_time_varying_input1d.jl
Sbozzolo: I think repeated values are fine, but better to be more restrictive. I moved this to the docstring.

src/shared_utilities/drivers.jl
kmdeck: let me double check.

src/shared_utilities/TimeVaryingInputs.jl
Sbozzolo: Added.

src/shared_utilities/analytic_time_varying_input.jl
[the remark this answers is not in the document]
Sbozzolo: Added

src/shared_utilities/interpolating_time_varying_input1d.jl
Sbozzolo: Fixed.

docs/tutorials/standalone/Canopy/canopy_tutorial.jl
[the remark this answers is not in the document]
Sbozzolo: Okay, thanks!

src/shared_utilities/drivers.jl
kmdeck: we can keep it for now and Ill open an issue? no need to go down a rabbit hole for something tangential.

Sbozzolo: Open #461

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
Sbozzolo: Got it, I was looking at this function
https://github.com/CliMA/ClimaLSM.jl/blob/b10b56d3a36af014f73625b524fb506be3e1aafc/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl#L472-L482

Sbozzolo: I replaced the function with the correct `boundary_flux`.

Sbozzolo:
> Looks great and unifies a lot for us in addition to the GPU support, thanks very much for your hard work on this.
> 
> A couple questions/commments left in the PR. Ill check buildkite next.

I think I took care of everything but the question about splitting `utils.jl`.

src/shared_utilities/general_utils.jl
kmdeck: this is fine! we can clean up our utils (callback_utils.jl might be its own thing) later.

kmdeck: or we can move the function into TimeVaryingInput if that makes sense (if other parts of ClimaLSM do not need it). not a big deal either way I think

src/shared_utilities/interpolating_time_varying_input1d.jl
```diff
@@ -0,0 +1 @@
+
```
kmdeck, line 1: could remove this file?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -565,6 +566,47 @@ function ClimaLSM.boundary_flux(
...
+        p::NamedTuple,
+        t,
+    )::ClimaCore.Fields.Field
+
+Sets the flux using the CO2 concentration to the atmospheric one, as read from the cache.
+
+"""
+function ClimaLSM.boundary_flux(
+    bc::AtmosCO2StateBC,
+    boundary::ClimaLSM.BottomBoundary,
```
kmdeck, line 591: oh, one small thing. This is at the top boundary, in which case we need to extract the top level (not 1, below)

so the arg should be TopBoundary() and `p_len = Spaces.nlevels(axes(p.soilco2.D))`, so we can do
```
    D_c = Fields.Field(
        Fields.field_values(Fields.level(p.soilco2.D, p_len)),
        axes(Δz),
    )
```

for both C_c and D_c

src/shared_utilities/general_utils.jl
Sbozzolo: I will keep it there for the time being. I think it is a useful function to have around.

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
Sbozzolo: Thanks! I copied the wrong block!

src/shared_utilities/interpolating_time_varying_input1d.jl
Sbozzolo: Thanks!

PR #452, opened 2024-01-16
accommodate files containing multiple vars

juliasloan25:
## Purpose 
Currently, our file reading and regridding works on files that contain one variable. For global runs, we will have input data for multiple variables, so we should extend our infrastructure to work for that case.

## To-do
- [x] each `PrescribedDataTemporal` object contains multiple `FileState` objects, not just one
  - Dict mapping `varname` to `FileState`
- [x] update FileReader functions:
  - PrescribedDataTemporal
  - FileInfo
  - FileState
  - read_data_fields!
  - interpolate_data (renamed to get_data_at_date)
- [x] update Regridder functions:
  - hdwrite_regridfile_rll_to_cgll
  - remove regrid_netcdf_to_field (was only used in `set_initial_parameter_field` for `BulkAlbedoStatic` - see last bullet point)
- [x] remove regridder tests (now empty)
- [x] add test for file with multiple varnames (use file that's smaller than ERA)
- [x] unify PrescribedDataTemporal and PrescribedDataStatic interfaces
  - before, temporal constructor downloaded data and called `hdwrite_regridfile_rll_to_cgll`, and static constructor just downloaded data
    - now both constructors call `hdwrite_regridfile_rll_to_cgll` so all data is regridded after PrescribedData objects are created
- [x] unify `set_initial_parameter_field`
  - before, `PrescribedDataTemporal` called `read_data_fields!` and `interpolate_data` (now `get_data_at_date`), and `PrescribedDataStatic` called `regrid_netcdf_to_field`
  - now, temporal calls `read_data_fields!` and `get_data_at_date`, and static calls `get_data_at_date`
  - can remove `regrid_netcdf_to_field` function now as unused

----


src/shared_utilities/Regridder.jl
```diff
@@ -187,9 +208,6 @@ function hdwrite_regridfile_rll_to_cgll(
     topology = ClimaCore.Topologies.Topology2D(
         cpu_context,
         ClimaCore.Spaces.topology(space).mesh,
-        ClimaCore.Topologies.spacefillingcurve(
```
juliasloan25, line 190: Using `spacefillingcurve` produced a different topology, so `space_undistributed != space`, even in the undistributed case. Removing this call makes `space_undistributed == space` in the undistributed case

src/standalone/Bucket/artifacts/artifacts.jl
```diff
@@ -49,3 +52,51 @@ function bareground_albedo_dataset_path()
     path = AW.get_data_folder(bareground_albedo_dataset)
     return joinpath(path, "bareground_albedo.nc")
 end
+
+"""
+    era5_t2m_sp_u10n_dataset_path()
+
```
kmdeck, line 58: should we note that these are used only in testing?

src/shared_utilities/FileReader.jl
```diff
@@ -392,32 +440,78 @@ return the same value unless `date_idx` is modified elsewhere in between.
...
+        nothing,
+        varname,
+        comms_ctx,
+    )
+    Regridder.swap_space!(field, space)
+    return nans_to_zero.(field)
+end
+
+"""
+    get_data_at_date(
```
kmdeck, line 484: not in this PR, but is the plan that this will become `evaluate!`? and PrescribedDataTemporal will contain a TimeVaryingInput type?

src/shared_utilities/FileReader.jl
```diff
@@ -132,42 +133,62 @@ struct SimInfo{D}
...
-    PrescribedDataStatic(
+    PrescribedDataStatic{FT}(
         get_infile::Function,
         regrid_dirpath::String,
-        varname::String,
-        commx_ctx::ClimaComms.AbstractCommsContext,
-    )
+        varnames::Vector{String},
+        surface_space::Spaces.AbstractSpace,
+        commx_ctx::ClimaComms.AbstractCommsContext;
```
kmdeck, line 141: update docstring for different kwargs (comms not needed?)

src/shared_utilities/FileReader.jl
```diff
@@ -254,20 +279,31 @@ function PrescribedDataTemporal{FT}(
...
+        file_states[varname] =
+            FileState(deepcopy(data_fields), copy(date_idx0), segment_length)
+    end
     sim_info = SimInfo(date_ref, t_start)
 
-    args = (file_info, file_state, sim_info)
-    return PrescribedDataTemporal{typeof.(args)...}(args...)
+    args = (file_info, file_states, sim_info)
+    type_args =
+        (typeof(file_info), typeof(first(file_states)[2]), typeof(sim_info))
```
kmdeck, line 297: maybe a comment explaining why we have `first` and `[2]`?

src/shared_utilities/FileReader.jl
```diff
@@ -289,89 +325,100 @@ function read_data_fields!(
...
+            file_state = pd_file_states[varname]
+
+            # Increment `date_idx` to use next date
+            date_idx = file_state.date_idx[1] += Int(1)
+            # Time between consecutive dates being stored gives `segment_length`
+            file_state.segment_length .=
+                (all_dates[Int(date_idx + 1)] - all_dates[Int(date_idx)]).value
+
+            # Read in data fields at both dates
+            file_state.data_fields[1] .= deepcopy(
```
kmdeck, line 396: did we need the deep copy here still?

src/shared_utilities/FileReader.jl
```diff
@@ -289,89 +325,100 @@ function read_data_fields!(
...
+                        regrid_dirpath,
+                        outfile_root,
+                        all_dates[Int(date_idx)],
+                        varname,
+                        comms_ctx,
+                    ),
+                    space,
+                ),
+            )
+            file_state.data_fields[2] .= deepcopy(
```
kmdeck, line 408: and here

src/shared_utilities/FileReader.jl
```diff
@@ -392,32 +440,78 @@ return the same value unless `date_idx` is modified elsewhere in between.
...
+
+    comms_ctx = ClimaComms.context(space)
+    field = Regridder.read_from_hdf5(
+        regrid_dirpath,
+        outfile_root,
+        nothing,
+        varname,
+        comms_ctx,
+    )
+    Regridder.swap_space!(field, space)
```
kmdeck, line 479: Im a little confused still on why we need this. isnt the regridded field created for the `space` already?

src/shared_utilities/FileReader.jl
```diff
@@ -392,32 +440,78 @@ return the same value unless `date_idx` is modified elsewhere in between.
...
+    return nans_to_zero.(field)
```
kmdeck, line 480: Probably longer term this shouldnt be in these functions, and we should assume the input data has been cleaned and is ready for use already.

kmdeck: (I think this specific replacement may be correct for albedo, for example, but not other fields)

src/shared_utilities/Regridder.jl
kmdeck: is it worth leaving this as a comment in the code?

src/standalone/Bucket/artifacts/artifacts.jl
kmdeck: should these live in `Bucket` or are they more general?

test/shared_utilities/file_reader.jl
```diff
@@ -113,51 +245,68 @@ if !Sys.iswindows()
...
-            "",             # varname
+            varnames,       # varnames
             "",             # outfile_root
             dummy_dates,    # all_dates
             date_idx0,      # date_idx0
         )
-        file_state = FileReader.FileState(
+        file_state1 = FileReader.FileState(
+            data_fields,        # data_fields
+            copy(date_idx0),    # date_idx
```
kmdeck, line 262: I guess date_idx and segment should be the same for each variable's FileState, so there is some duplicate into here. but it may be easier to store them for each? because then FileState is self contained.

kmdeck (commented):
I think this looks good! Added functionality plus removal of some code. I left a number of small comments

I am mainly wondering about how this will eventually tie in with TimeVaryingInput, since there are now interpolation and update (evaluate!) functions defined that do similar things to interpol and get_data_at_date. For static fields, we can discuss tomorrow!

src/shared_utilities/Regridder.jl
juliasloan25: good idea!

src/standalone/Bucket/artifacts/artifacts.jl
juliasloan25: Yes we should note that they're for testing. They're general enough to be used beyond just the bucket. What do you think about moving them to `test/`?

src/shared_utilities/FileReader.jl
```diff
@@ -254,20 +279,31 @@ function PrescribedDataTemporal{FT}(
...
         outfile_root,
         all_dates,
         date_idx0,
     )
-    file_state = FileState(data_fields, copy(date_idx0), segment_length)
+    file_states = Dict{String, FileState{typeof(data_fields)}}()
+    for i in 1:length(varnames)
+        varname = varnames[i]
+        file_states[varname] =
+            FileState(deepcopy(data_fields), copy(date_idx0), segment_length)
```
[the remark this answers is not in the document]
juliasloan25: `copy` isn't defined for tuples:

```
julia> tu = (1, 2, 3)
(1, 2, 3)

julia> typeof(tu)
Tuple{Int64, Int64, Int64}

julia> tu2 = copy(tu)
ERROR: MethodError: no method matching copy(::Tuple{Int64, Int64, Int64})

Closest candidates are:
  copy(::LinearAlgebra.Transpose{Bool, BitMatrix})
   @ LinearAlgebra ~/.julia/juliaup/julia-1.10.0+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/LinearAlgebra/src/bitarray.jl:240
  copy(::BitSet)
   @ Base bitset.jl:44
  copy(::Random.DSFMT.GF2X)
   @ Random ~/.julia/juliaup/julia-1.10.0+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Random/src/DSFMT.jl:110
  ...

Stacktrace:
 [1] top-level scope
   @ REPL[3]:1
```

or do you mean I can broadcast `copy` over the tuple?

src/shared_utilities/FileReader.jl
juliasloan25: I think something like that! I'm not exactly sure what the plan is for the interface, but that would make sense to me

src/shared_utilities/FileReader.jl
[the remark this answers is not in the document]
Sbozzolo: Ah no! Right! Tuples are immutable, you don't even need `copy`. Just pass them.

Sbozzolo:
Can you please describe the flow of the code and the interaction between the different structs somewhere?

src/shared_utilities/FileReader.jl
juliasloan25: good point! I was able to remove `swap_space!` altogether

src/shared_utilities/FileReader.jl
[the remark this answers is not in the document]
juliasloan25: We are modifying elements of the tuple (e.g. `data_fields[1]`, `data_fields[2]`). When I created them without `deepcopy`, the data fields for each variable were overwriting each other (e.g. `file_states[varname1].data_fields[1] == file_states[varname2].data_fields[1]` was true, which isn't what we want).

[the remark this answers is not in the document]
Sbozzolo: I don't think this should be possible:
```julia
julia> a = tuple(1, 2)
(1, 2)

julia> a[1] = 3
ERROR: MethodError: no method matching setindex!(::Tuple{Int64, Int64}, ::Int64, ::Int64)
```

Tuple should be immutable. Are converting them to Vectors somewhere?

docs/src/APIs/Regridder.md
```diff
@@ -7,5 +7,6 @@ CurrentModule = ClimaLand.Regridder
 ## Functions
 
 ```@docs
-ClimaLand.Regridder.regrid_netcdf_to_field
+ClimaLand.Regridder.hdwrite_regridfile_rll_to_cgll
+ClimaLand.Regridder.swap_space!
```
Sbozzolo, line 11: If you removed `swap_space!`, you can remove this line too.

src/shared_utilities/Regridder.jl
```diff
@@ -8,13 +8,11 @@ using Dates
 using JLD2
 using DocStringExtensions
 
-export regrid_netcdf_to_field
-
-nans_to_zero(v::T) where {T} = isnan(v) ? T(0) : v
-
+export swap_space!, hdwrite_regridfile_rll_to_cgll
```
Sbozzolo, line 11: Same here.

src/shared_utilities/Regridder.jl
```diff
@@ -243,18 +261,10 @@ function hdwrite_regridfile_rll_to_cgll(
             data_dates = strdate_to_datetime.(string.(Array(ds["date"])))
         else
             @warn "No dates available in file $datafile_rll"
-            data_dates = [Dates.DateTime(0)]
+            data_dates = [nothing]
```
Sbozzolo, line 264: This might lead to type instability.

test/shared_utilities/file_reader.jl
juliasloan25: Hmm yeah good point. It is really nice to have each `file_state` self-contained - we take advantage of this in `get_data_at_date` and it makes it easy to loop over `varname, file_state` pairs in `read_data_fields!`

Sbozzolo (commented):
I think it would be nice to check some of the implicit assumptions that you make here. For example, that all variables are defined on the same dates, that they all have the same floating point type, and so on.

juliasloan25:
> I think it would be nice to check some of the implicit assumptions that you make here. For example, that all variables are defined on the same dates, that they all have the same floating point type, and so on.

I added checks for these in `hdwrite_regridfile_rll_to_cgll`, where we open the input data file

src/shared_utilities/FileReader.jl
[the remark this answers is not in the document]
juliasloan25: When I print the type of `data_fields` in `read_data_fields!` right before its fields are updated ([here](https://github.com/CliMA/ClimaLand.jl/blob/main/src/shared_utilities/FileReader.jl#L306)), its type is
> typeof(file_state.data_fields) = Tuple{ClimaCore.Fields.Field{...}, ClimaCore.Fields.Field{...}}

I tried recreating this with a simple tuple of Fields but it fails with the tuple `MethodError`. I'm not sure how it's working right now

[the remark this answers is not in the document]
Sbozzolo: Can you paste when you tried and the full error?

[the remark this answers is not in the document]
juliasloan25: > using ClimaCore: Domains, Meshes, Topologies, Fields, Spaces
\# set up space
FT = Float64
radius = FT(6731e3)
Nq = 4
domain_sphere = Domains.SphereDomain(radius)
mesh = Meshes.EquiangularCubedSphere(domain_sphere, 4)
topology = Topologies.Topology2D(mesh)
quad = Spaces.Quadratures.GLL{Nq}()
surface_space_t = Spaces.SpectralElementSpace2D(topology, quad)
\# define data fields
data_field1 = ones(surface_space_t)
data_field2 = ones(surface_space_t) .* FT(2)
\# create tuple of data fields
data_fields = data_field1, data_field2
@show data_fields
\# try to update one field of tuple, but it fails
\# ERROR: MethodError: no method matching setindex!(::Tuple{…}, ::ClimaCore.Fields.Field{…}, ::Int64)
data_fields[1] = zeros(surface_space_t)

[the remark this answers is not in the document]
Sbozzolo: This is expected. You cannot change tuples. You have the same error with any type.
```julia
julia> a = tuple(1, 2)
(1, 2)

julia> a[1] = 3
ERROR: MethodError: no method matching setindex!(::Tuple{Int64, Int64}, ::Int64, ::Int64)
```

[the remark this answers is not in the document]
Sbozzolo: What you are doing in https://github.com/CliMA/ClimaLand.jl/blob/main/src/shared_utilities/FileReader.jl#L306 is modifying the value in place (`.=` vs `=`)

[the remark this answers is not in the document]
juliasloan25: I see. So that's why the values get overwritten if I don't use `deepcopy`

Sbozzolo (approved):

Issue #449, opened 2024-01-13
move coupling functions to ClimaLSM


juliasloan25:
**Is your feature request related to a problem? Please describe.**
Currently, the coupler implements quite a few functions that adapt the ClimaLSM internals to suit the coupler interface. It would be nice to move these into ClimaLSM so code that accesses ClimaLSM internals is more isolated to one place. This could also reduce code duplication, especially following the recent driver changes.

For example, [this](https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/AMIP/components/land/bucket_init.jl#L154-L157) `snow_precipitation` method in ClimaCoupler is identical to [this](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/shared_utilities/drivers.jl#L357-L359) method in ClimaLSM, except for the type dispatch.

**Describe the solution you'd like**
For cases like `snow_precipitation` described above, generalize the ClimaLSM version of the function, and remove the method in ClimaCoupler. Potentially also move other functions from ClimaCoupler's `experiments/AMIP/components/land/` to ClimaLSM.

PR #453, opened 2024-01-17
Move driver interface code from the Coupler to ClimaLSM

kmdeck:
## Purpose 
Add Coupled Driver types to ClimaLSM (currently they are defined in ClimaCoupler). There are also methods that are defined in ClimaCoupler for these types, which this PR moves into ClimaLSM.

Lastly, we unify use of `turbulent_fluxes` and `surface_air_density` between standalone models in ClimaLSM here, which slightly changes how these are stored in the Bucket model cache.

See issue #449 

## To-do
[] unit tests

## Content
I think self-explanatory from the description of purpose. some variables are present only in prescribed vs coupled runs, but ideally we can handle that with dispatch off of the Atmos type. (rho_sfc, e.g.)

Note that the `PrescribedAtmos` etc driver types have methods for `update_drivers!` which are called during callbacks. In coupled runs, we do not need to do any updating on the ClimaLSM side, so these methods do not exist and we would not set a callback.

Review checklist

I have:

----

### DIFF
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -8,7 +8,9 @@ import ..Parameters as LSMP
 export AbstractAtmosphericDrivers,
     AbstractRadiativeDrivers,
     PrescribedAtmosphere,
+    CoupledAtmosphere,
     PrescribedRadiativeFluxes,
+    CoupledRadiativeFluxes,
     compute_ρ_sfc,
     construct_atmos_ts,
     turbulent_fluxes,
@@ -95,6 +97,20 @@ struct PrescribedAtmosphere{
         return new{typeof(h), typeof.(args)...}(args..., h, gustiness)
     end
 end
+
+"""
+    CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT}
+
+To be used when coupling to an atmosphere model.
+"""
+struct CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT} end
+
+"""
+    CoupledAtmosphere{FT} <: AbstractAtmosphericDrivers{FT}
+
+To be used when coupling to an atmosphere model.
+"""
+struct CoupledAtmosphere{FT} <: AbstractAtmosphericDrivers{FT} end
 
 """
     compute_ρ_sfc(thermo_params, ts_in, T_sfc)
@@ -297,6 +313,33 @@ function turbulent_fluxes_at_a_point(
 end
 
 """
+    surface_fluxes(atmos::CoupledAtmosphere,
+                    model::AbstractModel,
+                    Y,
+                    p,
+                    t)
+
+Computes the turbulent surface fluxes terms at the ground for a coupled simulation.
+"""
+function ClimaLSM.turbulent_fluxes(
+    atmos::CoupledAtmosphere,
+    model::AbstractModel,
+    Y,
+    p,
+    t,
+)
+    # coupler has done its thing behind the scenes already
+    model_name = ClimaLSM.name(model)
+    model_cache = getproperty(p, model_name)
+    return (
+        lhf = model_cache.turbulent_fluxes.lhf,
+        shf = model_cache.turbulent_fluxes.shf,
+        vapor_flux = model_cache.turbulent_fluxes.vapor_flux,
+    )
+end
+
+
+"""
     PrescribedRadiativeFluxes{FT, SW, LW, DT, T} <: AbstractRadiativeDrivers{FT}
 
 Container for the prescribed radiation functions needed to drive land models in standalone mode.
@@ -337,12 +380,12 @@ longwave and shortwave radiation.
 This returns an energy flux.
 """
 function net_radiation(
-    radiation::PrescribedRadiativeFluxes{FT},
-    model::AbstractModel{FT},
+    radiation::PrescribedRadiativeFluxes,
+    model::AbstractModel,
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
     t,
-) where {FT}
+)
     LW_d = p.drivers.LW_d
     SW_d = p.drivers.SW_d
     earth_param_set = model.parameters.earth_param_set
@@ -358,20 +401,44 @@ function net_radiation(
 end
 
 """
-    liquid_precipitation(atmos::PrescribedAtmosphere, p, t)
+    net_radiation(radiation::CoupledRadiativeFluxes,
+                  model::AbstractModel,
+                  Y,
+                  p,
+                  t)
+
+Computes the net radiative flux at the ground for a coupled simulation.
+Your model cache must contain the field `R_n`.
+"""
+function ClimaLSM.net_radiation(
+    radiation::CoupledRadiativeFluxes,
+    model::AbstractModel,
+    Y::ClimaCore.Fields.FieldVector,
+    p::NamedTuple,
+    t,
+)
+    # coupler has done its thing behind the scenes already
+    model_name = ClimaLSM.name(model)
+    model_cache = getproperty(p, model_name)
+    return model_cache.R_n
+end
+
+
+"""
+    liquid_precipitation(atmos::AbstractAtmosphericDrivers, p, t)
 
 Returns the liquid precipitation (m/s) at the surface.
 """
-function liquid_precipitation(atmos::PrescribedAtmosphere{FT}, p, t) where {FT}
+function liquid_precipitation(atmos::AbstractAtmosphericDrivers, p, t)
     return p.drivers.P_liq
 end
 
 """
-    snow_precipitation(atmos::PrescribedAtmosphere, p, t)
+    snow_precipitation(atmos::AbstractAtmosphericDrivers, p, t)
 
 Returns the precipitation in snow (m of liquid water/s) at the surface.
 """
-function snow_precipitation(atmos::PrescribedAtmosphere{FT}, p, t) where {FT}
+function snow_precipitation(atmos::AbstractAtmosphericDrivers, p, t)
     return p.drivers.P_snow
 end
 
@@ -409,7 +476,7 @@ end
 
 """
     surface_air_density(
-                        atmos::AbstractAtmosphericDrivers,
+                        atmos::PrescribedAtmosphere,
                         model::AbstractModel,
                         Y,
                         p,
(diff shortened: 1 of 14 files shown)


src/shared_utilities/drivers.jl
```diff
@@ -87,6 +89,21 @@ struct PrescribedAtmosphere{FT, LP, SP, TA, UA, QA, RA, CA, DT} <:
     end
 end
 
+"""
+    CoupledGlobalRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT}
+
+To be used when coupling to an atmosphere model; internally, used for
```
kmdeck, line 95: update

src/shared_utilities/drivers.jl
```diff
@@ -281,6 +298,34 @@ function turbulent_fluxes_at_a_point(
...
+    t,
+)
+    # coupler has done its thing behind the scenes already
+    model_name = ClimaLSM.name(model)
+    model_cache = getproperty(p, model_name)
+    return (
+        lhf = model_cache.turbulent_fluxes.lhf,
+        shf = model_cache.turbulent_fluxes.shf,
+        vapor_flux = model_cache.turbulent_fluxes.vapor_flux,
+        r_ae = ClimaCore.Fields.similar(model_cache.turbulent_fluxes.lhf),
```
kmdeck, line 324: not all models store this, and, in particular, they need to store it only in non-coupled cases. clean this up!

src/shared_utilities/drivers.jl
```diff
@@ -341,20 +386,43 @@ function net_radiation(
...
 
 """
-    liquid_precipitation(atmos::PrescribedAtmosphere, p, t)
+    net_radiation(radiation::CoupledGlobalRadiativeFluxes,
+                  model::AbstractModel,
+                  Y,
+                  p,
+                  t)
+
+Computes the net radiative flux at the ground for a coupled simulation.
```
kmdeck, line 395: add doc string stating what the cache must contain

src/shared_utilities/drivers.jl
```diff
@@ -400,26 +467,55 @@ end
...
+    ClimaLSM.surface_air_density(
+                    atmos::CoupledGlobalAtmosphere,
+                    model::AbstractModel,
+                    Y,
+                    p,
+                    _...,
+                )
+Returns the air density at the surface in the case of a coupled simulation.
+
+This requires the field `ρ_sfc` to be present in the cache `p` under the name
```
kmdeck, line 505: ideally we should be able to handle this with dispatch when setting up individual model cache variables.

src/shared_utilities/drivers.jl
```diff
@@ -8,7 +8,9 @@ import ..Parameters as LSMP
 export AbstractAtmosphericDrivers,
     AbstractRadiativeDrivers,
     PrescribedAtmosphere,
+    CoupledGlobalAtmosphere,
```
kmdeck, line 11: After the discussion today, these would be renamed to CoupledAtmosphere, etc

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -262,7 +262,7 @@ boundary_var_domain_names(bc::AtmosDrivenFluxBC, ::ClimaLSM.TopBoundary) =
 """
     boundary_var_types(
```
juliasloan25, line 263: do we need to also implement `boundary_var_types` methods for the CoupledAtmosphere case?

juliasloan25 (approved):
This looks good to me!!

Maybe we want to hold off on merging until we've had a chance to think about whether we want to include the coupled driver case in the Interpolators work? @Sbozzolo

kmdeck:
> This looks good to me!!
> 
> Maybe we want to hold off on merging until we've had a chance to think about whether we want to include the coupled driver case in the Interpolators work? @Sbozzolo

for sure, we can hold off to be sure!

 My 2c is that we dont need any interpolation for the forcing variables (atmos/radiation) in coupled simulations, so it might be confusing to have them be related to each other conceptually. We are not doing any `update_drivers!` or interpolations internally to climalsm with coupled runs currently - the coupler just updates the values in `p`.

That said, it would make sense to include this in the Interpolators work if we wanted to store the multiple values of the fluxes that the coupler gets from Atmos in coupled runs, and then interpolate to the land simulation time in between the two atmos times, and write an `update_drivers!` method for `CoupledDrivers` that does the interpolation. But I dont think this is planned.




 and we wont have any update_drivers! function for these types (where the interpolations would be called). I think they might be different concepts.

kmdeck:
> > This looks good to me!!
> > Maybe we want to hold off on merging until we've had a chance to think about whether we want to include the coupled driver case in the Interpolators work? @Sbozzolo

fwiw, it is slightly confusing to call evaluating a forcing function f(t) an interpolation if it is an analytic function, but it's probably ok since it is only internal and we can document it well. But I think forcing/driving is a separate concept than interpolating data (even if some types of forcing require interpolations).

juliasloan25:
> That said, it would make sense to include this in the Interpolators work if we wanted to store the multiple values of the fluxes that the coupler gets from Atmos in coupled runs, and then interpolate to the land simulation time in between the two atmos times, and write an update_drivers! method for CoupledDrivers that does the interpolation. But I dont think this is planned.

I also don't think this is planned either - all component models are just stepped using the same timestep right now, and I think it will stay like that for a while.

I agree that the coupled vs prescribed atmos are different concepts. I'm just not sure if we want to have a uniform interface for both cases, or if we're okay with having different interfaces for the two.

Either way, I agree that the terminology of `interpolate` to update a forcing when interpolation is only one of the update options is perhaps misleading.

Sbozzolo:
I haven’t look at this yet, but I also wanted to bring up that
`interpolate` is not the best name and I wanted to solicit better names
that encode evaluating analytical functions and interpolated data (I
couldn’t find anything particularly suitable).

Sbozzolo:
The point of view I am taking is the one of model developers and my goal is to ensure that model developers do not have to think about the source of the data that they need to use.

For example, with my interpolator interface, the update drivers for atmosphere will always look like:
```julia
function make_update_drivers(a::PrescribedAtmosphere{FT}) where {FT}
    function update_drivers!(p, t)
        interpolate!(p.drivers.P_liq, a.liquid_precip, t)
        interpolate!(p.drivers.P_snow, a.snow_precip, t)
        interpolate!(p.drivers.T, a.T, t)
        interpolate!(p.drivers.P, a.P, t)
        interpolate!(p.drivers.u, a.u, t)
        interpolate!(p.drivers.q, a.q, t)
        interpolate!(p.drivers.c_co2, a.c_co2, t)
    end
    return update_drivers!
end
```
We don't have to worry if `a.T` is a function, a 1D data, or 2D data. 

I think we could extend this to also work with coupled runs. Internally, `interpolate!` will just fetch the correct information. This would unify the interface for coupled runs as well and make for a nice user/developer-experience (there's fewer types, only one way to do things, responsabilities are reasonably well segregrated, ...). There are a couple of technical challenges (+ unforseen ones) in implementing this.

In any case, what name could I pick instead of Interpolator? `DataEvaluator` ?

kmdeck:
> The point of view I am taking is the one of model developers and my goal is to ensure that model developers do not have to think about the source of the data that they need to use.
> 
> For example, with my interpolator interface, the update drivers for atmosphere will always look like:
> 
> ```julia
> function make_update_drivers(a::PrescribedAtmosphere{FT}) where {FT}
>     function update_drivers!(p, t)
>         interpolate!(p.drivers.P_liq, a.liquid_precip, t)
>         interpolate!(p.drivers.P_snow, a.snow_precip, t)
>         interpolate!(p.drivers.T, a.T, t)
>         interpolate!(p.drivers.P, a.P, t)
>         interpolate!(p.drivers.u, a.u, t)
>         interpolate!(p.drivers.q, a.q, t)
>         interpolate!(p.drivers.c_co2, a.c_co2, t)
>     end
>     return update_drivers!
> end
> ```
> 
> We don't have to worry if `a.T` is a function, a 1D data, or 2D data.
> 
> I think we could extend this to also work with coupled runs. Internally, `interpolate!` will just fetch the correct information. This would unify the interface for coupled runs as well and make for a nice user/developer-experience (there's fewer types, only one way to do things, responsabilities are reasonably well segregrated, ...). There are a couple of technical challenges (+ unforseen ones) in implementing this.
> 
> In any case, what name could I pick instead of Interpolator? `DataEvaluator` ?

What you have so far makes a lot of sense (prescribed data forcing the simulation, with the same interface if it is an anlytic function, 1d data, 2d, etc). `evaluate!` is a nice name over `interpolate!`, or even `update_field!`

But Im not sure coupled forcing needs to fit the same pattern because it has quite different requirements (aside from being "forcing"). With coupled runs/coupled forcing data, we do not need to have an `update_drivers!` function at all. With prescribed forcing, we need to call `update_drivers!` using a callback, and this is not required in coupled runs. 

I would propose that we merge this PR as it maintains the current state of how we run coupled simulations. it just moves the definition of these coupled types etc to ClimaLSM instead of ClimaCoupler. This would be a necessary step to changing how we handled coupled forcing, because moving forwards any changes we make (extending `evaluate!` to coupled forcing, e.g.) is internal to ClimaLSM. And then moving forwards we can figure out what is best - it would a separate PR to the GPU/interpolation work I think as well?

juliasloan25:
I do think that it would be nice to handle the coupled case since Interpolators (or whatever we name it) will handle all other atmos/radiative forcing cases. What technical challenges are you referring to? Would the coupled case add new ones, or are they more general?

I'm hesitant about `DataEvaluator` because in the case of an analytical function there's no data involved. Maybe just `Evaluator`, or `Updater`? But those might be too vague :/

Sbozzolo:
> I would propose that we merge this PR as it maintains the current state of how we run coupled simulations. it just moves the definition of these coupled types etc to ClimaLSM instead of ClimaCoupler. This would be a necessary step to changing how we handled coupled forcing, because moving forwards any changes we make (extending evaluate! to coupled forcing, e.g.) is internal to ClimaLSM. And then moving forwards we can figure out what is best - it would a separate PR to the GPU/interpolation work I think as well?

Yes, I agree.

> What technical challenges are you referring to? Would the coupled case add new ones, or are they more general?

I see at least three:
- If we add a fourth interpolator (`CoupledInterpolator`), the interpolator object will essentially have to contain pointers to the source of where we take the data from, so that we can the evaluate function just being `x .= y`, where `y` is the data independently handled by the coupler
- related, this might require `ClimaLSM` to know more about the details of the coupler, so it might end up that the ClimaLSM will have to depend on the coupler, which depends on ClimaLSM (not possible, we would have to have a 3rd package)
- it might be tricky to guarantee that when `evaluate!` is called the data is ready and up to date

I don't have the full implementation in mind, so I am not sure if these are really problems or not (and if there are even more problems)

> I'm hesitant about DataEvaluator because in the case of an analytical function there's no data involved. Maybe just Evaluator, or Updater? But those might be too vague :/

Yes, I am not fond of any of these either

src/standalone/Soil/boundary_conditions.jl
kmdeck: we will, yeah.. I think we will just need SHF, LHF, and vapor flux. 

the aerodynamic resistance `r_ae` is needed right now because of an issue in SurfaceFluxes.jl (it does not account for surface resistance, so we use both resistances to correct the fluxes surface fluxes returns right now). Longer term I think we can remove this from the cache for all models.

kmdeck: (for now, no additional methods for CoupledAtmosphere exist for any model but the bucket model, thought the bucket may rely on defaults and not have special methods like other models may need)

PR #454, opened 2024-01-17
Access temperature from integrator cache

Sbozzolo:
As instructed by @kmdeck


kmdeck (approved):
Looks good. as we discussed on Slack, the driver variables are now stored in the cache and should never be computed on the fly in functions used in the tendency, but instead updated via a callback and a single call to `update_drivers!`.

Issue #456, opened 2024-01-19
lib/ClimaLandSimulations functions instead of include


AlexisRenchon:
**Is your feature request related to a problem? Please describe.**
ClimaLandSimulations allows running ClimaLSM for FLUXNET sites, but currently include path to configuration files (domain, timestepper, parameters...). While it works, this is not very elegant. This SDI will refactor the code to use functions instead of includes. 

PR #457

PR #441, opened 2024-01-05
Add a lib folder containing packages extending ClimaLSM functionality 

AlexisRenchon:
## Purpose 

This PR adds a lib folder containing packages extending ClimaLSM functionality, for now ClimaLandSimulations with functions to run ClimaLSM at single site and in the future globally, and ClimaLandDashboards with web parameterisation dashboard and in the future single site (fluxnet) web dashboard

## Content

----
- [x] I have read and checked the items on the review checklist.

PR #457, opened 2024-01-19
ClimaLandSimulations: use functions instead of includes

AlexisRenchon:
## Purpose 

This PR refactor ClimaLandSimulations code to use functions instead of include(path). This does not change the functionality. #456 

## To-do

unresolved comments from last PR #441 :
- [x] functions instead of include
- [x] all sites (not just ozark) - need to update!
- [x] inputs and outputs DataFrame with Unitful should be outside of `fluxnet_simulation()` (for speed and maybe GPU compatibility)
- [x] Fluxnet module 
- [x] Allow changing 1 param (e.g., `fluxnet_simulation("US-MOz"; params = ozark_default_params(; hetero_resp = hetero_resp_ozark(; b = 2)))`) --- done 
- [ ] lib/ClimaLandSimulations/src/fluxnet_simulation.jl around line 266, why the +40? (this is in experiment/integrated as well)
```drivers.TA.values[1 + Int(round(t0 / DATA_DT))] + 40# Get soil temperature at t0```
- [ ] add tests for utilities (e.g., `getoutput`, `data_tools.jl`, etc)
- [ ] lib/ClimaLandSimulations/src/utilities/climalsm_output_dataframe.jl and others files, avoid hard-coding values
- [ ] lib/ClimaLandSimulations/src/utilities/inputs_dataframe.jl do we always have all variables from fluxnet input files? (note: I think so, even if there are -9999 they should be there - if not we can create it)
- [ ] Should LAI, RAI, plant_\nu be in met_drivers?
- [ ] Check that buildkite figures are similar to experiment, when it is, remove experiment folder
- [ ] Add modularity as an arg, for example, `fluxnet_simulation("US-MOz", conductance = Medlyn)`
- [ ] Make the list of parameters shorter (use more default values or constants)

## Content

----

### CODE REVIEW

**Comment by AlexisRenchon** [comment]
@juliasloan25 @kmdeck 
Making our files function, they will return named tuples. 
For example, 
```
julia> out = site_setup("US-MOz")
(dz_bottom = 1.5, dz_top = 0.025, n_stem = 1, n_leaf = 1, h_stem = 9.0, h_leaf = 9.5, t0 = 1.0368e7, dt = 120.0, n = 15)
```
In our current code, those variables are in global scope, so we have two options, unpack them
```
julia> dz_bottom, dz_top, n_stem, n_leaf, h_stem, h_leaf, t0, dt, n = out
(dz_bottom = 1.5, dz_top = 0.025, n_stem = 1, n_leaf = 1, h_stem = 9.0, h_leaf = 9.5, t0 = 1.0368e7, dt = 120.0, n = 15)

julia> n
15
```
or access them from the named tuple (but in that case, we'll need to adjust the code)
```
julia> out.n
15
```

Do you think we should unpack (but in that case, the parameter list will be long...) or use the nested values from the named tuples? (and in that case, we'll just have nested variable names in the code)

Just wondering what you think is "best practice" / "more readable" ...

**Comment by AlexisRenchon** [comment]
@juliasloan25 @kmdeck @Espeer5 

The PR is not ready to be merged, but I replaced all include of files with functions, as you can see in `src/fluxnet_simulation.jl`

You'll notice a `function fluxnet_simulation_test(...)` in that src/fluxnet_simulation.jl that I did to test the functions. - will need to delete it, but I thought it could be helpful for code review. 

Everything works (but only Ozark is implemented for now, adding the other sites following the same framework will be easy), but what do you think of the code design? There is a lot of nested variables in the code now (e.g., params.soil_heat.ν_ss_quartz). 

I didn't change the general file structure that we had previously in `experiments`, but maybe we could group some files? (not sure if there is a way to be less messy). Basically, I am open to any code or file restructuration. 

I think what will be important for functionality in future PRs will be to allow users to change modules (e.g., `fluxnet_simulation("US-MOz", conductance = Medlyn)`, where conductance could be any submodel implemented), and I am not sure yet how we'd implement something like this. Nothing urgent but to keep in mind. (parameters number and name could change, etc.)

Another functionality we mentioned last time is having a function like `run_simulation(met_data, params, domain, ...)` or similar, the idea being where users can use their own meteorological drivers and parameters, configuration, etc. (not especially fluxnet). Maybe that would be for another PR. (but I mention it if it is relevant for code design now). I think [Ed's PFT PR](https://github.com/CliMA/ClimaLSM.jl/pull/466) could be a start in that direction.

**Comment by Espeer5** [comment]
I will wait to try and merge #466 until this is done -> I will have the PFTs spit out parameters in NamedTuples like this instead of creating global scope variables for each. I like this, it's a lot cleaner to work with and should make writing different methods for getting different parameters from different places easier.

File: lib/ClimaLandSimulations/src/fluxnet/US-MOz/US-MOz_parameters.jl [hunk 1]
```diff
@@ -1,123 +1,125 @@
...
-retention_model = PlantHydraulics.LinearRetentionCurve{FT}(a)
-capacity = FT(10) # kg/m^2
-plant_S_s = FT(1e-2 * 0.0098) # m3/m3/MPa to m3/m3/m
-rooting_depth = FT(0.5) # from Natan
-z0_m = FT(0.13) * h_canopy
-z0_b = FT(0.1) * z0_m
+"""
+Site-specific model parameters for running CliMA LSM on the Ozark
+fluxtower site.
+"""
```
Espeer5, line 6: If we want to be able to support prescribing parameters/model type for one model and not for others, we may want to separate the prescription of each internal NamedTuple and then have a function that constructs them all together into the outer NamedTuple. So for example we could have variables for `default_hetero_resp`, `default_auto_resp`, etc., and then have a function that takes in a param set for each model and constructs them into an outer one.

Alternatively we could just overwrite the defaults inside the outer NamedTuple after we build the whole thing for whichever we want to prescribe special parameters for.

Espeer5: Or even possibly we could construct the different default params for each model in their own NT, and then in `site_parameters.jl` build each parameters struct from its own param NT and then we could just swap the default for a user defined NT there.

AlexisRenchon: Thanks!! These are great suggestions. Let's try it.

AlexisRenchon:
> I will wait to try and merge #466 until this is done -> I will have the PFTs spit out parameters in NamedTuples like this instead of creating global scope variables for each. I like this, it's a lot cleaner to work with and should make writing different methods for getting different parameters from different places easier.

@Espeer5 Sounds good! We can merge this PR (#457) ASAP (I think we need to wait 4 days because of renaming) and then start another PR for the rest. I think this one is already quite big.

AlexisRenchon:
@juliasloan25 @kmdeck @Espeer5 

I am wondering what's the best way for user parameters here. 
In this PR currently it is entered as a [NamedTuple](https://github.com/CliMA/ClimaLand.jl/blob/ar/climalandsimulations_functions/lib/ClimaLandSimulations/src/fluxnet/US-MOz/US-MOz_parameters.jl), but it is not convenient for two reasons: 
- modularity (but Ed explained a possible solution above)
- modifying just one param (what if, for example, user just want to change param.conductance.g1) --> it would be neat if one could do `run_fluxnet("US-MOz"; params = ozark_default_params(; conductance.g1 = 2.3)` for example

Is there a way we could use our model parameter constructor here? 
for example ClimaLand.Biogeochemistry.SoilCO2ModelParameters (or any)

lib/ClimaLandSimulations/src/fluxnet_simulation.jl
```diff
@@ -1,4 +1,33 @@
-export fluxnet_simulation
+export fluxnet_simulation, fluxnet_simulation_test
```
juliasloan25, line 1: I think it would be nice if the files in `src/` contained modules, so that namespaces are more clearly defined/isolated. This is similar to what we do currently in ClimaLSM, where we have different modules in `src/` for the different shared utilities, Bucket, Soil, etc.

We could have a `Fluxnet` module for `fluxnet_simulation.jl` and `met_drivers_FLUXNET.jl`, `InputOutput` for `climaland_output_dataframe.jl` and `inputs_dataframe.jl`, and then maybe `Setup` for `domain_setup.jl` and `timestepper_setup.jl` (we should make sure the domain_setup module doesn't have the same name as ClimaLand.Domains). What do you think?

lib/ClimaLandSimulations/src/ClimaLandSimulations.jl
```diff
@@ -17,6 +17,7 @@ using ClimaLand.Canopy
 using ClimaLand.Canopy.PlantHydraulics
 import ClimaLand.Parameters as LP
 import ClimaComms
+context = ClimaComms.context()
```
juliasloan25, line 20: usually the context is set by the driver file (so experiments/ozark.jl here) rather than in src/

juliasloan25 (commented):
I mainly had a suggestion about adding modules to the src code, which goes hand in hand with defining functions rather than including all the code. Other than that this looks good!!

lib/ClimaLandSimulations/src/fluxnet_simulation.jl
AlexisRenchon: Agreed!

lib/ClimaLandSimulations/experiments/README.md
```diff
@@ -9,22 +12,22 @@ julia> fluxnet_simulation("US-MOz")
 This will run ClimaLand with default parameters and domain. Optionally, you can modify these:
 
 ```jl
-julia> fluxnet_simulation("US-MOz"; customparams = path_to_parameters, customdomain = path_to_domain)
+julia> fluxnet_simulation("US-MOz"; site_setup_out = customsetup, params = customparams)
```
kmdeck, line 15: what does "out" refer to in the kwarg?

lib/ClimaLandSimulations/experiments/ozark.jl
```diff
@@ -1,6 +1,10 @@
 using ClimaLandSimulations
 
-sv, sol, Y, p, inputs, climaland = fluxnet_simulation("US-MOz");
+sv, sol, Y, p = fluxnet_simulation("US-MOz");
+
+# DataFrame for input and output
+inputs, inputs_SI, inputs_commonly_used = make_inputs_df("US-MOz")
+climaland = make_output_df(sv, inputs)
```
kmdeck, line 7: maybe "simulation_output" instead of `climaland` as a name?

lib/ClimaLandSimulations/src/ClimaLandSimulations.jl
kmdeck: ^^ yes, because I think it will change if on CPU or CPU

lib/ClimaLandSimulations/src/fluxnet/US-Ha1/US-Ha1_config.jl
```diff
@@ -0,0 +1,19 @@
+export harvard_default_configs
+
+"""
+    harvard_default_configs = (
+        data_link = "https://caltech.box.com/shared/static/xixaod6511cutz51ag81k1mtvy05hbol.csv",
+        time_offset = 5,
```
kmdeck, line 6: maybe this could be more descriptive? `local_to_UTC` or something?

lib/ClimaLandSimulations/src/fluxnet_simulation.jl
```diff
@@ -8,71 +37,44 @@ Run ClimaLand at a fluxnet site.
 function fluxnet_simulation(
     site_ID;
     FT = Float64,
-    path_to_sim = joinpath(
-        ClimaLandSimulations_dir,
-        "src",
-        "fluxnet",
-        "$site_ID",
-        "$(site_ID)_simulation.jl",
+    site_setup_out = site_setup(site_ID),
```
kmdeck, line 40: we could say "setup = site_setup()", "domain = site_domain(),"... etc instead of `site_setup_out`?

kmdeck: what is the difference between `setup` and `config`? :)

lib/ClimaLandSimulations/src/fluxnet/site_parameters.jl
```diff
@@ -0,0 +1,87 @@
...
+        plant_hydraulics.ψ63,
+        plant_hydraulics.Weibull_param,
+    )
+    retention_model =
+        PlantHydraulics.LinearRetentionCurve{FT}(plant_hydraulics.a)
+
+    return (
+        hetero_resp = hetero_resp,
+        auto_resp = auto_resp,
+        soil = soil,
```
kmdeck, line 53: should soil and soil heat be different? similarly, many of the params below (k solid, k_dry, etc) are for soil

lib/ClimaLandSimulations/src/fluxnet_simulation.jl
kmdeck: (could use more descriptive names?)

lib/ClimaLandSimulations/src/utilities/climaland_output_dataframe.jl
```diff
@@ -28,7 +28,7 @@ end
 
 Return a dataframe containing climaland outputs
 """
-function make_output_df(sv, inputs)
+function make_output_df(sv, inputs; N_days = 60, N_spinup_days = 30)
```
kmdeck, line 31: do we also need the state `u` and not just `sv`?

lib/ClimaLandSimulations/src/utilities/inputs_dataframe.jl
```diff
@@ -1,11 +1,25 @@
...
-    make_inputs_df(LOCAL_DATETIME, drivers)
+    make_inputs_df(site_ID)
 
 Returns DataFrames of inputs, one unitless, one in SI units, one in commonly used units. 
 """
-function make_inputs_df(LOCAL_DATETIME, drivers)
+function make_inputs_df(
+    site_ID;
+    FT = Float64,
+    site_setup_out = site_setup(site_ID),
```
kmdeck, line 11: I noticed several functions where make these - should they be made once and passed? not sure, just asking! right now they are not expensive to make

lib/ClimaLandSimulations/src/utilities/met_drivers_FLUXNET.jl
```diff
@@ -198,26 +196,28 @@ function setup_drivers(site_ID)
     LAI_dt = Second(MODIS_LAI[2, 1] - MODIS_LAI[1, 1]).value
     LAI_seconds = FT.(0:LAI_dt:((length(MODIS_LAI[:, 1]) - 1) * LAI_dt))
 
-    # LAI function for radiative transfer
+    # LAI spline for radiative transfer
```
kmdeck, line 199: no longer a spline :)

kmdeck:
> Everything works (but only Ozark is implemented for now, adding the other sites following the same framework will be easy), but what do you think of the code design? There is a lot of nested variables in the code now (e.g., params.soil_heat.ν_ss_quartz).

I am OK with nested variables! they are clear and internal. I would suggest though that we match the same naming convention that we use for `p` and `Y` - so just a single `soil` field (not soil and soil_heat), and follow the other names for canopy, canopy components, soil co2, etc.

> I didn't change the general file structure that we had previously in `experiments`, but maybe we could group some files? (not sure if there is a way to be less messy). Basically, I am open to any code or file restructuration.
For now I think this is OK. 

> I think what will be important for functionality in future PRs will be to allow users to change modules (e.g., `fluxnet_simulation("US-MOz", conductance = Medlyn)`, where conductance could be any submodel implemented), and I am not sure yet how we'd implement something like this. Nothing urgent but to keep in mind. (parameters number and name could change, etc.)
Agreed! it would be nice. FWIW, it will also always be an option to run the code like we currently do in experiments (a little more explicitly) for more custom runs that we find hard to incorporate into the interface.

> Another functionality we mentioned last time is having a function like `run_simulation(met_data, params, domain, ...)` or similar, the idea being where users can use their own meteorological drivers and parameters, configuration, etc. (not especially fluxnet). Maybe that would be for another PR. (but I mention it if it is relevant for code design now). I think [Ed's PFT PR](https://github.com/CliMA/ClimaLSM.jl/pull/466) could be a start in that direction.
in my opinion, being able to run with your own parameters and config is very important, but for now using different drivers is less important. Just thinking about our own use cases (simulating flux net sites with flux net data but with different parameters, which we need to do in calibration) as higher priority for now :)


Keep in mind that after Nat's work with moving things to parameters (for site level and global runs), and as we work to be able to handle globally varying parameters (for global runs), this interface will need to change a bit.

kmdeck:
> @juliasloan25 @kmdeck @Espeer5
> 
> I am wondering what's the best way for user parameters here. In this PR currently it is entered as a [NamedTuple](https://github.com/CliMA/ClimaLand.jl/blob/ar/climalandsimulations_functions/lib/ClimaLandSimulations/src/fluxnet/US-MOz/US-MOz_parameters.jl), but it is not convenient for two reasons:
> 
>     * modularity (but Ed explained a possible solution above)
> 
>     * modifying just one param (what if, for example, user just want to change param.conductance.g1) --> it would be neat if one could do `run_fluxnet("US-MOz"; params = ozark_default_params(; conductance.g1 = 2.3))` for example
> 
> 
> Is there a way we could use our model parameter constructor here? for example ClimaLand.Biogeochemistry.SoilCO2ModelParameters (or any)

Yes! our constructors usually already allow for this. We can splat named tuples into kwargs too:
```
julia> function foo(;a=2, b = 3, c = 4)
       @show a;
       @show b;
       @show c;
       end
foo (generic function with 1 method)

julia> foo()
a = 2
b = 3
c = 4
4

julia> x = (b = 5,)
(b = 5,)

julia> foo(;x...)
a = 2
b = 5
c = 4
4```
 so, it would be good to make use of the constructors we already have, maybe by splatting the passed NamedTuple (with a shorter number of params, optionally)?

**Comment by kmdeck** [comment]
FWIW, there is still something wrong in the ozark plots. The soil water content is not changing in time. Perhaps before merging, we can make sure the output is correct? not sure if it is is on the plotting side or from something in the sim itself.

File: lib/ClimaLandSimulations/src/fluxnet_simulation.jl (hunk 7 above)
Comment (AlexisRenchon) [reply lib/ClimaLandSimulations/src/fluxnet_simulation.jl]: I renamed as suggested, not sure about `setup` and `config` - I think it could possibly be done from the same file, we could definitely use some reorganization (it's on the to do list)

File: lib/ClimaLandSimulations/src/utilities/climaland_output_dataframe.jl (hunk 9 above)
Comment (AlexisRenchon) [reply lib/ClimaLandSimulations/src/utilities/climaland_output_dataframe.jl:31]: we could return `u` in `fluxnet_simulation()`

File: lib/ClimaLandSimulations/src/utilities/inputs_dataframe.jl (hunk 10 above)
Comment (AlexisRenchon) [reply lib/ClimaLandSimulations/src/utilities/inputs_dataframe.jl:11]: I understand it's a bit weird but I'll leave it like that for now - `make_inputs_df` is different because it uses unitful and creates input with SI or "commonly used" units, but I avoid this for `fluxnet_simulation()` where we might want efficiency (make_inputs_df is useful for e.g., plotting stuff). 
We could refactor later.

**Comment by AlexisRenchon** [comment]
@kmdeck 
I addressed most of your review comments, I think! 
Mostly changes are: Fluxnet is now a module, and user can change some parameters from the default. I also used our struct from src for the parameters when possible, and wrote a comment when we couldn't

File: lib/ClimaLandSimulations/src/ClimaLandSimulations.jl (hunk 3 above)
Comment (juliasloan25) [reply lib/ClimaLandSimulations/src/ClimaLandSimulations.jl]: If `include("Fluxnet/fluxnet_simulation.jl")` is called in the driver it can be there, but otherwise it needs to be in the driver itself. It's not good practice to include a src/ file in the driver. A good solution could be to define a function `get_context` in fluxnet_simulation.jl which returns `ClimaComms.context()`, then call that from the driver. Would that be okay with you?

File: lib/ClimaLandSimulations/src/utilities/timestepper_setup.jl [hunk 12]
```diff
@@ -1,7 +1,32 @@
...
+
+"""
+    timestepper_setup(site_setup_out;
+        N_spinup_days = 30,
+        N_days_sim = 30,
+        timestepper = CTS.RK4(),
+        ode_algo = CTS.ExplicitAlgorith(timestepper)
+        )
+
+Define the setup for the simulation timestepper.
```
juliasloan25, line 11: Could you explain the default timestepper (4th order Runge Kutta method), and that other timesteppers from ClimaTimeSteppers could be used? Also maybe explain what happens during spinup vs sim days

lib/ClimaLandSimulations/src/utilities/domain_setup.jl
```diff
@@ -1,14 +1,55 @@
...
+        zmin = FT(-10),
+        zmax = FT(0)
+        )
+
+setup the soil and canopy domain of the simulation. Returns:
+- h_canopy,
+- compartment_midpoints, 
+- compartment_surfaces,
+- land_domain,
+- canopy_domain,
```
juliasloan25, line 15: Could you briefly explain these? e.g. what is a component

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/site_setup.jl
```diff
@@ -0,0 +1,52 @@
...
+
+function site_setup(site_ID)
+    default_args = Dict(
+        "US-MOz" => ozark_default_args,
+        "US-Ha1" => harvard_default_args,
+        "US-NR1" => niwotridge_default_args,
+        "US-Var" => vairaranch_default_args,
+    )
+
+    if haskey(default_args, site_ID)
```
juliasloan25, line 49: We should add an else case here - maybe raise an error if we don't have default args for the provided `site_ID`, and give a message that this site isn't currently supported (or may have a typo)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/site_parameters.jl
```diff
@@ -0,0 +1,98 @@
...
+        println("This site_ID does not exist")
+    end
+end
+
+# The structs below are necessary and different from ClimaLand/src struct 
+# because of, for example, shared parameters (e.g., porosity in soil or soilco2)
+# or, another example, parameters calculated from other parameters (e.g., κ_solid)
+# the structs below only contain unique params, that are given as number (not computed FTwhere)
+
+struct hetero_resp # differs from SoilCO2ModelParameters because of porosity
```
juliasloan25, line 31: Julia convention says that struct names should use CamelCase, unlike functions https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/site_parameters.jl
```diff
@@ -0,0 +1,98 @@
...
+    PAR_albedo::FT
+    NIR_albedo::FT
+end
+
+struct vangenuchten # I don't think these have a struct in src
+    vg_n::FT
+    vg_α::FT
+end
+
+struct κ # same, these are seen in soil_heat_parameterizations.jl, but in no struct. May be fixed in PR SpaceVaryingInput #485
```
juliasloan25, line 65: I think a more descriptive name for the struct would be nice (what does kappa represent?)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/met_drivers_fluxnet.jl
```diff
@@ -198,26 +196,28 @@ function setup_drivers(site_ID)
...
-        maxLAI,
-        RAI,
-        plant_ν,
+        LOCAL_DATETIME = LOCAL_DATETIME,
+        atmos_co2 = atmos_co2,
+        DATA_DT = DATA_DT,
+        drivers = drivers,
+        atmos = atmos,
+        radiation = radiation,
+        LAIfunction = LAIfunction,
```
juliasloan25, line 218: will this always be a function, or should we rename this more generally to allow for other types of LAI input?

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/inputs_dataframe.jl
```diff
@@ -1,11 +1,25 @@
...
 """
-    make_inputs_df(LOCAL_DATETIME, drivers)
+    make_inputs_df(site_ID)
 
 Returns DataFrames of inputs, one unitless, one in SI units, one in commonly used units. 
 """
-function make_inputs_df(LOCAL_DATETIME, drivers)
+function make_inputs_df(
+    site_ID;
+    FT = Float64,
```
juliasloan25, line 10: I'm hesitant about redefining FT here. FT should be defined once in the driver to avoid Float32/Float64 conflicts. It's not used in this function so I think we can remove it here

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-Var/US-Var_parameters.jl
```diff
@@ -0,0 +1,122 @@
...
+        κ_om = FT(0.25), # W/m/K
+        κ_liq = FT(0.57), # W/m/K
+        κ_ice = FT(2.29), # W/m/K
+        κ_air = FT(0.025), #W/m/K
+        ρp = FT(2700), # kg/m^3
+        #κ_solid = Soil.κ_solid(ν_ss_om, ν_ss_quartz, κ_om, κ_quartz, κ_minerals)
+        #κ_dry_val = Soil.κ_dry(ρp, soil_ν, κ_solid, κ_air)
+        #κ_sat_frozen_val = Soil.κ_sat_frozen(κ_solid, soil_ν, κ_ice)
+        #κ_sat_unfrozen_val = Soil.κ_sat_unfrozen(κ_solid, soil_ν, κ_liq);
+        #ρc_ds = FT((1 - soil_ν) * 4e6); # J/m^3/K
```
juliasloan25, line 58: why are some params commented out in this file? (and in the other parameter files)

lib/ClimaLandSimulations/src/Fluxnet/Fluxnet.jl
```diff
@@ -0,0 +1,64 @@
...
+@compound CO₂
+
+# Directory paths
+climalandsimulations_dir = pkgdir(ClimaLandSimulations)
+climaland_dir = pkgdir(ClimaLand)
+
+# ClimaParameters
+include(joinpath(climaland_dir, "parameters", "create_parameters.jl"))
+const FT = Float64
+earth_param_set = create_lsm_parameters(FT)
```
juliasloan25, line 41: FT should be defined in the driver, not in src, so that we can run simulations with either Float32 or Float64. Could this setup be moved to fluxnet_simulation.jl?

lib/ClimaLandSimulations/src/Fluxnet/Fluxnet.jl
```diff
@@ -0,0 +1,64 @@
...
+using UnitfulMoles: molC
+using Unitful, UnitfulMoles
+using ArtifactWrappers
+using DelimitedFiles
+using Thermodynamics
+using Formatting
+using DataFrames
+
+# UnitfulMoles compound
+@compound CO₂
```
juliasloan25, line 32: could this be moved to a file within the module? maybe fluxnet_utilities/site_parameters.jl?

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_simulation.jl
```diff
@@ -0,0 +1,411 @@
+export fluxnet_simulation, fluxnet_simulation_test
+
+function fluxnet_simulation_test(
```
juliasloan25, line 3: what is `fluxnet_simulation` vs `fluxnet_simulation_test`?

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_simulation.jl
```diff
@@ -0,0 +1,411 @@
...
+end
+
+"""
+    fluxnet_simulation(site_ID; kwargs)
+
+Run ClimaLand at a fluxnet site.
+"""
+function fluxnet_simulation(
+    site_ID;
+    FT = Float64,
```
juliasloan25, line 57: this function is kind of the driver, so this should be the only place we define FT (in `fluxnet_simulation_test` is okay too I think since that seems like another driver)

lib/ClimaLandSimulations/src/ClimaLandSimulations.jl
```diff
@@ -1,57 +1,34 @@
...
+# Directory paths
+climalandsimulations_dir = pkgdir(ClimaLandSimulations)
 climaland_dir = pkgdir(ClimaLand)
-savedir =
-    joinpath(climaland_dir, "experiments", "integrated", "fluxnet/figures/")
+
+# ClimaParameters
 include(joinpath(climaland_dir, "parameters", "create_parameters.jl"))
+const FT = Float64
 earth_param_set = create_lsm_parameters(FT)
```
juliasloan25, line 23: FT shouldn't be defined here. If we move `earth_param_set` inside the driver `fluxnet_simulation` we won't need it here

lib/ClimaLandSimulations/src/ClimaLandSimulations.jl
```diff
@@ -1,57 +1,34 @@
...
 
-ClimaLandSimulations_dir = pkgdir(ClimaLandSimulations)
+include(joinpath("Fluxnet", "Fluxnet.jl"))
+using .Fluxnet
```
juliasloan25, line 26: this is really nice, thank you!!

lib/ClimaLandSimulations/experiments/ozark.jl
```diff
@@ -1,6 +1,21 @@
 using ClimaLandSimulations
+using ClimaLandSimulations.Fluxnet
 
-sv, sol, Y, p, inputs, climaland = fluxnet_simulation("US-MOz");
+# default parameters, except for custom parameter hetero_resp.b
+sv_test, sol_test, Y_test, p_test = fluxnet_simulation(
```
juliasloan25, line 5: maybe add a comment that this is where the simulation is actually solved (or rename fluxnet_simulation -> run_ fluxnet_simulation?)

lib/ClimaLandSimulations/experiments/README.md
juliasloan25: the variables could be `site_setup`, and the function could be called `make_site_setup` or `get_site_setup`. Same for the other `_out` variables - `domain_setup_out`, `site_config_out`, etc

juliasloan25 (commented):
This looks really good! I left some more comments

lib/ClimaLandSimulations/experiments/README.md
kmdeck: [can be later PR but I agree with Julia]

kmdeck: I think I mentioned this at another point too, but "setup" is a little vague - what is it setting up? [later PR ok]

lib/ClimaLandSimulations/experiments/ozark.jl
```diff
@@ -1,6 +1,21 @@
...
+    "US-MOz";
+    params = ozark_default_params(; hetero_resp = hetero_resp_ozark(; b = 2)),
```
kmdeck, line 7: should this be the default? if this is testing that we can add in other parameters, maybe that should go into a unit test. [later PR ok]

lib/ClimaLandSimulations/experiments/ozark.jl
```diff
@@ -1,6 +1,21 @@
...
+)
+
+# default parameters
+sv, sol, Y, p = fluxnet_simulation("US-MOz")
+
+# add custom param
```
kmdeck, line 13: can remove commented lines?

lib/ClimaLandSimulations/src/Fluxnet/Fluxnet.jl
kmdeck: These are small changes but if you dont mind, changing them in this PR might be better since some of these lines may not show up in later PRs, and it will be hard to keep a list of all of them on your own.

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-Ha1/US-Ha1_parameters.jl
```diff
@@ -0,0 +1,126 @@
...
+        # Autotrophic respiration parameters
+        ne = FT(8 * 1e-4),
+        ηsl = FT(0.01),
+        σl = FT(0.05),
+        μr = FT(1.0),
+        μs = FT(0.1),
+        f1 = FT(0.012),
+        f2 = FT(0.25),
+    ),
+    soil = (
```
kmdeck, line 34: soil and soil heat should be combined [later PR ok]

kmdeck: (parameters in soil_heat also greatly affect soil moisture, they arent really separate models)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-Ha1/US-Ha1_simulation.jl
```diff
@@ -0,0 +1,28 @@
...
+    harvard_default_args = (
+        dz_bottom = FT(1.5),
+        dz_top = FT(0.025),
+        n_stem = Int64(1),
+        n_leaf = Int64(1),
+        h_leaf = FT(12), # m
+        h_stem = FT(14), # m
+        t0 = Float64(120 * 3600 * 24),
+        dt = Float64(40),
+        n = 45,
```
kmdeck, line 13: is it ever mentioned what `n` is? or what these parameters mean?

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-MOz/US-MOz_parameters.jl
```diff
@@ -0,0 +1,230 @@
+export ozark_default_params,
+    hetero_resp_ozark,
+    auto_resp_ozark,
+    soil_ozark,
+    vangenuchten_ozark,
+    κ_ozark,
```
kmdeck, line 6: PR https://github.com/CliMA/ClimaLand.jl/pull/489 should make it so we can remove this (either before or after this merges)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-MOz/US-MOz_parameters.jl
```diff
@@ -0,0 +1,230 @@
...
+    energy_balance_ozark,
+    conductance_ozark,
+    photosynthesis_ozark,
+    plant_hydraulics_ozark
+
+function ozark_default_params(;
+    hetero_resp = hetero_resp_ozark(),
+    auto_resp = auto_resp_ozark(),
+    soil = soil_ozark(),
+    vangenuchten = vangenuchten_ozark(),
```
kmdeck, line 17: it feels odd that this is outside of the soil parameters [future PR ok]

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-MOz/US-MOz_parameters.jl
```diff
@@ -0,0 +1,230 @@
...
+    κ = κ_ozark(),
+    radiative_transfer = radiative_transfer_ozark(),
+    energy_balance = energy_balance_ozark(),
```
kmdeck, line 20: could be `canopy_energy_balance` for clarity?

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-Ha1/US-Ha1_parameters.jl
kmdeck: or we can discuss other ways to split - e.g. the roughness lengths are also used for evaporation

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-Ha1/US-Ha1_parameters.jl
```diff
@@ -0,0 +1,126 @@
...
+        soil_vg_n = FT(2.05), # unitless
+        soil_vg_α = FT(0.04), # inverse meters
+        θ_r = FT(0.067), # m3/m3, from Wang et al. 2021 https://doi.org/10.5194/gmd-14-6741-2021
+    ),
+    soil_heat = (
+        # Soil heat transfer parameters; not needed for hydrology only test
+        ν_ss_quartz = FT(0.1),
+        ν_ss_om = FT(0.1),
+        ν_ss_gravel = FT(0.0),
+        κ_quartz = FT(7.7), # W/m/K
```
kmdeck, line 48: for US-Ha, these params appear here, but for Ozark, they appear in a different struct. could make consistent [preferably this PR]

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/inputs_dataframe.jl
```diff
@@ -1,11 +1,25 @@
...
+    site_setup_out = site_setup(site_ID),
```
kmdeck, line 11: could this be "site_setup = make_site_setup"
"domain = make_domain"
"config = make_config"
etc? 

At another point, I had asked about the difference between config and setup. both somewhat vague terms

kmdeck: It's also confusing because we are using `setup` as a verb and a noun
"site_setup" (noun)
"setup_drivers" (verb)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/met_drivers_fluxnet.jl
kmdeck: good point! for global runs it will be read from a netcdf file. later PR ok!

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/site_parameters.jl
```diff
@@ -0,0 +1,98 @@
...
+    θ_r::FT
+    ν_ss_gravel::FT
+    z_0m::FT
+    z_0b::FT
+    emissivity::FT
+    PAR_albedo::FT
+    NIR_albedo::FT
+end
+
+struct vangenuchten # I don't think these have a struct in src
```
kmdeck, line 60: this does, also Brooks and Corey

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/site_parameters.jl
kmdeck, line 31: I'm still quite opposed to these structs since they re-define structs we already have. Maybe we can brainstorm other ways to fix the problems that they are addressing. future PR ok but this seems quite important

lib/ClimaLandSimulations/src/utilities/climaland_output_dataframe.jl
kmdeck: ok! we also want to plot the prognostic variables. so that sounds good!

lib/ClimaLandSimulations/src/utilities/domain_setup.jl
kmdeck: hmmm also we are redefining things that we already have defined and documented in ClimaLand/src.

AlexisRenchon: I think it is okay to double the docstrings definition, if there's different userbase for ClimaLand and ClimaLandSimulation (which just calls ClimaLand), but in the docs it should be defined just once, in ClimaLand. 
Alternatively, we could just link to ClimaLand docs in the docstrings. 

I will leave this for another PR, we can decide later

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/met_drivers_fluxnet.jl
AlexisRenchon: I'll leave this for later!

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-Var/US-Var_parameters.jl
AlexisRenchon: Only Ozark parameters file was done, I'll do the other ones now (and delete those commented out lines)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/inputs_dataframe.jl
AlexisRenchon: I actually had to put it back after removing FT definition from the module (and putting it fluxnet_simulation), it is used in domain_setup.jl 

(I think this function make_inputs_df is not a priority, it is just used for plotting. We can think more about that later)

lib/ClimaLandSimulations/src/Fluxnet/Fluxnet.jl
AlexisRenchon: done!

lib/ClimaLandSimulations/src/Fluxnet/Fluxnet.jl
AlexisRenchon: I put this line in inputs_dataframe.jl, where it is needed

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_simulation.jl
AlexisRenchon: I removed fluxnet_simulation_test, it was just to test things locally

lib/ClimaLandSimulations/src/ClimaLandSimulations.jl
AlexisRenchon: done

lib/ClimaLandSimulations/experiments/ozark.jl
AlexisRenchon: would run_fluxnet be a better name? (run_fluxnet_simulation becomes a bit long)

lib/ClimaLandSimulations/experiments/ozark.jl
AlexisRenchon: what do you mean by should this be the default? 

I agree this should eventually become a unit test - maybe later PR

lib/ClimaLandSimulations/experiments/ozark.jl
AlexisRenchon: why? the comments is to explain a bit this "experiment" code

This file should eventually be a unit test file though, it is just testing functionality... 

Let's think a bit more what to do in next PRs (what should be unit test, what could be an experiment)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-Ha1/US-Ha1_parameters.jl
AlexisRenchon: I modified the site params to match ozark, which was the only one up to date (sorry!)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-Ha1/US-Ha1_simulation.jl
AlexisRenchon: good point. I am not sure. (this is copied from experiments). 

More broadly, in future PRs, we definitely need to make better docstring throughout

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-MOz/US-MOz_parameters.jl
AlexisRenchon: future PR will fix this! =)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-Ha1/US-Ha1_parameters.jl
AlexisRenchon: sorry, only ozark was up to date. should be good now.

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/inputs_dataframe.jl
AlexisRenchon: I'll do the make_setup and make_config, etc., in this PR. 
I agree the code should be refactored a bit to avoid these vague terms (future PR)

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/site_parameters.jl
AlexisRenchon: Definitely! We need to find a way to use our src struct.

AlexisRenchon:
@kmdeck @juliasloan25 
I think I've addressed almost everything, besides some comments where I think we need to think a bit more. 
After we merge this PR, I will open another one listing the (long) to do list.

lib/ClimaLandSimulations/src/ClimaLandSimulations.jl
AlexisRenchon: I am not sure what you mean by the driver?

lib/ClimaLandSimulations/experiments/ozark.jl
juliasloan25: that sounds good to me!

lib/ClimaLandSimulations/src/ClimaLandSimulations.jl
juliasloan25: I just mean wherever the simulation is actually getting setup and run, so ozark.jl in this case I think

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/inputs_dataframe.jl
juliasloan25: Hmm, maybe we should define `FT` in experiments/ozark.jl and pass it to both `fluxnet_simulation` and `make_inputs_df`. It looks like domain isn't used in the function, so we could remove both `FT` and `domain` args

juliasloan25 (approved):

PR #460, opened 2024-01-22
dependency updates

juliasloan25:
## Purpose 
update to SurfaceFluxes v0.9, ClimaParameters v0.8, Insolation v0.9

## To-do
- [x] update dependencies
- [x] update for ClimaParameters breaking changes


Sbozzolo:
Can you also update GitHub action cache to v4?

kmdeck (approved):
It looks like the main changes were to Clima Parameters..I guess we dont have any test of our create_parameters.jl function because it is outside of src...but the buildkite runs look good and our tests which use the earth_param_set still run. so I think we are good.

thank you for taking care of this and keeping us up to date!

PR #465, opened 2024-01-23
Introduce `TimeVaryingInput2D`

Sbozzolo:
Okay, this is a massive commit (and I feel bad about it).

This commit splits PrescribedDataStatic and PrescibedDataTemporal in multiple independent modules.

Before, PrescribedData was:
- reading files
- regridding
- discovering dates available
- interpolating
- taking care of edge cases (e.g., what to do when reading outside of where the date is outside of the definition range)

This commit splits all the different functions in different independent modules so that they can be more easily maintained and upgraded.

This will soon be needed to efficiently support reading and buffering reads for GPU runs.

In all of this, we are still using ClimaCoreTempestRemap and doing everything ahead of time. At the moment, we are not efficiently reusing any informtion across variables that share the same remapping weights, but this can be added in the future.


test/standalone/Bucket/albedo_types.jl
```diff
@@ -1,460 +1,460 @@
...
-            end
-        end
-        rm(regrid_dirpath, recursive = true)
-    end
-end
-
-# Delete testing directory and files
-rm(regrid_dir_static; recursive = true, force = true)
-rm(regrid_dir_temporal; recursive = true, force = true)
+# using Test
```
kmdeck, line 1: forgot to uncomment?

experiments/standalone/Bucket/global_bucket_staticmap.jl
```diff
@@ -88,18 +88,9 @@ tf = 7 * 86400;
 Δt = 3600.0;
 
 # Construct albedo parameter object using static map
-# Use separate regridding directory for CPU and GPU runs to avoid race condition
-device_suffix =
```
kmdeck, line 92: device_suffix is needed for doing the gpu/cpu comparison after each sim runs.

experiments/standalone/Bucket/global_bucket_temporalmap.jl
```diff
@@ -87,17 +87,8 @@ tf = 50 * 86400;
 Δt = 3600.0;
 
 # Construct albedo parameter object using temporal map
-# Use separate regridding directory for CPU and GPU runs to avoid race condition
-device_suffix =
```
kmdeck, line 91: likewise here

experiments/standalone/Bucket/global_bucket_staticmap.jl
Sbozzolo: The new regridder system makes its own folders, so we don't need this anymore

kmdeck: maybe I am misunderstanding, but `device_suffix` is used in more places in these files. it is used to save the output to a file to then compare the output of the simulation (run on gpu) with the output of the simulation (run on cpu) to check that they agree

kmdeck: (this has nothing to do with regridding)

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,246 @@
...
+import ClimaCore
+
+import ..FileReader: NCFileReader
+import ..Regridder
+import ..Regridder:
+    AbstractRegridder, TempestRegridder, regrid, AVAILABLE_REGRIDDERS
+
+"""
+
+Note: It is best to have one file per variable with all the temporal data.
```
kmdeck, line 14: this may be a little inconvenient/ add a step - all the reanalysis data is in one file to start with.

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,246 @@
+module DataHandling
+
+using Dates
+
+import ClimaCore
+
+import ..FileReader: NCFileReader
+import ..Regridder
+import ..Regridder:
+    AbstractRegridder, TempestRegridder, regrid, AVAILABLE_REGRIDDERS
```
kmdeck, line 10: I dont see anything in the style guide for all caps in names - why is AVAILABLE_REGRIDDERS capitalized?

kmdeck: ah - I see some languages use that for constants. it doesnt seem like Julia follows that notation: https://docs.julialang.org/en/v1/base/base/#const

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,246 @@
...
+
+"""
+
+Note: It is best to have one file per variable with all the temporal data.
+
+Assumptions:
+- There is only one file with all the data in time
```
kmdeck, line 17: this seems to contradict the above Note?

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,246 @@
...
+"""
+struct DataHandler{
+    FR <: NCFileReader,
+    REG <: AbstractRegridder,
+    SPACE <: ClimaCore.Spaces.AbstractSpace,
+    REF_DATE <: Dates.DateTime,
+    TSTART <: AbstractFloat,
+    DATES <: AbstractArray{<:Dates.DateTime},
+    DIMS,
+    TIMES <: AbstractArray{<:AbstractFloat},
```
kmdeck, line 32: We discussed this last time, but`TIMES <: AbstractArray{<:AbstractFloat}` is the same as `AbstractArray{<:AbstractFloat}` in this context, right? 

If so, I dont see that we need to have these types as parametric types.

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,246 @@
...
+        :TempestRegridder =>
+            (target_space, mktempdir(), varname, file_path),
+    )
+    RegridderConstructor = getfield(Regridder, regridder_type)
+    regridder = RegridderConstructor(regridder_args[regridder_type]...)
+
+    # NOTE: this is not concretely typed
+    _cached_regridded_fields = Dict{Dates.DateTime, ClimaCore.Fields.Field}()
+
+    available_dates = file_reader.available_dates
```
kmdeck, line 109: available_dates and available_times do not have the same reference, that might be confusing? perhaps could document that (if i understand correctly) available_dates = list of dates in the file where there is data while available_times are the corresponding times of those dates in seconds since the start of the simulation?

kmdeck: ah ok, it is described below! all good.

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,246 @@
...
+If `time` is one of the snapshots, return the next time.
+"""
+function next_time(data_handler::DataHandler, time::AbstractFloat)
+    index = searchsortedfirst(data_handler.available_times, time)
+    time in data_handler.available_times && (index += 1)
+    return data_handler.available_times[index]
+end
+
+function next_time(data_handler::DataHandler, date::Dates.DateTime)
+    if date in data_handler.available_dates
```
kmdeck, line 191: maybe worth using the same code  as `next_time` for clarity (otherwise the reader wonders if `next_date` does something different than `next_time`)

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,246 @@
...
+    # Dates.DateTime(0) is the cache key for static maps
+    if date != Dates.DateTime(0)
+        date in data_handler.available_dates || error("date not available")
+    end
+
+    regridder_type = nameof(typeof(data_handler.regridder))
+
+    snapshot = get!(data_handler._cached_regridded_fields, date) do
+        # Add here the arguments when we have more regridders
+        regrid_args = Dict(:TempestRegridder => (date,))
```
kmdeck, line 222: I am confused why the regridding needs the date.

I am a little confused by the notation also. we get the value of `data_handler._cached_regridded_fields` corresponding do the date, but then what is that called in the `do` block/where is it used in the `do` block? is that what `date` then refers to (instead of the actual date)?

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,246 @@
...
+        regrid(data_handler.regridder, regrid_args[regridder_type]...)
```
kmdeck, line 223: Does this mean we need to regrid frame by frame, or is a future option to regrid batch by batch (many frames at once)? If regridding is matrix operations by a static weight matrix, we could process many snapshots at once?

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,246 @@
...
+    snapshot(data_handler::DataHandler)
+
+Return the snapshot from `data_handler` associated to the given `time` (if relevant).
+
+The `time` has to be available in the `data_handler`.
+
+`snapshot` potentially modifies the internal state of `data_handler` and it might be a very
+expensive operation.
+"""
+function snapshot(data_handler::DataHandler, date::Dates.DateTime)
```
kmdeck, line 211: should this be called `regridded_snapshot`? and the doc string can also reference regridding

src/shared_utilities/FileReader.jl
```diff
@@ -1,549 +1,100 @@
...
-    end
+function NCFileReader(file_path::AbstractString, varname::AbstractString)
+    dataset = NCDatasets.NCDataset(file_path)
 
-    # Find the index of the start file closest to this simulation's start date
-    # Like `segment_length`, store in an array so we can modify in struct
-    date_idx0 =
-        [argmin(abs.(Dates.value(date_start) .- Dates.value.(all_dates[:])))]
+    # NOTE: We are hardcoding a few things here!
+    dimensions = dataset["lon"][:], dataset["lat"][:]
```
kmdeck, line 51: is this not what `dimensions::DIMS` is for? if not, they could also pass in the dimension_names. then we dont need to hardcode

src/shared_utilities/FileReader.jl
```diff
@@ -1,549 +1,100 @@
...
-        file_states[varname] =
-            FileState(deepcopy(data_fields), copy(date_idx0), segment_length)
+    if !isempty(available_dates)
+        issorted(available_dates) ||
+            error("Cannot process files that are not sorted in time")
     end
-    sim_info = SimInfo(date_ref, t_start)
 
-    args = (file_info, file_states, sim_info)
+    _cached_reads = Dict{Dates.DateTime, Array}()
```
kmdeck, line 59: in the future, this is where we would pre-allocate the space for the batch of data we would read in?

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -137,29 +137,21 @@ function SpaceVaryingInput(
...
-    varname::N,
-    space::S,
-) where {
-    PDS <: FileReader.PrescribedDataStatic,
-    N <: String,
-    S <: ClimaCore.Spaces.SpectralElementSpace2D,
-}
-    return FileReader.get_data_at_date(data, space, varname)
+function SpaceVaryingInput(data_handler::DataHandler)
+    return snapshot(data_handler)
```
kmdeck, line 150: repeating myself here, but I think this could be `regridded_snapshot` or `regridded_field` or something else for clarity

kmdeck: especially here it is a little confusing (outside of the data handler module)

src/shared_utilities/TimeVaryingInputs.jl
```diff
@@ -81,19 +81,54 @@ When passing single-site data
 When a `times` and `vals` are passed, `times` have to be sorted and the two arrays have to
 have the same length.
 
+=======
+When the input is a function, the signature of the function can be `func(time, args...;
```
kmdeck, line 85: nice!!

src/shared_utilities/TimeVaryingInputs.jl
```diff
@@ -81,19 +81,54 @@ When passing single-site data
...
+kwargs...)`. The function will be called with the additional arguments and keyword arguments
+passed to `evaluate!`. This can be used to access the state and the cache and use those to
+set the output field.
+
+For example:
+```julia
+CO2fromp(time, Y, p) = p.atmos.co2
```
kmdeck, line 92: but maybe this isnt a good example? is this for the analytic time varying input case?

src/shared_utilities/analytic_time_varying_input.jl
```diff
@@ -1,15 +1,22 @@
 struct AnalyticTimeVaryingInput{F <: Function} <: AbstractTimeVaryingInput
-    # func here as to be GPU-compatible (e.g., splines are not)
+    # func here has to be GPU-compatible (e.g., splines are not)
```
kmdeck, line 2: is it worth saying that it should also be performant? not allocating big amounts of memory, not looking up variables in global scope, etc

src/shared_utilities/interpolating_time_varying_inputs.jl
```diff
@@ -2,24 +2,10 @@ InterpolatingTimeVaryingInput =
...
-struct LinearInterpolation <: AbstractInterpolationMethod end
-
-"""
-    in(time, itp::InterpolatingTimeVaryingInput0D)
+    in(time, itp::InterpolatingTimeVaryingInput)
 
 Check if the given `time` is in the range of definition for `itp`.
 """
-function Base.in(time, itp::InterpolatingTimeVaryingInput0D)
+function Base.in(time, itp::InterpolatingTimeVaryingInput)
```
kmdeck, line 9: is this still being used? Also, not seeing this "InterpolatingTimeVaryingInput" (without the 0D, 2D suffix)

src/shared_utilities/DataHandling.jl
Sbozzolo: ALL_CAPS is mostly a way to tell the reader that the variable is a "global fixed parameter", with the emphasis being in "global". We don't have to do this, but I personally find it an useful additional qualifier.

(`const` in Julia doesn't even ensure that the value doesn't change.)

src/shared_utilities/DataHandling.jl
Sbozzolo: It's probably poorly worded: instead of having multiple files with ranges of dates, it is best to have one single file with alle the dates ("all the data (for a given variable) in time").

src/shared_utilities/DataHandling.jl
Sbozzolo: TempestRegridder regrids everything ahead of time, saves everything into files, and then read the a filename that is constructed with the date. That's why we need the date.

This block does the following: if `date` is a key in `data_handler._cached_regridded_fields`, then return the associated value. If `date` is not yet a key, then execute the do-block and add the return value as key. So, we change `data_handler._cached_regridded_fields` to add a new `date` with the regridded value.

experiments/standalone/Bucket/global_bucket_staticmap.jl
Sbozzolo: If you look at the lines below, it is only `regrid_dir` that uses `device_suffix`.

src/shared_utilities/SpaceVaryingInputs.jl
Sbozzolo: Renamed to `regridded_snapshot`

src/shared_utilities/DataHandling.jl
Sbozzolo: Renamed to `regridded_snapshot`

src/shared_utilities/analytic_time_varying_input.jl
Sbozzolo: Yes

src/shared_utilities/interpolating_time_varying_inputs.jl
Sbozzolo: It is. It is defined at the beginning of this file:
```julia
InterpolatingTimeVaryingInput =
    Union{InterpolatingTimeVaryingInput0D, InterpolatingTimeVaryingInput2D}
```

experiments/standalone/Bucket/global_bucket_staticmap.jl
Sbozzolo: Okay, it depends on the file. I restored it where needed.

test/shared_utilities/data_handling.jl
```diff
@@ -0,0 +1,142 @@
...
+    for FT in (Float32, Float64)
+        target_space =
+            ClimaLand.Domains.SphericalShell(;
+                radius = FT(6731e3),
+                depth = FT(1.0),
+                nelements = (4, 4),
+                npolynomial = 4,
+            ).space.surface
+        PATH = temporal_data_path
+        varname = "sp"
```
juliasloan25, line 57: these could be moved outside of the for loop for consistency with the previous test (but it doesn't really matter)

test/shared_utilities/file_readers.jl
```diff
@@ -0,0 +1,64 @@
...
+    NCDataset(albedo_bareground_path) do nc
+        available_dates = FileReaders.read_available_dates(nc)
+        @test isempty(available_dates)
+
+        ncreader = FileReaders.NCFileReader(PATH, "sw_alb")
+
+        @test ncreader.dimensions[1] == nc["lon"][:]
+        @test ncreader.dimensions[2] == nc["lat"][:]
+
+        close(ncreader)
```
juliasloan25, line 62: we could check that `OPEN_NCFILES` is updated correctly in the case without time too

test/shared_utilities/time_varying_inputs.jl
```diff
@@ -28,9 +31,16 @@ context = ClimaComms.SingletonCommsContext(device)
...
+    # Check with args and kwargs
+    fun2 = (x, y; z) -> 2x * y * z
+    input2 = TimeVaryingInputs.TimeVaryingInput(fun2)
+
+    TimeVaryingInputs.evaluate!(column_field, input2, 10.0, 20.0; z = 30.0)
+    @test Array(parent(column_field))[1] == fun2(10.0, 20.0; z = 30.0)
 end
 
-@testset "Temporal TimeVaryingInput 1D" begin
+@testset "InteprolatingTimeVaryingInput0D" begin
```
juliasloan25, line 43: typo: Inteprolating -> Interpolating

test/shared_utilities/time_varying_inputs.jl
```diff
@@ -102,3 +112,106 @@ end
         end
     end
 end
+
+@testset "InteprolatingTimeVaryingInput2D" begin
```
juliasloan25, line 116: typo: Inteprolating -> Interpolating

test/shared_utilities/time_varying_inputs.jl
```diff
@@ -102,3 +112,106 @@ end
...
+                ),
+            ),
+        )
+
+        # Right nearest point
+        target_time = available_times[9] - 1
+        TimeVaryingInputs.evaluate!(dest, input_nearest, target_time)
+
+        @test isequal(
+            Array(parent(dest)),
```
juliasloan25, line 170: why do we need to use Array here but not in the "on node" case?

juliasloan25 (commented):
Thank you for all this work Gabriele! I left many comments, and I have a few more here:
- We're opening Netcdf files and keeping them open throughout the simulation, then closing at the end. What will happen if the simulation crashes?
- There are some things left to add in future PRs that we should open issues for:
  - allow regridder-specific args: mono, regrid_dir, dates to regrid
  - regrid only simulation dates (related to dates arg in previous point)
  - add FileReaders cache
  - add DataHandling cache cleanup function/implement LRU cache
- please update the NEWS.md file too!

src/shared_utilities/DataHandling.jl
Sbozzolo: Addressed in the ClimaUtilities version.

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,317 @@
...
+the simulation grid.
+
+This is no trivial task. Among the challenges:
+- data can be large and cannot be read all in one go and/or held in memory
+- regridding onto the simulation grid can be very expensive
+- IO can be very expensive
+- CPU/GPU communication can be a bottleneck
+
+The `DataHandling` takes the divide and conquer approach: the various core tasks and
+features and split into other independent modules (chiefly `FileReaders`, and `Regridders`).
```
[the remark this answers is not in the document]
Sbozzolo: Addressed in the ClimaUtilities version

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,317 @@
...
+The key struct in `DataHandling` is the `DataHandler`. The `DataHandler` contains a
+`FileReaders`, a `Regridders`, and other metadata necessary to perform its operations (e.g.,
+target `ClimaCore.Space`). The `DataHandler` can be used for static or temporal data, and
+exposes the following key functions:
+- `regridded_snapshot(time)`: to obtain the regridded field at the given `time`. `time` has to be
+                    available in the data.
+- `available_times` (`available_dates`): to list all the `times` (`dates`) over which the
+                    data is defined.
+- `previous_time(time/date)` (`next_time(time/date)`): to obtain the time of the snapshot
+                         before the given `time` or `date`. This can be used to compute the
```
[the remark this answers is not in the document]
Sbozzolo: Addressed in the ClimaUtilities version

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,317 @@
...
+                target_space::ClimaCore.Spaces.AbstractSpace;
+                reference_date::Dates.DateTime = Dates.DateTime(1979, 1, 1),
+                t_start::AbstractFloat = 0.0,
+                regridder_type = :TempestRegridder)
+
+Create a `DataHandler` to read `varname` from `file_path` and remap it to `target_space`.
+
+The DataHandler maintains a cache of Fields that were previously computed.
+
+TODO: Add function to clear cache, and/or CACHE_MAX_SIZE (this will probably require
```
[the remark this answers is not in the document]
Sbozzolo: https://github.com/CliMA/ClimaUtilities.jl/issues/21

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,317 @@
...
+===================
+
+Time/date information will be ignored for static input files. (They are still set to make
+everything more type stable.)
+
+- `reference_date`: Calendar date corresponding to the start of the simulation.
+- `t_start`: Simulation time at the beginning of the simulation. Typically this is 0
+             (seconds), but if might be different if the simulation was restarted.
+- `regridder_type`: What type of regridding to perform. Currently, the only one implemented
+                    is `:tempest` to use `TempestRemap`. `TempestRemap` regrids everything
```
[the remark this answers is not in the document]
Sbozzolo: Fixed in ClimaUtilities (there are two regridders now)

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,317 @@
...
+            (target_space, mktempdir(), varname, file_path),
+    )
+    RegridderConstructor = getfield(Regridders, regridder_type)
+    regridder = RegridderConstructor(regridder_args[regridder_type]...)
+
+    # NOTE: this is not concretely typed
+    _cached_regridded_fields = Dict{Dates.DateTime, ClimaCore.Fields.Field}()
+
+    available_dates = file_reader.available_dates
+    times_s = Second.(available_dates .- reference_date) ./ Second(1)
```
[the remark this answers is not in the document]
Sbozzolo: I copied the round from somewhere else. Maybe we don't need it at all.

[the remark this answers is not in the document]
Sbozzolo: I removed the round, we can put it back if it causes problems.

src/shared_utilities/FileReaders.jl
```diff
@@ -0,0 +1,163 @@
...
+import Dates
+import NCDatasets
+
+abstract type AbstractFileReader end
+
+# We allow multiple NCFileReader to share the same underlying NCDataset. For this, we put
+# all the NCDataset into a dictionary where we keep track of them. OPEN_NCFILES is
+# dictionary that maps file paths to a Tuple with the first element being the NCDataset and
+# the second element being a Set of Strings, the variables that are being read from that
+# file. Every time a NCFileReader is created, this Set is modified by adding or removing a
```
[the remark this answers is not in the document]
Sbozzolo: You remove it when you call `Base.close`

src/shared_utilities/FileReaders.jl
```diff
@@ -0,0 +1,163 @@
...
+            dimensions = (lon, lat)
+        else
+            error("$file_path does not contain information about dimensions")
+        end
+    end
+    available_dates = read_available_dates(dataset)
+
+    if !isempty(available_dates)
+        issorted(available_dates) ||
+            error("Cannot process files that are not sorted in time")
```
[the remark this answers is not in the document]
Sbozzolo: Added in ClimaUtilities.

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,317 @@
...
+`regridded_snapshot` potentially modifies the internal state of `data_handler` and it might be a very
+expensive operation.
+
+TODO: Add `regridded_snapshot!`
+"""
+function regridded_snapshot(data_handler::DataHandler, date::Dates.DateTime)
+
+    # Dates.DateTime(0) is the cache key for static maps
+    if date != Dates.DateTime(0)
+        date in data_handler.available_dates || error("date not available")
```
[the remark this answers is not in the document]
Sbozzolo: Added in ClimaUtilities

src/shared_utilities/Regridders.jl
```diff
@@ -1,11 +1,145 @@
...
+"""
+    Regridders
+
+The `Regridders` module implement structs and functions to remap datasets to simulation
+grids.
+
+Currently, the only scheme implemented is `TempestRegridder`, which uses
+`ClimaCoreTempestRemap`.
+
+The key function exposed by `Regridder` is the `regrid` method.
```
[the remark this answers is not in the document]
Sbozzolo: Fixed in ClimaUtilities

src/shared_utilities/Regridders.jl
```diff
@@ -1,11 +1,145 @@
...
+
+    target_space::SPACE
+    regrid_dir::STR1
+    outfile_root::STR2
+    varname::STR3
+    mono::Bool
+end
+
+"""
+    TempestRegridder(target_space::ClimaCore.Spaces.AbstractSpace,
```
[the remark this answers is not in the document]
Sbozzolo: Fixed in ClimaUtilities

src/shared_utilities/Regridders.jl
```diff
@@ -1,11 +1,145 @@
...
+    return read_from_hdf5(
+        regridder.regrid_dir,
+        regridder.outfile_root,
+        date,
+        regridder.varname,
+        regridder.target_space,
+    )
+end
 
 export hdwrite_regridfile_rll_to_cgll
```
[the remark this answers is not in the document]
Sbozzolo: Fixed in ClimaUtilities

src/shared_utilities/Regridders.jl
```diff
@@ -1,11 +1,145 @@
...
+"""
+module Regridders
+
 using Dates
-using DocStringExtensions
+
+import ClimaComms
+import ClimaCore
+import ClimaCore: Spaces
+import ClimaCoreTempestRemap
```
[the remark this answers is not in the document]
Sbozzolo: I am personally not a fan of this pattern. The main reason is that it adds one layer of indirection to what is happening and readers have to figure out what CCTR means. The module is called only five times, all in the same function, so I don't think there's good reasons to shorten its name.

src/shared_utilities/Regridders.jl
```diff
@@ -309,21 +435,27 @@ function hdwrite_regridfile_rll_to_cgll(
...
+            NCDatasets.NCDataset(datafile_cgll, "r") do ds_wt
+                (
+                    # read the data in, and remove missing type (will error if missing data is present)
+                    offline_outvector = nomissing(Array(ds_wt[varname])[:, :]), # ncol, times
+                    times = read_available_dates(ds_wt),
+                )
+            end
+
+        # Use dummy date when there are no real dates
+        isempty(times) && (times = [DateTime(0)])
```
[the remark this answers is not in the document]
Sbozzolo: You could make it an empty DateTime array, but the reason I added a number is to use the same caching mechanism used by temporal date. The cache is indexed over the date, so I add a fake date to retrieve that value.

src/shared_utilities/FileReaders.jl
```diff
@@ -0,0 +1,163 @@
+"""
+    FileReaders
```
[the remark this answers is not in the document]
Sbozzolo: Added in ClimaUtilities

src/shared_utilities/DataHandling.jl
```diff
@@ -0,0 +1,317 @@
+"""
+    DataHandling
```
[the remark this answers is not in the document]
Sbozzolo: Added in ClimaUtilities

src/standalone/Bucket/Bucket.jl
```diff
@@ -116,32 +116,24 @@ This particular method can only be used with global runs.
 """
 function PrescribedBaregroundAlbedo{FT}(
     α_snow::FT,
-    regrid_dirpath::String,
     surface_space::ClimaCore.Spaces.AbstractSpace;
     varnames = ["sw_alb"],
-    get_infile::Function = Bucket.bareground_albedo_dataset_path,
+    albedo_file_path::AbstractString = Bucket.bareground_albedo_dataset_path(),
```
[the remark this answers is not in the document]
Sbozzolo: This should no longer be the case after https://github.com/CliMA/ClimaCoupler.jl/commit/31adbb839182af54b3482af9867a7330fbd16eab

src/shared_utilities/FileReaders.jl
```diff
@@ -0,0 +1,163 @@
...
+
+    """A tuple of arrays with the various physical dimensions where the data is defined
+    (e.g., lon/lat)"""
+    dimensions::DIMS
+
+    """A vector of DateTime collecting all the available dates in the file"""
+    available_dates::DATES
+
+    """NetCDF file opened by NCDataset"""
+    dataset::NC
```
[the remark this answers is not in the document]
Sbozzolo: Added in ClimaUtilities

test/shared_utilities/file_readers.jl
Sbozzolo: It wouldn't test anything that we are not testing in the case with time

test/shared_utilities/time_varying_inputs.jl
Sbozzolo: Fixed in ClimaUtilities

test/shared_utilities/time_varying_inputs.jl
Sbozzolo: Fixed in ClimaUtilities

test/shared_utilities/time_varying_inputs.jl
Sbozzolo: I added the Array there too

Sbozzolo:
Implemented in #560

PR #466, opened 2024-01-24
Add PFTs to ClimaLandSimulations

Espeer5:
## Purpose 
Adds the ability for users to define PFT schemes which construct the required site-varying parameters to run the model, and defines a default PFT scheme mirroring the one in CLM5. Will allow `fluxnet_simulation` to be run with or without default or custom PFTs.

## Content
 - Adds a PFT struct which consists mainly of a `NamedTuple` which maps each required site-varying parameter to a value
   - Uses an inner constructor to validate all fields are populated
 - Provides a function `params_from_pfts` which given a PFT scheme (a `Vector` of `PFT` structs) and a `Vector` giving the percentage of cover of each PFT in the scheme for a given area, defines each required parameter variable as the linear combination of the parameters for each PFT, weighted by the cover fraction.
 - Provides default PFT scheme with (some) sourced parameter values

## To-do
 - [ ] Source parameter values for plant hydraulics parameters for each PFT
 - [ ] Separate site-specific non-PFT params, and site-to-site constants such that they are still read in when using PFTs
 - [ ] Use multiple dispatch to handle whether parameters are read from site file or from PFT
 - [ ] Create a tutorial for using PFTs in fluxnet sim
 - [ ] Change to dominant PFT type rather than dot product

----


lib/ClimaLandSimulations/src/utilities/pfts.jl
```diff
@@ -0,0 +1,614 @@
...
+        α_PAR_leaf = 0.07, # CLM5.0 Table 2.3.1
+        α_NIR_leaf = 0.35, # CLM5.0 Table 2.3.1
+        τ_PAR_leaf = 0.05, # CLM5.0 Table 2.3.1
+        τ_NIR_leaf = 0.10, # CLM5.0 Table 2.3.1
+
+        # Canopy Energy Model
+        ac_canopy = 2.5e3, # Taken from site params
+
+        # Stomatal Conductance Model
+        g1 = 2.35, # CLM5.0 Table 2.9.1
```
Espeer5, line 116: I need a little bit of help understanding the `g1` value units... In our code `g1` is around 150, but in the CLM tables `g1` is on the order of 1 but i don't see the unit with it: [CLM g1 table](https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/tech_note/Photosynthesis/CLM50_Tech_Note_Photosynthesis.html)

lib/ClimaLandSimulations/src/utilities/pfts.jl
```diff
@@ -0,0 +1,614 @@
...
+#    10.5194/gmd-2022-11. 
+#    abb. as (Oliver et al. 2022)
+
+# Define the default PFTs that are used in the model.
+
+NET_Temp = Pft(
+    "Temperate Needleleaf Evergreen Tree",
+    (
+        # Radiative Transfer Model
+        Ω = 0.74,          # Chen et al. 2012 Table 1
```
Espeer5, line 106: Something I was thinking about here is that our radiative transfer model doesn't distinguish between stem area and leaf area because the MODIS LAI contains both LAI and SAI, but we have been using the radiative transfer parameters from [CLM](https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/tech_note/Surface_Albedos/CLM50_Tech_Note_Surface_Albedos.html) for leaf area only and the stem values can vary quite widely from the leaf values. Perhaps we should be doing something with a linear combination of the steam and leaf values for each PFT? But then we don't really know the fraction of leaf to stem. Perhaps there is a characteristic stem to leaf ratio for each PFT and we could use this to weight the reflectances and transmittances?

lib/ClimaLandSimulations/src/utilities/pfts.jl
```diff
@@ -0,0 +1,614 @@
...
+        # Hydraulics Model and Plant Parameters
+        SAI = 2.50,                # Guess
+        f_root_to_shoot = 3.50,    # Guess
+        K_sat_plant = 2.78e-6,     # Guess
+        ψ63 = -4.0 / 0.0098,       # Guess
+        Weibull_param = 4.00,      # Taken from site params
+        a = 0.05 * 0.0098,         # Taken from site params
+        capacity = 15.00,          # Guess
+        plant_S_s = 1e-2 * 0.0098, # Taken from site params
+        rooting_depth = 3.90,      # Bonan Table 2.3
```
Espeer5, line 130: Any help finding sources for plant hydraulics parameters or accurate guesses of the values would be really helpful. I haven't found anywhere in the literature matching these values to PFTs. 

Similarly for the `ac_canopy`.

Espeer5:
i was able to run the Ozark experiment using PFTs this morning setting the Ozark site to a pure temperate deciduous broadleaf forest, but using prior values for the `g1`. 
[diurnals.pdf](https://github.com/CliMA/ClimaLSM.jl/files/14043313/diurnals.pdf)
[fingerprint.pdf](https://github.com/CliMA/ClimaLSM.jl/files/14043325/fingerprint.pdf)
[timeseries_fluxes.pdf](https://github.com/CliMA/ClimaLSM.jl/files/14043326/timeseries_fluxes.pdf)
[timeseries_H2O.pdf](https://github.com/CliMA/ClimaLSM.jl/files/14043328/timeseries_H2O.pdf)
looks encouraging but I still need to find the correct unit conversion for `g1` from CLM and correct values for the energy and hydraulics parameters.

AlexisRenchon:
just a comment, I think it's good practice to add your initial in front of the branch name, i.e., `es/PFTs` (but no need to rename it, not essential)

lib/ClimaLandSimulations/src/fluxnet_simulation.jl
```diff
@@ -1,4 +1,5 @@
 export fluxnet_simulation
+include(joinpath(ClimaLandSimulations_dir, "src", "utilities", "pfts.jl"))
```
AlexisRenchon, line 2: I think you should include in the module instead lib/ClimaLandSimulations/src/ClimaLandSimulations.jl

lib/ClimaLandSimulations/src/fluxnet_simulation.jl
```diff
@@ -34,11 +35,15 @@ function fluxnet_simulation(
         "utilities",
         "timestepper_setup.jl",
     ),
+    pft_pcts = "None",
 ) # why is there 2?
```
AlexisRenchon, line 39: uh I need to delete my comments - note to self

lib/ClimaLandSimulations/src/utilities/pfts.jl
```diff
@@ -0,0 +1,614 @@
...
+            if param ∉ defined_param_names
+                error("PFT $name is missing parameter $param")
+            end
+        end
+        new(name, parameters)
+    end
+end
+
+
+# The default PFTs used in the model use parameter values drawn from defferent
```
AlexisRenchon, line 71: typo: different. 

Also I think it's okay to have this text in the docstring (what do you think @kmdeck)

AlexisRenchon (commented):
Hey Ed, this is a really neat PR! Thanks a lot! I will go back to it a bit later. I like that everything is in functions or struct, as we are trying to write clean code now (as opposed to experiment which was more hard-coding friendly). We could discuss more with Kat and Julia about best practice for this PR.

lib/ClimaLandSimulations/src/fluxnet_simulation.jl
```diff
@@ -34,11 +35,15 @@ function fluxnet_simulation(
         "utilities",
         "timestepper_setup.jl",
     ),
+    pft_pcts = "None",
```
Espeer5, line 38: P.S this is just a temporary fix to be able to run the model with some PFTs right now, will clean this up later.

lib/ClimaLandSimulations/src/utilities/pfts.jl
kmdeck: I double checked. the typical units in the literature are sqrt(kPa). We use SI units throughout simulations which would mean we use sqrt(Pa). 1 sqrt(kPa) = 1 sqrt(kPa) * (Pa/kPa)^(1/2) sqrt(Pa) = sqrt(1000) sqrt(Pa).

lib/ClimaLandSimulations/src/utilities/pfts.jl
kmdeck: We can also get rooting depth from this beta parameter: CLM Table 2.11.1, in the chapter on plant hydraulics. I can look this up tomorrow (a little math to convert from beta to rooting depth) if Bonan doesnt have it

plant_S_s can be constant. It applies in a regime that Im not sure is super physical in plants - when the plant pores are saturated. im not sure how important it is but I dont think we have data on it.

CLM says that the root to shoot ratio is a PFT parameter...root-to-shoot ratio (PFT parameter) 
and I think a lot of these plant hydraulics params should be too. Maybe @braghiere knows if there is a file with them if they are not published in the docs?

lib/ClimaLandSimulations/src/utilities/pfts.jl
```diff
@@ -0,0 +1,614 @@
...
+        # to work and ensure the ordering of the parameters.
+        push!(pft_df, pft.parameters)
+    end
+    
+    # Instantiate a variable with the correct name for the weighted average of 
+    # each parameter.
+    for param in param_list
+        # each param value is the dot product of the PFT cover percentages and
+        # the parameter values
+        param_value = pft_cover ⋅ pft_df[:, param]
```
kmdeck, line 607: can change to just taking the params from the max(pft_cover) pft type as discussed today?

lib/ClimaLandSimulations/src/utilities/pfts.jl
braghiere: I think it is fine if we consider the optical depth is made only of LAI. Other LSMs do it without losing much.

lib/ClimaLandSimulations/src/utilities/pfts.jl
braghiere: The acronym for this parameter is froot_leaf    = allocation parameter: new fine root C per new leaf C (gC/gC). Refer to: https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/tech_note/Plant_Hydraulics/CLM50_Tech_Note_Plant_Hydraulics.html#vertical-root-distribution

See CLM PFT file example: https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/lnd/clm2/pftdata/pft-physiology-cn.c070601.readme

AlexisRenchon:
@Espeer5 is this PR outdated? (as you merged the other one), if so we can close this one?

Espeer5:
Closing - this is outdated now as PFTs were added to source in #493

PR #469, opened 2024-01-26
Add a global bucket run to the buildkite pipeline, on GPU

kmdeck:
## Purpose 
Adds a simple global land model run on GPU. 

## To-do
Future PR: add a version of this with land/sea mask, time and spatially varying input
Future PR: use existing Atmos diagnostics and analysis tools

## Content
- adds CairoMakie to experiments' Project.toml
- uses analytic TimeVaryingInput0D
- does not employ a land/sea mask
- the test is mainly that it runs and looks reasonable. We could also in principle test performance here, since this is a very simple system with no file IO

I also fixed a bug in that the drivers in the bucket tutorial were not being updated (no driver callback). There is now a driver callback.

Review checklist

I have:

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -23,7 +23,7 @@ steps:
       - "julia --project -e 'using Pkg; Pkg.status()'"
 
       - echo "--- Instantiate experiments"
-      - "julia --project=experiments -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
+      - "julia --project=experiments -e 'using Pkg; Pkg.develop(;path=\".\"); Pkg.instantiate(;verbose=true)'"
       - "julia --project=experiments -e 'using Pkg; Pkg.status()'"
 
       - echo "--- Instantiate lib/ClimaLandSimulations"
@@ -78,6 +78,17 @@ steps:
         command: "julia --color=yes --project=experiments experiments/standalone/Soil/water_conservation.jl"
         artifact_paths: "experiments/standalone/Soil/water_conservation*png"
 
+      - label: "Global Bucket on CPU"
+        key: "global_bucket_cpu"
+        command: "julia --color=yes --project=experiments experiments/standalone/Bucket/global_bucket.jl"
+        artifact_paths: "experiments/standalone/Bucket/artifacts/*cpu*"
+
+      - label: "Compare GPU bucket with CPU bucket"
+        command: "julia --color=yes --project=experiments experiments/standalone/Bucket/compare_gpu_cpu_output.jl"
+        depends_on:
+          - "global_bucket_cpu"
+          - "global_bucket_gpu"
+
   - group: "lib/ClimaLandSimulations"
     steps:
       - label: "Ozark figures Makie"
@@ -91,6 +102,14 @@ steps:
         agents:
           slurm_mem_per_cpu: 8G
           slurm_gpus: 1
+
+      - label: "Global Bucket on GPU"
+        key: "global_bucket_gpu"
+        command: "julia --color=yes --project=experiments experiments/standalone/Bucket/global_bucket.jl"
+        agents:
+          slurm_mem: 8G
+          slurm_gpus: 1
+        artifact_paths: "experiments/standalone/Bucket/artifacts/*gpu*"
 
   - wait
 
--- a/docs/tutorials/standalone/Bucket/bucket_tutorial.jl
+++ b/docs/tutorials/standalone/Bucket/bucket_tutorial.jl
@@ -327,14 +327,18 @@ prob = SciMLBase.ODEProblem(
 );
 
 # We need a callback to get and store the auxiliary fields, as they
-# are not stored by default.
+# are not stored by default. We also need a callback to update the
+# drivers (atmos and radiation)
 saveat = collect(t0:Δt:tf);
 saved_values = (;
     t = Array{Float64}(undef, length(saveat)),
     saveval = Array{NamedTuple}(undef, length(saveat)),
 );
-
-cb = ClimaLSM.NonInterpSavingCallback(saved_values, saveat);
+saving_cb = ClimaLSM.NonInterpSavingCallback(saved_values, saveat);
+updateat = copy(saveat)
+updatefunc = ClimaLSM.make_update_drivers(bucket_atmos, bucket_rad)
+driver_cb = ClimaLSM.DriverUpdateCallback(updateat, updatefunc)
+cb = SciMLBase.CallbackSet(driver_cb, saving_cb)
 
 sol = SciMLBase.solve(prob, ode_algo; dt = Δt, saveat = saveat, callback = cb);
 
--- a/experiments/Manifest.toml
+++ b/experiments/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.10.0"
 manifest_format = "2.0"
-project_hash = "c1c208a541a75df9e58a2ec32c1ca80f6eaf200a"
+project_hash = "50cbdda495063d9b4566ea0057c1b351f04893e0"
 
 [[deps.ADTypes]]
 git-tree-sha1 = "41c37aa88889c171f1300ceac1313c06e891d245"
@@ -26,6 +26,16 @@ weakdeps = ["ChainRulesCore", "Test"]
     AbstractFFTsChainRulesCoreExt = "ChainRulesCore"
     AbstractFFTsTestExt = "Test"
 
+[[deps.AbstractLattices]]
+git-tree-sha1 = "222ee9e50b98f51b5d78feb93dd928880df35f06"
+uuid = "398f06c4-4d28-53ec-89ca-5b2656b7603d"
+version = "0.3.0"
+
+[[deps.AbstractTrees]]
+git-tree-sha1 = "faa260e4cb5aba097a73fab382dd4b5819d8ec8c"
+uuid = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
+version = "0.4.4"
+
 [[deps.Adapt]]
 deps = ["LinearAlgebra", "Requires"]
 git-tree-sha1 = "cde29ddf7e5726c9fb511f340244ea3481267608"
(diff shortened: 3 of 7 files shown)


.buildkite/pipeline.yml
```diff
@@ -92,6 +96,13 @@ steps:
           slurm_mem_per_cpu: 8G
           slurm_gpus: 1
 
+      - label: "Global Bucket on GPU"
+        command: "julia --color=yes --project=experiments experiments/standalone/Bucket/global_bucket.jl"
+        agents:
+          slurm_mem_per_cpu: 8G
+          slurm_gpus: 1
+        artifact_paths: "experiments/standalone/Bucket/anim*.mp4"
+
```
Sbozzolo, line 105: We could also add an extra step to compare CPU and GPU results as we do in ClimaAtmos:

https://github.com/CliMA/ClimaAtmos.jl/blob/fb7b56018491f1d457f708920ca209ecd08b717e/.buildkite/pipeline.yml#L607-L619

(We would probably have to save something that is meaningful to compare)

docs/tutorials/standalone/Bucket/bucket_tutorial.jl
```diff
@@ -327,14 +327,18 @@ prob = SciMLBase.ODEProblem(
...
+# drivers (atmos and radiation)
 saveat = collect(t0:Δt:tf);
 saved_values = (;
     t = Array{Float64}(undef, length(saveat)),
     saveval = Array{NamedTuple}(undef, length(saveat)),
 );
-
-cb = ClimaLSM.NonInterpSavingCallback(saved_values, saveat);
+saving_cb = ClimaLSM.NonInterpSavingCallback(saved_values, saveat);
+updateat = deepcopy(saveat)
```
Sbozzolo, line 338: I think we want `copy`, not `deepcopy` in this particular case.

```julia
julia> a = [1, 2, 3]
3-element Vector{Int64}:
 1
 2
 3

julia> b = copy(a)
3-element Vector{Int64}:
 1
 2
 3

julia> b[2] = 10
10

julia> b
3-element Vector{Int64}:
  1
 10
  3

julia> a
3-element Vector{Int64}:
 1
 2
 3
```

experiments/Manifest.toml
Sbozzolo, line 1: Do want to check this in?

experiments/standalone/Bucket/global_bucket.jl
```diff
@@ -0,0 +1,200 @@
...
+    make_set_initial_cache,
+    PrescribedAtmosphere,
+    PrescribedRadiativeFluxes
+using ClimaLSM.TimeVaryingInputs
+
+function compute_extrema(v)
+    maxes = [maximum(u) for u in v]
+    mins = [minimum(u) for u in v]
+    return (minimum(mins), maximum(maxes))
+end
```
Sbozzolo, line 27: Can you please add a docstring to explain what is the expected input?

experiments/standalone/Bucket/global_bucket.jl
```diff
@@ -0,0 +1,200 @@
...
+    TimeVaryingInput(P_atmos),
+    ref_time,
+    h_atmos,
+);
+
+# Prescribed radiation -- a prescribed downwelling SW diurnal cycle, with a
+# peak at local noon, and a prescribed downwelling LW radiative
+# flux, assuming the air temperature is on average 275 degrees
+# K with a diurnal amplitude of 5 degrees K:
+SW_d = (t) -> @. max(1361 * sin(2π * t / 86400 + 7200));
```
Sbozzolo, line 95: Why do you need `@.` here? Is `SW_d` going to be evaluated on an array of `t`s?

experiments/standalone/Bucket/global_bucket.jl
```diff
@@ -0,0 +1,200 @@
...
+    p,
+);
+
+saveat = collect(t0:(Δt * 3):tf);
+saved_values = (;
+    t = Array{Float64}(undef, length(saveat)),
+    saveval = Array{NamedTuple}(undef, length(saveat)),
+);
+saving_cb = ClimaLSM.NonInterpSavingCallback(saved_values, saveat);
+updateat = deepcopy(saveat)
```
Sbozzolo, line 136: Same as before, I think `copy` should be used here.

experiments/standalone/Bucket/global_bucket.jl
```diff
@@ -0,0 +1,200 @@
...
+    outfile = joinpath(outdir, string("anim_", field_name, ".mp4"))
+    record(fig, outfile, 1:nframes; framerate = framerate) do frame
+        CairoMakie.heatmap!(
+            longpts,
+            latpts,
+            field_ts[frame],
+            colorrange = clims,
+        )
+    end
+end
```
Sbozzolo, line 200: Atmos found animations to be difficult to inspect and debug, maybe we can also make some of the frames accessible?

Sbozzolo (commented):
Overall it looks good. Two comments for future directions:
- Maybe the next PR could be add a "everything" test, where we turn on all the components in their most complex iteration (e.g., reading data from file) and we can keep expanding as we build new components
- For plotting and diagnostics, it would be nice to use nice to use the same Atmos interface in the future (to avoid reinventig the wheel)

For this PR, I'd suggest adding some minimal timing information so that we can have a sense of how fast/slow things are. At this point, just adding `@time` might be enough.

experiments/Manifest.toml
kmdeck: we have been checking it in, because it depends on the dev version of ClimaLSM, not the latest registered version.

experiments/standalone/Bucket/global_bucket.jl
kmdeck: no need for it, thanks!

docs/tutorials/standalone/Bucket/bucket_tutorial.jl
kmdeck: thanks! i changed this.

experiments/standalone/Bucket/global_bucket.jl
kmdeck: likewise, changed this.

.buildkite/pipeline.yml
kmdeck: I will save the remapped/interpolated state at the final step. I agree 100% that we should be using the analysis and diagnostics packages you built for ClimaAtmos, but would like to implement that in another PR.

kmdeck:
> Overall it looks good. Two comments for future directions:
> 
>     * Maybe the next PR could be add a "everything" test, where we turn on all the components in their most complex iteration (e.g., reading data from file) and we can keep expanding as we build new components
> 
>     * For plotting and diagnostics, it would be nice to use nice to use the same Atmos interface in the future (to avoid reinventig the wheel)
> 
> 
> For this PR, I'd suggest adding some minimal timing information so that we can have a sense of how fast/slow things are. At this point, just adding `@time` might be enough.

Thank you! I like the idea for the next PR, agree on the 2nd part, and added in @time.

experiments/standalone/Bucket/global_bucket.jl
kmdeck: good call. they really are hard to inspect. In this case the output is extremely boring (constant everywhere, no lat/lon dependence in forcing), so I will also make timeseries of the (single valued except for roundoff) value of the state vars as a function of time.

kmdeck: and turn off the anims for now

experiments/standalone/Bucket/compare_gpu_cpu_output.jl
```diff
@@ -0,0 +1,9 @@
+using DelimitedFiles
+using Statistics
+import ClimaLSM
+outdir = joinpath(pkgdir(ClimaLSM), "experiments/standalone/Bucket/artifacts")
+cpu_state = readdlm(joinpath(outdir, "tf_state_cpu.txt"), ',')
+gpu_state = readdlm(joinpath(outdir, "tf_state_gpu.txt"), ',')
+@show mean(cpu_state .- gpu_state)
+@show eps(Float64)
```
juliasloan25, line 8: do we want to print `eps(Float64)` here? I think it would get printed if the assertion fails which is maybe all we need (not a big deal either way though)

.buildkite/pipeline.yml
```diff
@@ -92,6 +103,14 @@ steps:
           slurm_mem_per_cpu: 8G
           slurm_gpus: 1
 
+      - label: "Global Bucket on GPU"
+        key: "global_bucket_gpu"
+        command: "julia --color=yes --project=experiments experiments/standalone/Bucket/global_bucket.jl"
+        agents:
+          slurm_mem_per_cpu: 8G
```
juliasloan25, line 110: should we specify `slurm_mem_per_cpu` for GPU runs? I see in ClimaAtmos they just specify `slurm_mem`, I'm not sure what's best or if they're equivalent https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/pipeline.yml#L691-L692

juliasloan25 (approved):
This looks great to me! Thank you Kat :)

experiments/standalone/Bucket/compare_gpu_cpu_output.jl
kmdeck: It actually wasnt printing it! (also, apologies for all the failed buildkite runs. no GPU locally so i am trial and error debugging this using buildkite. not very efficient, I should have just worked on the cluster!)

and, I dont know what a good threshold is. the value of the error is 6e-15. eps(Float64) is ~2e-16. The mean error will be lower than the individual value error, so perhaps this is artificially low, but in any case it seems like good agreement.

kmdeck: I am using 100*eps(Float64) now.

.buildkite/pipeline.yml
kmdeck: Ill update this

experiments/standalone/Bucket/compare_gpu_cpu_output.jl
juliasloan25: I think 100*eps(Float64) or 50*eps(Float64) (~1e-14) would be good!

PR #475, opened 2024-02-01
fix Bucket snow/soil warnings

juliasloan25:
## Purpose 
I noticed that we're getting a bunch of warnings in the Bucket snow and soil tests that look like this:

> ┌ Warning: Replacing docs for `Main.var"##Bucket soil tests#507".ref_time :: Union{}` in module `Main.var"##Bucket soil tests#507"`
└ @ Base.Docs docs/Docs.jl:243
┌ Warning: Replacing docs for `Main.var"##Bucket soil tests#507".precip :: Union{}` in module `Main.var"##Bucket soil tests#507"`
└ @ Base.Docs docs/Docs.jl:243

(see GHA run [here](https://github.com/CliMA/ClimaLand.jl/actions/runs/7747825838/job/21129149749))
I think this is because we had some lines containing `"Atmos"` or `"Radiation"`, which seem like they were meant to be comments but are being interpreted by Julia as docstrings. This PR changes them to comments, which removes all the warnings from those tests.

### DIFF
--- a/test/standalone/Bucket/snow_bucket_tests.jl
+++ b/test/standalone/Bucket/snow_bucket_tests.jl
@@ -57,8 +57,7 @@ for FT in (Float32, Float64)
     init_temp(z::FT, value::FT) where {FT} = FT(value)
     for i in 1:3
         @testset "Conservation of water and energy I (snow present), FT = $FT" begin
-            "Radiation"
-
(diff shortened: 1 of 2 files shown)


kmdeck (approved):
good catch and nice fix :)

PR #479, opened 2024-02-05
add CoupledAtmos methods

### LINKED ISSUE #476: turbulent_fluxes failure with CoupledAtmos
## Describe the bug
When running the bucket model with a CoupledAtmosphere, turbulent_fluxes returns a `NamedTuple`. This is what ClimaCoupler was doing previously and isn't a bug itself, but this change was made at the same time that `update_aux` was changed to broadcast over `turbulent_fluxes` (see change [here](https://github.com/CliMA/ClimaLand.jl/compare/v0.6.0...v0.7.0#diff-da195e37edf7ac71968ff125c60ee4eef7886667488ff08b6ecef0618c904734L558-R569)). Broadcasting over NamedTuples isn't allowed, so this gives an error.

When the bucket is run with a PrescribedAtmosphere, `turbulent_fluxes` returns a field, which we can broadcast over, which is why we didn't catch this sooner. It would be nice to add a test that sets up the bucket with a CoupledAtmosphere and CoupledRadiativeFluxes and checks that everything can run smoothly.

failing buildkite run from coupler: https://buildkite.com/clima/climacoupler-ci/builds/2547

fixed by #479

juliasloan25:
## Purpose 
Add methods for compatibility with ClimaCoupler:
- turbulent_fluxes(::CoupledAtmosphere, ...)
- initialize_drivers(::CoupledAtmosphere, ...)

This will allow us to remove these functions from ClimaCoupler

closes #476

### DIFF
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -313,7 +313,7 @@ function turbulent_fluxes_at_a_point(
 end
 
 """
-    surface_fluxes(atmos::CoupledAtmosphere,
+    turbulent_fluxes(atmos::CoupledAtmosphere,
                     model::AbstractModel,
                     Y,
                     p,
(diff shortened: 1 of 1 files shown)


src/shared_utilities/drivers.jl
```diff
@@ -660,6 +656,29 @@ function initialize_drivers(a::PrescribedAtmosphere{FT}, coords) where {FT}
     return vars.drivers
 end
 
+"""
+    initialize_drivers(r::CoupledAtmosphere{FT}, coords) where {FT}
```
kmdeck, line 660: r -> a in doc string argument

kmdeck (approved):
clean fix! thank you.

PR #480, opened 2024-02-05
Refactor AutotrophicRespirationParameters to use ClimaParameters

nefrathenrici:
This is the first in a series of PRs changing the parameter constructors to use ClimaParameters.

### Content 
The changes here are straightforward, since the values for `AutotrophicRespirationParameters` never change.
The core changes are in `src/standalone/Vegetation/autotrophic_respiration.jl`
The former `AutotrophicRespirationParameters` constructor has been removed and replaced by two constructors. For both constructors, parameters can be passed in via kwargs.
1. `AutotrophicRespirationParameters(FT; kwargs...)` takes a float type and uses the param values found in the default climaparameters toml dict.
2. `AutotrophicRespirationParameters(toml_dict; kwargs...)` takes a toml dictionary, and populates the struct from the dict.

Since all of the struct values are constant, I have replaced calls to the old constructor with the new `AutotrophicRespirationParameters(FT)` constructor.

### DIFF
--- a/ext/CreateParametersExt.jl
+++ b/ext/CreateParametersExt.jl
@@ -1,11 +1,14 @@
 module CreateParametersExt
 
-import ClimaLand.Parameters.LandParameters
 import Thermodynamics.Parameters.ThermodynamicsParameters
 import Insolation.Parameters.InsolationParameters
 import SurfaceFluxes.Parameters.SurfaceFluxesParameters
 import SurfaceFluxes.UniversalFunctions as UF
 import CLIMAParameters as CP
+
+# Parameter structs to override
+import ClimaLand.Parameters.LandParameters
+import ClimaLand.Canopy.AutotrophicRespirationParameters
 
 LandParameters(::Type{FT}) where {FT <: AbstractFloat} =
     LandParameters(CP.create_toml_dict(FT))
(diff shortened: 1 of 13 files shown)


ext/CreateParametersExt.jl
```diff
@@ -51,4 +54,52 @@ function LandParameters(toml_dict::CP.AbstractTOMLDict)
...
+    AutotrophicRespirationParameters(toml_dict; kwargs...)
+    AutotrophicRespirationParameters(;
+        ne,
+        ηsl,
+        σl,
+        μr,
+        μs,
+        f1,
+        f2,
+    )
```
juliasloan25, line 69: Is this constructor meant to be in this docstring?

ext/CreateParametersExt.jl
```diff
@@ -51,4 +54,52 @@ function LandParameters(toml_dict::CP.AbstractTOMLDict)
     )
 end
 
+# TODO: Should this be organized into multiple files?
```
juliasloan25, line 57: would be good to remove this TODO before merging! I think it's okay to keep everything in this file but you have a better vision of how large this might become so whatever you think is best works for me

ext/CreateParametersExt.jl
juliasloan25: Actually it probably is good to document what's included in AutotrophicRespirationParameters, so users know what they can overwrite, but this sort of makes it look like a third constructor option which could be misleading

juliasloan25 (approved):
Looks good! Just a couple comments

ext/CreateParametersExt.jl
juliasloan25: Or maybe it's better for users to just look at `AutotrophicRespirationParameters` in src/standalone/Vegetation/autotrophic_respiration.jl if they want to see what they can overwrite

nefrathenrici: It's probably better to remove it - as you said, users can look at the fieldnames if they want to circumvent these constructors.

ext/CreateParametersExt.jl
nefrathenrici: We can always split it up easily later if needed, so I will just remove this for now.

PR #481, opened 2024-02-06
Update (Optimality)FarquharParameters to use ClimaParameters

nefrathenrici:
Similar to #480.

Core changes are to the constructors in `src/standalone/Vegetation/photosynthesis.jl` and `src/standalone/Vegetation/optimality_farquhar.jl`.
The other changes are to update to the new constructors. 
I have removed all of the calls that explicitly pass in values because they only pass in default values (see `docs/tutorials/standalone/Canopy/canopy_tutorial.jl`), but overriding the default parameters via kwargs is still possible.

New constructors:
```

import ClimaLand
import CLIMAParameters as CP
toml_dict= CP.create_toml_dict(Float32)

ClimaLand.Canopy.FarquharParameters(Float64, ClimaLand.Canopy.C3())
ClimaLand.Canopy.FarquharParameters(Float64, ClimaLand.Canopy.C3(); Vcmax25 = 99999)
ClimaLand.Canopy.FarquharParameters(toml_dict, ClimaLand.Canopy.C3(); Vcmax25 = 99999999, pc = 444444444)
```

### DIFF
--- a/docs/tutorials/standalone/Canopy/canopy_tutorial.jl
+++ b/docs/tutorials/standalone/Canopy/canopy_tutorial.jl
@@ -179,26 +179,7 @@ stomatal_model = MedlynConductanceModel{FT}(cond_params);
 
 # Arguments for photosynthesis model:
 
-photo_params = FarquharParameters{FT}(
-    Canopy.C3();
-    oi = FT(0.209),
-    ϕ = FT(0.6),
-    θj = FT(0.9),
-    f = FT(0.015),
-    sc = FT(5e-6),
-    pc = FT(-2e5),
-    Vcmax25 = FT(5e-5),
-    Γstar25 = FT(4.275e-5),
-    Kc25 = FT(4.049e-4),
-    Ko25 = FT(0.2874),
-    To = FT(298.15),
-    ΔHkc = FT(79430),
-    ΔHko = FT(36380),
-    ΔHVcmax = FT(58520),
-    ΔHΓstar = FT(37830),
-    ΔHJmax = FT(43540),
-    ΔHRd = FT(46390),
-)
+photo_params = FarquharParameters(FT, Canopy.C3(); Vcmax25 = FT(5e-5))
 
 photosynthesis_model = FarquharModel{FT}(photo_params);
 
--- a/src/standalone/Vegetation/optimality_farquhar.jl
+++ b/src/standalone/Vegetation/optimality_farquhar.jl
@@ -8,7 +8,7 @@ The required parameters for the optimality Farquhar photosynthesis model.
 Currently, only C3 photosynthesis is supported.
 $(DocStringExtensions.FIELDS)
 """
-struct OptimalityFarquharParameters{FT <: AbstractFloat}
+Base.@kwdef struct OptimalityFarquharParameters{FT <: AbstractFloat}
     "Photosynthesis mechanism: C3 only"
     mechanism::C3
     "Γstar at 25 °C (mol/mol)"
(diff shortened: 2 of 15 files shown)


docs/tutorials/standalone/Canopy/canopy_tutorial.jl
```diff
@@ -181,25 +181,8 @@ stomatal_model = MedlynConductanceModel{FT}(cond_params);
...
 
-photo_params = FarquharParameters{FT}(
-    Canopy.C3();
-    oi = FT(0.209),
-    ϕ = FT(0.6),
-    θj = FT(0.9),
-    f = FT(0.015),
-    sc = FT(5e-6),
-    pc = FT(-2e5),
-    Vcmax25 = FT(5e-5),
```
kmdeck, line 192: This one is read in from a map and should not move to CP

src/standalone/Vegetation/optimality_farquhar.jl
```diff
@@ -71,46 +71,35 @@ Base.eltype(::OptimalityFarquharParameters{FT}) where {FT} = FT
...
+    (;
+    :Jmax_activation_energy             => :ΔHJmax,
+    :intercellular_CO2_concentration    => :oi,
+    :CO2_compensation_point_25c         => :Γstar25,
+    :Farquhar_curvature_parameter       => :θj,
+    :kelvin_25C                         => :To,
+    :photosystem_II_quantum_yield       => :ϕ,
+    :O2_michaelis_menten                => :Ko25,
+    :CO2_michaelis_menten               => :Kc25,
+    :maximum_dry_soil_layer_depth       => :f,
```
kmdeck, line 88: incorrect name here, will look up correct one

src/standalone/Vegetation/photosynthesis.jl
```diff
@@ -87,46 +89,34 @@ end
...
+    name_map = (;
+    :Jmax_activation_energy => :ΔHJmax,
+    :intercellular_CO2_concentration => :oi,
+    :CO2_compensation_point_25c => :Γstar25,
+    :Farquhar_curvature_parameter => :θj,
+    :kelvin_25C => :To,
+    :photosystem_II_quantum_yield => :ϕ,
+    :O2_michaelis_menten => :Ko25,
+    :CO2_michaelis_menten => :Kc25,
+    :maximum_dry_soil_layer_depth => :f,
```
kmdeck, line 105: likewise here

kmdeck (commented):
let's go over how it works, but aside from the comments I left, it looks fine

src/standalone/Vegetation/optimality_farquhar.jl
nefrathenrici: In the description of `FarquharParameters`: f is the constant factor appearing the dark respiration term, equal to 0.015.
Looks like the `maximum_dry_soil_layer_depth` has the same default value. I will add it in as a separate parameter.

docs/tutorials/standalone/Canopy/canopy_tutorial.jl
nefrathenrici: I will keep the default value in the constructor!

juliasloan25: Is that true for all the parameters that are staying here? (`sc` and `pc` too)

juliasloan25 (approved):
This looks good to me!

docs/tutorials/standalone/Canopy/canopy_tutorial.jl
nefrathenrici: `sc` and `pc` each have different values for some test cases in the repo, so I thought it would be best not to change them. They do have default values in climaparameters, so I could try removing the override values and see if the CI passes.
`sc` has values 5e-6 and 2e-6, `pc` has value -2e5 and -2e6

nefrathenrici: It seems like they don't make a difference to the tests. There are still some constructors in ClimaLandSimulations that have hardcoded values, but they seem to have different requirements.

PR #482, opened 2024-02-06
add bucket albedo map experiments

juliasloan25:
## Purpose 
Add global bucket experiments using albedo as a static map and as a temporal map. Run these on CPU and GPU in CI pipeline. This will test if our current static and temporal albedo types can be used globally on GPU.

These runs are copied almost directly from the functional albedo global bucket run, but the albedo types are changed, and the simulation runtime is increased in the temporal map case to exercise the code path used when reading in the next data from a file.

follows #469
closes #447

## Content
- [x] add 2 new experiments: global bucket with static map albedo, global bucket with temporal map albedo
- [x] add experiments to buildkite pipeline on CPU and GPU
- [x] modify Regridder so we can use same input file with different float types in different runs
- [x] add new temporal albedo dataset with NaN converted to 1, update artifact path
- [x] use NCDatasets' `nomissing` function to convert `Union{Missing, FT}` type to `FT` when reading  from input datasets
  - fixes `ERROR: cannot convert `Missing` to `Float64` which occasionally pops up on buildkite

## Status
- function: slightly different numbers, plots look the same on CPU + GPU
- static map: slightly different numbers, plots look the same on CPU + GPU
- temporal map: NaNs on both CPU + GPU, nothing plotted

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -78,24 +78,30 @@ steps:
         command: "julia --color=yes --project=experiments experiments/standalone/Soil/water_conservation.jl"
         artifact_paths: "experiments/standalone/Soil/water_conservation*png"
 
-      - label: "Global Bucket on CPU"
-        key: "global_bucket_cpu"
-        command: "julia --color=yes --project=experiments experiments/standalone/Bucket/global_bucket.jl"
-        artifact_paths: "experiments/standalone/Bucket/artifacts/*cpu*"
-
-      - label: "Compare GPU bucket with CPU bucket"
-        command: "julia --color=yes --project=experiments experiments/standalone/Bucket/compare_gpu_cpu_output.jl"
-        depends_on:
-          - "global_bucket_cpu"
-          - "global_bucket_gpu"
-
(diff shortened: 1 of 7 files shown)


kmdeck (approved):
Looks good, Julia! The plot for the temporal albedo looks pretty wacky. The mean surface fluxes are very large and positive. I think it is because the high latitudes have an albedo of 1, which reflects all shortwave radiation back, and this affects the mean.

I think over time we will move away from the temporal map (because it makes more sense to compute the albedo consistently using our snow model+bareground albedo), so im not too worried about it. But otherwise we need to find a better map without the missing data.

You added a `nomissing` function. Does this replace the values? or does the simulation run with missing values?

juliasloan25:
> You added a nomissing function. Does this replace the values? or does the simulation run with missing values?

By default, data read in from a NetCDF file using NCDatasets will have type `Union{Missing, DataType}` (where DataType here for us is Float64). There are no missing values in the data, so `NCDatasets.nomissing` just converts all the types to `DataType`. If there were a missing value, it would give an error message and exit the simulation. I'm not sure why this is coming up now and didn't come up before

PR #484, opened 2024-02-06
store bareground albedo

### LINKED ISSUE #472: next_albedo bug with snow
## Describe the bug
In the function and static map albedo cases, we initialize `p.bucket.alpha_sfc` as the fixed ground albedo, then update it with the [next_albedo](https://github.com/CliMA/ClimaLSM.jl/blob/ab2ee9b20879694ea9c6310a43edfbbe57739179/src/standalone/Bucket/Bucket.jl#L592) function. However, in `next_albedo`, `alpha_sfc` is calculated as a function of `alpha_sfc` and `alpha_snow` when it should be a function of `alpha_ground` and `alpha_snow`.

This would cause a problem, e.g. if it were to snow for a while, where `alpha_sfc` would always increase towards `alpha_snow` and never decrease below its current value. Even if the snow melted entirely, the albedo wouldn't decrease below `alpha_sfc`, though we would want it to decrease to `alpha_ground`

juliasloan25:
## Purpose 
Store bareground albedo for albedo types that require bareground and snow albedo (`BulkAlbedoFunction` and `BulkAlbedoStatic`, but not `BulkAlbedoTemporal`). This allows us to remove the surface albedo that was previously stored in these two objects.

closes #472

## Content
- [x] add `\alpha_bareground` to `BulkAlbedoFunction/Static`
- [x] remove `\alpha_sfc` from `BulkAlbedoFunction/Static`
- [x] set `\alpha_bareground` (a Field) in albedo object constructor, using either function or input file
- [x] use `\alpha_bareground` to set `p.bucket.\alpha_sfc` in set_initial_parameter_field
- [x] use `\alpha_bareground` instead of `\alpha_sfc` in `next_albedo`
- [x] update for `BulkAlbedoFunction` constructor interface change (needs space now)
- [x] remove `set_initial_parameter_field` - calculations handled in `next_albedo`
- [x] update tests to check correct snow fraction update

----

### DIFF
--- a/docs/src/APIs/Bucket.md
+++ b/docs/src/APIs/Bucket.md
@@ -23,7 +23,6 @@ ClimaLand.Bucket.beta_factor
 ## Methods for handling parameters read in from NetCDF files
 
 ```@docs
-ClimaLand.Bucket.set_initial_parameter_field!
 ClimaLand.Bucket.update_soil_albedo
 ```
 
--- a/docs/src/APIs/Bucket.md
+++ b/docs/src/APIs/Bucket.md
@@ -20,12 +20,6 @@ ClimaLand.Bucket.surface_albedo
 ClimaLand.Bucket.beta_factor
 ```
 
-## Methods for handling parameters read in from NetCDF files
-
-```@docs
-ClimaLand.Bucket.update_soil_albedo
-```
-
 ## Artifact Path Functions
 
 ```@docs
--- a/docs/tutorials/standalone/Bucket/bucket_tutorial.jl
+++ b/docs/tutorials/standalone/Bucket/bucket_tutorial.jl
@@ -168,6 +168,17 @@ import ClimaLand
 include(joinpath(pkgdir(ClimaLand), "parameters", "create_parameters.jl"));
 earth_param_set = create_lsm_parameters(FT);
 
+# Set up the model domain. At every surface coordinate point, we'll solve
+# an ODE for `W` and `Ws`, and for every subsurface point, we solve for `T`.
+# In coupled simulations run at the same
+# resolution as the atmosphere, the bucket horizontal resolution would match the
+# horizontal resolution at the lowest level of the atmosphere model. In general, however, the two
+# resolutions do not need to match. Here we just set up something
+# simple - a Column.
+soil_depth = FT(3.5);
+bucket_domain = Column(; zlim = (-soil_depth, FT(0.0)), nelements = 10);
+surface_space = bucket_domain.space.surface
+
 # Define our `BulkAlbedoFunction` model using a constant surface and snow albedo:
 # The surface albedo is a function of coordinates, which would be
 # (x,y) on a plane, and (lat,lon) on a sphere. Another albedo
@@ -176,7 +187,7 @@ earth_param_set = create_lsm_parameters(FT);
 # These options only applies when coordinates are (lat,lon).
 α_sfc = (coordinate_point) -> 0.2;
 α_snow = FT(0.8);
-albedo = BulkAlbedoFunction(α_snow, α_sfc);
+albedo = BulkAlbedoFunction{FT}(α_snow, α_sfc, surface_space);
 # The critical snow level setting the scale for when we interpolate between
 # snow and surface albedo
 σS_c = FT(0.2);
@@ -206,17 +217,6 @@ bucket_parameters = BucketModelParameters(
     τc,
     earth_param_set,
 );
-
-# Set up the model domain. At every surface coordinate point, we'll solve
-# an ODE for `W` and `Ws`, and for every subsurface point, we solve for `T`.
-# In coupled simulations run at the same
-# resolution as the atmosphere, the bucket horizontal resolution would match the
-# horizontal resolution at the lowest level of the atmosphere model. In general, however, the two
-# resolutions do not need to match. Here we just set up something
-# simple - a Column.
-
-soil_depth = FT(3.5);
-bucket_domain = Column(; zlim = (-soil_depth, FT(0.0)), nelements = 10);
 
 
 # The PrescribedAtmosphere and PrescribedRadiation need to take in a reference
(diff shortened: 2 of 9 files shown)


src/standalone/Bucket/Bucket.jl
```diff
@@ -65,51 +65,60 @@ abstract type AbstractBucketModel{FT} <: AbstractExpModel{FT} end
...
+function BulkAlbedoFunction{FT}(
+    α_snow::FT,
+    α_bareground::F,
+) where {FT <: AbstractFloat, F}
+    BulkAlbedoFunction(α_snow, α_bareground)
 end
 
 function BulkAlbedoFunction{FT}(
     α_snow::FT,
     α_sfc::Function,
```
kmdeck, line 92: maybe we should also call this alpha_bareground for clarity?

kmdeck: i guess we cant quite call it that, but maybe something similar instead of `sfc`

src/standalone/Bucket/Bucket.jl
```diff
@@ -389,8 +402,8 @@ function set_initial_parameter_field!(
     p,
     surface_coords,
 ) where {FT}
-    (; α_sfc) = albedo
-    @. p.bucket.α_sfc = α_sfc(surface_coords)
+    # TODO should we include snow here?
```
kmdeck, line 405: I think yes, for the Function and the Static cases. the surface albedo in those cases is the linear combination of bareground & snow albedo

kmdeck: wait - can we get rid of the `set_initial_parameter_field` altogether? In update aux we will compute the linear combination and update p.bucket.alpha_surface for all models.

kmdeck: I think we can for the Static and Function cases, because we have already set the bareground parameter field in the constructor. but we have to see if `next_albedo` does the right thing in this case: https://github.com/CliMA/ClimaLand.jl/blob/f05f861a8eed75d45cc0686333333b912137260b/src/standalone/Bucket/Bucket.jl#L452

src/standalone/Bucket/Bucket.jl
juliasloan25: good point! what about `alpha_bareground_func`?

src/standalone/Bucket/Bucket.jl
juliasloan25: I think we can! I changed the logic for next_albedo to handle the initialization case and verified locally that it produces the same field as the current `set_initial_parameter_field!`

docs/tutorials/standalone/Bucket/bucket_tutorial.jl
```diff
@@ -168,6 +168,17 @@ import ClimaLand
...
+# resolution as the atmosphere, the bucket horizontal resolution would match the
+# horizontal resolution at the lowest level of the atmosphere model. In general, however, the two
+# resolutions do not need to match. Here we just set up something
+# simple - a Column.
+soil_depth = FT(3.5);
+bucket_domain = Column(; zlim = (-soil_depth, FT(0.0)), nelements = 10);
+surface_space = bucket_domain.space.surface
+
 # Define our `BulkAlbedoFunction` model using a constant surface and snow albedo:
 # The surface albedo is a function of coordinates, which would be
```
kmdeck, line 183: small change - should these `surface` references be `bareground` now?

experiments/standalone/Bucket/global_bucket.jl
```diff
@@ -41,9 +41,22 @@ context = ClimaComms.context()
...
+    radius = FT(6.3781e6),
+    depth = soil_depth,
+    nelements = (10, 10),
+    npolynomial = 1,
+    dz_tuple = FT.((1.0, 0.05)),
+);
+surface_space = bucket_domain.space.surface
+
+# Set up parameters
 α_sfc = (coordinate_point) -> 0.2;
```
kmdeck, line 57: likewise here

src/standalone/Bucket/Bucket.jl
```diff
@@ -601,11 +524,12 @@ function next_albedo(
...
     t,
 ) where {FT}
-    (; α_snow) = model_albedo
+    (; α_snow, α_bareground) = model_albedo
     (; σS_c) = parameters
-    α_sfc = p.bucket.α_sfc
     σS = Y.bucket.σS
-    return @. ((1 - σS / (σS + σS_c)) * α_sfc + σS / (σS + σS_c) * α_snow)
+    return @. (
+        (1 - σS / (σS + σS_c)) * α_bareground + σS / (σS + σS_c) * α_snow
```
kmdeck, line 531: yay! this is much clearer than how we had it before :)

test/standalone/Bucket/snow_bucket_tests.jl
```diff
@@ -28,7 +28,6 @@ for FT in (Float32, Float64)
     earth_param_set = create_lsm_parameters(FT)
     α_sfc = (coordinate_point) -> 0.2 # surface albedo, spatially constant
```
kmdeck, line 29: could be alpha_bareground_func now

docs/src/APIs/Bucket.md
```diff
@@ -23,7 +23,6 @@ ClimaLand.Bucket.beta_factor
 ## Methods for handling parameters read in from NetCDF files
 
 ```@docs
-ClimaLand.Bucket.set_initial_parameter_field!
 ClimaLand.Bucket.update_soil_albedo
```
kmdeck, line 26: is this defined?

kmdeck (approved):
looks good, Julia! thank you.

docs/tutorials/standalone/Bucket/bucket_tutorial.jl
juliasloan25: good point! I'll change it in all the drivers + tests

PR #485, opened 2024-02-07
SpaceVaryingInput

kmdeck:
## Purpose 
Adds a "SpaceVaryingInput" function which has the following methods
- 2D - returns a regridded field
- 1D - returns a regridded field
- 1D - returns a field of structs with the regridded values
- analytic. Takes in a function of (coords) and can return whatever you like

Future extensions could be for depth resolved global data (3D), or to extend the struct case to the 2D case if needed

For the struct case, this is because: some parameters get packaged into a struct which we then dispatch on. For example, the soil retention model (vanGenuchten, BrooksCorey). Each has a few parameters which will spatially vary. If we have a field of structs, we can still use dispatch +broadcasting. Im not sure if this is also needed for the BucketModel and for the Canopy Model subparameterizations. If it is very niche, we can put these methods in the Soil module. This may also lead to performance issues as per @Sbozzolo 

The goal is to unify the way we handle parameters for spatially varying parameters regardless of the simulation domain.  Note that non-spatially varying parameters are scalars and do not use the `SpaceVaryingInput` function.

## To-do
- Write more tests?
- see discussion point below
## Content
- add SpaceVaryingInputs module
- defined linear interpolation function, moved to `general_utilities`. Now Space and Time VaryingInputs share this. Slightly modified to account for edge case (double check if needed)
- generalize Richards and EnergyHydrology parameters struct to allow for FT or Fields as arguments
- BucketAlbedoStatic and Function use SpaceVaryingInput under the hood now.
- 1D case is in the layered soil tutorial experiment
- update soil boundary condition functions to use some helper functions (extract the top or bottom level of a center field, put onto the corresponding 2D face field). 
- extract the top or bottom level ^^ as described above for the parameters if they are fields vs scalars. *Discussion*: if they are passed into the constructor as scalars, should we create a field of constant value? then we would not need this if statement (or method).

Note that many files have changed for two reasons which are seemingly unrelated to this PR:
- `RichardsParameters{FT, typeof(hcm)}() -> RichardsParameters()`. Before we had scalars (FT) and a scalar struct (hcm). Now we can support scalars or fields. we could do `typeof(field_parameter), typeof(hcm)`, but Im not sure what we are gaining from this, so I ended up just removing it. We may want to do the same for other parameter structs. For EnergyHydrology, I kept the `{FT}`, since there are some default global scalar parameters in that struct. in the long term, I think this will need to update too. But im a little unsure on all of this. 
- I had to add in a parametric type to `vanGenuchten{FT}`'s inner constructor. Im not sure, but without it, the SpaceVaryingInput call in the 1D case does not have an inferrable eltype.
## Future PRs
1. simplify BulkAlbedoFunction and BulkAlbedoStatic to be PrescribedBaregroundAlbedo, and we can compute the bareground albedo in the driver using the "SpaceVaryingInput" function.
2. test 1D depth varying input with energy hydrology model - update parametric type in that model's parameter struct too
3. set up global soil run using mask + soil parameters from a file, add in 3D case
4. propagate to canopy..

----

### DIFF
--- a/docs/tutorials/standalone/Soil/layered_soil.jl
+++ b/docs/tutorials/standalone/Soil/layered_soil.jl
@@ -1,0 +1,172 @@
+# This shows how to run single colum soil model, in standalone mode
+# with spatially varying properties. We are mimicking the experiment
+# carried out in Huang et. al.
+# Can. J. Soil Sci. (2011) 91: 169183 doi:10.4141/CJSS09118,
+# which measured the infiltration of layered soil in Fort McMurray,
+# Alberta, Canada. We thank Mingbin Huang and S. Lee Barbour for
+# correspondence and support, including sharing of data, with us
+# Note that all data used in this tutorial is available in their
+# publication.
+
+using Plots
+import SciMLBase
+import ClimaTimeSteppers as CTS
+using ClimaCore
+import CLIMAParameters as CP
+using ArtifactWrappers
+using DelimitedFiles: readdlm
+
+using ClimaLand
+using ClimaLand.Domains: Column
+using ClimaLand.Soil
+import ClimaLand
+FT = Float64;
+
+# Define simulation times
+t0 = Float64(0)
+tf = Float64(60 * 60)
+dt = Float64(30);
+
+# Define the domain 
+zmax = FT(0)
+zmin = FT(-1.1)
+nelems = 75
+Δ = FT((zmax - zmin) / nelems / 2)
+soil_domain = Column(; zlim = (zmin, zmax), nelements = nelems);
+
+# Download the parameter data. This has been obtained from
+# Table 1b of
+# Infiltration and drainage processes in multi-layered coarse soils
+# Mingbin Huang et. al.
+# Can. J. Soil Sci. (2011) 91: 169183 doi:10.4141/CJSS09118
+af = ArtifactFile(
+    url = "https://caltech.box.com/shared/static/qvbt37xkwz8gveyi6tzzbs0e18trpcsq.csv",
+    filename = "sv_62.csv",
+)
+dataset = ArtifactWrapper(@__DIR__, "sv62", ArtifactFile[af]);
+dataset_path = get_data_folder(dataset);
+data = joinpath(dataset_path, af.filename)
+parameter_data = readdlm(data, ',');
+# Our model treats z as increasing in the upwards direction.
+# Values below the surface are negative.
+# Because of this, we convert the (positive-valued) depth
+# of the data into a monotonically increasing z coordinate value.
+# using a negative sign and the reverse function.
+depth = reverse(-parameter_data[1, :] .* 0.01) # convert to m
+ksat = reverse(parameter_data[6, :] .* 1 / 100.0 / 60.0) # convert cm/min to m/s
+vgα = reverse(parameter_data[4, :] .* 100 * 2) # they report αᵈ; αʷ = 2αᵈ. This experiment is for infiltration (wetting).
+vgn = reverse(parameter_data[5, :])
+residual_frac = reverse(parameter_data[2, :])
+porosity = reverse(parameter_data[3, :]);
+
+# Create fields corresponding to the parameter
+ν = SpaceVaryingInput(depth, porosity, soil_domain.space.subsurface)
+K_sat = SpaceVaryingInput(depth, ksat, soil_domain.space.subsurface)
+θ_r = SpaceVaryingInput(depth, residual_frac, soil_domain.space.subsurface);
+# The specific storativity is not something we have data on, so we
+# approximate it as being constant in depth, and create the parameter field directly:
+S_s = ClimaCore.Fields.zeros(soil_domain.space.subsurface) .+ 1e-3;
+
+# The retention model is a vanGenuchten model with α and n as a function of
+# depth, read from the data:
+hcm = SpaceVaryingInput(
+    depth,
+    (; α = vgα, n = vgn),
+    soil_domain.space.subsurface,
+    vanGenuchten{FT},
+);
+
+# The parameter struct:
+params = ClimaLand.Soil.RichardsParameters(;
+    ν = ν,
+    hydrology_cm = hcm,
+    K_sat = K_sat,
+    S_s = S_s,
+    θ_r = θ_r,
+);
+
+# From here on out, everything should look familiar if you've already gone
+# through the other soil tutorials.
+# Set Boundary conditions:
+# At the top, we use the observed value of Ksat at the top of the domain.
+# Setting the flux to be -Ksat is approximating the top as saturated.
+function top_flux_function(p, t)
+    return -0.0001033
+end
+top_bc = ClimaLand.Soil.FluxBC(top_flux_function)
+bottom_bc = ClimaLand.Soil.FreeDrainage()
+boundary_fluxes = (; top = (water = top_bc,), bottom = (water = bottom_bc,))
+soil = Soil.RichardsModel{FT}(;
+    parameters = params,
+    domain = soil_domain,
+    boundary_conditions = boundary_fluxes,
+    sources = (),
+);
+
+# Initial the state vectors, and set initial conditions
+Y, p, cds = initialize(soil);
+
+# Initial conditions
+Y.soil.ϑ_l .= 0.0353; # read from Figure 4 of Huang et al. 
+
+# We also set the initial conditions of the auxiliary state here:
+set_initial_cache! = make_set_initial_cache(soil)
+set_initial_cache!(p, Y, t0);
+
+# Timestepping:
+stepper = CTS.ARS111()
+@assert FT in (Float32, Float64)
+err = (FT == Float64) ? 1e-8 : 1e-4
+convergence_cond = CTS.MaximumError(err)
+conv_checker = CTS.ConvergenceChecker(norm_condition = convergence_cond)
+ode_algo = CTS.IMEXAlgorithm(
+    stepper,
+    CTS.NewtonsMethod(
+        max_iters = 50,
+        update_j = CTS.UpdateEvery(CTS.NewNewtonIteration),
+        convergence_checker = conv_checker,
+    ),
+)
+exp_tendency! = make_exp_tendency(soil)
+imp_tendency! = make_imp_tendency(soil)
+update_jacobian! = make_update_jacobian(soil)
+jac_kwargs =
+    (; jac_prototype = RichardsTridiagonalW(Y), Wfact = update_jacobian!)
+prob = SciMLBase.ODEProblem(
+    CTS.ClimaODEFunction(
+        T_exp! = exp_tendency!,
+        T_imp! = SciMLBase.ODEFunction(imp_tendency!; jac_kwargs...),
+        dss! = ClimaLand.dss!,
+    ),
+    Y,
+    (t0, tf),
+    p,
+)
+saveat = [0.0, 8.0, 16.0, 24.0, 32.0, 40.0, 60.0] .* 60 # chosen to compare with data in plots in paper
+sol = SciMLBase.solve(prob, ode_algo; dt = dt, saveat = saveat);
+
+z = parent(ClimaCore.Fields.coordinate_field(soil_domain.space.subsurface).z)
+ϑ_l = [parent(sol.u[k].soil.ϑ_l) for k in 1:length(sol.t)]
+plot(ϑ_l[1], z, label = "initial", color = "grey", aspect_ratio = 0.8)
+plot!(ϑ_l[2], z, label = "8min", color = "orange")
+plot!(ϑ_l[3], z, label = "16min", color = "red")
+plot!(ϑ_l[4], z, label = "24min", color = "teal")
+plot!(ϑ_l[5], z, label = "32min", color = "blue")
+plot!(ϑ_l[6], z, label = "40min", color = "purple")
+plot!(ϑ_l[7], z, label = "60min", color = "green")
+scatter!(porosity, depth, label = "Porosity")
+plot!(legend = :bottomright)
+
+plot!(xlim = [0, 0.7])
+
+plot!(
+    ylim = [-1.1, 0],
+    yticks = [-1.1, -1, -0.9, -0.8, -0.7, -0.6, -0.5, -0.4, -0.3, -0.2, -0.1],
+)
+
+plot!(ylabel = "Depth (m)")
+
+plot!(xlabel = "Volumeteric Water Content")
+
+savefig("./sv62_alpha_2_inf_updated_data_climaland.png")
+# ![](sv62_alpha_2_inf_updated_data_climaland.png)
--- a/src/ClimaLand.jl
+++ b/src/ClimaLand.jl
@@ -11,10 +11,12 @@ include("shared_utilities/general_utils.jl")
 include("shared_utilities/TimeVaryingInputs.jl")
 using .TimeVaryingInputs
 export TimeVaryingInput, evaluate!
-
 include("shared_utilities/Regridder.jl")
 include("shared_utilities/Domains.jl")
 include("shared_utilities/FileReader.jl")
+include("shared_utilities/SpaceVaryingInputs.jl")
+using .SpaceVaryingInputs
+export SpaceVaryingInput
 using .Domains
 include("shared_utilities/utils.jl")
 include("shared_utilities/models.jl")
--- a/src/shared_utilities/SpaceVaryingInputs.jl
+++ b/src/shared_utilities/SpaceVaryingInputs.jl
@@ -1,0 +1,165 @@
+# SpaceVaryingInputs.jl
+#
+# This module contains methods to process external data, regrid it onto the
+# model grid, and return the corresponding fields for use in the simulation.
+# This module only concerns with external data which varies in space,
+# and not time. For temporally varying input, we refer you to `TimeVaryingInputs.jl`.
+
+# All spatially varying parameter fields are assumed to fit into memory,
+# and on GPU runs, they have underlying CuArrays on the GPU.
+
+# The planned parameter underlying arrays are:
+# - one-dimensional (values prescribed as a function of depth at a site),
+# - two-dimensional (values prescribed globally at each lat/lon),
+# - three-dimensional (values prescribed as a function of depth globally)
+# - analytic (functions of the coordinates of the space)
+
+module SpaceVaryingInputs
+using ClimaCore
+using ClimaComms
+using DocStringExtensions
+import ..searchsortednearest
+import ..linear_interpolation
+using ClimaLand.FileReader
+export SpaceVaryingInput
+
+# Analytic case
+"""
+    SpaceVaryingInput(data_function::Function, space::ClimaCore.Spaces.AbstractSpace) 
+  
+Returns the parameter field to be used in the model; appropriate when
+a parameter is defined using a function of the coordinates of the space.
+
+Pass the ``data" as a function `data_function` which takes coordinates as arguments, 
+and  the ClimaCore space of the model simulation.
+
+This returns a scalar field.
+Note that data_function is broadcasted over the coordinate field. Internally, inside
+your function, this must be unpacked (coords.lat, coords.lon, e.g.) for 
+use of the coordinate values directly.
+"""
+function SpaceVaryingInput(
+    data_function::Function,
+    space::ClimaCore.Spaces.AbstractSpace,
+)
+    model_value = ClimaCore.Fields.zeros(space)
+    coords = ClimaCore.Fields.coordinate_field(space)
+    return model_value .= data_function.(coords)
+end
+
+# 1-D Case
+"""
+    function SpaceVaryingInput(
+        data_z::AbstractArray,
+        data_values::AbstractArray,
+        space::S,
+    ) where {S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace}
+
+Given a set of depths `data_z` and the observed values `data_values` 
+at those depths, create an interpolated field of values at each value
+of z in the model grid - defined implicitly by `space`.
+
+Returns a ClimaCore.Fields.Field of scalars.
+"""
+function SpaceVaryingInput(
+    data_z::AbstractArray,
+    data_values::AbstractArray,
+    space::S,
+) where {S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace}
+    model_value = ClimaCore.Fields.zeros(space)
+    # convert the passed arrays to the appropriate type for the device
+    device = ClimaComms.device(space)
+    AT = ClimaComms.array_type(device)
+    data_values = AT(data_values)
+    data_z = AT(data_z)
+    zvalues = ClimaCore.Fields.coordinate_field(space).z
+    #now create the parameter field
+    model_value .= map(zvalues) do z
+        linear_interpolation(data_z, data_values, z)
+    end
+    return model_value
+end
+
+
+"""
+    SpaceVaryingInputs.SpaceVaryingInput(
+        data_z::AbstractArray,
+        data_values::NamedTuple,
+        space::S,
+        dest_type::Type{DT},
+    ) where {
+        S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace,
+        DT,
+    }
+
+Returns a field of parameter structs to be used in the model; 
+appropriate when the parameter struct values vary in depth; 
+the `dest_type` argument is the struct type - we assumed that
+your struct as a constructor which accepts the values of its arguments
+by kwarg,
+- `data_z` is where the measured values were obtained,
+- `data_values` is a NamedTuple with keys equal to the argument names
+of the struct, and with values equal to an array of measured values,
+- `space` defines the model grid.
+
+As an example, we can create a field of vanGenuchten structs as follows. This struct
+requires two parameters, `α` and `n`. Let's assume that we have measurements of these
+as a function of depth at the locations given by `data_z`, called `data_α` and `data_n`.
+Then we can write
+`vG_field = SpaceVaryingInput(data_z, (;α = data_α, n = data_n), space, vanGenuchten{Float32})`.
+Under the hood, at each point in the model grid, we will create
+`vanGenuchten{Float32}(;α = interp_α, n = interp_n)`, where `interp` indicates
+the interpolated value at the model depth.
+
+Returns a ClimaCore.Fields.Field of type DT.
+
+"""
+function SpaceVaryingInput(
+    data_z::AbstractArray,
+    data_values::NamedTuple,
+    space::S,
+    dest_type::Type{DT},
+) where {S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace, DT}
+    zvalues = ClimaCore.Fields.coordinate_field(space).z
+    # convert the passed arrays to the appropriate type for the device
+    device = ClimaComms.device(space)
+    AT = ClimaComms.array_type(device)
+    data_z = AT(data_z)
+    data_values_AT = map(AT, data_values)
+    # now create the field of structs
+    model_value = map(zvalues) do z
+        args = map(data_values_AT) do value
+            return linear_interpolation(data_z, value, z)
+        end
+        DT(; args...)
+    end
+    return model_value
+end
+
+"""
+    SpaceVaryingInput(data::PDS, varname::N; space::S) 
+    where {PDS <: FileReader.PrescribedDataStatic, N <: String, S <: ClimaCore.Spaces.SpectralElement2D}
+  
+Returns the parameter field to be used in the model; appropriate when
+a parameter is defined on the surface of the Earth.
+
+Pass the data as a `FileReader.PrescribedDataStatic` object
+as well as the variable name of the variable in the data file, and the ClimaCore space
+of the model simulation.
+
+Returns a ClimaCore.Fields.Field of scalars; analogous to the 1D case which also
+returns a ClimaCore.Fields.Field of scalars.
+"""
+function SpaceVaryingInput(
+    data::PDS,
+    varname::N,
+    space::S,
+) where {
+    PDS <: FileReader.PrescribedDataStatic,
+    N <: String,
+    S <: ClimaCore.Spaces.SpectralElementSpace2D,
+}
+    return FileReader.get_data_at_date(data, space, varname)
+end
+
+end
--- a/src/shared_utilities/general_utils.jl
+++ b/src/shared_utilities/general_utils.jl
@@ -16,3 +16,30 @@ function searchsortednearest(a, x)
         return abs(a[i] - x) < abs(a[i - 1] - x) ? i : i - 1
     end
 end
+
+
+"""
+    linear_interpolation(indep_vars, dep_vars, indep_value)
+
+Carries out linear interpolation to obtain a value at
+location `indep_value`, using a independent variable
+1-d vector `indep_vars` and a dependent variable 
+1-d vector `dep_vars`.
+
+If the `indep_value` is outside the range of `indep_vars`, this
+returns the endpoint value closest.
+"""
+function linear_interpolation(indep_vars, dep_vars, indep_value)
+    N = length(indep_vars)
+    id = searchsortedfirst(indep_vars, indep_value)
+    if id == 1
+        dep_vars[begin]
+    elseif id == N + 1
+        dep_vars[end]
+    else
+        id_prev = id - 1
+        x0, x1 = indep_vars[id_prev], indep_vars[id]
+        y0, y1 = dep_vars[id_prev], dep_vars[id]
+        y0 + (y1 - y0) / (x1 - x0) * (indep_value - x0)
+    end
+end
--- a/src/standalone/Soil/boundary_conditions.jl
+++ b/src/standalone/Soil/boundary_conditions.jl
@@ -10,6 +10,73 @@ export TemperatureStateBC,
     boundary_var_types
 
 """
+    get_top_surface_field(
+        center_field::ClimaCore.Fields.Field,
+        surface_space,
+    )
+
+A helper function to extract the top level of a center field and
+cast it onto the surface face space.
+"""
+function get_top_surface_field(
+    center_field::ClimaCore.Fields.Field,
+    surface_space,
+)
+    # TODO: Find cleaner way to do this
+    nz = Spaces.nlevels(axes(center_field))
+    return Fields.Field(
+        Fields.field_values(Fields.level(center_field, nz)),
+        surface_space,
+    )
+end
+
+"""
+    get_top_surface_field(
+        center_val,
+        _,
+    )
+
+A helper function for the case where we use `get_top_surface_field` on
+a parameter that is a scalar rather than a field. Returns the scalar value.
+"""
+function get_top_surface_field(center_val, _)
+    return center_val
+end
+
+"""
+    get_bottom_surface_field(
+        center_field::ClimaCore.Fields.Field,
+        bottom_space,
+    )
+
+A helper function to extract the bottom level of a center field and
+cast it onto the bottom face space.
+"""
+function get_bottom_surface_field(
+    center_field::ClimaCore.Fields.Field,
+    bottom_space,
+)
+    # TODO: Find cleaner way to do this
+    return Fields.Field(
+        Fields.field_values(Fields.level(center_field, 1)),
+        bottom_space,
+    )
+end
+
+"""
+    get_bottom_surface_field(
+        center_val,
+        _,
+    )
+
+A helper function for the case where we use `get_bottom_surface_field` on
+a parameter that is a scalar rather than a field. Returns the scalar value.
+"""
+function get_bottom_surface_field(center_val, _)
+    return center_val
+end
+
+"""
     AbstractSoilBC <: ClimaLand. AbstractBC
 
 An abstract type for soil-specific types of boundary conditions, like free drainage.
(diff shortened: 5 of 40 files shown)


src/standalone/Soil/energy_hydrology.jl
```diff
@@ -63,22 +57,23 @@ struct EnergyHydrologyParameters{
...
+    ν_ss_quartz::F,
+    ν_ss_gravel::F,
     hydrology_cm::C,
-    K_sat::FT,
-    S_s::FT,
-    θ_r::FT,
+    K_sat::F,
+    S_s::F,
+    θ_r::F,
     PAR_albedo = FT(0.2),
```
kmdeck, line 68: These can vary spatially, but the default is a scalar (not needed in configurations). When it is a scalar, we broadcast.

experiments/standalone/Soil/zettl_sv62.jl
```diff
@@ -0,0 +1,230 @@
...
+
+# Domain
+zmax = FT(0)
+zmin = FT(-1.1)
+nelems = 75
+Δ = FT((zmax - zmin) / nelems / 2)
+soil_domain = Column(; zlim = (zmin, zmax), nelements = nelems)
+
+parameter_data = readdlm(
+    "/Users/katherinedeck/Desktop/code/ClimaLand.jl/experiments/Standalone/Soil/sv_62.csv",
```
kmdeck, line 27: Ill move these to box when we add this to buildkite. itll be part of this PR too

src/shared_utilities/general_utils.jl
```diff
@@ -16,3 +16,21 @@ function searchsortednearest(a, x)
         return abs(a[i] - x) < abs(a[i - 1] - x) ? i : i - 1
     end
 end
+
+
+"""
```
kmdeck, line 21: doc string

src/shared_utilities/general_utils.jl
```diff
@@ -16,3 +16,21 @@ function searchsortednearest(a, x)
...
+"""
+function linear_interpolation(indep_vars, dep_vars, indep_value)
+    N = length(indep_vars)
+    id = searchsortedfirst(indep_vars, indep_value)
+    if id == 1
+        dep_vars[id]
+    elseif id == N
```
kmdeck, line 28: edge case I ran across, I think the time case has an @assert that `min(indep_vars) <= indep_value <= max(indep_vars)`

kmdeck:
for the GPU test, I need to use CuArrays, I think. will fix.

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -9,6 +9,45 @@ export TemperatureStateBC,
...
+"""
+function get_top_surface_field(
+    center_field::ClimaCore.Fields.Field,
+    surface_space,
+)
+    nz = Spaces.nlevels(axes(center_field))
+    return Fields.Field(
+        Fields.field_values(Fields.level(center_field, nz)),
+        surface_space,
+    )
```
Sbozzolo, line 29:
```suggestion
   # TODO: Find cleaner way to do this
    nz = Spaces.nlevels(axes(center_field))
    return Fields.Field(
        Fields.field_values(Fields.level(center_field, nz)),
        surface_space,
    )
```

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -9,6 +9,45 @@ export TemperatureStateBC,
...
+cast it onto the bottom face space.
+"""
+function get_bottom_surface_field(
+    center_field::ClimaCore.Fields.Field,
+    bottom_space,
+)
+    return Fields.Field(
+        Fields.field_values(Fields.level(center_field, 1)),
+        bottom_space,
+    )
```
Sbozzolo, line 48:
```suggestion
    # TODO Find cleaner way to do this
    return Fields.Field(
        Fields.field_values(Fields.level(center_field, 1)),
        bottom_space,
    )
```

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -418,21 +457,27 @@ function ClimaLand.boundary_flux(
...
-    )
+    surface_space = axes(Δz)
+    K_c = get_top_surface_field(p.soil.K, surface_space)
+    ψ_c = get_top_surface_field(p.soil.ψ, surface_space)
 
     # Calculate pressure head using boundary condition
+    # Extract parameters; if these are fields, get the values at the surface
     (; hydrology_cm, θ_r, ν, S_s) = model.parameters
+    if typeof(hydrology_cm) <: ClimaCore.Fields.Field
+        hcm_bc = get_top_surface_field(hydrology_cm, surface_space)
```
Sbozzolo, line 468: `get_top_surface_field` could be get a method for scalars that just returns itself so that you can remove the if here.

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -463,13 +508,28 @@ function ClimaLand.boundary_flux(
...
 
     # Calculate pressure head using boundary condition
     (; hydrology_cm, θ_r, ν, S_s) = model.parameters
+    if typeof(hydrology_cm) <: ClimaCore.Fields.Field
+        hcm_bc = get_bottom_surface_field(hydrology_cm, surface_space)
+        θ_r_bc = get_bottom_surface_field(θ_r, surface_space)
+        ν_bc = get_bottom_surface_field(ν, surface_space)
+        S_s_bc = get_bottom_surface_field(S_s, surface_space)
+    else
+        hcm_bc = hydrology_cm
```
Sbozzolo, line 523: Same as above

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -463,13 +508,28 @@ function ClimaLand.boundary_flux(
...
+        θ_r_bc = θ_r
+        ν_bc = ν
+        S_s_bc = S_s
+    end
+
     θ_bc = FT.(rre_bc.bc(p, t))
-    ψ_bc = @. pressure_head(hydrology_cm, θ_r, θ_bc, ν, S_s)
+    ψ_bc = @. pressure_head(hcm_bc, θ_r_bc, θ_bc, ν_bc, S_s_bc)
+    ψ_bc = Fields.Field(Fields.field_values(Fields.level(ψ_bc, 1)), axes(Δz))
```
Sbozzolo, line 531: The purpose of this `Field` gymnastics is hard to understand. I don't know exactly what's the better way to this, but if you package this into a function a reader can guess your intent from the function name.

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,134 @@
...
+of the model simulation.
+"""
+function SpaceVaryingInput(
+    data::Function,
+    space::S,
+) where {S <: ClimaCore.Spaces.AbstractSpace}
+    model_value = ClimaCore.Fields.zeros(space)
+    coords = ClimaCore.Fields.coordinate_field(space)
+    @. model_value = data(coords)
+    return model_value
```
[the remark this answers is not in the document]
kmdeck: Yes, to the question. I was trying to keep the arguments and structure similar for all these functions (they get data, they return the model_value). Ill change to data_function (?). let me know if you prefer any other changes with respect to this

docs/tutorials/standalone/Soil/layered_soil.jl
```diff
@@ -0,0 +1,161 @@
...
+    depth,
+    (; α = vgα, n = vgn),
+    soil_domain.space.subsurface,
+    vanGenuchten{FT},
+)
+params = ClimaLand.Soil.RichardsParameters(;
+    ν = ν,
+    hydrology_cm = hcm,
+    K_sat = K_sat,
+    S_s = ClimaCore.Fields.zeros(soil_domain.space.subsurface) .+ 1e-3,## FIX
```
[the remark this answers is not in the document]
kmdeck: ah, we dont have data for this. so I am treating it as constant. But it's easier to just make the field directly, rather than go through the SpaceVaryingInput function.

The limitation Im working with (why it has to be a field rather than a scalar) is because the type of all parameters \nu, K_sat, S_s, \theta_r are the same in the parameter struct definition. so if some are fields, they all need to be.

I left some comments so that it is clearer why I am doing this. if there is a better way of handling the struct def to be more flexible im open to changing it

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,134 @@
...
+    varname::N,
+    space::S,
+) where {
+    PDS <: FileReader.PrescribedDataStatic,
+    N <: String,
+    S <: ClimaCore.Spaces.SpectralElementSpace2D,
+}
+    model_value = ClimaCore.Fields.zeros(space)
+    model_value .= FileReader.get_data_at_date(data, space, varname)
+    return model_value
```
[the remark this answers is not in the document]
kmdeck: Up to you on this. Ill keep it the way it is (so that all these methods take data and return model_value) until hearing back.

src/standalone/Soil/boundary_conditions.jl
kmdeck: thanks! did so

src/standalone/Soil/boundary_conditions.jl
kmdeck: I think this is no longer needed. This is what `get_bottom_surface_field` does. ill make sure it runs as expected.

src/ClimaLand.jl
```diff
@@ -11,10 +11,12 @@ include("shared_utilities/general_utils.jl")
 include("shared_utilities/TimeVaryingInputs.jl")
 using .TimeVaryingInputs
```
juliasloan25, line 12: not for this PR, but should we have `using .ModuleName` instead of `include("path-to-module.jl")` for all the files we include in this file?

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+  
+Returns the parameter field to be used in the model; appropriate when
+a parameter is defined using a function of the coordinates of the space.
+
+Pass the ``data" as a function `data_function` which takes coordinates as arguments, 
+and  the ClimaCore space of the model simulation.
+"""
+function SpaceVaryingInput(
+    data_function::Function,
+    space::ClimaCore.Spaces.AbstractSpace,
```
juliasloan25, line 37: function and docstring don't match - maybe we want to use `where {S <: ...}` in both cases?

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+using ClimaLand.FileReader
+export SpaceVaryingInput
+
+# Analytic case
+"""
+    SpaceVaryingInput(data_function::Function, space::S) 
+    where {S <: ClimaCore.Spaces.AbstractSpace}
+  
+Returns the parameter field to be used in the model; appropriate when
+a parameter is defined using a function of the coordinates of the space.
```
juliasloan25, line 30: This can be used for both the 0D (point space) and 2D cases, right?

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+ClimaCore.Fields.Field.
+"""
+function SpaceVaryingInput(
+    data_z::AbstractArray,
+    data_value::AbstractArray,
+    space::S,
+) where {S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace}
+    model_value = ClimaCore.Fields.zeros(space)
+    zvalues = ClimaCore.Fields.coordinate_field(space).z
+    N = length(data_z)
```
juliasloan25, line 64: `N` isn't used here

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+# 1-D Case
+"""
+    function SpaceVaryingInput(
+        data_z::AbstractArray,
+        data_value::AbstractArray,
+        space::S,
+    ) where {S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace}
+
+Interpolates the variable with values given in `data_value` as a function of
+depths `data_z` to the model grid defined implicitly by `space`. Returns a
```
juliasloan25, line 54: could you just explain that `data_z` is an array of the depths `data_value` is defined on? Also, maybe `data_value` should be called `data_values` since it's an array of multiple values

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+        data_values::NamedTuple,
+        space::S,
+        dest_type::Type{DT},
+    ) where {
+        S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace,
+        DT,
+    }
+
+Returns a field of parameter structs to be used in the model; 
+appropriate when the parameter struc values vary in depth.
```
juliasloan25, line 84: typo: struc -> struct

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+
+Note that your struct *must* have a constructor which accepts the 
+values of the named tuple by keyword argument, without any other arguments.
```
juliasloan25, line 87: Maybe it would help to give an example of a struct we use this for (e.g. van Genuchten)

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+        DT(; args...)
+    end
+    return model_value
+end
+
+
+# 2D case (struct version not supported yet)
+"""
+    SpaceVaryingInput(data::PDS, varname::N; space::S) 
+    where {PDS <: FileReader.PrescribedDataStatic, N <: Symbol, S <: ClimaCore.Spaces.SpectralElement2D}
```
juliasloan25, line 110: docstring has `N <: Symbol` but function has `N <: String`

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+"""
+    function SpaceVaryingInput(
+        data_z::AbstractArray,
+        data_value::AbstractArray,
+        space::S,
+    ) where {S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace}
+
+Interpolates the variable with values given in `data_value` as a function of
+depths `data_z` to the model grid defined implicitly by `space`. Returns a
+ClimaCore.Fields.Field.
```
juliasloan25, line 55: maybe note that this returns a Field of scalars (as opposed to the field of structs returned below)

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+    space::S,
+    dest_type::Type{DT},
+) where {S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace, DT}
+    zvalues = ClimaCore.Fields.coordinate_field(space).z
+    data_keys = keys(data_values)
+    model_value = map(zvalues) do z
+        args = map(data_keys) do key
+            key => linear_interpolation(data_z, getproperty(data_values, key), z)
+        end
+        DT(; args...)
```
juliasloan25, line 101: why do we need `dest_type` in this case but not in the non-struct 1D case?

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+    SpaceVaryingInput(data::PDS, varname::N; space::S) 
+    where {PDS <: FileReader.PrescribedDataStatic, N <: Symbol, S <: ClimaCore.Spaces.SpectralElement2D}
+  
+Returns the parameter field to be used in the model; appropriate when
+a parameter is defined on the surface of the Earth.
+
+Pass the data as a `FileReader.PrescribedDataStatic` object
+as well as the variable name of the variable in the data file, and the ClimaCore space
+of the model simulation.
+"""
```
juliasloan25, line 118: could add a note that this returns a Field of scalars and is analogous to the non-struct 1D case

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+Pass the ``data" as a function `data_function` which takes coordinates as arguments, 
+and  the ClimaCore space of the model simulation.
+"""
+function SpaceVaryingInput(
+    data_function::Function,
+    space::ClimaCore.Spaces.AbstractSpace,
+)
+    model_value = ClimaCore.Fields.zeros(space)
+    coords = ClimaCore.Fields.coordinate_field(space)
+    @. model_value = data_function(coords)
```
juliasloan25, line 41: before, we wrapped the function call in `FT`. Do we still need to do that here?

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,133 @@
...
+    data::PDS,
+    varname::N,
+    space::S,
+) where {
+    PDS <: FileReader.PrescribedDataStatic,
+    N <: String,
+    S <: ClimaCore.Spaces.SpectralElementSpace2D,
+}
+    model_value = ClimaCore.Fields.zeros(space)
+    model_value .= FileReader.get_data_at_date(data, space, varname)
```
juliasloan25, line 129: before, we wrapped the `get_data_at_date` call in `FT`. This was needed because that function returns whatever's stored in `data.file_state.data_fields`, which isn't currently converted to `FT` as it's regridded. I have a fix for that in the global bucket PR ([here](https://github.com/CliMA/ClimaLand.jl/pull/482/files#diff-f9e76ad8c8a7afdffe40057e6e9cefae9e0dca3fe839c98f5ffae5088c7fe4e6R302-R306)), but we could also make that change in this PR, or just continue to wrap the call in `FT` for now

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -418,21 +485,20 @@ function ClimaLand.boundary_flux(
...
-    K_c = Fields.Field(
-        Fields.field_values(Fields.level(p.soil.K, p_len)),
-        axes(Δz),
-    )
-    ψ_c = Fields.Field(
-        Fields.field_values(Fields.level(p.soil.ψ, p_len)),
-        axes(Δz),
-    )
+    surface_space = axes(Δz)
+    K_c = get_top_surface_field(p.soil.K, surface_space)
```
juliasloan25, line 489: should these be named `K_f` and `\psi_f` since they're actually on the face space? or `K_bc` and `\psi_bc` for consistency w the parameters?

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -626,12 +693,12 @@ function ClimaLand.∂tendencyBC∂Y(
         top = Operators.Extrapolate(),
     )
     K = Fields.level(interpc2f_op.(p.soil.K), face_len)
-    ϑ_l = Fields.level(interpc2f_op.(Y.soil.ϑ_l), face_len)
+    dψ = Fields.level(
+        interpc2f_op.(dψdϑ.(hydrology_cm, Y.soil.ϑ_l, ν, θ_r, S_s)),
```
juliasloan25, line 697: does it matter if we interpolate before or after the call to `dψdϑ`?

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -2,54 +2,48 @@
...
-of the soil are constant.
+ equation system. 
 $(DocStringExtensions.FIELDS)
 """
-struct EnergyHydrologyParameters{
-    FT <: AbstractFloat,
-    C <: AbstractSoilHydrologyClosure,
-    PSE,
-}
+struct EnergyHydrologyParameters{FT <: AbstractFloat, F, C, PSE}
```
juliasloan25, line 8: for RichardsParameters we specify `F <: Union{AbstractFloat, ClimaCore.Fields.Field}` - it might be nice to do that here too

test/shared_utilities/space_varying_inputs.jl
```diff
@@ -0,0 +1,75 @@
...
+        xlim = xlim,
+        ylim = ylim,
+        zlim = zlim,
+        nelements = nelements,
+        npolynomial = 0,
+    )
+    column = Domains.Column(; zlim = zlim, nelements = nelements[3])
+
+    domains = [shell, box, column]
+    analytic_func = (coords) -> 2.0
```
juliasloan25, line 42: we could define this as `analytic_func = (lat, lon) -> 2.0` to make sure that the coords are of the right form for each domain, but like this is also fine :)

test/shared_utilities/space_varying_inputs.jl
```diff
@@ -0,0 +1,75 @@
...
+using ClimaLand.SpaceVaryingInputs: SpaceVaryingInput
+using ClimaLand: Domains
+
+using ClimaCore: Fields
+using ClimaComms
+
+device = ClimaComms.device()
+context = ClimaComms.SingletonCommsContext(device)
+
+@testset "SpaceVaryingInput" begin
```
juliasloan25, line 12: Could you add a comment for which cases this tests (0D and 2D without structs, I think)?

docs/tutorials/standalone/Soil/layered_soil.jl
```diff
@@ -0,0 +1,171 @@
...
+    CTS.ClimaODEFunction(
+        T_exp! = exp_tendency!,
+        T_imp! = SciMLBase.ODEFunction(imp_tendency!; jac_kwargs...),
+        dss! = ClimaLand.dss!,
+    ),
+    Y,
+    (t0, tf),
+    p,
+)
+saveat = [0.0, 8.0, 16.0, 24.0, 32.0, 40.0, 60.0] .* 60 # chosen to compare with data in plots in paper
```
juliasloan25, line 144: It might be nice to also plot the data from the paper for comparison in the same plot like we do with the RRE Bonan comparison, but I don't think we have to :)

src/shared_utilities/SpaceVaryingInputs.jl
Sbozzolo: I would suggest using `space::ClimaCore.Spaces.AbstractSpace` instead of `wehre {S ....`.

The difference is that the former compiles one single method for every space, the other one compiles a method for each different type. We don't need this level of specialization given that this is only run at initialization and it is not performance critical.

src/shared_utilities/SpaceVaryingInputs.jl
Sbozzolo: I think that this is not needed. `model_value` will have its own float type already, so the cast should be implicit.

src/shared_utilities/SpaceVaryingInputs.jl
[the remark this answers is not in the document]
Sbozzolo: This is what goes in the mind of a reader: you are creating a new Field, just to fill with with the values provided by another function and returning that? Why not returning the other function directly?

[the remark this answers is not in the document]
kmdeck: Ill change it, and the other case where i did something similar. I had thought being more verbose would be clearer but if it just looks odd, there is no reason to keep it.

src/shared_utilities/SpaceVaryingInputs.jl
kmdeck: this can be used for any space, yes! whatever your function is just needs to unpack the coordinates correctly (coords.lat, coords.lon, coords.x, etc.)

docs/tutorials/standalone/Soil/layered_soil.jl
kmdeck: I'd like to do that too! and in the soil paper we can because the people who collected the data would be co-authors. but I need to check with them if we can share it publicly. The data that is currently in this tutorial is in a table in their paper and so is public.

src/shared_utilities/SpaceVaryingInputs.jl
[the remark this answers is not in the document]
Sbozzolo: A reason you might want to keep it this is for the FT as discussed somewhere else, but then you should say so in a comment.

src/ClimaLand.jl
kmdeck: instead of having both the include and the using statements?

juliasloan25: yeah

src/shared_utilities/SpaceVaryingInputs.jl
kmdeck: in the non struct case, the dest type is a scalar/float so it is automatic, because the data we are interpolating is also scalar/float

src/shared_utilities/SpaceVaryingInputs.jl
kmdeck: Good question. I removed it here, but then in the tests I had to add it back. For example, if we have f = (coords) -> 2, we cant broadcast that over coordinates which correspond to a floating point field:

```
julia> column = Domains.Column(; zlim = zlim, nelements = nelements[3])
ClimaLand.Domains.Column{Float32}((-1.0f0, 0.0f0), (10,), nothing, (:bottom, :top), (surface = ClimaCore.Spaces.PointSpace{ClimaComms.SingletonCommsContext{ClimaComms.CPUSingleThreaded}, ClimaCore.DataLayouts.DataF{ClimaCore.Geometry.LocalGeometry{(3,), ClimaCore.Geometry.ZPoint{Float32}, Float32, StaticArraysCore.SMatrix{1, 1, Float32, 1}}, SubArray{Float32, 1, Matrix{Float32}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}}(ClimaComms.SingletonCommsContext{ClimaComms.CPUSingleThreaded}(ClimaComms.CPUSingleThreaded()), ClimaCore.DataLayouts.DataF{ClimaCore.Geometry.LocalGeometry{(3,), ClimaCore.Geometry.ZPoint{Float32}, Float32, StaticArraysCore.SMatrix{1, 1, Float32, 1}}, SubArray{Float32, 1, Matrix{Float32}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}
  Float32[0.0, 0.1, 0.05, 10.0, 0.1, 10.0, 100.0, 0.01]), subsurface = CenterFiniteDifferenceSpace:
  context: SingletonCommsContext using CPUSingleThreaded
  mesh: 10-element IntervalMesh of IntervalDomain: z ∈ [-1.0,0.0] (:bottom, :top)))

julia> f = (coords) -> 2
#1 (generic function with 1 method)

julia> space = column.space.subsurface
CenterFiniteDifferenceSpace:
  context: SingletonCommsContext using CPUSingleThreaded
  mesh: 10-element IntervalMesh of IntervalDomain: z ∈ [-1.0,0.0] (:bottom, :top)

julia> coords = Fields.coordinate_field(space)
ClimaCore.Geometry.ZPoint{Float32}-valued Field:
  z: Float32[-0.95, -0.85, -0.75, -0.65, -0.55, -0.45, -0.35, -0.25, -0.15, -0.05]

julia> f.(coords)
ERROR: Struct type Int64 cannot be represented using base type Float32
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] macro expansion
   @ ~/.julia/packages/ClimaCore/wYSFf/src/DataLayouts/struct.jl:34 [inlined]
 [3] check_basetype(::Type{Float32}, ::Type{Int64})
   @ ClimaCore.DataLayouts ~/.julia/packages/ClimaCore/wYSFf/src/DataLayouts/struct.jl:33
 [4] (ClimaCore.DataLayouts.VF{Int64})(array::Matrix{Float32})
   @ ClimaCore.DataLayouts ~/.julia/packages/ClimaCore/wYSFf/src/DataLayouts/DataLayouts.jl:900
 [5] similar(bc::Base.Broadcast.Broadcasted{ClimaCore.DataLayouts.VFStyle{Array{…}}, Nothing, var"#1#2", Tuple{ClimaCore.DataLayouts.VF{…}}}, ::Type{Int64})
   @ ClimaCore.DataLayouts ~/.julia/packages/ClimaCore/wYSFf/src/DataLayouts/broadcast.jl:279
 [6] similar(bc::Base.Broadcast.Broadcasted{ClimaCore.Fields.FieldStyle{…}, ClimaCore.Spaces.FiniteDifferenceSpace{…}, var"#1#2", Tuple{…}}, ::Type{Int64})
   @ ClimaCore.Fields ~/.julia/packages/ClimaCore/wYSFf/src/Fields/broadcast.jl:153
 [7] copy(bc::Base.Broadcast.Broadcasted{ClimaCore.Fields.FieldStyle{…}, ClimaCore.Spaces.FiniteDifferenceSpace{…}, var"#1#2", Tuple{…}})
   @ ClimaCore.Fields ~/.julia/packages/ClimaCore/wYSFf/src/Fields/broadcast.jl:106
 [8] materialize(bc::Base.Broadcast.Broadcasted{ClimaCore.Fields.FieldStyle{ClimaCore.DataLayouts.VFStyle{…}}, Nothing, var"#1#2", Tuple{ClimaCore.Fields.Field{…}}})
   @ Base.Broadcast ./broadcast.jl:903
 [9] top-level scope
   @ REPL[21]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

So, I will add the FT back into src

src/shared_utilities/SpaceVaryingInputs.jl
kmdeck: will do!

src/standalone/Soil/boundary_conditions.jl
kmdeck: hmmm well we actually do want the center value of psi for the finite difference approx of the gradient (Y_bc - Y_c). but we need to put it on the right space in order to carry out the computation (because clima core wont let us extract a level of a center valued field and combine with a face valued field).

Ill add some documentation

src/standalone/Soil/boundary_conditions.jl
kmdeck: it would in general matter if dψdϑ is nonlinear (and it is). I think we might still want to do it in this order (compute the field and then interpolate, instead of computing interpolated arguments and then computing the result), but im not sure. 

But, "interpc2f_op" just takes the top value as the face value, so for this particular use case (getting the top value of dψdϑ) it shouldnt matter, right?

test/shared_utilities/space_varying_inputs.jl
kmdeck: I wanted to keep this general, because the function gets the coordinate field (which has the values stored as coords.lat, coords.lon, or coords.lat, coords.lon, coords.z, or just coords.z, etc). then we only need one method that expects f(coords) instead of one that unpacks them correctly for each space (and the order of arguments is not fixed).

test/shared_utilities/space_varying_inputs.jl
kmdeck: added!

src/ClimaLand.jl
kmdeck: https://discourse.julialang.org/t/how-to-include-my-module-file-as-a-package/75682/12

ill try it out now. if it works it's just a couple of changes and might get done faster than if it is an issue

src/shared_utilities/SpaceVaryingInputs.jl
kmdeck: this is interesting! but if I do `model_value .= f.(coords)`, the broadcasting also does the type conversion. @Sbozzolo so it seems like, at least for the analytic input, we need to do 
```
    model_value = ClimaCore.Fields.zeros(space)
    coords = ClimaCore.Fields.coordinate_field(space)
    return model_value .= data_function.(coords)
```

then we can remove the FT from src

Sbozzolo: Please, add comments explaining why you do that :)

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,152 @@
...
+    data_values::NamedTuple,
+    space::S,
+    dest_type::Type{DT},
+) where {S <: ClimaCore.Spaces.CenterFiniteDifferenceSpace, DT}
+    zvalues = ClimaCore.Fields.coordinate_field(space).z
+    data_keys = keys(data_values)
+    model_value = map(zvalues) do z
+        args = map(data_keys) do key
+            key => linear_interpolation(data_z, getproperty(data_values, key), z)
+        end
```
Sbozzolo, line 125: Probably something like would work
```suggestion
args = map(data_values) do value
    return linear_interpolation(data_z, value, z)
end
```
`data_z` has to be GPU-compatible (`CuArray` or `SVector`)

Sbozzolo: This works because you are not passing the `key` to the GPU. Everything is run on GPU.

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,169 @@
...
+        key => AT(getproperty(data_values, key))
+    end
+    data_values_AT = NamedTuple(data_values_AT)
+    =#
+    # now create the field of structs
+    model_value = map(zvalues) do z
+        args = map(data_values) do AT(value)
+            return linear_interpolation(data_z, value, z)
+        end
+        DT(; args...)
```
Sbozzolo, line 138: Okay, now the issue is that `DT` is not isbits.

Sbozzolo: @dennisYatunin can you help us here?

What we are trying to do here is to create a Field of struct where the type of struct is dynamically determined. This function `SpaceVaryingInput` will be used to set up general space-varying parameter fields.

dennisYatunin: Oh actually this just looks like invalid syntax. You can't write `do AT(value)`; you need to write `do value` and then pass `AT(value)` to the `linear_interpolation`.

kmdeck: DT should be ok, I think. Just a struct of floats

we are converting the depth coordinates (data_z) to AT, and then also each value of a NamedTuple of Vectors. let me try what you suggested, and happy to refactor too

kmdeck: I dont think we can call AT(value) within the loop because AT itself is not isbits

kmdeck: yay thank you @dennisYatunin !

src/shared_utilities/SpaceVaryingInputs.jl
juliasloan25: I think your second comment explains it! Because the broadcasting when we set the new value of `model_value` in `model_value .= f.(coords)` is what does the typecasting

src/standalone/Soil/boundary_conditions.jl
juliasloan25: That makes sense to me

juliasloan25 (approved):

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -2,54 +2,53 @@
...
- equation system. In this simplest
-form, we assume the conductivity and volumetric heat capacity
-of the soil are constant.
+ equation system. 
 $(DocStringExtensions.FIELDS)
 """
 struct EnergyHydrologyParameters{
     FT <: AbstractFloat,
-    C <: AbstractSoilHydrologyClosure,
+    F <: Union{AbstractFloat, ClimaCore.Fields.Field},
```
Sbozzolo, line 10: This doesn't guarantee that the struct is concretely typed. F could be `AbstractFloat`, which is not concrete. What you want is `F <: Union{<: Real, <: ClimaCore.Fields.Field}` I think.
Or
`F <: Union{FT, FIELD} where {FIELD <: ClimaCore.Fields.Field}`

src/standalone/Soil/rre.jl
```diff
@@ -1,33 +1,30 @@
...
+    RichardsParameters{F <: Union{AbstractFloat, ClimaCore.Fields.Field}, C <: AbstractSoilHydrologyClosure}
 
 A struct for storing parameters of the `RichardModel`.
 $(DocStringExtensions.FIELDS)
 """
-struct RichardsParameters{
-    FT <: AbstractFloat,
-    C <: AbstractSoilHydrologyClosure,
-}
+struct RichardsParameters{F <: Union{AbstractFloat, ClimaCore.Fields.Field}, C}
```
Sbozzolo, line 7: Same comment as before about `F`.

docs/tutorials/standalone/Soil/layered_soil.jl
```diff
@@ -0,0 +1,171 @@
...
+# Initial conditions
+Y.soil.ϑ_l .= 0.0353; # read from Figure 4 of Huang et al. 
+
+# We also set the initial conditions of the auxiliary state here:
+set_initial_cache! = make_set_initial_cache(soil)
+set_initial_cache!(p, Y, t0);
+
+# Timestepping:
+stepper = CTS.ARS111()
+err = (FT == Float64) ? 1e-8 : 1e-4
```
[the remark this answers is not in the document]
dennisYatunin: `Dict`s are not type-stable. Something like `@assert FT in (Float32, Float64)` would be better for this, as it could be optimized out by the compiler.

src/shared_utilities/general_utils.jl
```diff
@@ -16,3 +16,30 @@ function searchsortednearest(a, x)
...
+1-d vector `dep_vars`.
+
+If the `indep_value` is outside the range of `indep_vars`, this
+returns the endpoint value closest.
+"""
+function linear_interpolation(indep_vars, dep_vars, indep_value)
+    N = length(indep_vars)
+    id = searchsortedfirst(indep_vars, indep_value)
+    if id == 1
+        dep_vars[id]
```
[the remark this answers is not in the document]
kmdeck: fixed!

src/standalone/Soil/energy_hydrology.jl
kmdeck: Based off of this: https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type, this seems correct now.

I went with `F <: Union{<: AbstractFloat, <: ClimaCore.Fields.Field}`, because `Real` is an abstract type that includes ints as concrete examples, which isnt what we want.

kmdeck: good catch!

src/shared_utilities/SpaceVaryingInputs.jl
```diff
@@ -0,0 +1,164 @@
...
+function SpaceVaryingInput(
+    data::PDS,
+    varname::N,
+    space::S,
+) where {
+    PDS <: FileReader.PrescribedDataStatic,
+    N <: String,
+    S <: ClimaCore.Spaces.SpectralElementSpace2D,
+}
+    return FileReader.get_data_at_date(data, space, varname)
```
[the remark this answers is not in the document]
kmdeck: @juliasloan25 this is OK now, I think? from this PR/commit:
https://github.com/CliMA/ClimaLand.jl/commit/ebb59735e9a3dacac22081c5abec4190611fb044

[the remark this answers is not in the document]
kmdeck: I added a test to the current albedo test using it, and when FT = Float32 the returned field is Float32.

src/standalone/Soil/energy_hydrology.jl
dennisYatunin: In Julia, `Union` and `Tuple` are special in terms of how their type parameters are treated. Here are some examples
```
julia> 1 isa Union{Real, Fields.Field}
true
julia> (1,) isa Tuple{Real}
true
julia> [1] isa Vector{<:Real}
true
julia> [1] isa Vector{Real}
false
```

PR #486, opened 2024-02-07
Clarify role of n_layers in radiative transfer

### LINKED ISSUE #270: Revisit n_layers in plant hydraulics vs radiation model
Just a note to keep track of the fact that right now, we dont enforce the number of layers in the two stream model to be the same as the number of layers in the plant hydraulics model. I think this is fine right now, but we may need to revisit this if we move to a proper multi-layer canopy (leaf temp, potential, etc modeled in each layer).

Espeer5:
## Purpose 
This PR clarifies the role of `n_layers` in the `TwoStream` radiative transfer model and differentiates it from the vertical discretization of the canopy. Will close #270

In the TwoStream model, total absorbed radiation is computed by integrating over the total LAI in the canopy in vertical layers. The `n_layers` parameter specifies the size of the partition for integrating. Without a vertical profile for LAI, this is just a simple constant whose value beyond 20 layers doesn't change the resulting value very much. This is therefore unrelated to the number of vertical layers in the canopy discretization used in the hydraulics model.

## Content

 - Remove `n_layers` from any places where parameters are specified -> it should be constant everywhere
 - Clarify the purpose of `n_layers` in the `TwoStreamParameters` documentation

----

### DIFF
--- a/docs/tutorials/integrated/soil_canopy_tutorial.jl
+++ b/docs/tutorials/integrated/soil_canopy_tutorial.jl
@@ -287,7 +287,6 @@ radiative_transfer_args = (;
         Ω = FT(0.69),
         λ_γ_PAR = FT(5e-7),
         λ_γ_NIR = FT(1.65e-6),
-        n_layers = UInt64(20),
     )
 )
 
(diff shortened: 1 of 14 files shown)


kmdeck (approved):
thank you!!

PR #487, opened 2024-02-07
Add package extension for CLIMAParameters

nefrathenrici:
This PR adds a package extension in `ext/CreateParametersExt.jl`, which provides constructors for `LandParameters` via a FT or a toml dict. I have removed the `create_lsm_parameters(FT)` function and replaced it with  a function`LandParameters(FT)`

This should be merged before the other parameter PRs, since those changes should also go in the extension to avoid adding CLIMAParameters as a direct dependency

Two issues:

- Insolation is needed as a direct dependency for the package extension
- `Aqua.test_stale_deps` doesn't work well with package extensions. Insolation is never used directly in src, so Aqua thinks that it is a stale dependency. See [here](https://juliatesting.github.io/Aqua.jl/dev/stale_deps/#Stale-dependencies) for more info.


Sbozzolo:
Extensions require julia 1.9, so you have to bump that too.

experiments/integrated/ozark/conservation/ozark_conservation.jl
```diff
@@ -14,13 +13,14 @@ using ClimaLand.Canopy
 using ClimaLand.Canopy.PlantHydraulics
 import ClimaLand
 import ClimaLand.Parameters as LP
+import CLIMAParameters
```
juliasloan25, line 16: is CLIMAParameters being used in this file?

experiments/standalone/Biogeochemistry/experiment.jl
```diff
@@ -9,15 +9,15 @@ using ClimaLand.Soil.Biogeochemistry: MicrobeProduction
 using Dates
 
 import ClimaLand.Parameters as LP
-include(joinpath(pkgdir(ClimaLand), "parameters", "create_parameters.jl"))
+import CLIMAParameters
```
juliasloan25, line 12: same here

experiments/standalone/Bucket/global_bucket.jl
```diff
@@ -7,10 +7,10 @@ using Statistics
 import ClimaTimeSteppers as CTS
 using ClimaCore
 using ClimaCore: Remapping, Geometry
-import CLIMAParameters as CP
 import ClimaComms
 import ClimaLand
-include(joinpath(pkgdir(ClimaLand), "parameters", "create_parameters.jl"));
+import ClimaLand.Parameters as LP
+import CLIMAParameters
```
juliasloan25, line 13: same here too

docs/tutorials/shared_utilities/driver_tutorial.jl
```diff
@@ -77,7 +76,7 @@ SW_d = TimeVaryingInput(seconds, SW_d);
 # Finally, for many models we also need to specify the function
 # for computing the zenith angle as a function of simulation time.
 # To do so, we use the `Insolation` package as follows:
-earth_param_set = create_lsm_parameters(Float64);
+earth_param_set = ClimaLand.Parameters.LandParameters(Float64);
```
juliasloan25, line 79: can use LP.LandParameters here

experiments/integrated/ozark/conservation/ozark_conservation.jl
nefrathenrici: It's needed to load the package extension code, basically replacing the old `include(joinpath(pkgdir(ClimaLand), "parameters", "create_parameters.jl"))`

juliasloan25 (approved):

PR #491, opened 2024-02-09
rm ClimaCore deprecated functions

### LINKED ISSUE #490: remove deprecated function calls
We have at least one deprecated function call in our code. ClimaCore is planning a release today (2/9/24) or next week that will remove support for these, so we need to fix them

juliasloan25:
## Purpose 
update deprecated ClimaCore function calls

closes #490

### DIFF
--- a/src/shared_utilities/Domains.jl
+++ b/src/shared_utilities/Domains.jl
@@ -195,7 +195,8 @@ end
         ylim::Tuple{FT,FT},
         nelements::Tuple{Int,Int},
         periodic::Tuple{Bool,Bool},
-        npolynomial::Int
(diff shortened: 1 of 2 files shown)


kmdeck (approved):
Looks fine, thank you Julia!

Just one question - will we need to eventually pass in the comms_context from the driver? If that is the case, should we add it to the arguments of the domain constructors instead of making internally?

juliasloan25:
> Just one question - will we need to eventually pass in the comms_context from the driver? If that is the case, should we add it to the arguments of the domain constructors instead of making internally?

Good point - I'll add it as optional parameters to the domain constructors

PR #493, opened 2024-02-12
Add PFTs to ClimaLand src

Espeer5:
## Purpose 
Adds PFT struct and default scheme to ClimaLand src and adds an experiment running Ozark site with PFTs.

## Todo:
 - [x] Add PFT Ozark run to buildkite
 - [x] Pass all tests before marking ready for review
 - [x] Add tests for proper PFT construction, proper assignment of parameters
 - [x] Repair race condition in buildkite
 - [x] Repair bug in assignment
 - [x] Model Stability issue? After Rebase?

## Content
 - pfts.jl added to Canopy.jl along with a function for loading PFT parameters into the namespace by most dominant PFT in a grid cell
 - Experiment added which runs Ozark site as a deciduous broadleaf PFT type.

----


src/standalone/Vegetation/pfts.jl
```diff
@@ -0,0 +1,491 @@
+"""
+This file contains the PFT struct used to prescribe vegetation parameters to the 
+model when running using PFTs, as well as the default PFT scheme. ClimaLand 
+supports running using default PFTs, user defined PFTs, or fully prescribed 
+vegetation parameters.
+
+The default PFTs used in the model use parameter values drawn from defferent
```
kmdeck, line 7: defferent -> different

src/standalone/Vegetation/pfts.jl
```diff
@@ -0,0 +1,491 @@
...
+    :α_PAR_leaf, # PAR Leaf Reflectance
+    :α_NIR_leaf, # Near-infrared Leaf Reflectance
+    :τ_PAR_leaf, # PAR Leaf Transmittance
+    :τ_NIR_leaf, # Near-infrared Leaf Transmittance
+
+    # Conductance Model
+    :g1, # Medlyn CO2 Sensitivity
+
+    # Photosynthesis model
+    :Vcmax25, # Max rate of carboxylation of Rubisco
```
kmdeck, line 42: Small comment we could add - when running the the Optimality Farquhar model, this is not used.

src/standalone/Vegetation/pfts.jl
```diff
@@ -0,0 +1,491 @@
...
+using DataFrames
+
+# List of parameters that PFTs need to define in order to be valid
+param_list = [
+    # TwoStreamModel parameters
+    :Ω,          # Clumping Index
+    :α_PAR_leaf, # PAR Leaf Reflectance
+    :α_NIR_leaf, # Near-infrared Leaf Reflectance
+    :τ_PAR_leaf, # PAR Leaf Transmittance
+    :τ_NIR_leaf, # Near-infrared Leaf Transmittance
```
kmdeck, line 36: do we include emissivity?

kmdeck: or is that constant over all PFTs

kmdeck: Also the leaf angle distribution l_d

src/standalone/Vegetation/pfts.jl
```diff
@@ -0,0 +1,491 @@
...
+
+    # Conductance Model
+    :g1, # Medlyn CO2 Sensitivity
```
kmdeck, line 39: I think we also need something related to the specific heat of the canopy. The model uses ac_canopy, or the heat capacity per unit area (heat capacity per unit volume *height of canopy). I will try and find a table of values.

src/standalone/Vegetation/pfts.jl
```diff
@@ -0,0 +1,491 @@
...
+
+    # Photosynthesis model
+    :Vcmax25, # Max rate of carboxylation of Rubisco
+
+    # Plant Hydraulics and general plant parameters
+    :SAI,             # Avg Stem Area Index
```
kmdeck, line 45: I believe this is site specific, not PFT specific. perhaps a PFT specific parameter is LAI/SAI? @braghiere

src/standalone/Vegetation/pfts.jl
```diff
@@ -0,0 +1,491 @@
...
+    :f_root_to_shoot, # Root to Shoot Ratio
+    :K_sat_plant,     # Hydraulic Conductivity at Saturation
+    :ψ63,             # Xylem water potential at 63% loss of conductivity
+    :capacity,        # Water Storage Capacity
```
kmdeck, line 49: I think that this is a derived parameter, and the fundamental parameter is the plant porosity,
rather than:
    `plant_ν =
        params.plant_hydraulics.capacity / (maxLAI * setup.h_leaf) / FT(1000)`
we should solve for capacity given nu, max LAI, h_leaf.

src/standalone/Vegetation/pfts.jl
```diff
@@ -0,0 +1,491 @@
...
+ensure all required parameters are defined for the PFT.
+"""
+struct Pft
+    name::String
+    parameters::NamedTuple
+
+    # Inner constructor to verify all required parameters are defined
+    function Pft(name, parameters)
+        defined_param_names = keys(parameters)
+        for param in param_list
```
kmdeck, line 67: Im not sure if it matters, but is param_list globally accessible? should it be defined within the constructor?

Espeer5: My thinking was that we should export the param_list from the pft module so that if a user wants to know which parameters are defined by PFT structs then they could potentially view that list but if you don't think that useful I'm happy to move it into the constructor.

src/standalone/Vegetation/pfts.jl
Espeer5: Got it

src/standalone/Vegetation/pfts.jl
braghiere: Have a look at this paper: https://link.springer.com/article/10.1007/s00382-019-04924-9 (Table 1) Screenshot 2024-02-21 at 2 50 00 PM

braghiere: Also, we talked about implementing the leaf angular distribution later on, instead of using 0.5. Can we quickly add what CLM does? It is pretty straightforward. See Eq. 3.3 (https://www2.cesm.ucar.edu/models/cesm2/land/CLM50_Tech_Note.pdf).

𝐺 (𝜇) = 𝜑1 + 𝜑2𝜇 (3.3)

where 𝜑1 = 0.5 − 0.633𝜒𝐿 − 0.33𝜒𝐿2
 and 𝜑2 = 0.877 (1 − 2𝜑1) for −0.4 ≤ 𝜒𝐿 ≤ 0.6. 

𝜒𝐿 is a PFT parameter.

src/standalone/Vegetation/pfts.jl
braghiere: There is a parameter called: 
dsladlai      = dSLA/dLAI: rate of change of SLA with LAI through canopy, projected area basis (m^2/gC)

The link to all CLM input files is here: https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/lnd/clm2/

PFTs here: https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/lnd/clm2/pftdata/
Surface data (maps) here: https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/lnd/clm2/surfdata/

src/standalone/Vegetation/pfts.jl
Espeer5: I think I should add this change to `ld` in a short follow up PR... Since the formula for `ld` requires the inverse optical depth and the solar zenith angle it will need to be computed within the radiative transfer model itself. So here it this PR I will include XL as a PFT param, and then I will create a second PR that adds this change to ld everywhere.

src/standalone/Vegetation/pfts.jl
Espeer5: Following Land Team meeting on 2/21, leaving SAI as a site parameter and not parameterizing it in terms of PFTs for the moment. Something I didn't mention in the meeting but that I think is worth noting is that our MODIS LAI = SAI + LAI is read in automatically based on lat/long and computing an approximation for SAI from that value rather than as a PFT param can probably result in higher accuracy while still getting the value automatically as would including as a PFT value.

Espeer5:
@kmdeck after rebase, I see failures in the unit tests, but it seems like the same tests failed on the most recent run on main, is something broken here?

Otherwise this PR should be ready to go

kmdeck:
this was something that came up from some changes made to an upstream repo. If you pin to Insolation 0.9.1 in the main Project.toml, it should work.
`Insolation = "=0.9.1"`

experiments/integrated/fluxnet/met_drivers_FLUXNET.jl
```diff
@@ -217,4 +217,4 @@ LAIfunction = TimeVaryingInput(LAI_seconds, FT.(MODIS_LAI[:, 2]); context)
 # defined in the parameters file for each site
 maxLAI = FT(maximum(MODIS_LAI[:, 2]))
 RAI = maxLAI * f_root_to_shoot
-plant_ν = capacity / (maxLAI * h_leaf) / FT(1000)
+capacity = plant_ν * maxLAI * h_leaf / FT(1000)
```
kmdeck, line 220: small mistake: nu = c/(LAI*h)/1000 -> c = 1000*LAI*h*nu

src/standalone/Vegetation/pfts.jl
```diff
@@ -0,0 +1,534 @@
...
+        g1 = 74.31, # kPa^(1/2) ⋅ CLM5.0 Table 2.9.1
+
+        # Photosynthesis Model
+        Vcmax25 = 5.1e-5, # mol CO2/m^2/s ⋅ CLM2 PFT Data
+
+        # Plant Hydraulics and general plant parameters
+        f_root_to_shoot = 1.0, # Unitless ⋅ CLM2 PFT Data
+        K_sat_plant = 5e-9,    # m/s
+        ψ63 = -4 / 0.0098,     # / MPa to m
+        plant_ν = 2e-4,        # Unitless
```
kmdeck, line 115: maybe just double check the order of magnitude on this given the /1000 vs *1000 typo above?
fine to set these all equal for now :)

kmdeck (approved):
looks good to me once CI passes, 1000 factor fix is made, and a quick look at the CI plots for the PFT ozark run to make sure it looks reasonable.

thanks very much!!

src/standalone/Vegetation/pfts.jl
kmdeck: I was maybe confused and thinking of the case of a script that has something like:
```
a = [1,2,3]
function this_gets_called_every_timestep(x)
    x .= a
end
```
which would have a slowdown because `a` is defined only in global scope. But I think that isnt the same case when something is defined within a module, you have with` param_list`? Also, Pft() is only called once, not every step, but still would be good to avoid if it *is* an issue.

braghiere (approved):

AlexisRenchon (approved):
Hey Ed, 
It looks great! Thanks!

PR #494, opened 2024-02-13
ClimaLandSimulations make all figures of experiments

AlexisRenchon:
## Purpose 

This PR main goal is to be up to date with experiments/integrated/fluxnet, meaning it will make all the same figures, and ensure they are doing exactly the same thing. 

## To-do

- [x] Fixed README link to install unregistered package ClimaLandSimulations
- [x] Climaland_output_dataframe.jl - output by layers
- [x] Diurnals: better legend subplot 1, GPP, ER, AR, HR?
- [x] Plot cumulative P and ET
- [x] Index_t_start and index_t_start removed from Makie plot args
- [x] Model start and end is set in make_setup() which gets site default values (which can be modified), e.g., ozark_default_args 
- [x] CI workflow (.yml) for ClimaLandSimulation test
- [x] Test: ozark, default args
- [x] Test: ozark, change default start time (t0) 
- [x] Test: ozark, change default parameters 
- [x] Site default as function returning tuples instead of tuple, for more user flexibility
- [x] Timeseries: lines instead of scatter
- [x] WIP: added more fingerplots
- [ ] Test: all sites, default args
- [ ] Makie_plots.jl, make sure to use inputs and outputs in units that we want, instead of hard-coding conversion factors which can be confusing (currently error in cumulative water units and timeseries water units)
- [ ] In experiments/integrated/fluxnet, we compare model to data only when driver observation are available (not gap-filled), but we don't filter for gap-filled NEE / GPP / Reco. We should do so here.
- [ ] Use FLUXNET2015 datasets, drivers gapfilled with ERA
- [ ] climaland_output_dataframe.jl don't hardcode 20 (n layers)
- [ ] Inputs_dataframe.jl -> make a function to convert units (from, to) of a DataFrame, which can be called on input and output dataframe
- [ ] Flux timeseries: better legend radiation, SW_in LW_in, LW_out, G...
- [ ] Plot: only when met data was available (don't use gapfilled index) 
- [ ] Add widget to plot timeseries or diurnal etc. of use chosen variable 
- [ ] (!high prio) make all plots that /experiment/integrated makes, and make sure they are the same
- [ ] (!high prio) docstrings for all struct/functions
- [ ] (!high prio) better README.md
- [ ] (!high prio) docs (ClimaLandSimulation page in ClimaLand docs)
- [ ] (!high prio) move earth_param_set out of module
- [ ] (!high prio) move const FT = FLoat64 out of module
- [ ] use our src constructor for parameters (still need to fix hetero_resp and plant_hydraulics)
- [ ] less functions or clearer names (make_config(), make_setup()...) make_domain and make_timestepper can live in the same file (maybe setup_utils.jl) 
- [ ] modules as an arg (e.g., photosynthesis = Farquhar or photosynthesis = optimality)
- [ ] no hard-coded values 
- [ ] shorter list of params (use more defaults or constants)
- [ ] lib/ClimaLandSimulations/src/fluxnet_simulation.jl around line 266, why the +40? (this is in experiment/integrated as well)
- [ ] Should LAI, RAI, plant_\nu be in met_drivers?
- [ ] lib/ClimaLandSimulations/src/utilities/inputs_dataframe.jl do we always have all variables from fluxnet input files? (note: I think so, even if there are -9999 they should be there - if not we can create it)
- [ ] add else case for if haskey(...) in make_config
- [ ] rename LAIfunction to allow for other types of LAI input (e.g. netcdf)
- [ ] add PFTs functionality 
- [ ] Add Mitra gap-filling code in data_tools.jl (or another gap-filling method, maybe use ERA from FLUXNET)
- [ ] Experiment: all sites, not just Ozark

## Content

----

### DIFF
--- a/.github/workflows/ClimaLandSimulations.yml
+++ b/.github/workflows/ClimaLandSimulations.yml
@@ -1,0 +1,31 @@
+name: ClimaLandSimulations CI
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+permissions:
+  actions: write
+  contents: read
+
+jobs:
+  lib-climalandsimulations:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 45
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: '1.10'
+      - uses: julia-actions/cache@v1
+      - name: Install Julia dependencies
+        run: >
+          julia --project= -e 'using Pkg; Pkg.develop(path="$(pwd())"); Pkg.develop(path="$(pwd())/lib/ClimaLandSimulations")'
+      - name: Run the tests
+        continue-on-error: true
+        env:
+            CI_OUTPUT_DIR: output
+        run: >
+          julia --project= -e 'using Pkg; Pkg.test("ClimaLandSimulations")'
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,6 +23,8 @@ v0.11.1
 -------
 - ![][badge-✨feature] Add option to profile albedo job. PR
   [#505](https://github.com/CliMA/ClimaLand.jl/pull/551)
+- ![][badge-✨feature] ClimaLandSimulations: better plots (legend and style tweaks, added cumulative ET and P), unit tests, start and end time as an optional argument. PR [#494](https://github.com/CliMA/ClimaLand.jl/pull/494)
+- ![][badge-🐛bugfix] ClimaLandSimulations: installation readme. PR [#494](https://github.com/CliMA/ClimaLand.jl/pull/494)
 
 v0.11.0
 -------
--- a/lib/ClimaLandSimulations/README.md
+++ b/lib/ClimaLandSimulations/README.md
@@ -1,16 +1,27 @@
 # ClimaLandSimulations
 
 A library of methods to run ClimaLand:
-- at single sites, such as FLUXNET
+- at single sites 
 - globally (WIP)
 
 ## Installation
 
 ```jl
 julia> ] # to go into pkg mode
-pkg> add https://github.com/CliMA/ClimaLand.jl/tree/main/lib/ClimaLandSimulations # because unregistered for now
+pkg> add https://github.com/CliMA/ClimaLand.jl:lib/ClimaLandSimulations # because unregistered for now
 pkg> *backspace* # press backspace button to go back into julia mode
 julia> using ClimaLandSimulations
 ```
 
 For examples, see the [experiments](https://github.com/CliMA/ClimaLand.jl/lib/ClimaLandSimulations/experiments) folder
+
+### Development of the `ClimaLandSimulations` subpackage
+
+    cd ClimaCore/lib/ClimaLandSimulations
+
+    # Add ClimaCore to subpackage environment
+    julia --project -e 'using Pkg; Pkg.develop(path="../../")'
+
+    # Instantiate ClimaCoreMakie project environment
+    julia --project -e 'using Pkg; Pkg.instantiate()'
+    julia --project -e 'using Pkg; Pkg.test()'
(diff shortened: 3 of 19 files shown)


AlexisRenchon:
@Sbozzolo Could you just check if `.github/workflows/ClimaLandSimulations.yml` looks good? It works, but I am not sure if it is exactly what we want.

.github/workflows/ClimaLandSimulations.yml
```diff
@@ -0,0 +1,34 @@
...
+    timeout-minutes: 45
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: '1.10'
+      - uses: actions/cache@v3
+        env:
+          cache-name: cache-artifacts
```
Sbozzolo, line 17: There's a cache action specific for julia, see it here: https://github.com/CliMA/ClimaDiagnostics.jl/blob/main/.github/workflows/CI.yml

It also requires the permissions block:
```
# Needed to allow julia-actions/cache to delete old caches that it has created
permissions:
  actions: write
  contents: read
```

.github/workflows/ClimaLandSimulations.yml
```diff
@@ -0,0 +1,34 @@
+name: ClimaLandSimulations CI
+on:
+  pull_request:
```
Sbozzolo, line 3: This will no run on main. You also need:
```yml
  push:
    branches:
      - main
```

.github/workflows/ClimaLandSimulations.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install Julia dependencies
+        run: >
+          julia --project=monorepo -e 'using Pkg; Pkg.develop(path="$(pwd())"); Pkg.develop(path="$(pwd())/lib/ClimaLandSimulations")'
+      - name: Run the tests
+        continue-on-error: true
+        env:
+            CI_OUTPUT_DIR: output
+        run: >
+          julia --project=monorepo -e 'using Pkg; Pkg.test("ClimaLandSimulations")'
+          && echo "TESTS_SUCCESSFUL=true" >> $GITHUB_ENV
```
Sbozzolo, line 34: There's a missing newline. Also, I don't you need `&& echo "TESTS_SUCCESSFUL=true" >> $GITHUB_ENV`

.github/workflows/ClimaLandSimulations.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - name: Install Julia dependencies
+        run: >
+          julia --project=monorepo -e 'using Pkg; Pkg.develop(path="$(pwd())"); Pkg.develop(path="$(pwd())/lib/ClimaLandSimulations")'
```
Sbozzolo, line 27: Why the `monorepo` project?

AlexisRenchon: I am not sure... It is what ClimaCore.jl does for its ci of projects in /lib folders, but I am not sure what it is for or if it is needed.  https://github.com/CliMA/ClimaCore.jl/blob/main/.github/workflows/ClimaCoreMakie.yml#L28

.github/workflows/ClimaLandSimulations.yml
AlexisRenchon: What do you mean by missing newline?

Sbozzolo: The last line of the file does not contain a newline (or carriage return) character as last character.
On GitHub, you can see this as a red circle after the last line

.github/workflows/ClimaLandSimulations.yml
Sbozzolo: I don't know why it's there, but you don't need it.

Sbozzolo (approved):

PR #497, opened 2024-02-13
negative snow/water test, beta

kmdeck:
## Purpose 
May close https://github.com/CliMA/ClimaCoupler.jl/issues/472

The default parameters of bucket model make it so the beta function recovers the SLIM model (linear beta, switch occurs at 0.75*bucket capacity, and no snow beta).

## To-do
TBD: make beta_snow and beta_bucket different functions
TBD: make them functions the user can pass in (for now, the user can control this with parameters of the function)

## Content
- Adds several checks of the following form: `W = max(W, 0)` in functions where being negative would return an incorrect answer and we want to treat the negative number like zero
- adds tests that ensure the tendencies and water fluxes are zero in the case of zero precip, snow melt, but negative snow and water content (evap = sublimation = 0)
- Employs a beta factor for snow. This has the same form as for the bucket, namely:
```
if x < x_c
    (x/x_c)^p
else
    1
end
```
This is what we had before for water, with `x_c = 0.75*W_f`, `x = W`, and `p=1`. We've now set p=2 to damp nonlinearly to zero. The branch is used for the bucket because the land evaporates at the potential rate if the bucket is full enough (beta=1), but then drops as it gets drier. The transition happens at x_c [SLIM supplement text Equation 36].

For snow, the sublimation rate should be the potential rate (beta = 1) until the snow vanishes. To achieve this, we use the same function for beta, but use `x_c = 0.1 S_c`, where `S_c` is the parameter used to control the albedo of the land surface. For snow levels about S_c, the albedo is the snow albedo, as S drops far below S_c, it becomes the bareground albedo. My only concern here is that the physical situation could be S<<S_c (patchy, thin snow), and the albedo-> bareground albedo, but the vapor flux -> 0, because we treat the snow cover fraction as zero or 1 in all cases but the albedo. In reality, in this situation, the vapor flux should tend towards the bareground vapor flux.
beta

Review checklist

I have:

----

### DIFF
--- a/src/standalone/Bucket/Bucket.jl
+++ b/src/standalone/Bucket/Bucket.jl
@@ -54,7 +54,7 @@ export BucketModelParameters,
     BulkAlbedoStatic,
     BulkAlbedoTemporal,
     surface_albedo,
-    partition_surface_fluxes
+    partition_snow_surface_fluxes
 
 include(
     joinpath(pkgdir(ClimaLand), "src/standalone/Bucket/artifacts/artifacts.jl"),
@@ -248,8 +248,14 @@ struct BucketModelParameters{
     albedo::AAM
     "Critical σSWE amount (m) where surface transitions from to snow-covered"
     σS_c::FT
+    "Fraction of critical amount of snow at which sublimation β begins to decay to zero (unitless)"
+    f_snow::FT
     "Capacity of the land bucket (m)"
     W_f::FT
+    "Fraction of bucket capacity at which evaporation β begins to decay to zero (unitless)"
+    f_bucket::FT
+    "Exponent used in β decay (unitless)"
+    p::FT
     "Roughness length for momentum (m)"
     z_0m::FT
     "Roughness length for scalars (m)"
@@ -269,13 +275,19 @@ BucketModelParameters(
     z_0m::FT,
     z_0b::FT,
     τc::FT,
-    earth_param_set::PSE,
+    earth_param_set::PSE;
+    f_snow = FT(0.0),
+    f_bucket = FT(0.75),
+    p = FT(1),
 ) where {FT, AAM, PSE} = BucketModelParameters{FT, AAM, PSE}(
     κ_soil,
     ρc_soil,
     albedo,
     σS_c,
+    f_snow,
     W_f,
+    f_bucket,
+    p,
     z_0m,
     z_0b,
     τc,
@@ -399,17 +411,17 @@ function make_compute_exp_tendency(model::BucketModel{FT}) where {FT}
         # the returned fluxes are computed correctly for the cell
         # regardless of snow-coverage.
 
-        # The below is NOT CORRECT if we want the snow
-        # cover fraction to be intermediate between 0 and 1.
+        # The below must be adjusted to compute F_sfc over snow and over soil
+        # if we want the snow cover fraction to be intermediate between 0 and 1.
         (; turbulent_fluxes, R_n) = p.bucket
         F_sfc = @. (turbulent_fluxes.shf .+ turbulent_fluxes.lhf + R_n) # Eqn (21)
         _T_freeze = LP.T_freeze(model.parameters.earth_param_set)
         _LH_f0 = LP.LH_f0(model.parameters.earth_param_set)
         _ρ_liq = LP.ρ_cloud_liq(model.parameters.earth_param_set)
         _ρLH_f0 = _ρ_liq * _LH_f0 # Latent heat per unit volume.
-        # partition energy fluxes
+        # partition energy fluxes for snow covered area
         partitioned_fluxes =
-            partition_surface_fluxes.(
+            partition_snow_surface_fluxes.(
                 Y.bucket.σS,
                 p.bucket.T_sfc,
                 model.parameters.τc,
@@ -419,7 +431,9 @@ function make_compute_exp_tendency(model::BucketModel{FT}) where {FT}
                 _ρLH_f0,
                 _T_freeze,
             )
-        (; F_melt, F_into_snow, G) = partitioned_fluxes
+        (; F_melt, F_into_snow, G_under_snow) = partitioned_fluxes
+        G = @. F_sfc * (1 - snow_cover_fraction) +
+           G_under_snow * snow_cover_fraction # Equation 22
         # Temperature profile of soil.
         gradc2f = ClimaCore.Operators.GradientC2F()
         divf2c = ClimaCore.Operators.DivergenceF2C(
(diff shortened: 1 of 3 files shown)


src/standalone/Bucket/bucket_parameterizations.jl
```diff
@@ -55,20 +55,26 @@ a helper function which computes and returns the surface evaporative scaling
  factor for the bucket model.
 """
 function ClimaLand.surface_evaporative_scaling(model::BucketModel, Y, p)
-    beta = beta_factor.(Y.bucket.W, Y.bucket.σS, model.parameters.W_f)
+    beta =
+        beta_factor.(
+            Y.bucket.W,
+            Y.bucket.σS,
+            model.parameters.W_f,
+            model.parameters.σS_c,
```
LenkaNovak, line 63: Just checking, do all of the params (incl. S_c and p) come from the SLIM model?

src/standalone/Bucket/bucket_parameterizations.jl
```diff
@@ -55,20 +55,26 @@ a helper function which computes and returns the surface evaporative scaling
...
 
-Computes the beta factor which scales the evaporation from the potential
+Computes the beta factor which scales the evaporation/sublimation from the potential
 rate.
 """
-function beta_factor(W::FT, σS::FT, W_f::FT) where {FT}
+function beta_factor(W::FT, σS::FT, W_f::FT, σS_c::FT) where {FT}
     snow_cover_fraction = heaviside(σS)
-    return (1 - snow_cover_fraction) * β(W, W_f) + snow_cover_fraction * 1
+    return (1 - snow_cover_fraction) * β(W, FT(0.75 * W_f)) +
```
LenkaNovak, line 76: Do we want to make 0.75 and 0.1 tunable parameters? It would be good to comment on these choices in the doc string.

src/standalone/Bucket/Bucket.jl
```diff
@@ -526,7 +526,7 @@ function next_albedo(
 ) where {FT}
     (; α_snow, α_bareground) = model_albedo
     (; σS_c) = parameters
-    σS = Y.bucket.σS
+    σS = max.(Y.bucket.σS, FT(0))# if σS is small and negative, clip to zero for albedo estimation
```
LenkaNovak, line 529: Should this be eps(FT), so it's consistent with the heaviside function? (also below)

test/standalone/Bucket/snow_bucket_tests.jl
```diff
@@ -54,6 +54,107 @@ for FT in (Float32, Float64)
...
+                ),
+            ) .== FT(0.0),
+        )
+        exp_tendency!(dY, Y, p, t0)
+        # Test that all water tendencies are zero (no precip, no snowmelt, no vapor loss)
+        @test dY.bucket.W == ClimaCore.Fields.zeros(surface_space)
+        @test dY.bucket.Ws == ClimaCore.Fields.zeros(surface_space)
+        @test dY.bucket.σS == ClimaCore.Fields.zeros(surface_space)
+    end
+
```
LenkaNovak, line 157: may it be worth adding a test for the `partition_surface_fluxes` as well, just to ensure all fluxes are the expected signs?

LenkaNovak (commented):
LGTM, thank you, @kmdeck ! I only had a couple of minor comments.

src/standalone/Bucket/bucket_parameterizations.jl
kmdeck: SLIM uses the linear model we were using before for beta (beta = W/(0.75*Wf) for W < 0.75Wf, 1 otherwise). adding in the nonlinearity is our modification. They use W_f of 200 mm or 0.2m, sigmaS_c of 50 mm (0.05m). There is no separate beta for snow. They write:
```
Note that if the latent heat flux term attempts to evaporate (or sublimate)
139 more water than is available in the combined water and snow buckets, the latent heat flux term is
140 set to the total water plus snow available ... and the excess energy that would have
141 been used by LH if more water were available is instead partitioned to SH.
```

Other parameters:
volumetric heat capacity of soil is 2e6 J/m^3/K
soil K is 1.5 W/m/K
soil depth of 3.5m.

I think the main differences are:
- SLIM also uses an additional resistance to evaporation,
- SLIM modifies the MO fluxes for vegetation (vegetation height affects roughness lengths). 
- SLIM has some representation of glaciers

src/standalone/Bucket/bucket_parameterizations.jl
kmdeck: We can make these tunable params - and `p` also?

src/standalone/Bucket/Bucket.jl
kmdeck: Ill change this to be consistent.

test/standalone/Bucket/snow_bucket_tests.jl
kmdeck: added some in!

src/standalone/Bucket/Bucket.jl
LenkaNovak: I meant more for the case of `0 < σS < eps(FT)`, whether it should be treated as 0 or nonzero. It also appears in `partition_surface_fluxes()`. That said, it looks like the heaviside is only used for `snow_cover_fraction` which is then only used to scale evaporation, so I think that's ok, as long as the user doesn't get confused how to treat small values of snow. Maybe a comment would do. :)

LenkaNovak: But in that case, should the comment on Line 457 (`# Equation (3) of the text, snow melt already multipied by snow_cover_fraction`) say `max(0, σS)` 🤔

LenkaNovak (approved):
LGTM, @kmdeck , thank you! I only had one more clarification comment.

src/standalone/Bucket/bucket_parameterizations.jl
LenkaNovak: Sounds great, thanks!

src/standalone/Bucket/Bucket.jl
kmdeck: ah, I think I understand. You're right that we are being inconsistent in the `0 < σS < eps(FT)` regime. In principle, the snow_cover_fraction function can be anything (e.g. sigmoid function), and clipping negative values of σS to zero should still be a valid thing to do (they shouldnt depend on each other). I think they do because some of parameterizations depend on σS (e.g. the albedo, the beta function) and some depend on snow cover fraction. `partition_surface_fluxes` uses both in the same function.

Here is what I did to try and fix this:
- `partition_surface_fluxes` now works in units of emitting/absorbing area (snow). This is consistent with how we handle fluxes in the documentation and everywhere else - they are fluxes per emitting area and we multiply by emitting area fraction (1-σ) or σ to get fluxes per unit ground area when used in setting tendencies. σ is now applied outside the partition_surface_fluxes now, in Bucket.jl lines 457, 468, 474, and 437.
- The albedo model (linear interpolation between snow and bareground albedo) fundamentally conflicts with a heaviside function for snow cover fraction, so I left this alone for now. The inconsistency remains but it is O(eps) and doesnt affect any conservation. I left a note in the documentation.
- The other inconsistency arises in the beta function, but this function is independent of the snow cover fraction and implemented for numerical reasons only, so I dont know that they need to be consistent. if 0 < σS < eps(FT), this function wont be invoked for snow because we will use the beta of soil instead.

If we want to use the bucket model as a stepping stone for the full land model for testing different fractional coverages within a grid cell, we can make this consistent and use the SCF for the albedo function, at least.

LenkaNovak: That seems reasonable to me. Thank you for digging into this, Kat!

PR #501, opened 2024-02-14
Update dependencies

nefrathenrici:
Peel off from #481 

A few notes
- Thermodynamics, SurfaceFluxes, ClimaParameters, ClimaCore and Insolation all have updated compat
- Thermodynamics had breaking changes, and we need to ensure FT-correctness in a few places
- I have removed a few repeated compat entries from `Project.toml`s that dev ClimaLand. For example, since ClimaLand already enforces Insolation = 0.9, the experiments Project does not need its own compat for Insolation.


Project.toml
```diff
@@ -38,11 +38,11 @@ CreateParametersExt = "CLIMAParameters"
 Adapt = "3, 4"
 ArtifactWrappers = "0.2"
 CFTime = "0.1"
-CLIMAParameters = "0.8.6"
+CLIMAParameters = "0.9"
```
juliasloan25, line 41: could we keep the previous compat (e.g. `CLIMAParameters = "0.8.6, 0.9"`), or do these updates not have backwards compatibility? (for all the packages, not just climaparameters)

lib/ClimaLandSimulations/Project.toml
```diff
@@ -37,17 +37,15 @@ cuDNN = "02a925ec-e4fe-4b08-9a7e-0d78e3d38ccd"
 
 [compat]
 ArtifactWrappers = "0.2"
-CLIMAParameters = "0.8"
 CairoMakie = "0.11"
 ClimaComms = "0.5.6"
-ClimaCore = "0.11.7"
+ClimaCore = "0.12"
```
juliasloan25, line 42: maybe we can leave this out since it's in the ClimaLand Project.toml?

juliasloan25 (approved):
Looks good! I just had a couple comments

Project.toml
nefrathenrici: Older versions of CLIMAParameters are compatible for now, but #481 will require v0.9 or greater for the new parameters. I could change the compat again in in that PR, I'm not sure what's preferable.

The other packages all require CP v0.9. We could keep their older versions in the compat, but I think the restrictions with climaparameters will force the package manager to grab the newer versions.

juliasloan25: Okay! Changing it to just the latest versions in this PR is fine

PR #504, opened 2024-02-17
Refactor `SoilCO2ModelParameters` constructor

nefrathenrici:
This PR uses ClimaParams for `SoilCO2ModelParameters`.

As before, the new constructors are:
```julia
import ClimaLand
import ClimaParams as CP
toml_dict= CP.create_toml_dict(Float32)

ClimaLand.Soil.Biogeochemistry.SoilCO2ModelParameters(Float64)
ClimaLand.Soil.Biogeochemistry.SoilCO2ModelParameters(Float64; ν = 99999)
ClimaLand.Soil.Biogeochemistry.SoilCO2ModelParameters(toml_dict)
ClimaLand.Soil.Biogeochemistry.SoilCO2ModelParameters(toml_dict; ν = 99999)
```

### DIFF
--- a/docs/src/standalone/apps/hetero_resp.jl
+++ b/docs/src/standalone/apps/hetero_resp.jl
@@ -21,19 +21,7 @@ function ParamViz.parameterisation(
     D_oa,
 ) # constants
 
-    params = SoilCO2ModelParameters{FT}(;
-        ν = ν,
-        θ_a100 = θ_a100,
-        D_liq = D_liq,
-        α_sx = α_sx,
-        Ea_sx = Ea_sx,
-        kM_sx = kM_sx,
-        kM_o2 = kM_O₂,
-        O2_a = O₂_a,
-        D_oa = D_oa,
-        p_sx = p_sx,
-        earth_param_set = earth_param_set,
-    )
+    params = SoilCO2ModelParameters(FT; ν)
     # θ has to be lower than porosity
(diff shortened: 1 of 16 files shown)


AlexisRenchon:
Our soilCO2 model is a combination of these two papers: 
https://onlinelibrary.wiley.com/doi/abs/10.1111/j.1365-2486.2011.02546.x
for the co2 production model 
and https://gmd.copernicus.org/articles/11/1909/2018/ for the diffusion model 
they provide example of values for the parameters, we could pick any for the default... note that they are not in SI and need to be converted. 
Happy to help more!

nefrathenrici:
Thank you @AlexisRenchon !

AlexisRenchon (approved):
Hi Nat, 
The PR looks good... is there anything in particular I should look at?

nefrathenrici:
> Hi Nat, The PR looks good... is there anything in particular I should look at?

Are these [defaults](https://github.com/CliMA/ClimaLand.jl/pull/504/files#diff-7523e89d54ef402f8bf7a06b83cd224a114d9109723b362d3bf68886bb99c147R470-R473) correct?

AlexisRenchon:
Yes the defaults are correct! 
Maybe for porosity we could just say 0.5, if we do need a default (porosity of the soil is going to be different everywhere so it's a bit weird to have a very specific default)

PR #505, opened 2024-02-21
Refactor EnergyHydrologyParameters

nefrathenrici:
Same as #481 and #480, adds two new constructors for EnergyHydrologyParameters that use ClimaParams. One takes an FT, one takes a toml.
I also updated the experiment Manifest, for some reason Insolation was not updated.
```julia
  EnergyHydrologyParameters(
        ::Type{FT};
        ν,
        ν_ss_om,
        ν_ss_quartz,
        ν_ss_gravel,
        hydrology_cm,
        K_sat,
        S_s,
        θ_r,
        kwargs...,
)

    EnergyHydrologyParameters(
        toml_dict;
        ν,
        ν_ss_om,
        ν_ss_quartz,
        ν_ss_gravel,
        hydrology_cm,
        K_sat,
        S_s,
        θ_r,
        kwargs...,
)
```

### DIFF
--- a/docs/tutorials/standalone/Soil/freezing_front.jl
+++ b/docs/tutorials/standalone/Soil/freezing_front.jl
@@ -88,11 +88,10 @@ using ClimaLand.Soil
 import ClimaLand
 import ClimaLand.Parameters as LP
 
-# # Preliminary set-up
+# Preliminary set-up
 
 # Choose a floating point precision, and get the parameter set, which holds constants used across CliMA models:
 FT = Float32
-earth_param_set = LP.LandParameters(FT);
 
 # Set the values of other parameters required by the model:
 ν = FT(0.535)
@@ -100,7 +99,7 @@ K_sat = FT(3.2e-6) # m/s
 S_s = FT(1e-3) #inverse meters
 vg_n = FT(1.48)
 vg_α = FT(1.11) # inverse meters
-hcm = vanGenuchten{FT}(; α = vg_α, n = vg_n);
+hydrology_cm = vanGenuchten{FT}(; α = vg_α, n = vg_n);
 # You could also try the Brooks and Corey model:
 #ψb = FT(-0.6)
 #c = FT(0.43)
@@ -109,16 +108,16 @@ hcm = vanGenuchten{FT}(; α = vg_α, n = vg_n);
 ν_ss_om = FT(0.3)
 ν_ss_quartz = FT(0.7)
 ν_ss_gravel = FT(0.0)
-params = Soil.EnergyHydrologyParameters{FT}(;
-    ν = ν,
-    ν_ss_om = ν_ss_om,
-    ν_ss_quartz = ν_ss_quartz,
-    ν_ss_gravel = ν_ss_gravel,
-    hydrology_cm = hcm,
-    K_sat = K_sat,
-    S_s = S_s,
-    θ_r = θ_r,
-    earth_param_set = earth_param_set,
+params = Soil.EnergyHydrologyParameters(
+    FT;
+    ν,
+    ν_ss_om,
+    ν_ss_quartz,
+    ν_ss_gravel,
+    hydrology_cm,
+    K_sat,
+    S_s,
+    θ_r,
 );
 # Choose the domain and discretization:
 zmax = FT(0)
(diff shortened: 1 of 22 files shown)


docs/tutorials/standalone/Soil/freezing_front.jl
```diff
@@ -109,16 +108,16 @@ hcm = vanGenuchten{FT}(; α = vg_α, n = vg_n);
...
-    ν_ss_quartz = ν_ss_quartz,
-    ν_ss_gravel = ν_ss_gravel,
-    hydrology_cm = hcm,
-    K_sat = K_sat,
-    S_s = S_s,
-    θ_r = θ_r,
-    earth_param_set = earth_param_set,
+params = Soil.EnergyHydrologyParameters(
+    FT;
+    ν,
```
juliasloan25, line 113: I wonder if it would be better to continue using the keyword arguments (e.g. passing `\nu = \nu`) so that we don't rely on the positional order of the arguments. I'm okay with either way, just a thought!

docs/tutorials/standalone/Soil/soil_energy_hydrology.jl
```diff
@@ -116,18 +116,18 @@ Ksat = FT(4.42 / 3600 / 100) # m/s
...
+hydrology_cm = vanGenuchten{FT}(; α = vg_α, n = vg_n)
+θ_r = FT(0.0)
+params = Soil.EnergyHydrologyParameters(
+    FT;
+    ν,
+    ν_ss_om,
+    ν_ss_quartz,
+    ν_ss_gravel,
+    hydrology_cm,
     K_sat = Ksat,
```
juliasloan25, line 128: only K_sat is set with `=` here - is that intentional?

ext/CreateParametersExt.jl
```diff
@@ -220,4 +225,154 @@ function OptimalityFarquharParameters(toml_dict; kwargs...)
...
+    ν_ss_quartz::F,
+    ν_ss_gravel::F,
+    hydrology_cm::C,
+    K_sat::F,
+    S_s::F,
+    θ_r::F,
+    PAR_albedo = 0.2,
+    NIR_albedo = 0.4,
+    kwargs...,
+) where {F <: Union{<:AbstractFloat, ClimaCore.Fields.FieldVector}, C}
```
juliasloan25, line 297: is it correct that these will be either floats or fieldvectors? I'm not as familiar with the parameters, but I would have expected floats or fields

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-NR1/US-NR1_parameters.jl
```diff
@@ -87,21 +87,21 @@ function soil_niwotridge(;
...
-        S_s = S_s,
-        θ_r = θ_r,
-        PAR_albedo = PAR_albedo,
-        NIR_albedo = NIR_albedo,
-        emissivity = emissivity,
-        z_0m = z_0m,
-        z_0b = z_0b,
-        earth_param_set = earth_param_set,
+    return EnergyHydrologyParameters(
+        FT; # here, calls the src function, not the src struct
```
juliasloan25, line 91: I think we can get rid of this comment

juliasloan25 (approved):
Looks good! I left a few comments but I also think it's good to merge as-is

docs/tutorials/standalone/Soil/freezing_front.jl
nefrathenrici: I think that the semicolon after the FT is enough to ensure that they are read as kwargs if the variable name matches the keyword:

```julia
f(a;b = 2) = a*b
f(2;b = 3)
b = 4
f(2; b)
```

juliasloan25: ahh I didn't realize! nice

ext/CreateParametersExt.jl
nefrathenrici: I'm not sure, this is just copied over from the old EnergyHydrologyParameters [constructor](https://github.com/CliMA/ClimaLand.jl/blob/main/src/standalone/Soil/energy_hydrology.jl#L84).

PR #511, opened 2024-02-23
Topmodel runoff

kmdeck:
## Purpose 
Add the option to use the TOPMODEL runoff parameterization for surface and subsurface runoff.

## Content
1. Adds in TOPMODEL runoff parameterization. This requires adding both a new method for computing the infiltration into the soil, while accounting for surface runoff, as well as a new source term to account for subsurface runoff. 
2. Adds this option for both Richards and Energy Hydrology model
3. Adds a global simulation of Richards model with TOPMODEL runoff + PrescribedPrecipitation to buildkite
Model equations are taken from: https://agupubs.onlinelibrary.wiley.com/doi/pdfdirect/10.1029/2005JD006111

Review checklist

I have:

----


src/shared_utilities/drivers.jl
```diff
@@ -656,6 +672,21 @@ function initialize_drivers(a::PrescribedAtmosphere{FT}, coords) where {FT}
     return vars.drivers
 end
 
+"""
+"""
```
kmdeck, line 676: doc string

src/standalone/Soil/Runoff/Runoff.jl
```diff
@@ -0,0 +1,143 @@
...
+"""
+    subsurface_runoff_source(runoff::AbstractRunoffModel)::Union{Nothing, AbstractSoilSource} 
+A function which returns the soil source for the runoff model 
+`runoff`; the default returns nothing in which case no source is added.
+"""
+subsurface_runoff_source(
+    runoff::AbstractRunoffModel,
+)::Union{Nothing, AbstractSoilSource} = nothing
+
+# TOPMODEL
```
kmdeck, line 58: docstrings, also put equations

src/shared_utilities/drivers.jl
```diff
@@ -98,6 +99,19 @@ struct PrescribedAtmosphere{
     end
 end
 
+"""
+    PrescribedPreciptiation{FT, LP} <: AbstractAtmosphericDrivers{FT}
```
kmdeck, line 103: typo

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -213,7 +211,7 @@ struct FreeDrainage <: AbstractSoilBC end
 
 
 """
-   RichardsAtmosDrivenFluxBC{F <: Function, R <: AbstractRunoffModel} <: AbstractSoilBC
+   RichardsAtmosDrivenFluxBC{F <: AbstractTimeVaryingInput, R <: AbstractRunoffModel} <: AbstractSoilBC
```
kmdeck, line 214: update this

src/standalone/Soil/Soil.jl
```diff
@@ -157,10 +157,23 @@ function horizontal_components!(
     _...,
 ) end
 
+"""
```
kmdeck, line 160: These were previously defined in runoff.jl, but are general enough to move here, I thought. We allow users to pass in a tuple of source terms, but in some cases, some boundary conditions require source terms to be added. Using runoff is one example (it affects the BC and the source terms), sublimation (source) + evaporation(boundary flux) is another example. so this is used in the model constructors to be consistent. if the user chooses to use runoff, this will add the correct source term.

https://github.com/CliMA/ClimaLand.jl/blob/1be8238de46bdd233ee838fcdec2e3d35f933680/src/standalone/Soil/runoff.jl#L55

src/standalone/Soil/Runoff/Runoff.jl
```diff
@@ -0,0 +1,257 @@
...
+subsurface_runoff_source(runoff::AbstractRunoffModel) = runoff.subsurface_source
+
+
+"""
+    NoRunoff <: AbstractRunoffModel
+A concrete type of soil runoff model; the 
+default choice, which does not include any
+runoff.
+"""
+struct NoRunoff <: AbstractRunoffModel
```
kmdeck, line 55: previously defined here https://github.com/CliMA/ClimaLand.jl/blob/1be8238de46bdd233ee838fcdec2e3d35f933680/src/standalone/Soil/runoff.jl#L29

experiments/standalone/Soil/richards_runoff.jl
```diff
@@ -0,0 +1,246 @@
...
+        S = FT((FT(1) + (α * (z - z_∇))^n)^(-m))
+        ϑ_l = S * (ν - θ_r) + θ_r
+    else
+        ϑ_l = -S_s * (z - z_∇) + ν
+    end
+    return FT(ϑ_l)
+end
+t0 = 0.0
+tf = 360.0 * 10
+dt = 360.0 * 2
```
AlexisRenchon, line 136: dt = 12 minutes
tf-t0 = 2 hours?

experiments/standalone/Soil/richards_runoff.jl
```diff
@@ -0,0 +1,246 @@
...
+    saveval = Array{NamedTuple}(undef, length(saveat)),
+)
+saving_cb = ClimaLand.NonInterpSavingCallback(sv, saveat)
+updateat = deepcopy(saveat)
+updatefunc = ClimaLand.make_update_drivers(atmos, nothing)
+driver_cb = ClimaLand.DriverUpdateCallback(updateat, updatefunc)
+cb = SciMLBase.CallbackSet(driver_cb, saving_cb)
+@time sol = SciMLBase.solve(prob, ode_algo; dt = dt, saveat = dt, callback = cb)
+
+longpts = range(-180.0, 180.0, 101)
```
AlexisRenchon, line 183: That's ~ 3 degree spatial resolution? (slightly different for lon and lat)

AlexisRenchon (commented):
I have looked through the code, it looks good to me!
I see you're using CairoMakie =) 
I see the even the CPU experiment fails on buildkite, it looks like a netcdf file reading error? Does it work locally?
It is the first time I look at a global run code with ClimaLand, once this works I can add some functionality in ClimaLandSimulations

experiments/standalone/Soil/richards_runoff.jl
kmdeck: yes! partly because create the TOPMODEL parameter map at higher resolution is very slow. but eventually I think we should make a map at higher resolution (1/2 degree?) and have that be our benchmark run, instead of 3 degrees.

experiments/standalone/Soil/richards_runoff.jl
kmdeck: Ill make it a little longer. we have some issues with speed :( as you have noticed too

AlexisRenchon (approved):

PR #525, opened 2024-02-27
Add logos

Sbozzolo:
The got lost in translation.


kmdeck (approved):
looks good. One very minor thing is that part of the dot on top of the `i` in Clima is not covered by the leaf in the logo.svg file. this may not be visible when it is rendered on the page.

kmdeck:
Note that we need to pin to Insolation 0.9.1,
in the main Project.toml
`Insolation = "=0.9.1"`

and then rebuild the Manifests (I think) for docs and experiments.

AlexisRenchon (approved):

Sbozzolo:
> looks good. One very minor thing is that part of the dot on top of the `i` in Clima is not covered by the leaf in the logo.svg file. this may not be visible when it is rendered on the page.

`logo.svg` renders differently on different computers, but it should look like logo.png

kmdeck:
> > looks good. One very minor thing is that part of the dot on top of the `i` in Clima is not covered by the leaf in the logo.svg file. this may not be visible when it is rendered on the page.
> 
> `logo.svg` renders differently on different computers, but it should look like logo.png

ok! all good then.

I have the smallest preference to make the blue leaf a teardrop of water instead of a leaf, but definitely not a priority nor does that need to happen without discussing with the team first/be part of this PR

Sbozzolo:
logo_drop

This is what is looks like with the teardrop. I like it. :)

PR #534, opened 2024-02-29
Cleanup richards equation BC specification

kmdeck:
## Purpose 
Simplifies RichardsModel BC specification from a nested named tuple `(top = (;water = BC,), bottom = (;water = BC,))` to a simpler named tuple `(top = = BC, bottom = BC,)`, where `typeof(BC) <: AbstractSoilBC`.

This is cleaner IMO and it proved useful when working with the `RichardsAtmosDrivenFluxBC` (reduced the number of methods I needed to define)

## Content
- Change type of `p.soil.top_bc` and `p.soil.bottom_bc` from a NamedTuple to a scalar
- propagate
- I also made a small change to `RichardsAtmosFluxBC` to have it take the new PrescribedPrecipitation type instead of a function (uses the TimeVaryingInput infrastructure, updates via callbacks like all of our drivers).

Review checklist

I have:

----

### DIFF
--- a/docs/tutorials/standalone/Soil/layered_soil.jl
+++ b/docs/tutorials/standalone/Soil/layered_soil.jl
@@ -95,7 +95,7 @@ function top_flux_function(p, t)
 end
 top_bc = ClimaLand.Soil.FluxBC(top_flux_function)
 bottom_bc = ClimaLand.Soil.FreeDrainage()
-boundary_fluxes = (; top = (water = top_bc,), bottom = (water = bottom_bc,))
+boundary_fluxes = (; top = top_bc, bottom = bottom_bc)
 soil = Soil.RichardsModel{FT}(;
     parameters = params,
     domain = soil_domain,
--- a/docs/tutorials/standalone/Soil/richards_equation.jl
+++ b/docs/tutorials/standalone/Soil/richards_equation.jl
@@ -114,8 +114,7 @@ soil_domain = Column(; zlim = (zmin, zmax), nelements = nelems);
 
 surface_flux = Soil.FluxBC((p, t) -> 0.0)
 bottom_flux = Soil.FluxBC((p, t) -> 0.0)
-boundary_conditions =
-    (; top = (water = surface_flux,), bottom = (water = bottom_flux,));
+boundary_conditions = (; top = surface_flux, bottom = bottom_flux);
 
 # Lastly, in this case we don't have any sources, so we pass an empty tuple:
 sources = ();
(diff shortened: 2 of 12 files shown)


juliasloan25:
I don't think it's great that we're dispatching off of `Union{NamedTuple, AbstractSoilBC}` in a lot of places - I think it's a little confusing that `RichardsModel` and `EnergyHydrology` use different types for their BCs. What if we instead extended `AbstractSoilBC` to have subtypes `RichardsBC`, `EnergyHydrologyBC` which could contain the fields they each need. Then we can dispatch off of these types instead of `Union{NamedTuple, AbstractSoilBC}`

kmdeck:
> I don't think it's great that we're dispatching off of `Union{NamedTuple, AbstractSoilBC}` in a lot of places - I think it's a little confusing that `RichardsModel` and `EnergyHydrology` use different types for their BCs. What if we instead extended `AbstractSoilBC` to have subtypes `RichardsBC`, `EnergyHydrologyBC` which could contain the fields they each need. Then we can dispatch off of these types instead of `Union{NamedTuple, AbstractSoilBC}`

Hmm that is more or less the way we have it currently. both have boundary condition types that are NamedTuples, but the richards case only has the `water` key while the energy/hydrology model has both `water` and `heat`.  The types of the values are `AbstractSoilBC`. However, it's not as clean as it could be because we also support the `AtmosDriven` bc types, which are also subtypes of `AbstractSoilBC`, but in this case, the user doesnt supply a named tuple, but the BC directly. (it has to be this way, for EH, because it specifies BOTH heat and water BC).

We can make it cleaner maybe by doing the following:
`AbstractRichardsModelBC <: AbstractSoilBC` and `AbstractEHModelBC<:AbstractSoilBC`, with:
(no subfield of `water`)
`RichardsAtmosDrivenFluxBC <: AbstractRichardsModelBC`
`MoistureBC <: AbstractRichardsModelBC`
`FreeDrainage <: AbstractRichardsModelBC`
`WaterFluxBC <: AbstractRichardsModelBC`

(no subfield of `water` or `heat`)
`AtmosDrivenFluxBC <: AbstractEHModelBC`
(more general case has those subfields)
```
struct SoilEnergyHydrologyBC <: AbstractEHModelBC
    water::Union{MoistureBC, FreeDrainage, WaterFluxBC}
    heat::Union{TemperatureBC, HeatFluxBC}
end
```
and I guess also define an AbstractHeatEquationBC type with `TemperatureBC` and `HeatFluxBC` as subtypes...

Im not sure. this doesnt feel like a high priority unless it is a blocker for the TOPMODEL PR. I will try and work around our existing BC infrastructure for now.

kmdeck:
closing unless needed

PR #535, opened 2024-02-29
Refactor boundary condition types

kmdeck:
## Purpose 
Refactor boundary condition types.

What we had in place was a little confusing (you could either pass a NamedTuple with values of type "AbstractSoilBC" or something of type "AbstractSoilBC" as your boundary condition), had very little unification between how we handled BC between soil +soil co2, and had nested named tuples for models with only a single boundary condition needed.

I was also finding it very hard to use the RichardsAtmosDrivenBC in PR #511 due to the design.

After this PR, we would have:
Instead of:
```
boundary_fluxes = (top = RichardsAtmosDrivenFluxBC(), bottom = (water = WaterFluxBC(),))
```
you now have 
```
boundary_fluxes = (top = RichardsAtmosDrivenFluxBC(), bottom = WaterFluxBC()).
```
instead of
```
boundary_fluxes = (top = (water = MoistureBC(), bottom = (water = WaterFluxBC(),))
```
you now have
```
boundary_fluxes = (top = MoistureBC(), bottom = WaterFluxBC())
```
instead of :
```
boundary_fluxes = (top = (water = MoistureBC(), heat = TemperatureBC()), bottom = (water = WaterFluxBC(), heat = HeatFluxBC(),))
```
you now have:
```
boundary_fluxes = (top = WaterHeatBC(; water = MoistureBC(), heat = TemperatureBC()),
                   bottom = WaterHeatBC(; water = WaterFluxBC(), heat = HeatFluxBC()))
```
instead of
```
boundary_fluxes = (top = AtmosDrivenFluxBC(), bottom = (water = WaterFluxBC(), heat = HeatFluxBC(),)) 
```

now, we would have
```
boundary_fluxes = (top = AtmosDrivenFluxBC(), bottom = WaterHeatBC(; water = WaterFluxBC(), heat = HeatFluxBC())) 
```
## To-do
review

## Content
This PR refactors the code so that 
1. You can make types of `AbstractWaterBC` which can be used for RichardsModel or for the water equation in the EnergyHydrologyModel. All of our old BC types are now subtyped from this.
2. You can also define types of `AbstractHeatBC` for the soil heat equation. All of our old BC types are now subtyped from this.
3. To create BC for the EnergyHydrologyModel, you can combined an AbstractHeat and AbstractWater BC if these are treated independently, or use a BC that indicates that the water and heat BC are set in way that depend on each other. The former is a `WaterHeatBC <: AbstractEnergyHydrologyBC`, the second has one example currently: `AtmosDrivenFluxBC <: AbstractEnergyHydrologyBC`. (change: introduce WaterHeatBC struct instead of using a NamedTuple, introduce abstract type)
4. The SoilCO2 model follows the same conventions as RichardsModel (single equation, single BC).
5. Since boundary condition types may require additional variables, all PDE models now use and extend the `boundary_var` methods instead of this being done haphazardly/only for soil.
6. I removed the nested named tuples with redundant information for soil co2 and richards equation. This means we specify BC like `top = WaterFluxBC()` and not `top = (;water = WaterFluxBC()))`
7. `soil_boundary_fluxes` now updates in place (easy change to make but not directly related to this PR)

Review checklist

I have:

----

### DIFF
--- a/docs/src/APIs/Soil.md
+++ b/docs/src/APIs/Soil.md
@@ -72,16 +72,15 @@ ClimaLand.Soil.soil_surface_infiltration
 ## Soil BC Methods and Types
 
 ```@docs
-ClimaLand.Soil.AbstractSoilBC
 ClimaLand.Soil.MoistureStateBC
-ClimaLand.Soil.FluxBC
+ClimaLand.Soil.HeatFluxBC
+ClimaLand.Soil.WaterFluxBC
 ClimaLand.Soil.TemperatureStateBC
 ClimaLand.Soil.FreeDrainage
 ClimaLand.Soil.RichardsAtmosDrivenFluxBC
 ClimaLand.Soil.AtmosDrivenFluxBC
-ClimaLand.Soil.boundary_vars
-ClimaLand.Soil.boundary_var_domain_names
-ClimaLand.Soil.boundary_var_types
+ClimaLand.Soil.WaterHeatBC
+ClimaLand.Soil.soil_boundary_fluxes!
 ```
 
 ## Soil Source Types
--- a/src/shared_utilities/boundary_conditions.jl
+++ b/src/shared_utilities/boundary_conditions.jl
@@ -4,8 +4,16 @@ export boundary_flux,
     TopBoundary,
     BottomBoundary,
     diffusive_flux,
-    get_Δz
+    get_Δz,
+    boundary_var_domain_names,
+    boundary_var_types,
+    boundary_vars
 
+
+### These methods and types are not required for all models. They
+### are only useful for models using ClimaCore's differential operators
+### for solving PDEs. Your model may not need them.
+## Currently, only the soil and soil CO2 models use these.
 """
     AbstractBC
 
@@ -13,6 +21,7 @@ An abstract type for types of boundary conditions, which will include
 prescribed functions of space and time as Dirichlet conditions or
 Neumann conditions, in addition to other 
 convenient conditions.
+
 """
 abstract type AbstractBC end
 
@@ -30,7 +39,7 @@ abstract type AbstractBoundary end
     TopBoundary{} <: AbstractBoundary{}
 
 A simple object which should be passed into a function to
-indicate that we are considering the top boundary of the soil.
+indicate that we are considering the top boundary.
 """
 struct TopBoundary <: AbstractBoundary end
 
@@ -38,9 +47,13 @@ struct TopBoundary <: AbstractBoundary end
     BottomBoundary{} <: AbstractBoundary{}
 
 A simple object which should be passed into a function to
-indicate that we are considering the bottom boundary of the soil.
+indicate that we are considering the bottom boundary.
 """
 struct BottomBoundary <: AbstractBoundary end
+
+
+bc_name(::BottomBoundary) = :bottom_bc
+bc_name(::TopBoundary) = :top_bc
 
 """
     get_Δz(z::ClimaCore.Fields.Field)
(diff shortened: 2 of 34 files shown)


kmdeck:
@juliasloan25 investigating cleaning up BC here....will tag for a review if it pans out!

src/shared_utilities/boundary_conditions.jl
```diff
@@ -84,3 +87,69 @@ function boundary_flux(
...
+
+"""
+    boundary_vars(::AbstractBC , ::ClimaLand.TopBoundary)
+
+The list of symbols for additional variables to add to the
+model auxiliary state, which defaults to adding storage for the
+ top boundary flux fields,
+but which can be extended depending on the type of boundary condition used.
+
+Note that `:top_bc` must be present, with the default type and domain name,
```
kmdeck, line 100: I'll note that this is not required for all models and currently used by soil & soil co2

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -77,143 +82,79 @@ function get_bottom_surface_field(center_val, _)
 end
 
 """
-    AbstractSoilBC <: ClimaLand. AbstractBC
+    AbstractRichardsBC <: ClimaLand.AbstractBC
 
-An abstract type for soil-specific types of boundary conditions, like free drainage.
+An abstract type for boundary conditions for the RichardsModel.
```
kmdeck, line 87: should say for "Richards equation"

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -77,143 +82,79 @@ function get_bottom_surface_field(center_val, _)
...
-A simple concrete type of boundary condition, which enforces a
-state boundary condition T = f(p,t) at either the top or bottom of the domain.
-"""
-struct TemperatureStateBC{F <: Function} <: AbstractSoilBC
-    bc::F
-end
-
-"""
-   FluxBC <: AbstractSoilBC
+   WaterFluxBC <: AbstractSoilModelBC
```
kmdeck, line 137: update AbstractRichardsBC

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -245,12 +191,45 @@ end
...
+   HeatFluxBC <: AbstractHeatBC
+
+A simple concrete type of boundary condition, which enforces a
+normal flux boundary condition f(p,t) at either the top or bottom of the domain.
+"""
+struct HeatFluxBC{F <: Function} <: AbstractHeatBC
+    bc::F
+end
+
+struct EnergyHydrologyBC{W <: AbstractRichardsBC, H <: AbstractHeatBC} <:
```
kmdeck, line 216: doc string

src/standalone/Soil/rre.jl
```diff
@@ -480,3 +480,23 @@ function ClimaLand.make_update_jacobian(model::RichardsModel{FT}) where {FT}
...
+
+"""
+    ClimaLand.get_drivers(model::RichardsModel)
+
+Returns the driver variable symbols for the RichardsModel; these
+depend on the boundary condition type and currently only are required
+for the RichardsAtmosDrivenFluxBC, which is driven by
+a prescribed time and space varying precipitation.
+"""
+function ClimaLand.get_drivers(model::RichardsModel)
```
juliasloan25, line 492: we could also pass in `model.boundary_conditions.top` as an argument and dispatch on its type to avoid the if/else branching in this function

test/integrated/soil_energy_hydrology_biogeochemistry.jl
```diff
@@ -1,5 +1,6 @@
 using Test
 using ClimaCore
+import ClimaParams
```
juliasloan25, line 3: do we need this?

docs/src/APIs/Soil.md
```diff
@@ -72,16 +72,14 @@ ClimaLand.Soil.soil_surface_infiltration
 ## Soil BC Methods and Types
 
 ```@docs
-ClimaLand.Soil.AbstractSoilBC
 ClimaLand.Soil.MoistureStateBC
-ClimaLand.Soil.FluxBC
+ClimaLand.Soil.HeatFluxBC
+ClimaLand.Soil.WaterFluxBC
```
juliasloan25, line 77: we should add `WaterHeatBC` too

juliasloan25 (approved):
This looks great, Kat! I think this interface makes a lot more sense. It might be good to explain the BC interface somewhere in the tutorials (maybe `richards_equation.jl` or `soil_energy_hydrology.jl`?) - just briefly explaining that we expect a NamedTuple of (; top, bottom), where each field is an instance of a BC type.

test/integrated/soil_energy_hydrology_biogeochemistry.jl
kmdeck: I think we somehow have some leakages in our tests with the environments? because this runs in CI without this change, but if I just try and run it locally by itself, it fails because it doesnt have ClimaParams.

docs/src/APIs/Soil.md
kmdeck: good catch!

src/shared_utilities/boundary_conditions.jl
```diff
@@ -84,3 +97,69 @@ function boundary_flux(
...
+    boundary_vars(::AbstractBC, ::ClimaLand.BottomBoundary)
+
+The list of symbols for additional variables to add to the
+model auxiliary state, which defaults to adding storage for the
+ bottom boundary flux fields,
+but which can be extended depending on the type of boundary condition used.
+
+Use this function in the exact same way you would use `auxiliary_vars`.
+
+Note that `:bottom_bc` must be present, with the default type and domain name,
```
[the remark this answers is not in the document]
kmdeck:
```
The list of symbols for additional variables to add to the
model auxiliary state, for models solving PDEs, which defaults 
to adding storage for the bottom boundary flux fields,
but which can be extended depending on the type of 
boundary condition used.

For the Soil and SoilCO2 models - which solve PDEs - the 
tendency functions and 
update_boundary_fluxes functions are coded to access the field `:bottom_bc` 
to be present in the model cache, which is why this is the default.
 If this is not your (PDE) model's 
desired behavior, you can extend this function with a new method.

Use this function in the exact same way you would use `auxiliary_vars`.
```

a bit clearer?

src/standalone/Soil/rre.jl
kmdeck: I think youre right, but different models store the drivers in different places (this perhaps is something we can unify later on?), so for now I've gone with the function signature of`get_drivers(model)` for all models types, and then each method can unpack the location of the drivers as needed.

kmdeck:
I added a tutorial for soil boundary conditions! good call. it'll help us down the line too :D

PR #536, opened 2024-03-01
Update to ClimaCore 0.13

Sbozzolo:
`ClimaCore` 0.13 removes some deprecated functions, this PR updates Land to remove those functions.


docs/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.1"
+julia_version = "1.10.0"
```
juliasloan25, line 3: can you update to julia v1.10.1 so that the manifest doesn't get downgraded?

juliasloan25 (approved):
please just update to julia v1.10.1 for the manifests, then this is good to go

PR #537, opened 2024-03-02
Add depot

Sbozzolo:
Our buildkite pipeline spends 50% of its time in initializing the environment, so adding a depot should appreciably reduce CI time.


juliasloan25:
The initialization for this buildkite run still took as long as the most recent run on main :/ (both 30 mins). We can still add the depot though

Sbozzolo:
Yes, the first time it takes the same amount of time, but subsequently it
should be faster.

Sbozzolo:
Here it took only 3 minutes:

https://buildkite.com/clima/climaland-ci/builds/3511#018e19a4-7fdf-4f85-ac95-a733e3973123

So, I am merging this.

juliasloan25 (approved):

PR #540, opened 2024-03-06
Varying Leaf Angular Distribution

Espeer5:
## Purpose 

Vary the leaf angular distribution according to a leaf orientation index. This is the same parameterization for ld as CLM uses.

## To-do

- [x] Fix Two-Stream model tests to verify output using PySellersTwoStream code

## Content

 - Leaf orientation index added to `TwoStreamParameters` in place of the leaf angular distribution
 - Several formulae changed in TwoStream source code
 - Testing file for TwoStream test swapped to use one set up with leaf orientation index
 - Restore files post-rebase (several errors introduced)

----


Espeer5:
I worked on this PR a lot over break and there are some issues.

Referring to the test run [here](https://github.com/CliMA/ClimaLand.jl/pull/540/files#diff-acea5676c3cb31c2a243d1fef2b0a637c1169db71a7f887fa6d671ee61113ba4), the output using the CLM method of specifying the leaf angular distribution by orientation index matches the output of another repository which can be configured to run the CLM radiative transfer model found [here](https://github.com/tquaife/pySellersTwoStream) to within 1%. This is the same way that we tested the TwoStreamModel before. Notice that our model uses a clumping index while this one does not.

However, despite the agreement of the two models, when our code is run with the new changes and the leaf orientation index that should correspond to a temperate broadleaf deciduous forest on Ozark site, the radiative transfer plots get noticeably worse, for example the SW_OUT plot looks much worse than before:

Screenshot 2024-04-01 at 10 44 49 AM

I've double and triple checked all of the equations and the test correctness seems to suggest the equations are entered correctly, so I am not sure what to do here.

Any suggestions @braghiere?

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -201,23 +201,31 @@ function plant_absorbed_pfd(
...
 
-    (; ld, Ω, n_layers) = RT.parameters
+    (; χl, Ω, n_layers) = RT.parameters
 
+    ϕ1 = 0.5 - 0.633 * χl - 0.33 * χl^2
+    ϕ2 = 0.877 * (1 - 2 * ϕ1)
+    μ = cos(θs)
     # Compute μ̄, the average inverse diffuse optical length per LAI
-    μ̄ = 1 / (2 * ld)
+    μ̄ = 1 / ϕ2 * (1 - (ϕ1 / ϕ2) * log((ϕ1 + ϕ2) / ϕ1))
```
braghiere, line 210: I don't understand this part. I think adding ld(χl) still holds:

μ̄ = 1 / (2 * ld)

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -201,23 +201,31 @@ function plant_absorbed_pfd(
...
+    # Compute the leaf angular distribution based on the leaf orientation index
+    ld = ϕ1 + ϕ2 * μ
 
     ω = α_leaf + τ_leaf
 
+    divisor = isnan(μ * ϕ2 + ld) ? 1e-6 : min(μ * ϕ2 + ld, 1e-6)
+
     # Compute aₛ, the single scattering albedo
-    aₛ = 0.5 * ω * (1 - cos(θs) * log((abs(cos(θs)) + 1) / abs(cos(θs))))
+    aₛ = 0.5 * ω * (ld / divisor) * (1 - (μ * ϕ1 / divisor) * log((μ * ϕ1 + divisor) / (μ * ϕ1)))
```
braghiere, line 220: Also this part. Shouldn't it be 

aₛ = 0.5 * ω * (1 - cos(θs) * log((abs(cos(θs)) + 1) / abs(cos(θs))))?

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -201,23 +201,31 @@ function plant_absorbed_pfd(
...
 
     # Compute β₀, the direct upscattering parameter
     β₀ = (1 / ω) * aₛ * (1 + μ̄ * K) / (μ̄ * K)
 
     # Compute β, the diffuse upscattering parameter
     diff = α_leaf - τ_leaf
-    # With uniform distribution, Dickinson integral becomes following:
-    c²θ̄ = pi * ld / 4
+    # Using the CLM approach with χl
+    c²θ̄ = ((1 + χl) / 2)^2
```
braghiere, line 228: And this: 

c²θ̄ = pi * ld / 4

braghiere (commented):
So, I didn't understand a few equations here, but to me the only change in the new version is that ld = ϕ1 + ϕ2 * μ, instead of ld = 0.5. The ld also becomes a function of the sun zenith angle.

src/standalone/Vegetation/canopy_parameterizations.jl
Espeer5: This is a good point, I should try this way as opposed to changing all the equations. When I implemented this I just changed all of the relevant equations to match section 2.3.1 [here](https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/tech_note/Surface_Albedos/CLM50_Tech_Note_Surface_Albedos.html#canopy-radiative-transfer)

Espeer5:
Closing; Decided on a different approach to solving this problem.

PR #542, opened 2024-03-07
update implicit solver interface

### LINKED ISSUE #528: use new interface for implicit RRE
Use the most recent implicit solver interface, as described in https://github.com/CliMA/ClimaCore.jl/issues/1230.

Atmos PR updating to new interface: https://github.com/CliMA/ClimaAtmos.jl/pull/2044

**Notes on interface changes**
- majority of changes in ClimaCore MatrixFields
- `ldiv!` now in ClimaCore src/Fields.fieldvector.jl
  - can remove from ClimaLand.jl
- `thomas_algorithm` now in ClimaCore src/Operators/thomas_algorithm.jl
  - can remove from ClimaLand.jl
- `linsolve` not moved to ClimaCore ?

**Atmos update**
- struct SchurComplementW -> ImplicitEquationJacobian
- Operators.Operator2Stencil -> MatrixFields.operator_matrix
- Wfact! remains 
- ldiv! remains (I thought this was moved to ClimaCore ?)

**QA**
- verify RRE experiment results don't change

juliasloan25:
## Purpose 
Update implicit solver interface for RichardsModel to use `ClimaCore.MatrixFields`.

Comparing the buildkite timing to the most recent `main` [build](https://github.com/CliMA/ClimaLand.jl/blob/c5426109a299533964a87d28a0ecdcc3842e0b89/experiments/standalone/Soil/richards_runoff.jl#L72), the RichardsModels runs see a ~25% speedup from this PR.

closes #528
https://github.com/CliMA/ClimaLand.jl/pull/602 has to be merged first for update to ClimaCore v0.14 

## To Do
- [x] GPU compatibility - requires https://github.com/CliMA/ClimaCore.jl/pull/1684 and ClimaCore patch release
- [x] compare allocations/performance to previous implementation

## Content
- delete `thomas_algorithm`
handling multiple vars
- the full soil case will contain a variable that isn't stepped implicitly (theta_i), so we can't just loop over everything in the state
  - when extending, will add if statement, so that we use tridiagonal for theta_l and heat, and identity for theta_i
- boundary conditions
- [x] allocate space in cache for `dtopBCdY` (extend `boundary_vars`, `boundary_var_domain_names`, `boundary_var_types`, etc for `MoistureStateBC`)
- [x] update cache var with expression currently in `dtendencyBCdY` (except multiply by dz factor)
- [x] update `dtendencyBCdY`: update in place, rename to `set_dfluxBCdY!`, multiply by dz factor
- [x] update `dtheta_resdtheta` setting: add if/else branch where if default use expression currently in code; if `p.soil.dtopBCdtheta` exists, add `dtopBCdtheta` term before `div` applied (use `Operators.SetBoundaryOperators`, then construct Bidiagonal(?)Matrix of zeros with this BC)

## Status 4/17/24
The implicit solver has been updated and works on CPU. It doesn't work on GPU because of a couple of problems. First, one of the operators produces a non-`isbits` object. This is because the space is not being stripped out of the input object, which should happen via a call to `strip_space`. We need to extend `adapt_structure` for the `SetBoundaryOperator` type (and perhaps other operator types too), but even with that the solver fails on GPU.

It seems like the second bug may be from FT not being inferred correctly when it crosses the function boundary from `make_update_jacobian` to `update_jacobian!`

----


src/standalone/Soil/rre.jl
```diff
@@ -504,9 +478,25 @@ end
...
+"""
+    add_dtopBCdY_to_cache(p::NamedTuple, model::RichardsModel)
+
+Allocate space in `p` for the derivative of the top boundary condition with
+respect to the state variables, and merges it into `p` under the key `dtopBCdY`.
+This is used in the implicit solver for
+`RichardsModel`, where we need to handle the boundary specially in computing
+the Jacobian.
+"""
+function ClimaLand.add_dtopBCdY_to_cache(p::NamedTuple, model::RichardsModel)
```
juliasloan25, line 498: do we always want to add this for RichardsModel, or only if using MoistureStateBC?

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -487,13 +489,14 @@ function ClimaLand.∂tendencyBC∂Y(
         interpc2f_op.(dψdϑ.(hydrology_cm, Y.soil.ϑ_l, ν, θ_r, S_s)),
         face_len,
     )
-    return ClimaCore.Fields.FieldVector(;
-        :soil => (; :ϑ_l => @. -K / Δz * dψ / (2 * Δz)),
-    )
+    # Update `dfluxBCdY` in place
+    p.dfluxBCdY .=
+        ClimaCore.Fields.FieldVector(; soil = (; ϑ_l = @. -K * dψ / Δz))
```
kmdeck, line 494: there may be a sign error here

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -516,7 +519,8 @@ layer cell index.
 If `F_bc` can be approximated as independent of `Y_N`, the derivative
 is zero.
 """
-function ClimaLand.∂tendencyBC∂Y(
+# TODO do we even need to define this method? won't be called if not using MoistureStateBC at TopBoundary
```
kmdeck, line 522: agreed - cant we fall back on https://github.com/CliMA/ClimaLand.jl/pull/542/files#diff-a450f17176137e18c75988ec46e63ededf61ce96a207f9d816a7777cf8432d2bL60?

kmdeck: Also, we can fall back on the default and not use the `if haskey()` statement!

src/shared_utilities/models.jl
```diff
@@ -404,6 +404,15 @@ function get_drivers(model::AbstractModel)
     return (nothing, nothing)
 end
 
+"""
+    add_dfluxBCdY_to_cache(p::NamedTuple, model::AbstractModel, _, _)
+
+We only need to add this field in the case of a `RichardsModel`, so we
+provide a fallback method for all other models which does nothing.
+"""
+function add_dfluxBCdY_to_cache(p::NamedTuple, model::AbstractModel, _, _)
```
kmdeck, line 413: I dont think we need this because we have the `boundary_vars` (and types, domain_names) functions which have the capability to add these variables already.

src/shared_utilities/models.jl
```diff
@@ -418,5 +427,11 @@ function initialize(model::AbstractModel{FT}) where {FT}
     Y = initialize_prognostic(model, coords)
     p = initialize_auxiliary(model, coords)
     p = add_drivers_to_cache(p, model, coords)
+    p = add_dfluxBCdY_to_cache(
```
kmdeck, line 430: similar, would not need this because the methods for `boundary_vars` change for the model/bc combo

kmdeck: note that the `boundary_var` interface is only used for models with PDEs currently, where have/want the capability to set flexible boundary conditions, and the different types of BCs might require different variables. The canopy model is solving a set of ODEs without the same flexibility. If we want to ever implicitly timestep those equations, we would need some extra machinery, possibly like what you sketched out here. but what we have should also work if we need to implicitly timestep the soil & account for dfluxBCdY in the integrated soil/canopy case, since the soil has a different BC type in that case.

src/standalone/Soil/rre.jl
```diff
@@ -324,88 +326,81 @@ function ClimaLand.make_update_aux(model::RichardsModel)
             effective_saturation(ν, Y.soil.ϑ_l, θ_r),
         )
         @. p.soil.ψ = pressure_head(hydrology_cm, θ_r, Y.soil.ϑ_l, ν, S_s)
+
+        # If p.soil.dfluxBCdY is present, update it
```
kmdeck, line 330: I dont think we need the if statement since we have a default `set_dfluxBCdY!` method for any model/any BC
I think this should not go in `update_aux!`, but instead here https://github.com/CliMA/ClimaLand.jl/blob/d60e2b27f0b87a37792d984ea08c7f6d810fb81d/src/standalone/Soil/rre.jl#L102

kmdeck: because update_boundary_fluxes! is called after all the other cache variables are updated. In the soil/canopy (integrated model) case, the boundary flux will depend on cache variables in another component. if we need to account for dfluxBCdY in that case, we would need to set it here instead of in update_aux! to be sure the other cache variables are up to date.

src/standalone/Soil/rre.jl
```diff
@@ -420,57 +415,71 @@ to using the modified Picard scheme of Celia et al. (1990).
...
-            model.boundary_conditions.top,
-            ClimaLand.TopBoundary(),
-            Δz_top,
-            Y,
-            p,
-            t,
-        )
+
+
+        ∂T_bc∂YN = p.dfluxBCdY
```
kmdeck, line 481: I believe we can get rid of this + code that follows as we added the BC contribution above, but we can discuss!

src/standalone/Soil/rre.jl
```diff
@@ -504,9 +513,44 @@ end
...
+respect to the state variables, and merges it into `p` under the key `dfluxBCdY`.
+Specifically, this stores the derivative of the top face flux with respect to
+the state variable at the top cell center.
+This is used in the implicit solver for
+`RichardsModel` when using a `MoistureStateBC` at the top boundary, in which
+case we need to handle the boundary specially in computing the Jacobian.
+
+"""
+# TODO should this go in p.dfluxBCdY or p.soil.dfluxBCdY?
+function ClimaLand.add_dfluxBCdY_to_cache(
```
kmdeck, line 541: not sure if we need this method because we added the variable to the cache by adding it to `boundary_vars`. In this case, it is added to `p.soil`, not to `p` directly.

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -487,13 +489,14 @@ function ClimaLand.∂tendencyBC∂Y(
         interpc2f_op.(dψdϑ.(hydrology_cm, Y.soil.ϑ_l, ν, θ_r, S_s)),
         face_len,
     )
-    return ClimaCore.Fields.FieldVector(;
-        :soil => (; :ϑ_l => @. -K / Δz * dψ / (2 * Δz)),
-    )
+    # Update `dfluxBCdY` in place
+    p.dfluxBCdY .=
```
juliasloan25, line 493: store in p.soil

test/runtests.jl
```diff
@@ -25,9 +25,9 @@ end
 end
 
 # Standalone Bucket model tests
-@safetestset "Bucket albedo types tests" begin
-    include("standalone/Bucket/albedo_types.jl")
-end
+# @safetestset "Bucket albedo types tests" begin
```
kmdeck, line 28: need to uncomment?

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -466,7 +470,7 @@ variable, this is given by
 `∂T_N∂Y_N = [-∂/∂z(∂F_bc/∂Y_N)]_N`, where `N` indicates the top
```
kmdeck, line 470: I think we need to update the doc string

src/standalone/Soil/rre.jl
```diff
@@ -118,6 +122,19 @@ function make_update_boundary_fluxes(model::RichardsModel)
             p,
             t,
         )
+
+        # If p.soil.dfluxBCdY is present, update it
+        if haskey(p.soil, :dfluxBCdY)
```
kmdeck, line 127: we have defined a default version of `set_dfluxBCdY!` which does `nothing` if `model::AbstractModel`. We have a more specific method defined below. Do we still need the if statement? or can we remove that, and call set_dfluxBCdY!, but in all cases but this one, it does nothing?

src/standalone/Soil/rre.jl
```diff
@@ -330,84 +347,76 @@ end
 
 
 """
-    RichardsTridiagonalW{R, J, W, T} <: ClimaLand.AbstractTridiagonalW
+    ImplicitEquationJacobian{M, S}
```
kmdeck, line 350: https://github.com/CliMA/ClimaLand.jl/blob/ec37455a0e58d6662c44a7216a84cc16042fa8f5/src/standalone/Soil/rre.jl#L364

I see this is defined only in rre.jl. Should this be defined in Soil.jl (the same struct will also be used for the full soil model)? Or, if not, should this be defined something more specific (RichardsImplicitEquationJacobian)?

src/shared_utilities/implicit_tendencies.jl
```diff
@@ -1,5 +1,4 @@
-export make_tendency_jacobian,
-    make_update_jacobian, AbstractTridiagonalW, ∂tendencyBC∂Y
+export make_tendency_jacobian, make_update_jacobian, set_dfluxBCdY!
```
kmdeck, line 1: I wonder if this should have a clearer name, or if we need this function at all. We can get rid of this function if we move the `update_aux!` and `update_boundary_fluxes!` function calls that `make_tendency_jacobian` does before it calls update jacobian into update_jacobian.

kmdeck (commented):
This looks good! I left a couple of minor comments. nice job @juliasloan25 @dennisYatunin !!

src/standalone/Soil/rre.jl
```diff
@@ -330,84 +347,76 @@ end
...
-    return RichardsTridiagonalW(
-        dtγ_ref,
-        ∂ϑₜ∂ϑ,
-        W_column_arrays,
-        temp1,
-        temp2,
-        transform,
-        ones_face_space,
+
+    # Cosntruct a tridiagonal matrix that will be used as the Jacobian
```
Sbozzolo, line 385:
```suggestion
    # Construct a tridiagonal matrix that will be used as the Jacobian
```

src/standalone/Soil/rre.jl
juliasloan25: Good point! `set_dfluxBCdY!` has to dispatch on the model type and also the boundary condition type to only update the cache for the combination of RichardsModel with a MoistureStateBC at the top boundary. But that's what the non-nothing method dispatches off of so I think we can just remove the if statement

src/standalone/Soil/rre.jl
juliasloan25: I think the same struct will be used for the full soil model, since they'll both have tridiagonal jacobians, and this struct is generalized for any soil model rather than just richards

src/shared_utilities/implicit_tendencies.jl
juliasloan25: Good point! I think it makes more sense to get rid of `make_tendency_jacobian`, and just have `make_update_jacobian`

juliasloan25: I realized that we're updating the aux and boundary fluxes in `exp_tendency!`, `imp_tendency!`, and `update_jacobian!`. I think this is fine for now, but we might want to avoid recomputing these multiple times per timestep. Dennis mentioned that ClimaTimeSteppers has the capability to take in an update cache function, which will be called once per timestep (Atmos leverages this to call `set_precomputed_quantities!` as few times as possible)., so we could later on switch to using that as a performance improvement

juliasloan25: After removing `tendency_jacobian!`, we can't directly compare the jacobian entries in `test/shared_utilities/implicit_timestepping/richards_model.jl`, since `update_jacobian!` also calls `update_aux` and `boundary_fluxes`. I wonder if it's still worth keeping separate functions for cleaner structure and more granular testing

juliasloan25:
failing because of a change introduced in https://github.com/CliMA/ClimaCore.jl/pull/1653/commits/31e69efe212afcff92356278d87c671db29a790a - need to debug this

juliasloan25:
> failing because of a change introduced in [CliMA/ClimaCore.jl#1653](https://github.com/CliMA/ClimaCore.jl/pull/1653) - need to debug this

fixed in https://github.com/CliMA/ClimaCore.jl/pull/1750/files#diff-fd85ec06a7d1ce92773b2c873ef104edc2471ecf15908935906af32125bb38a1

kmdeck (approved):
looks great! thank you!!

PR #546, opened 2024-03-12
fix CPU global bucket state

juliasloan25:
## Purpose 
After https://github.com/CliMA/ClimaLand.jl/pull/536, we begin to see that output for the global bucket on CPU is incorrect (e.g. build [here](https://buildkite.com/clima/climaland-ci/builds/3500#018e10f5-aef9-4509-889e-724db4be3bd8)). This wasn't caught before because in our comparison of cpu/gpu means, we weren't taking the absolute value.

## To-do
- [x] add absolute value to comparison
- [x] revert changes to find which caused invalid CPU results
- [x] apply fix to avoid overwriting CPU results
  - this involves wrapping the result of interpolation in `Array(...)` so that memory is allocated for it. A fix for this will be included in ClimaCore v0.13.2 (see ClimaCore [PR](https://github.com/CliMA/ClimaCore.jl/pull/1643#pullrequestreview-1932233793))


.buildkite/pipeline.yml
```diff
@@ -23,6 +23,7 @@ steps:
 
       - echo "--- Instantiate experiments"
       - "julia --project=experiments -e 'using Pkg; Pkg.develop(;path=\".\"); Pkg.instantiate(;verbose=true)'"
+      - "julia --project=experiments -e 'using Pkg; Pkg.add(url=\"https://github.com/CliMA/ClimaCore.jl\")'"
```
juliasloan25, line 26: Why do we need this?

Project.toml
```diff
@@ -42,7 +42,7 @@ ClimaParams = "0.10.2"
 CSV = "0.10"
 CUDA = "5"
 ClimaComms = "0.5.6"
-ClimaCore = "0.12, 0.13"
+ClimaCore = "0.13"
```
juliasloan25, line 45: maybe we should put 0.13.2

.buildkite/pipeline.yml
Sbozzolo: I was testing to make sure it worked.

experiments/standalone/Bucket/compare_gpu_cpu_output.jl
```diff
@@ -4,6 +4,6 @@ import ClimaLand
 outdir = joinpath(pkgdir(ClimaLand), "experiments/standalone/Bucket/artifacts")
 cpu_state = readdlm(joinpath(outdir, "tf_state_cpu.txt"), ',')
 gpu_state = readdlm(joinpath(outdir, "tf_state_gpu.txt"), ',')
-@show mean(cpu_state .- gpu_state)
+@show maximum(cpu_state .- gpu_state)
 @show eps(Float64)
-@assert mean(cpu_state .- gpu_state) < eps(Float64) * 5e2
+@assert abs(maximum(cpu_state .- gpu_state)) < eps(Float64) * 1e3
```
Sbozzolo, line 9: I was planning on using `isapprox` instead

experiments/standalone/Bucket/compare_gpu_cpu_output.jl
```diff
@@ -4,6 +4,7 @@ import ClimaLand
 outdir = joinpath(pkgdir(ClimaLand), "experiments/standalone/Bucket/artifacts")
 cpu_state = readdlm(joinpath(outdir, "tf_state_cpu.txt"), ',')
 gpu_state = readdlm(joinpath(outdir, "tf_state_gpu.txt"), ',')
-@show mean(cpu_state .- gpu_state)
-@show eps(Float64)
-@assert mean(cpu_state .- gpu_state) < eps(Float64) * 5e2
+@show abs(maximum(cpu_state .- gpu_state))
+@show abs(median(cpu_state .- gpu_state))
+@show abs(mean(cpu_state .- gpu_state))
+@assert isapprox(cpu_state, gpu_state)
```
Sbozzolo, line 10: @juliasloan25 what do you think of this?

Instead of comparing only one value, we compare the entire state.

Sbozzolo (approved):
LGTM

experiments/standalone/Bucket/compare_gpu_cpu_output.jl
juliasloan25: That looks good! Is there a way to set the tolerance for `isapprox` in case they're too different?

experiments/Manifest.toml
```diff
@@ -1,13 +1,13 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.1"
+julia_version = "1.10.0"
```
juliasloan25, line 3: we shouldn't downgrade julia

experiments/standalone/Bucket/compare_gpu_cpu_output.jl
Sbozzolo: Yes, you can set `atol` and `rtol` if you want to

PR #551, opened 2024-03-13
Add benchmark for global bucket

Sbozzolo:
This PR adds flamegraphs for allocations and compute for the global temporal bucket runs.

### DIFF
--- a/experiments/standalone/Bucket/global_bucket_temporalmap.jl
+++ b/experiments/standalone/Bucket/global_bucket_temporalmap.jl
@@ -38,6 +38,14 @@ using ClimaLand:
     PrescribedRadiativeFluxes
 using ClimaLand.TimeVaryingInputs
 
+PROFILING = false
+try
+    import Profile, ProfileCanvas
+    global PROFILING = true
+    @info "ProfileCanvas found, running with profiler"
+catch
+end
+
 """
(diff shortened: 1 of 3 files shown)


experiments/standalone/Bucket/global_bucket_temporalmap.jl
```diff
@@ -38,6 +38,14 @@ using ClimaLand:
     PrescribedRadiativeFluxes
 using ClimaLand.TimeVaryingInputs
 
+PROFILING = false
+try
```
juliasloan25, line 42: why is this in a try catch? so we only profile on buildkite when we've added ProfileCanvas to the environment?

juliasloan25 (approved):
This looks good to me! It'll be nice to have it for the most complex albedo case

experiments/standalone/Bucket/global_bucket_temporalmap.jl
Sbozzolo: I don't want to add ProfileCanvas to experiments environment, and users should be able to run the code outside of buildkite

juliasloan25 (commented):
I almost forgot! Please also update the NEWS.md to explain these changes :) I'll send a message to the land channel

PR #557, opened 2024-03-22
Update StatsBase compat

akshaysridhar:
Updates compat restriction on `StatsBase`.


Project.toml
```diff
@@ -55,7 +55,6 @@ LinearAlgebra = "1"
 NCDatasets = "0.11, 0.12, 0.13, 0.14"
 SciMLBase = "1, 2"
 StaticArrays = "1"
-StatsBase = "0.34"
```
nefrathenrici, line 58: I think the compat should be `StatsBase = "0.33, 0.34"`. Atmos has the same [compat](https://github.com/CliMA/ClimaAtmos.jl/blob/main/Project.toml#L80C1-L80C25)

akshaysridhar: Fixed.

nefrathenrici:
Should we make a patch release with the updated compat?

juliasloan25:
> Should we make a patch release with the updated compat?

Sure, we can! I was hoping to make patch releases of ClimaLand and ClimaAtmos today, so if we can get this in soon we can make a release with it

juliasloan25 (approved):

PR #561, opened 2024-03-26
Add profiling of the land surface model.

kmdeck:
## Purpose 
- Adds flame graphs for the soil/canopy land model global run
- changes a `sum`, which worked for a column, to a column integral
- changes the zenith angle driver update to work for the case where the zenith angle function returns a field. see Issue #574 as well
- updates the dss! helper function to work for Tuple valued fields, without adding another buffer for the tuple valued field. See issue #573 for future work
- Increases the number of points in the global bucket runs by 5x (unrelated, but a very small change, and easy to merge in this one)

## Content
- update NEWS
- add script
- update buildkite

Review checklist

I have:

----


Sbozzolo:
Is this run different from `experiments/integrated/fluxnet/run_fluxnet.jl US-MOz` (besides profiling)?

kmdeck:
> Is this run different from `experiments/integrated/fluxnet/run_fluxnet.jl US-MOz` (besides profiling)?

hmm slightly. We have the run `experiments/integrated/fluxnet/run_fluxnet.jl US-MOz` which runs for a month? and produces science plots. it saves output every hour I believe. That is true for every fluxnet site we test at currently.

and then we have a one-off run for the Ozark site which tests conservation of water and energy (saves output every step, runs for a shorter amount of time). This is also runs with FT = Float64 and Float32 and makes sure the type is preserved by the simulation

and this adds a very similar one off test as above ^^ but doesnt save any output, and runs only for a half day.

I agree that three scripts may not be needed, as there is a lot of duplicate code, but they test three-four different things, so I found it conceptually simpler to have them be different scripts...

Sbozzolo:
Maybe we could consolidate the last two?

Having three nearly identical scripts increases the maintenance costs

kmdeck:
> Maybe we could consolidate the last two?
> 
> Having three nearly identical scripts increases the maintenance costs

ok, Ill try and address this. it was a little ugly and unreadable when I had done that originally (if PROFILING and FT ==Float64 do one thing, else do another, and if Float32 do yet another).

experiments/integrated/ozark/allocations/ozark_allocations.jl
```diff
@@ -0,0 +1,128 @@
...
+catch
+end
+global FT = Float64
+climaland_dir = pkgdir(ClimaLand)
+global site_ID = "US-MOz"
+savedir = joinpath(climaland_dir, "experiments/integrated/ozark/allocations")
+global earth_param_set = LP.LandParameters(FT)
+
+# Create model and setup some timestepping parameters
+include(
```
kmdeck, line 32: duplicate code is in the simulation setup and in the initial conditions. Note that we should reset initial conditions in between runs, in particular `p` may be updated in place during the sim. so that is why that is in a separate file from the rest of the model setup.

Ideally these would not be `include` statements, but instead be calling functions. but both included scripts require global variables and I had a world age issue when I tried it. Im not sure. maybe there is an easy fix

kmdeck:
> > Maybe we could consolidate the last two?
> > Having three nearly identical scripts increases the maintenance costs
> 
> ok, Ill try and address this. it was a little ugly and unreadable when I had done that originally (if PROFILING and FT ==Float64 do one thing, else do another, and if Float32 do yet another).

Maybe a good middle ground is to have separate scripts but have them rely on the same functions (defined in included scripts). Then code duplication is reduced. Maintenance would still be an issue. That is what I tried in the current version.

Having a single script do multiple things (involving 5 different solve commands with different settings) feels messy, *especially* because I am having trouble putting all the model/state setup into functions that I can call in each of these 5 situations.

experiments/integrated/performance/conservation/ozark_conservation.jl
```diff
@@ -15,252 +15,25 @@ import ClimaLand
...
+global climaland_dir = pkgdir(ClimaLand)
+global site_ID = "US-MOz"
 
 for float_type in (Float32, Float64)
     # Make these global so we can use them in other ozark files
     global FT = float_type
     global earth_param_set = LP.LandParameters(FT)
-    global site_ID = "US-MOz"
 
-    # Utility functions for reading in and filling fluxnet data
```
kmdeck, line 26: This stuff, along with model setup, is all in a single included file now.

experiments/integrated/performance/profile_allocations.jl
```diff
@@ -0,0 +1,443 @@
...
+
+    Insolation.instantaneous_zenith_angle.(
+        d,
+        δ,
+        η_UTC,
+        sfc_cds.long,
+        sfc_cds.lat,
+    ).:1
+end
+function SW_IN(
```
kmdeck, line 157: this is allocating, revert to simple analytic function for this test

NEWS.md
```diff
@@ -3,6 +3,7 @@ ClimaLand.jl Release Notes
 
 main
 --------
+-  Add profiling of soil/anopy model to buildkite pipeline PR [#561](https://github.com/CliMA/ClimaLand.jl/pull/561)
```
Sbozzolo, line 6: Are there more news entries to be added here?

.buildkite/pipeline.yml
```diff
@@ -71,8 +71,12 @@ steps:
...
 
       - label: "ozark_conservation"
-        command: "julia --color=yes --project=experiments experiments/integrated/ozark/conservation/ozark_conservation.jl"
-        artifact_paths: "experiments/integrated/ozark/conservation/*png"
+        command: "julia --color=yes --project=experiments experiments/integrated/performance/conservation/ozark_conservation.jl"
+        artifact_paths: "experiments/integrated/performance/conservation/*png"
+
+      - label: "soil/canopy lsm performance"
+        command: "julia --color=yes --project=experiments experiments/integrated/performance/profile_allocations.jl"
+        artifact_paths: "experiments/integrated/performance/*html"
```
Sbozzolo, line 79: Can we run this on GPU too?

experiments/integrated/performance/profile_allocations.jl
```diff
@@ -0,0 +1,416 @@
...
+    soil_model_type = soil_model_type,
+    soil_args = soil_args,
+    canopy_component_types = canopy_component_types,
+    canopy_component_args = canopy_component_args,
+    canopy_model_args = canopy_model_args,
+)
+exp_tendency! = make_exp_tendency(land)
+# Set up timestepping and simulation callbacks
+dt = Float64(150)
+tf = Float64(t0 + dt * 100)
```
Sbozzolo, line 336:
```suggestion
tf = Float64(t0 + 100dt)
```
cuter :)

src/shared_utilities/utils.jl
```diff
@@ -165,7 +165,18 @@ function dss_helper!(
     p::NamedTuple,
 )
     buffer = p.dss_buffer_2d
-    ClimaCore.Spaces.weighted_dss!(field, buffer)
+    field_element_type = typeof(field).types[1].parameters[1]
```
Sbozzolo, line 168: Can you add two lines of comment to explain this?

src/shared_utilities/utils.jl
```diff
@@ -165,7 +165,18 @@ function dss_helper!(
...
+    if field_element_type <: AbstractFloat
+        ClimaCore.Spaces.weighted_dss!(field, buffer)
+    elseif field_element_type <: Tuple
+        n = length(field_element_type.types)
+        for i in 1:1:n
```
Sbozzolo, line 173:
```suggestion
        for i in 1:n
```

src/shared_utilities/utils.jl
kmdeck: yes! will do

NEWS.md
kmdeck: Ill note the other changes

.buildkite/pipeline.yml
kmdeck: added!

src/shared_utilities/utils.jl
```diff
@@ -165,7 +165,23 @@ function dss_helper!(
     p::NamedTuple,
 )
     buffer = p.dss_buffer_2d
-    ClimaCore.Spaces.weighted_dss!(field, buffer)
+    # The buffer is set up expecting a single scalar at each point in
+    # the space. if the field element type is a scalar, we are ready to
+    # compute the dss. However, if the field is Tuple-valued,
+    # we need to carry out the dss for each element of the tuple (each
+    # of which is equivalent to a scalar valued field)
+    field_element_type = typeof(field).types[1].parameters[1]
```
Sbozzolo, line 173: Can't we use eltype here?

Sbozzolo (approved):

src/shared_utilities/utils.jl
kmdeck: yes! thanks, much better

PR #564, opened 2024-03-26
update script to have lon/lat format and to save lon/lat values

kmdeck:
## Purpose 
Update the topmodel preprocessing script to include the lat and lon values to store them in lon/lat order. The file in box has already been adjusted to reflect this.

## To-do
make sure the file in box works.


Sbozzolo (approved):
Thank you!

PR #565, opened 2024-03-27
Paper plots

kmdeck:
## Purpose 
- Updates evaporation to be less dependent on resolution in the vertical (note: results look worse now)
- ameliorates resolution dep by linearly interpolating to the surface (not efficient as written/allocates a lot)
- Adds two tutorials which make plots included in the paper (moved one from experiments/)
- moves get_deltaz to Domains, as the linear_interpolation_to_surface needs it

## Content

----

### DIFF
--- a/docs/tutorials/standalone/Soil/evaporation_gilat_loess.jl
+++ b/docs/tutorials/standalone/Soil/evaporation_gilat_loess.jl
@@ -1,0 +1,358 @@
+# This sets up the simulation that mimicks the
+# lab experiment
+# presented in Gardener 1970b and modeled
+# also by Lehmann and Or, 2024.
+
+# For further details on how to setup a simulation, please
+# see our other Soil tutorials. This one is very terse
+# and does not provide complete explanations
+
+# The same experiment is carried out 3 times
+# 1. No evaporation (zero flux boundary conditions)
+# 2. With evaporation but no drainage (Ksat = 0)
+# 3. With evaporation and drainage
+
+using CairoMakie
+import SciMLBase
+import ClimaTimeSteppers as CTS
+using Thermodynamics
+
+using ClimaCore
+import ClimaParams as CP
+using SurfaceFluxes
+using StaticArrays
+using Dates
+using ArtifactWrappers
+using DelimitedFiles: readdlm
+
+using ClimaLand
+using ClimaLand.Domains: Column
+using ClimaLand.Soil
+import ClimaLand
+import ClimaLand.Parameters as LP
+import SurfaceFluxes.Parameters as SFP
+
+FT = Float64;
+earth_param_set = LP.LandParameters(FT)
+thermo_params = LP.thermodynamic_parameters(earth_param_set);
+
+# Parameters
+K_sat = FT(0.01 / 3600 / 24)
+vg_n = FT(1.45)
+vg_α = FT(1.5)
+hcm = vanGenuchten{FT}(; α = vg_α, n = vg_n)
+ν = FT(0.4)
+θ_r = FT(0.04)
+S_s = FT(1e-3)
+ν_ss_om = FT(0.0)
+ν_ss_quartz = FT(0.3)
+ν_ss_gravel = FT(0.0)
+emissivity = FT(1.0)
+PAR_albedo = FT(0.2)
+NIR_albedo = FT(0.4)
+z_0m = FT(1e-3)
+z_0b = FT(1e-4)
+d_ds = FT(0.01)# 10mm
+params = ClimaLand.Soil.EnergyHydrologyParameters(
+    FT;
+    ν,
+    ν_ss_om,
+    ν_ss_quartz,
+    ν_ss_gravel,
+    hydrology_cm = hcm,
+    K_sat,
+    S_s,
+    θ_r,
+    PAR_albedo,
+    NIR_albedo,
+    emissivity,
+    z_0m,
+    z_0b,
+    earth_param_set,
+    d_ds,
+);
+
+ref_time = DateTime(2005)
+SW_d = (t) -> 0
+LW_d = (t) -> 294.15^4 * 5.67e-8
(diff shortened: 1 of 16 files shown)


src/integrated/soil_canopy_model.jl
```diff
@@ -417,7 +417,7 @@ function soil_boundary_fluxes!(
     t,
 ) where {FT}
     bc = soil.boundary_conditions.top
-    soil_conditions = turbulent_fluxes(bc.atmos, soil, Y, p, t)
+    p.soil.turbulent_fluxes .= turbulent_fluxes(bc.atmos, soil, Y, p, t)
```
kmdeck, line 420: bug fix 👀

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -492,17 +491,22 @@ function ClimaLand.surface_resistance(
...
-            Y.soil.θ_i,
-            model.parameters,
-        ),
+    cds = ClimaCore.Fields.coordinate_field(model.domain.space.subsurface)
+    (; ν, θ_r, hydrology_cm) = model.parameters
+    S_l = effective_saturation.(ν, p.soil.θ_l, θ_r)
+    S_l_sfc = p.soil.sfc_scratch
+    ClimaLand.Domains.linear_interpolation_to_surface!(S_l_sfc, S_l, cds.z)
+    θ_l_sfc = @. S_l_sfc * (ν - θ_r) + θ_r
+    θ_i_sfc = ClimaLand.Domains.top_center_to_surface(Y.soil.θ_i)
```
juliasloan25, line 500: Maybe we can change this to use `linear_interpolation_to_surface` now too, since we know that function isn't allocating a new field. Or we could have a separate PR to go through and replace `top_center_to_surface` with `linear_interpolation_to_surface!` throughout the code

docs/tutorials/standalone/Soil/freezing_front.jl
```diff
@@ -106,7 +106,7 @@ hydrology_cm = vanGenuchten{FT}(; α = vg_α, n = vg_n);
 #hcm = BrooksCorey(;ψb = ψb, c = c);
 θ_r = FT(0.05)
 ν_ss_om = FT(0.3)
-ν_ss_quartz = FT(0.7)
+ν_ss_quartz = FT(0.3)
```
juliasloan25, line 109: Why is this changed? I also noticed it has value FT(1) in evaporation.jl. Should it be unified across files?

docs/tutorials/standalone/Soil/evaporation_gilat_loess.jl
```diff
@@ -0,0 +1,356 @@
...
+    0.34 / (1 + exp(-(z + 0.165) / 0.005)) + 0.05
+end
+function init_soil!(Y, z, params)
+    FT = eltype(Y.soil.ϑ_l)
+    Y.soil.ϑ_l .= estimated_ic.(z)
+    Y.soil.θ_i .= 0
+    T = FT(294.15)
+    ρc_s = @. Soil.volumetric_heat_capacity(Y.soil.ϑ_l, FT(0), params)
+    Y.soil.ρe_int =
+        Soil.volumetric_internal_energy.(FT(0), ρc_s, T, Ref(params))
```
juliasloan25, line 164: Is the `FT(0)` in these two lines `Ksat`? If so, it would be good to set it on another line to clarify

docs/tutorials/standalone/Soil/evaporation_gilat_loess.jl
```diff
@@ -0,0 +1,356 @@
...
+    top = top_bc,
+    bottom = WaterHeatBC(; water = zero_water_flux, heat = zero_heat_flux),
+)
+
+# Without drainage - wet boundary at bottom
+bottom_water_bc = MoistureStateBC((p, t) -> 0.35)
+no_drainage_boundary_fluxes = (;
+    top = top_bc,
+    bottom = WaterHeatBC(; water = bottom_water_bc, heat = zero_heat_flux),
+)
```
juliasloan25, line 132: I think it would be more clear/readable to define each boundary flux in the case it's used for, but this is personal preference :)

juliasloan25 (approved):
Looks good! I had a few comments, but I think it's good to merge

src/standalone/Soil/energy_hydrology.jl
kmdeck: I will change it here, and then open an issue for the other ones

kmdeck: oh wait. it will be a different issue. because we are in the same function, but we only have one allocated scratch spot to hold the values (instead of alllocating), we cant reuse it in the same function. I think we need multiple surface scratch fields...

docs/tutorials/standalone/Soil/freezing_front.jl
kmdeck: I changed it back - I dont think we have measurements for it, so we dont really know what it will be, but I also dont think it makes a big difference. so might as well keep it the same.

In general this parameter is soil specific (clay vs sand, etc) , so it is OK for it to change across files if those files represent different experiments.

docs/tutorials/standalone/Soil/evaporation_gilat_loess.jl
kmdeck: I defined it in another line, good call!

docs/tutorials/standalone/Soil/evaporation_gilat_loess.jl
kmdeck: good idea! I changed it

src/standalone/Soil/energy_hydrology.jl
kmdeck: noted here: https://github.com/CliMA/ClimaLand.jl/issues/615

PR #567, opened 2024-03-28
Move neural snow to an extension

### LINKED ISSUE #559: Move neural snow to Julia extension
The ClimaLand module for neural-network based snow brings a lot of additional dependencies. In turn, these dependency double the load time for ClimaLand (even when they are not used).

We should turn the neural-network based snow model into an extension, as done by [CloudMicrophysics](https://github.com/CliMA/CloudMicrophysics.jl)

Sbozzolo:
The neural snow module is a major source is latency. This PR moves it to an extension, so that we don't pay the cost of loading it if we are not using it.

Closes #559


src/standalone/Snow/Snow.jl
```diff
@@ -81,9 +81,4 @@ end
 
 include("./snow_parameterizations.jl")
 
-include("./neural_snow/DataTools.jl")
```
kmdeck, line 84: in the future, if we want to make a SnowModel struct that uses neural snow, do we just need to load the extensions in the script/driver like you do in the docs examples? What if there are situations where we need the types defined in ModelTools.jl inside ClimaLand/src?

kmdeck (commented):
I think this is fine, my only question is with respect to future snow models, which may need to dispatch off of the depth parameterization type (of which the neural snow model is one type). For example, with the neural snow depth model, we have another prognostic variable, which we can add easily with our existing utilities with dispatch, but it seems like ClimaLand/src would need to know about the neural snow types in that case.

src/standalone/Snow/Snow.jl
Sbozzolo: No, in the future, we will add to ClimaLand the functions that are relevant to ClimaLand and add their implementation in the extension. Suppose the function you care about is `compute_swe`, we will have `compute_swe` in ClimaLand, and it will be "activated" once add the required modules are loaded.

But I think that we don't need any of the functions defined in ModelTools and DataTools directly in ClimaLand, so I have to fetch them directly from the extension.

Sbozzolo:
> I think this is fine, my only question is with respect to future snow models, which may need to dispatch off of the depth parameterization type (of which the neural snow model is one type). For example, with the neural snow depth model, we have another prognostic variable, which we can add easily with our existing utilities with dispatch, but it seems like ClimaLand/src would need to know about the neural snow types in that case.

Yes, we will add all those types in ClimaLand, and leave the implementation of the infrastructure in the extension.

You can look at ClimaUtilities to get a sense of how this would look like.

Sbozzolo:
@kmdeck anything else here?

kmdeck (approved):
Discussed offline regarding using the NeuralSnow model in our snow model and dispatching off of the type, should all be ok!

thanks!

PR #575, opened 2024-03-29
initial cache method removal

kmdeck:
## Purpose 
Removes special methods of `set_initial_cache`. All models can use the default

## To-do
review

## Content
- bucket model method was a duplicate of the default -> easy to remove
- canopy model was a default except for `set_prescribed_canopy_field` call. Now we call this in update_aux and can use the default initial cache setting method.
- LSM model should have been calling update drivers, then update aux for each model, and then update boundary fluxes for each model, since this is what we do each timestep. it was instead caling update drivers, and then set initial cache for each model. Now we use the default, which does what we want.

Review checklist

I have:

----

### DIFF
--- a/src/standalone/Vegetation/component_models.jl
+++ b/src/standalone/Vegetation/component_models.jl
@@ -132,45 +132,15 @@ end
 """
      set_canopy_prescribed_field!(component::AbstractCanopyComponent,
                                   p,
-                                  t0,
+                                  t,
                                  ) end
 
-Sets the spatially and temporally varying prescribed fields of the `component`
-with their initial values.
-
-These fields are stored in the aux-state and *should not* depend on the prognostic
-state `Y` or other diagnostic variables stored in `p`; this allows them
-to be updated first, prior to updating the rest of the aux state and prognostic state.
-
-However, there is no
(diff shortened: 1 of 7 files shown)


src/standalone/Vegetation/component_models.jl
```diff
@@ -132,30 +132,9 @@ end
...
-
-These fields are stored in the aux-state and *should not* depend on the prognostic
-state `Y` or other diagnostic variables stored in `p`; this allows them
-to be updated first, prior to updating the rest of the aux state and prognostic state.
-
-However, there is no
-guarantee on the order of operations in terms of when diagnostic auxiliary
-variables are updated vs. prescribed field auxiliary variables; any required
-order of operations must be enforced by the developer who writes the update_aux
-function.
```
juliasloan25, line 149: Should we keep this docstring?

juliasloan25 (approved):
This is a great change, thank you Kat!

src/standalone/Vegetation/component_models.jl
kmdeck: I cleaned it up a bit, but I think it still holds, so keeping. let me know if you had something in mind for not including it!

PR #576, opened 2024-04-03
Convert ld to Function type and Parameterize by Leaf Orientation Index

Espeer5:
## Purpose 

Converts `ld` in the TwoStreamModel from a floating point number to a Function pointer, and adds a method for getting `ld` in terms of `χl` as is done in CLM. See [here](https://www2.cesm.ucar.edu/models/cesm2/land/CLM50_Tech_Note.pdf)

## To-do

- [x] Change `ld` in `BeerLambert` model to function for compatibility
- [x] Fix all tests, experiments, and tutorials for use with functional `ld`

## Content

 - Changes `ld` to a submodel `AbstractGFunction`
 - Creates 2 types of G Functions, for each the CLM method and the constant G function we used before
 - Propagates changes throughout tests, tutorials, and experiments
 - Adds methods dispatching off the type of GFunction model used to compute the current value of the G function for the current time step based on the solar zenith angle

----

### DIFF
--- a/experiments/integrated/fluxnet/US-MOz/US-MOz_parameters.jl
+++ b/experiments/integrated/fluxnet/US-MOz/US-MOz_parameters.jl
@@ -34,7 +34,8 @@ soil_α_NIR = FT(0.2)
 
 # TwoStreamModel parameters
 Ω = FT(0.69)
-ld = FT(0.5)
+χl = FT(0.1)
+G_Function = CLMGFunction(χl)
 α_PAR_leaf = FT(0.1)
 λ_γ_PAR = FT(5e-7)
 λ_γ_NIR = FT(1.65e-6)
--- a/experiments/integrated/fluxnet/run_fluxnet.jl
+++ b/experiments/integrated/fluxnet/run_fluxnet.jl
@@ -140,11 +140,11 @@ radiative_transfer_args = (;
     parameters = TwoStreamParameters(
         FT;
         Ω,
-        ld,
         α_PAR_leaf,
         τ_PAR_leaf,
         α_NIR_leaf,
         τ_NIR_leaf,
+        G_Function,
     )
 )
 # Set up conductance
--- a/src/standalone/Vegetation/canopy_parameterizations.jl
+++ b/src/standalone/Vegetation/canopy_parameterizations.jl
@@ -27,6 +27,35 @@ export plant_absorbed_pfd,
 # 1. Radiative transfer
 
 """
+    compute_G(
+        G::ConstantGFunction{FT},
+        _::FT,
+    )
+
+Returns the constant leaf angle distribution value for the given G function.
+Takes in an arbitrary value for the solar zenith angle, which is not used.
+"""
+function compute_G(G::ConstantGFunction{FT}, _::FT) where {FT}
+    return G.ld
+end
+
+"""
+    compute_G(
+        G::CLMGFunction{FT},
+        θs::FT,
+    )
+
+Returns the leaf angle distribution value for CLM G function as a function of the 
+solar zenith angle and the leaf orientation index. See section 3.1 of 
+https://www2.cesm.ucar.edu/models/cesm2/land/CLM50_Tech_Note.pdf
+"""
+function compute_G(G::CLMGFunction{FT}, θs::FT) where {FT}
+    ϕ1 = 0.5 - 0.633 * G.χl - 0.33 * G.χl^2
+    ϕ2 = 0.877 * (1 - 2 * ϕ1)
+    return FT(ϕ1 + ϕ2 * cos(θs))
+end
+
+"""
     compute_absorbances(
         RT::BeerLambertModel{FT},
         PAR,
@@ -201,10 +230,13 @@ function plant_absorbed_pfd(
     frac_diff::FT,
 ) where {FT}
 
-    (; ld, Ω, n_layers) = RT.parameters
+    (; G_Function, Ω, n_layers) = RT.parameters
+
+    # Get the current leaf angular distribution value based on the solar zenith angle
+    G = compute_G(G_Function, θs)
 
     # Compute μ̄, the average inverse diffuse optical length per LAI
-    μ̄ = 1 / (2 * ld)
+    μ̄ = 1 / (2G)
 
     ω = α_leaf + τ_leaf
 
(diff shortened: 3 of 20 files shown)


src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -203,8 +203,11 @@ function plant_absorbed_pfd(
 
     (; ld, Ω, n_layers) = RT.parameters
 
+    # Get the current leaf angular distribution value based on the solar zenith angle
+    G = FT(ld(cos(θs)))
```
kmdeck, line 207: I wonder if we can use a better name than G? or is G common notation for this problem?

kmdeck (commented):
One bigger question (that might also help with the GPU question) - would it be worthwhile to think of two different models for leaf_angle_distribution, one being `Constant` and the other `CLMStyle` (needs better name). and then we have two methods for it?

this is almost like what you are doing but it seems the above is the more Julian way. (I cant speak to the preference to one vs the other besides that, except that having two types might make it a little clearer what is supported).

maybe @Sbozzolo also has an opinion on this

src/standalone/Vegetation/canopy_parameterizations.jl
braghiere: G-funtion is the name. Ross, J. (1981), The Radiation Regime and Architecture of Plant Stands, Dr. W. Junk, The Hague, Netherlands.

Espeer5:
> One bigger question (that might also help with the GPU question) - would it be worthwhile to think of two different models for leaf_angle_distribution, one being `Constant` and the other `CLMStyle` (needs better name). and then we have two methods for it?
> 
> this is almost like what you are doing but it seems the above is the more Julian way. (I cant speak to the preference to one vs the other besides that, except that having two types might make it a little clearer what is supported).
> 
> maybe @Sbozzolo also has an opinion on this

So I thought about this approach, and the reason I opted for this approach rather than using two different models was just so that it would be very easy for users to extend our code to use any arbitrary ld function, as there are many different parameterizations in the literature (see Table 1 [here](https://besjournals.onlinelibrary.wiley.com/doi/full/10.1111/2041-210X.13273)) without having to set up an entire new model to plug in. But I am certainly open to taking the other approach, particularly if it would be the solution to the GPU issue.

experiments/integrated/fluxnet/US-MOz/US-MOz_parameters.jl
```diff
@@ -34,7 +34,10 @@ soil_α_NIR = FT(0.2)
 
 # TwoStreamModel parameters
 Ω = FT(0.69)
-ld = FT(0.5)
+# function ld(_)::FT 
+#     return FT(0.5)
+# end
```
Sbozzolo, line 39: This come is commented our

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -136,6 +136,9 @@ canopy_component_types = (;
 autotrophic_respiration_args =
     (; parameters = AutotrophicRespirationParameters(FT))
 # Set up radiative transfer
+if @isdefined(χl)
+    ld = ld_from_χl(χl)
+end
```
Sbozzolo, line 141: This is a rather fragile way to define `ld`. If you want to control this behavior, best to be specific about it by adding a flag that determines if ld should computed from chi or not.

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -203,8 +203,11 @@ function plant_absorbed_pfd(
...
+
     # Compute μ̄, the average inverse diffuse optical length per LAI
-    μ̄ = 1 / (2 * ld)
+    μ̄ = 1 / (2 * G)
```
Sbozzolo, line 210:
```suggestion
    μ̄ = 1 / (2G)
```

src/standalone/Vegetation/radiation.jl
```diff
@@ -29,6 +28,8 @@ Base.@kwdef struct BeerLambertParameters{FT <: AbstractFloat}
     λ_γ_PAR::FT
     "Typical wavelength per NIR photon (m)"
     λ_γ_NIR::FT
+    "Leaf angle distribution function (unitless)"
+    ld::Function
 end
```
Sbozzolo, line 33:
```suggestion
Base.@kwdef struct BeerLambertParameters{FT <: AbstractFloat, F <: Function}
    "PAR leaf reflectance (unitless)"
    α_PAR_leaf::FT
    "NIR leaf reflectance"
    α_NIR_leaf::FT
    "Emissivity of the canopy"
    ϵ_canopy::FT
    "Clumping index following Braghiere (2021) (unitless)"
    Ω::FT
    "Typical wavelength per PAR photon (m)"
    λ_γ_PAR::FT
    "Typical wavelength per NIR photon (m)"
    λ_γ_NIR::FT
    "Leaf angle distribution function (unitless)"
    ld::F
end
```

src/standalone/Vegetation/radiation.jl
```diff
@@ -95,6 +95,38 @@ A constructor supplying default values for the TwoStreamParameters struct.
...
+passes in the parameter χl, which is used to calculate the leaf angular 
+distribution function as a function of the solar zenith angle.
+"""
+
+"""
+A method for constructing a leaf angular distribution function from a leaf 
+orientation index, to parameterize ld in the same way as CLM. 
+See section 3.1 of 
+https://www2.cesm.ucar.edu/models/cesm2/land/CLM50_Tech_Note.pdf
+"""
```
Sbozzolo, line 123: The docstring doesn't match the function

Espeer5:
Thanks for all the feedback. I reworked the PR to split up G functions into 2 models with their own method as @kmdeck suggested. Let me know how this looks.

braghiere:
How do the test look like for the new G-funtion for SW albedo?

Espeer5:
> How does the test look like for the new G-funtion for SW albedo?

It looks good! Practically the same as running with the constant G function from before. For the Ozark site running with the new G function:
SW OUT_avg

braghiere:
> > How does the test look like for the new G-funtion for SW albedo?
> 
> It looks good! Practically the same as running with the constant G function from before. For the Ozark site running with the new G function: SW OUT_avg

Great job! :)

braghiere (approved):

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -201,10 +230,13 @@ function plant_absorbed_pfd(
     frac_diff::FT,
 ) where {FT}
 
-    (; ld, Ω, n_layers) = RT.parameters
+    (; G_Function, Ω, n_layers) = RT.parameters
+
+    # Get the current leaf angular distribution value based on the solar zenith angle
+    G = FT(compute_G(G_Function, θs))
```
kmdeck, line 236: I think you're all set without this additional FT

PR #589, opened 2024-04-23
Preallocate thermal state

### LINKED ISSUE #588: `construct_atmos_ts` allocates at every step
This is probably in 
```julia
function construct_atmos_ts(
    atmos::PrescribedAtmosphere{FT},
    p,
    thermo_params,
) where {FT}
    P = p.drivers.P
    T = p.drivers.T
    q = p.drivers.q
    ts_in = Thermodynamics.PhaseEquil_pTq.(thermo_params, P, T, q)
    return ts_in
end
```

Sbozzolo:
Fixes #588 

Leads to x2 improvement:

https://buildkite.com/clima/climaland-benchmark/builds/65#018f3174-1371-42ee-8dc2-eca52228dc0c

Average time: 5.3 s

https://buildkite.com/clima/climaland-benchmark/builds/53#018f3104-e10b-4e7e-af80-0a2476646201

Average time: 10 s


kmdeck (commented):
I think we should discuss where the ts is allocated/stored. In this particular case, I think it could make sense to store it/add it to `p` in:
 - `p.drivers.ts_atmos`, and update not with `evaluate!` like the other driver variables, but instead update with the `set_atmos_ts!` function you wrote. We already have an `update_drivers` function which this would be easy to add to. This is consistent also in that it will update  only update when `update_drivers` is called (as a callback), which is when the other drivers are updated, and not every timestep/stage (even though the drivers may not be changing each step).
 

More generally - if we need to add scratch variables - we can do so pretty easily via our existing `auxiliary_vars` and `boundary_vars` functionality. We may already have some, actually. then they would be stored in `p.soil.scratch` (e.g.) and we can choose which models need them/for which boundary conditions. We could also make a new function like you did and add scratch variables for a model to `p.scratch`, like we do for driver variables. But that is a little bigger of a job and I dont think it is needed for this PR.

juliasloan25 (approved):
This looks great! I agree with Kat's comment, and it looks like those changes have already been incorporated.

Can we fully remove the `construct_atmos_ts` function and remaining calls to it now? (e.g. in exp/Standalone/Soil/evaporation.jl, and a couple of the tests)

src/shared_utilities/models.jl
```diff
@@ -418,5 +426,6 @@ function initialize(model::AbstractModel{FT}) where {FT}
     Y = initialize_prognostic(model, coords)
     p = initialize_auxiliary(model, coords)
     p = add_drivers_to_cache(p, model, coords)
+    p = add_parameters_to_cache(p, model)
```
kmdeck, line 429: I just have a question here - should we store model.parameters in the cache, and then always access it this way (it is also stored in model.parameters, and usually we have accessed it that way in all of our RHS functions)?

Also, should we just store model.parameters.earth_param_set in the cache? Some parameters (e.g. soil spatially varying parameters, which are not in the earth_param_set but are in the model.parameters object) are large - multiple 3D fields. we probably dont want two copies

kmdeck: What work comes next on the parameters/what issues should we open? are we moving towards not storing parameters in model.parameters, but rather in p.parameters?

Sbozzolo: I haven't looked at this code for 3 weeks and forgot everything.

I think you are right: we only want the earth_param_set in the cache, and the model parameters being only the parameters for the model (ie, without the earth_param_set)

kmdeck (approved):
This looks good to me, thank you! nice improvement to the code.

PR #590, opened 2024-04-25
Increase tolerance for soil_test_3d

Sbozzolo:
I don't necessarily understand how, but a test fails [here](https://github.com/CliMA/ClimaUtilities.jl/actions/runs/8792303663/job/24166749566). This runner is a ClimaUtilities "downstream" test: every time we push to ClimaUtilities, we can that it doesn't break ClimaAtmos and ClimaLand.

Somehow, `eps(FT)` is smaller than most other cases in this test. Its value is `3.130974164378131e-17` instead of the usual `2.220446049250313e-16`. It might be machine/julia dependent. I increase the tolerance of that test so that it passes there, too.

It's not the best solution, but I don't have much better ideas.

### DIFF
--- a/test/standalone/Soil/soil_test_3d.jl
+++ b/test/standalone/Soil/soil_test_3d.jl
@@ -227,7 +227,7 @@ for FT in (Float32, Float64)
                 K(θX) * dψdθ(θX) * d2θdx2(XX, myZ) +
                 K(θX) * dθdx(XX, myZ)^2.0 * d2ψdθ2(θX)
             )
-            @test maximum(abs.(dYX .- expected)) < 10^10 * eps(FT)
+            @test maximum(abs.(dYX .- expected)) < 10^11 * eps(FT)
         end
     end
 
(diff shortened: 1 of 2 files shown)


test/standalone/Soil/soil_test_3d.jl
```diff
@@ -227,7 +227,7 @@ for FT in (Float32, Float64)
                 K(θX) * dψdθ(θX) * d2θdx2(XX, myZ) +
                 K(θX) * dθdx(XX, myZ)^2.0 * d2ψdθ2(θX)
             )
-            @test maximum(abs.(dYX .- expected)) < 10^10 * eps(FT)
+            @test maximum(abs.(dYX .- expected)) < 10^11 * eps(FT)
```
kmdeck, line 230: looks fine! The 10^10(11) is arbitrary, but I realized it's not an easy thing to compute a priori -  like we discussed the other day, when making many floating point computations per point in the field, and then taking a statistic of many samples (like a mean), knowing what "floating point error" translates to is hard - but this is even harder, because we are taking a maximum as our statistic, and order statistics depend on the number of samples.

kmdeck (approved):

PR #591, opened 2024-04-25
Global run of complex land model with ERA5 forcing

kmdeck:
## Purpose 
Carry out global run of soil+canopy model on GPU and CPU
## To-do
- Future PR: add in runoff. [This requires altering our land model constructor, which will then involve changing a number of files more (to supply runoff instead of using the default). I will follow up this PR with a quick one adding that.](https://github.com/CliMA/ClimaLand.jl/pull/669)

## Content
1. Previously, we had only ever simulated Richards model globally. Because the soil properties vary spatially, certain functions that are only used by EnergyHydrology needed to be updated to be broadcastable. The primary change is because the parameter struct contains fields, but is not a field of a parameter structs. this means that functions that we broadcasted over a Ref of the parameter struct before now must be supplied each parameter field as an argument. This led to a LOT of changes (volumetric_heat_capacity, volumetric_internal_energy, volumetric_internal_energy_liq, phase_change_source, soil_resistance)
2. Parameters that were marked as FT in the EnergyHydrologyParameter struct now are a unionof FT and Field. Some parameters are only defined on the surface, rather than throughout the soil, so they need a different parametric type
3. added benchmark - on GPU
4. added experiment - GPU/CPU

Note: For the benchmark -  currently about 1/4 of the time is in setup rather than solve, so maybe we need to change the setup of the simulation. however, making it longer leads to failures, so Im not sure yet.

Note: The experiment does not run on GPU in CI - on caltech - , but it does on GPU on clima (benchmark). For now I removed the run on caltech

----

### DIFF
--- a/.buildkite/target/pipeline.yml
+++ b/.buildkite/target/pipeline.yml
@@ -45,3 +45,13 @@ steps:
         agents:
           slurm_mem: 8GB
           slurm_gpus: 1
+
+      - label: "Soil/Canopy"
+        command: "julia --color=yes --project=.buildkite experiments/benchmarks/land.jl"
+        artifact_paths:
+          - "land_benchmark_gpu/*html"
+        env:
+          CLIMACOMMS_DEVICE: CUDA
+        agents:
+          slurm_mem: 8GB
+          slurm_gpus: 1
--- a/experiments/benchmarks/land.jl
+++ b/experiments/benchmarks/land.jl
@@ -1,0 +1,676 @@
+# # Global run of land model
+
+# The code sets up and runs the soil/canopy model for 6 hours on a spherical domain,
+# using ERA5 data. In this simulation, we have
+# turned lateral flow off because horizontal boundary conditions and the
+# land/sea mask are not yet supported by ClimaCore.
+
+# This code runs the model multiple times and collects statistics for execution time and
+# allocations
+#
+# When run with buildkite on clima, this code also compares with the previous best time
+# saved at the bottom of this file
+
+# Simulation Setup
+# Number of spatial elements: 101 in horizontal, 15 in vertical
+# Soil depth: 50 m
+# Simulation duration: 6 hours
+# Timestep: 180 s
+# Timestepper: ARS343
+# Maximum iterations: 1
+# Convergence criterion: 1e-8
+# Jacobian update: Every Newton iteration
+# Precipitation data update: every timestep
+import SciMLBase
+import ClimaComms
+@static pkgversion(ClimaComms) >= v"0.6" && ClimaComms.@import_required_backends
+import ClimaTimeSteppers as CTS
+using ClimaCore
+using ClimaUtilities.ClimaArtifacts
+import Interpolations
+using Insolation
+
+import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput
+import ClimaUtilities.SpaceVaryingInputs: SpaceVaryingInput
+import ClimaUtilities.Regridders: InterpolationsRegridder
+import ClimaUtilities.ClimaArtifacts: @clima_artifact
+import ClimaParams as CP
+
+using ClimaLand
+using ClimaLand.Soil
+using ClimaLand.Canopy
+import ClimaLand
+import ClimaLand.Parameters as LP
+
+using Statistics
+using Dates
+import NCDatasets
+import Profile, ProfileCanvas
+
+const FT = Float64;
+
+regridder_type = :InterpolationsRegridder
+context = ClimaComms.context()
+device = ClimaComms.device()
+device_suffix = device isa ClimaComms.CPUSingleThreaded ? "cpu" : "gpu"
+
+outdir = "land_benchmark_$(device_suffix)"
+!ispath(outdir) && mkpath(outdir)
+
+function setup_prob(t0, tf, Δt; nelements = (101, 15))
+
+    earth_param_set = LP.LandParameters(FT)
+    radius = FT(6378.1e3)
+    depth = FT(50)
+    domain = ClimaLand.Domains.SphericalShell(;
+        radius = radius,
+        depth = depth,
+        nelements = nelements,
+        npolynomial = 1,
+        dz_tuple = FT.((10.0, 0.1)),
+    )
+    surface_space = domain.space.surface
+    subsurface_space = domain.space.subsurface
+
+    ref_time = DateTime(2021)
+    t_start = 0.0
+
+    # Forcing data
+    era5_artifact_path =
+        ClimaLand.Artifacts.era5_land_forcing_data2021_folder_path(; context)    # Precipitation:
+    precip = TimeVaryingInput(
+        joinpath(era5_artifact_path, "era5_2021_0.9x1.25_clima.nc"),
+        "rf",
+        surface_space;
+        reference_date = ref_time,
+        t_start,
+        regridder_type,
+        file_reader_kwargs = (; preprocess_func = (data) -> -data / 3600,),
+    )
+
+    snow_precip = TimeVaryingInput(
+        joinpath(era5_artifact_path, "era5_2021_0.9x1.25.nc"),
+        "sf",
+        surface_space;
+        reference_date = ref_time,
+        t_start,
+        regridder_type,
+        file_reader_kwargs = (; preprocess_func = (data) -> -data / 3600,),
+    )
+
+    u_atmos = TimeVaryingInput(
+        joinpath(era5_artifact_path, "era5_2021_0.9x1.25_clima.nc"),
+        "ws",
+        surface_space;
+        reference_date = ref_time,
+        t_start,
+        regridder_type,
+    )
+    q_atmos = TimeVaryingInput(
+        joinpath(era5_artifact_path, "era5_2021_0.9x1.25_clima.nc"),
+        "q",
+        surface_space;
+        reference_date = ref_time,
+        t_start,
+        regridder_type,
+    )
+    P_atmos = TimeVaryingInput(
+        joinpath(era5_artifact_path, "era5_2021_0.9x1.25.nc"),
+        "sp",
+        surface_space;
+        reference_date = ref_time,
+        t_start,
+        regridder_type,
+    )
+
+    T_atmos = TimeVaryingInput(
+        joinpath(era5_artifact_path, "era5_2021_0.9x1.25.nc"),
+        "t2m",
+        surface_space;
+        reference_date = ref_time,
+        t_start,
+        regridder_type,
+    )
+    h_atmos = FT(10)
+
+    atmos = PrescribedAtmosphere(
+        precip,
+        snow_precip,
+        T_atmos,
+        u_atmos,
+        q_atmos,
+        P_atmos,
+        ref_time,
+        h_atmos,
+        earth_param_set,
+    )
+
+    # Prescribed radiation
+    SW_d = TimeVaryingInput(
+        joinpath(era5_artifact_path, "era5_2021_0.9x1.25.nc"),
+        "ssrd",
+        surface_space;
+        reference_date = ref_time,
+        t_start,
+        regridder_type,
+        file_reader_kwargs = (; preprocess_func = (data) -> data / 3600,),
+    )
+    LW_d = TimeVaryingInput(
+        joinpath(era5_artifact_path, "era5_2021_0.9x1.25.nc"),
+        "strd",
+        surface_space;
+        reference_date = ref_time,
+        t_start,
+        regridder_type,
+        file_reader_kwargs = (; preprocess_func = (data) -> data / 3600,),
+    )
+
+    function zenith_angle(
+        t,
+        ref_time;
+        latitude = ClimaCore.Fields.coordinate_field(surface_space).lat,
+        longitude = ClimaCore.Fields.coordinate_field(surface_space).long,
+        insol_params::Insolation.Parameters.InsolationParameters{FT} = earth_param_set.insol_params,
+    ) where {FT}
+        # This should be time in UTC
+        current_datetime = ref_time + Dates.Second(round(t))
+
+        # Orbital Data uses Float64, so we need to convert to our sim FT
+        d, δ, η_UTC =
+            FT.(
+                Insolation.helper_instantaneous_zenith_angle(
+                    current_datetime,
+                    ref_time,
+                    insol_params,
+                )
+            )
+
+        Insolation.instantaneous_zenith_angle.(
+            d,
+            δ,
+            η_UTC,
+            longitude,
+            latitude,
+        ).:1
+    end
+    radiation =
+        PrescribedRadiativeFluxes(FT, SW_d, LW_d, ref_time; θs = zenith_angle)
+
+    soil_params_artifact_path =
+        ClimaLand.Artifacts.soil_params_artifact_folder_path(; context)
+    extrapolation_bc = (
+        Interpolations.Periodic(),
+        Interpolations.Flat(),
+        Interpolations.Flat(),
+    )
+    soil_params_mask = SpaceVaryingInput(
+        joinpath(
+            soil_params_artifact_path,
+            "vGalpha_map_gupta_etal2020_2.5x2.5x4.nc",
+        ),
+        "α",
+        subsurface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+        file_reader_kwargs = (; preprocess_func = (data) -> data > 0,),
+    )
+    oceans_to_value(field, mask, value) =
+        mask == 1.0 ? field : eltype(field)(value)
+
+    vg_α = SpaceVaryingInput(
+        joinpath(
+            soil_params_artifact_path,
+            "vGalpha_map_gupta_etal2020_2.5x2.5x4.nc",
+        ),
+        "α",
+        subsurface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+    )
+    vg_n = SpaceVaryingInput(
+        joinpath(
+            soil_params_artifact_path,
+            "vGn_map_gupta_etal2020_2.5x2.5x4.nc",
+        ),
+        "n",
+        subsurface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+    )
+    x = parent(vg_α)
+    μ = mean(log10.(x[x .> 0]))
+    vg_α .= oceans_to_value.(vg_α, soil_params_mask, 10.0^μ)
+
+    x = parent(vg_n)
+    μ = mean(x[x .> 0])
+    vg_n .= oceans_to_value.(vg_n, soil_params_mask, μ)
+
+    vg_fields_to_hcm_field(α::FT, n::FT) where {FT} =
+        ClimaLand.Soil.vanGenuchten{FT}(; @NamedTuple{α::FT, n::FT}((α, n))...)
+    hydrology_cm = vg_fields_to_hcm_field.(vg_α, vg_n)
+
+    θ_r = SpaceVaryingInput(
+        joinpath(
+            soil_params_artifact_path,
+            "residual_map_gupta_etal2020_2.5x2.5x4.nc",
+        ),
+        "θ_r",
+        subsurface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+    )
+
+    ν = SpaceVaryingInput(
+        joinpath(
+            soil_params_artifact_path,
+            "porosity_map_gupta_etal2020_2.5x2.5x4.nc",
+        ),
+        "ν",
+        subsurface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+    )
+    K_sat = SpaceVaryingInput(
+        joinpath(
+            soil_params_artifact_path,
+            "ksat_map_gupta_etal2020_2.5x2.5x4.nc",
+        ),
+        "Ksat",
+        subsurface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+    )
+
+    x = parent(K_sat)
+    μ = mean(log10.(x[x .> 0]))
+    K_sat .= oceans_to_value.(K_sat, soil_params_mask, 10.0^μ)
+
+    ν .= oceans_to_value.(ν, soil_params_mask, 1)
+
+    θ_r .= oceans_to_value.(θ_r, soil_params_mask, 0)
+
+
+    S_s =
+        oceans_to_value.(
+            ClimaCore.Fields.zeros(subsurface_space) .+ FT(1e-3),
+            soil_params_mask,
+            1,
+        )
+    ν_ss_om =
+        oceans_to_value.(
+            ClimaCore.Fields.zeros(subsurface_space) .+ FT(0.1),
+            soil_params_mask,
+            0,
+        )
+    ν_ss_quartz =
+        oceans_to_value.(
+            ClimaCore.Fields.zeros(subsurface_space) .+ FT(0.1),
+            soil_params_mask,
+            0,
+        )
+    ν_ss_gravel =
+        oceans_to_value.(
+            ClimaCore.Fields.zeros(subsurface_space) .+ FT(0.1),
+            soil_params_mask,
+            0,
+        )
+    PAR_albedo = ClimaCore.Fields.zeros(surface_space) .+ FT(0.2)
+    NIR_albedo = ClimaCore.Fields.zeros(surface_space) .+ FT(0.2)
+    soil_params = Soil.EnergyHydrologyParameters(
+        FT;
+        ν,
+        ν_ss_om,
+        ν_ss_quartz,
+        ν_ss_gravel,
+        hydrology_cm,
+        K_sat,
+        S_s,
+        θ_r,
+        PAR_albedo = PAR_albedo,
+        NIR_albedo = NIR_albedo,
+    )
+
+
+    # TwoStreamModel parameters
+    Ω = FT(0.69)
+    ld = FT(0.5)
+    α_PAR_leaf = FT(0.1)
+    τ_PAR_leaf = FT(0.05)
+    α_NIR_leaf = FT(0.45)
+    τ_NIR_leaf = FT(0.25)
+
+    # Energy Balance model
+    ac_canopy = FT(2.5e5)
+
+    # Conductance Model
+    g1 = FT(141) # Wang et al: 141 sqrt(Pa) for Medlyn model; Natan used 300.
+
+    #Photosynthesis model
+    Vcmax25 = FT(9e-5) # from Yujie's paper 4.5e-5 , Natan used 9e-5
+
+    # Plant Hydraulics and general plant parameters
+    SAI = FT(0.0) # m2/m2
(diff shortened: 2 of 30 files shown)


src/standalone/Soil/energy_hydrology.jl
```diff
@@ -650,15 +691,22 @@ function ClimaLand.surface_resistance(
...
 ) where {FT}
     cds = ClimaCore.Fields.coordinate_field(model.domain.space.subsurface)
-    (; ν, θ_r, hydrology_cm) = model.parameters
-    θ_l_sfc = p.soil.sfc_scratch
-    ClimaLand.Domains.linear_interpolation_to_surface!(
-        θ_l_sfc,
-        p.soil.θ_l,
+    (; ν, θ_r, d_ds, earth_param_set, hydrology_cm) = model.parameters
+    resistance = p.soil.sfc_scratch
+    return ClimaLand.Domains.linear_interpolation_to_surface!(
```
kmdeck, line 696: revert this change?

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -12,6 +12,7 @@ $(DocStringExtensions.FIELDS)
 Base.@kwdef struct EnergyHydrologyParameters{
     FT <: AbstractFloat,
     F <: Union{<:AbstractFloat, ClimaCore.Fields.Field},
+    SF <: Union{<:AbstractFloat, ClimaCore.Fields.Field},
```
kmdeck, line 15: Add comment as to why we need both

ext/CreateParametersExt.jl
```diff
@@ -297,10 +298,14 @@ function EnergyHydrologyParameters(
     K_sat::F,
     S_s::F,
     θ_r::F,
-    PAR_albedo = 0.2,
-    NIR_albedo = 0.4,
+    PAR_albedo::SF = 0.2,
```
kmdeck, line 301: comment here too about SF vs F

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -321,7 +321,7 @@ function ClimaLand.Canopy.set_canopy_prescribed_field!(
     t,
 ) where {FT}
     (; LAIfunction, SAI, RAI) = component.parameters.ai_parameterization
-    evaluate!(p.canopy.hydraulics.area_index.leaf, LAIfunction, t)
+    evaluate!(p.canopy.hydraulics.area_index.leaf, LAIfunction, floor(t))
```
kmdeck, line 324: why do we need this here but not in for drivers?

experiments/benchmarks/land.jl
```diff
@@ -0,0 +1,670 @@
...
+@info "Saved compute flame to $flame_file"
+
+Profile.Allocs.@profile sample_rate = 0.01 setup_and_solve_problem()
+results = Profile.Allocs.fetch()
+profile = ProfileCanvas.view_allocs(results)
+alloc_flame_file = joinpath(outdir, "alloc_flame_$device_suffix.html")
+ProfileCanvas.html_file(alloc_flame_file, profile)
+@info "Saved allocation flame to $alloc_flame_file"
+
+if ClimaComms.device() isa ClimaComms.CUDADevice
```
kmdeck, line 652: @Sbozzolo do you know what this extra CUDA profile call does? It doesnt upload anything. When we run in buildkite, it successfully got through line 650, and then ran out of memory.

experiments/benchmarks/land.jl
```diff
@@ -0,0 +1,670 @@
...
+    SciMLBase.solve(prob, ode_algo; dt = Δt, callback = cb)
+    return nothing
+end
+
+# Warm up and greet
+setup_and_solve_problem(; greet = true)
+
+@info "Starting profiling"
+# Stop when we profile for MAX_PROFILING_TIME_SECONDS or MAX_PROFILING_SAMPLES
+MAX_PROFILING_TIME_SECONDS = 500
```
kmdeck, line 616: can up this to make the simulations longer and keep number of samples reasonable (~>8)

experiments/benchmarks/land.jl
Sbozzolo: It should provide detailed information about the cost of kernels, but we can remove it if it is a problem (we can put it back when things are more optimized)

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: trying without...

experiments/integrated/global/global_soil_canopy.jl
```diff
@@ -332,7 +332,9 @@ Y.soil.ρe_int .=
         soil_params.earth_param_set,
     )
 Y.soilco2.C .= FT(0.000412) # set to atmospheric co2, mol co2 per mol air
-Y.canopy.hydraulics.ϑ_l.:1 .= plant_ν
+# On the Caltech cluster, this is not permitted, so we allowscalar.
+# This operation is fine on clima cluster
+@allowscalar Y.canopy.hydraulics.ϑ_l.:1 .= plant_ν
```
Sbozzolo, line 337: What's the error on the caltech cluster?

kmdeck:
```
julia --color=yes --project=.buildkite experiments/integrated/global/global_soil_canopy.jl
ERROR: LoadError: Scalar indexing is disallowed.
Invocation of setindex! resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.
If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] errorscalar(op::String)
    @ GPUArraysCore /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:155
  [3] _assertscalar(op::String, behavior::GPUArraysCore.ScalarIndexing)
    @ GPUArraysCore /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:128
  [4] assertscalar(op::String)
    @ GPUArraysCore /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:116
  [5] setindex!(A::CUDA.CuArray{Float64, 4, CUDA.DeviceMemory}, v::Float64, I::Int64)
    @ GPUArrays /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/GPUArrays/OqrUV/src/host/indexing.jl:56
  [6] setindex!
    @ ./subarray.jl:343 [inlined]
  [7] _setindex!
    @ ./abstractarray.jl:1419 [inlined]
  [8] setindex!
    @ ./abstractarray.jl:1396 [inlined]
  [9] set_struct!
    @ /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/ClimaCore/Bf3Rx/src/DataLayouts/struct.jl:257 [inlined]
 [10] setindex!
    @ /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/ClimaCore/Bf3Rx/src/DataLayouts/DataLayouts.jl:784 [inlined]
 [11] fill!(data::ClimaCore.DataLayouts.IJF{Float64, 2, SubArray{Float64, 3, CUDA.CuArray{Float64, 4, CUDA.DeviceMemory}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, Int64}, true}}, val::Float64)
    @ ClimaCore.DataLayouts /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/ClimaCore/Bf3Rx/src/DataLayouts/DataLayouts.jl:734
 [12] fill!(data::ClimaCore.DataLayouts.IJFH{Float64, 2, SubArray{Float64, 4, CUDA.CuArray{Float64, 4, CUDA.DeviceMemory}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, Base.Slice{Base.OneTo{Int64}}}, false}}, val::Float64)
    @ ClimaCore.DataLayouts /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/ClimaCore/Bf3Rx/src/DataLayouts/DataLayouts.jl:316
 [13] copyto!
    @ /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/ClimaCore/Bf3Rx/src/DataLayouts/broadcast.jl:479 [inlined]
 [14] copyto!
    @ /central/scratch/esm/slurm-buildkite/climaland-ci/depot/default/packages/ClimaCore/Bf3Rx/src/Fields/broadcast.jl:408 [inlined]
 [15] materialize!
    @ ./broadcast.jl:914 [inlined]
 [16] materialize!(dest::ClimaCore.Fields.Field{ClimaCore.DataLayouts.IJFH{Float64, 2, SubArray{Float64, 4, CUDA.CuArray{Float64, 4, CUDA.DeviceMemory}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, Base.Slice{Base.OneTo{Int64}}}, false}}, ClimaCore.Spaces.SpectralElementSpace2D{ClimaCore.Grids.SpectralElementGrid2D{ClimaCore.Topologies.Topology2D{ClimaComms.SingletonCommsContext{ClimaComms.CUDADevice}, ClimaCore.Meshes.EquiangularCubedSphere{ClimaCore.Domains.SphereDomain{Float64}, ClimaCore.Meshes.NormalizedBilinearMap}, CartesianIndices{3, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}}, LinearIndices{3, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}}, CUDA.CuArray{Tuple{Int64, Int64, Int64, Int64, Bool}, 1, CUDA.DeviceMemory}, Vector{Tuple{Int64, Int64, Int64, Int64, Bool}}, CUDA.CuArray{Tuple{Int64, Int64}, 1, CUDA.DeviceMemory}, CUDA.CuArray{Int64, 1, CUDA.DeviceMemory}, CUDA.CuArray{Tuple{Bool, Int64, Int64}, 1, CUDA.DeviceMemory}, CUDA.CuArray{Int64, 1, CUDA.DeviceMemory}, CUDA.CuArray{Int64, 1, CUDA.DeviceMemory}, @NamedTuple{}, CUDA.CuArray{Tuple{Int64, Int64}, 1, CUDA.DeviceMemory}}, ClimaCore.Quadratures.GLL{2}, ClimaCore.Geometry.SphericalGlobalGeometry{Float64}, ClimaCore.DataLayouts.IJFH{ClimaCore.Geometry.LocalGeometry{(1, 2), ClimaCore.Geometry.LatLongPoint{Float64}, Float64, StaticArraysCore.SMatrix{2, 2, Float64, 4}}, 2, CUDA.CuArray{Float64, 4, CUDA.DeviceMemory}}, ClimaCore.DataLayouts.IJFH{Float64, 2, CUDA.CuArray{Float64, 4, CUDA.DeviceMemory}}, ClimaCore.DataLayouts.IFH{ClimaCore.Geometry.SurfaceGeometry{Float64, ClimaCore.Geometry.UVVector{Float64}}, 2, CUDA.CuArray{Float64, 3, CUDA.DeviceMemory}}, @NamedTuple{}}}}, bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0}, Nothing, typeof(identity), Tuple{Float64}})
    @ Base.Broadcast ./broadcast.jl:911
 [17] top-level scope
    @ /central/scratch/esm/slurm-buildkite/climaland-ci/4335/climaland-ci/experiments/integrated/global/global_soil_canopy.jl:335
in expression starting at /central/scratch/esm/slurm-buildkite/climaland-ci/4335/climaland-ci/experiments/integrated/global/global_soil_canopy.jl:335
🚨 Error: The command exited with status 1
```

the same exact line runs on clima

Sbozzolo: Weird, might be a depot thing. I don't think they should behave differently

kmdeck: Im not sure. Im just going to remove the GPU run on caltech I think.

src/standalone/Vegetation/PlantHydraulics.jl
juliasloan25: did it work without the floor?

kmdeck: No :/ Im not sure. TimeVaryingInputs work fine in the site level runs, and work fine in global runs except here. I can look into it but would prefer to open an issue and tackle separately, if that is OK. (even next week, so I dont forget about it)

Sbozzolo: What's the issue?

kmdeck: @juliasloan25 Gabriele found the bug, it is in Clima Utilities, where we convert the simulation time into seconds. We must have lucked out before that dt*gamma was always an integer number of seconds, now with ARS343 that is not the case for the global run at least.

The atmos `evaluate!` commands may not trigger the error because (1) they are called only when we update via the callback, which is every 3 hours (integer number of seconds from the start time) or maybe because (2) the LAI data is coming from MODIS, not era5, and so may have a different start time in the file...or another reason TBD


Gabriele will open an issue in ClimaUtilities

Sbozzolo: https://github.com/CliMA/ClimaUtilities.jl/issues/70

I think I removed the floor when porting to ClimaUtilities. We should find the best way to deal with this

kmdeck: Yes, I think the parameters gamma and a42 may be the culprit? https://github.com/CliMA/ClimaTimeSteppers.jl/blob/d10a7d0a55d6edd8f21f6e5852beb54ec416c08d/src/solvers/imex_tableaus.jl#L200

vs RK4: https://github.com/CliMA/ClimaTimeSteppers.jl/blob/d10a7d0a55d6edd8f21f6e5852beb54ec416c08d/src/solvers/explicit_tableaus.jl#L90

which didnt require this (before I rebased the PR for the implicit solver changes)

juliasloan25: Interesting! Thank you for tracking this down :)

ext/CreateParametersExt.jl
```diff
@@ -259,7 +260,12 @@ end
 
 EnergyHydrologyParameters has two constructors: float-type and toml dict based.
 Additional parameters must be added manually: ν, ν_ss_om, ν_ss_quartz, ν_ss_gravel, hydrology_cm, K_sat, S_s, and θ_r
-All parameters can be manually overriden via keyword arguments.
+All parameters can be manually overriden via keyword arguments. Note, however,
+that certain parameters must have the same type (e.g, if a field is 
+supplied for porosity, it must be supplied for all other parameters
+defined in the interior of the domain). Some parameters are defined only
+on the surface of the domain (e.g albedo), while other are defined everywhere
+(e.g. porosity). These are indicated with types `F` and `SF`.
```
juliasloan25, line 268: It would be good to include somewhere in the documentation a list of surface parameters and a list of subsurface parameters

.buildkite/target/pipeline.yml
```diff
@@ -45,3 +45,13 @@ steps:
         agents:
           slurm_mem: 8GB
           slurm_gpus: 1
+
+      - label: ":tree: Land"
+        command: "julia --color=yes --project=.buildkite experiments/benchmarks/land.jl"
```
juliasloan25, line 50: I don't think "land" is a very descriptive name here - maybe we could use "soil_canopy.jl" ?

experiments/integrated/global/global_soil_canopy.jl
juliasloan25: Can we open an issue and just copy the error and where it comes from? Even if we don't address/prioritize a fix, it would be nice to have this documented

docs/Manifest.toml
```diff
@@ -87,6 +87,12 @@ version = "2.3.0"
 uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
 version = "1.1.1"
 
+[[deps.ArnoldiMethod]]
```
juliasloan25, line 90: can we remove the docs update from this PR? It was updated in the PR Alexis just merged, #668

experiments/Manifest.toml
```diff
@@ -5,9 +5,9 @@ manifest_format = "2.0"
 project_hash = "973443a0a609649a9289253b6f5bc3de181d9006"
 
 [[deps.ADTypes]]
```
juliasloan25, line 7: can we also remove the `experiments/Manifest` update? I think we should remove this file entirely now that we have `.buildkite/Manifest.toml`, but that's a separate issue

experiments/integrated/global/global_parameters.jl
```diff
@@ -0,0 +1,210 @@
+using ArtifactWrappers
+#currently not being used
```
juliasloan25, line 2: It looks like this is being used by `global_soil_canopy.jl`! Could you add a comment that this file is meant to be used with that simulation (or others like it)?

juliasloan25: oh, maybe you meant the runoff section, not the whole file :) 

do you plan to include runoff in this PR?

experiments/benchmarks/land.jl
```diff
@@ -0,0 +1,671 @@
...
+            T_exp! = exp_tendency!,
+            T_imp! = SciMLBase.ODEFunction(imp_tendency!; jac_kwargs...),
+            dss! = ClimaLand.dss!,
+        ),
+        Y,
+        (t0, tf),
+        p,
+    )
+
+    updateat = Array(t0:(3600 * 3):tf)
```
juliasloan25, line 572: do we want this to be in terms of `dt` or `tf`? It doesn't have to be, but it is for the other benchmarks

experiments/benchmarks/land.jl
```diff
@@ -0,0 +1,671 @@
...
+    compartment_surfaces =
+        n_stem > 0 ? [zmax, h_stem, h_canopy] : [zmax, h_leaf]
+
+    z0_m = FT(0.13) * h_canopy
+    z0_b = FT(0.1) * z0_m
+
+
+    soilco2_ps = Soil.Biogeochemistry.SoilCO2ModelParameters(
+        FT;
+        ν = 1.0,# INCORRECT!
```
juliasloan25, line 382: incorrect?

experiments/benchmarks/land.jl
```diff
@@ -0,0 +1,671 @@
...
+    # Set up photosynthesis
+    photosynthesis_args = (;
+        parameters = Canopy.FarquharParameters(
+            FT,
+            Canopy.C3();
+            Vcmax25 = Vcmax25,
+        )
+    )
+    # Set up plant hydraulics
+    # Not ideal
```
juliasloan25, line 453: why not ideal?

experiments/integrated/global/global_soil_canopy.jl
```diff
@@ -0,0 +1,444 @@
...
+        T_exp! = exp_tendency!,
+        T_imp! = SciMLBase.ODEFunction(imp_tendency!; jac_kwargs...),
+        dss! = ClimaLand.dss!,
+    ),
+    Y,
+    (t0, tf),
+    p,
+)
+
+saveat = [t0, tf]#Array(t0:3600*3:tf)
```
juliasloan25, line 372: delete comment?

juliasloan25 (commented):
This is awesome, thank you Kat! It would be nice to reduce some of the code duplication between the `land.jl` and `global_soil_canopy.jl` drivers (and maybe for the other benchmark runs too), but this is fine for now :)

experiments/integrated/global/global_soil_canopy.jl
kmdeck: I'll do this!

src/integrated/soil_canopy_model.jl
```diff
@@ -92,7 +92,7 @@ function SoilCanopyModel{FT}(;
     (; atmos, radiation) = land_args
     # These should always be set by the constructor.
     sources = (RootExtraction{FT}(), Soil.PhaseChange{FT}())
-    # add heat BC
+    # How to add TOPMODEL??
```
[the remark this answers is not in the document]
kmdeck: yes! I will do that in a followon PR

.buildkite/target/pipeline.yml
kmdeck: changed!

experiments/integrated/global/global_parameters.jl
kmdeck: I opened PR #669 to do so, and yes, I meant just the runoff section

experiments/benchmarks/land.jl
kmdeck: I thought we could do every 3 hours because that is what we had discussed doing in real simulations (or at most, every hour). I think we had left another benchmark as updating every timestep, and that is helpful for working on optimizing the updating/file IO code, but I dont know that we need to do that in all files.

experiments/benchmarks/land.jl
kmdeck: yes - this should be using the soil porosity, but right now, the Soil Biogeochemistry model does not allow for fields in its parameter struct. I was thinking we could tackle that in another PR (there may be additional change in its' source code). But this is a good reminder, I will open an issue. FWIW, the only spatially varying canopy parameter right now is LAI, so this is not the ideal global model yet for that reason too.

experiments/benchmarks/land.jl
kmdeck: I updated the comment to explain why.

docs/Manifest.toml
kmdeck: yes! ill rebase

experiments/Manifest.toml
kmdeck: good catch!

ext/CreateParametersExt.jl
kmdeck: good idea! I added a description in the doc string for EnergyHydrologyParameters.

experiments/benchmarks/land.jl
juliasloan25: That makes sense! A more descriptive comment here might be helpful, but if it's just temporary and we keep track of the issue I think it's okay

ext/CreateParametersExt.jl
```diff
@@ -259,7 +260,14 @@ end
...
+
+Please see the EnergyHydrologyParameters documentarion for a complete list.
```
juliasloan25, line 270: typo: documentarion -> documentation

juliasloan25 (approved):
LGTM! Thank you for addressing the review so thoroughly :) I left a few more minor comments

kmdeck:
> LGTM! Thank you for addressing the review so thoroughly :) I left a few more minor comments

thank you for the quick and thorough review!

experiments/benchmarks/land.jl
```diff
@@ -0,0 +1,676 @@
...
+    g1 = FT(141) # Wang et al: 141 sqrt(Pa) for Medlyn model; Natan used 300.
+
+    #Photosynthesis model
+    Vcmax25 = FT(9e-5) # from Yujie's paper 4.5e-5 , Natan used 9e-5
+
+    # Plant Hydraulics and general plant parameters
+    SAI = FT(0.0) # m2/m2
+    f_root_to_shoot = FT(3.5)
+    RAI = FT(1.0)
+    K_sat_plant = 5e-9 # m/s # seems much too small?
```
Sbozzolo, line 355:
```suggestion
    K_sat_plant = FT(5e-9) # m/s # seems much too small?
```

experiments/benchmarks/land.jl
```diff
@@ -0,0 +1,676 @@
...
+        Canopy.PlantHydraulics.Weibull{FT}(K_sat_plant, ψ63, Weibull_param)
+    retention_model = Canopy.PlantHydraulics.LinearRetentionCurve{FT}(a)
+    plant_ν = FT(1.44e-4)
+    plant_S_s = FT(1e-2 * 0.0098) # m3/m3/MPa to m3/m3/m
+    rooting_depth = FT(0.5) # from Natan
+    n_stem = 0
+    n_leaf = 1
+    h_stem = 0.0
+    h_leaf = 1.0
+    zmax = 0.0
```
Sbozzolo, line 369: Are there `FT` missing here?

experiments/benchmarks/land.jl
```diff
@@ -0,0 +1,676 @@
...
+
+    # Soil microbes model
+    soilco2_type = Soil.Biogeochemistry.SoilCO2Model{FT}
+
+    # soil microbes args
+    Csom = (z, t) -> eltype(z)(5.0)
+
+    # Set the soil CO2 BC to being atmospheric CO2
+    soilco2_top_bc = Soil.Biogeochemistry.AtmosCO2StateBC()
+    soilco2_bot_bc = Soil.Biogeochemistry.SoilCO2FluxBC((p, t) -> 0.0) # no flux
```
Sbozzolo, line 396: There's probably a FT missing around `0.0`

experiments/benchmarks/land.jl
```diff
@@ -0,0 +1,676 @@
...
+        canopy_component_types = canopy_component_types,
+        canopy_component_args = canopy_component_args,
+        canopy_model_args = canopy_model_args,
+    )
+
+    Y, p, cds = initialize(land)
+
+    init_soil(ν, θ_r) = θ_r + (ν - θ_r) / 2
+    Y.soil.ϑ_l .= init_soil.(ν, θ_r)
+    Y.soil.θ_i .= FT(0.0)
```
Sbozzolo, line 534: Typically, you don't need `FT` when you `.=` into a Field

src/standalone/Vegetation/PlantHydraulics.jl
Sbozzolo, line 324: Maybe using `round` is better?

kmdeck: Youre right, but I think LAI data is updated every 8 days, so floor vs round makes no real difference here. I am leaving this as is since we will remove it entirely when the bug fix is in.

experiments/benchmarks/land.jl
kmdeck: we convert the BC into the FT type internally here: https://github.com/CliMA/ClimaLand.jl/blob/0683b8107f0e7af622a2c6881e9e5597e7d26c47/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl#L456

(these are very allocation heavy and we need to update this code anyways!)

experiments/benchmarks/land.jl
kmdeck: youre right! I will go through the file and update it in this PR #669

PR #592, opened 2024-04-26
Add benchmark pipeline

### LINKED ISSUE #587: Add target GPU run
The first step in improving performance is measuring it. We started adding some jobs, but I think it would be useful to have one single GPU job as target. For the time being, this could be the ERA5-forced global bucket, but it should grow in complexity as new capabilities are added to ClimaLand.

https://github.com/CliMA/ClimaLand.jl/pull/577
#591

### LINKED ISSUE #637: Update CUDA dependencies
We forced CUDA to 0.5.3 in release #636 , (with ]add CUDA@0.5.3) , due to a bug with above version in ClimaCore. Once this is fixed, we need to make CUDA up to date again

Sbozzolo:
Closes #587 

There's a new GitHub label now "Run benchmarks". When this label is activated, the standard benchmark is run for the given PR.

Closes #637 
Closes #639


experiments/benchmarks/bucket.jl
```diff
@@ -0,0 +1,214 @@
+# # Global bucket run using temporal map albedo
```
kmdeck, line 1: This looks good. Do we want to remove the profile/flame graph stuff here then? https://github.com/CliMA/ClimaLand.jl/blob/main/experiments/standalone/Bucket/global_bucket_temporalmap.jl

I am guessing we may still want to keep this running in experiments/standalone/Bucket, since it also uploads plots there, but we could also consider removing it entirely from there?


More broadly, should any script that we do the profile/allocations on should be moved here?

Sbozzolo: I am thinking that this benchmark job should be the ground truth of performance/allocation, but it doesn't have to be the only job that provides information about performance. We can remove the steps from the other bucket jobs, but I think it is useful to have them run all the time so that we can track regressions if they occur.

kmdeck: ok! we can leave as is then. Thank you!

kmdeck (approved):

PR #603, opened 2024-04-30
Move to non-conservative regridder as default

### LINKED ISSUE #569: Consider using InterpolationsRegridder instead of TempestRegridder
Now that we are using `ClimaUtilities` for our regridding needs, we have a new option. Instead of using TempestRemap for regridding, we can use a non-conservative linear interpolation.

The advantages of this are:
- MPI/GPU compatible
- no extra files are created
- it's online, so no work has to be done ahead of time

The disadvantages are:
- it's online, so no work is saved across runs
- it's not conservative

Sbozzolo:
ClimaCoreTempestRemap has a number of small usability issues, from not being compatible with GPUs to creating extra files and regridding everything ahead-of-time. This PR removes ClimaCoreTempestRemap from the dependency tree and let ClimaLand use the non-conservative Interpolations remapping instead.

Closes #569


test/standalone/Bucket/albedo_types.jl
```diff
@@ -91,298 +91,286 @@ FT = Float32
     @test p.bucket.α_sfc == next_alb_manual
 end
 
-# Add tests which use TempestRemap here -
```
juliasloan25, line 94: It's hard to read the diff, but the main change here is removing the `if !Sys.iswindows` check, right? I think everything else is independent of the remapper type

juliasloan25 (approved):

test/standalone/Bucket/albedo_types.jl
Sbozzolo: Yes, correct!

PR #620, opened 2024-05-21
Add missing @static

Sbozzolo:
ClimaComms.@import_required_backends exists only in ClimaComms 0.6, so the macro cannot be evaluated for previous versions. To ensure that the macro evaluation doesn't occur, we need to add a `@static`

As found in: https://github.com/CliMA/ClimaUtilities.jl/actions/runs/8901747705/job/25240331174


juliasloan25 (approved):

juliasloan25:
It would be good to add a note to the [transition wiki](https://github.com/CliMA/ClimaComms.jl/wiki/Transition-to-ClimaComms.jl-0.6)

PR #622, opened 2024-05-22
regridder_type in albedo constructor

juliasloan25:
## Purpose 
Right now, the coupler can't change the regridder type used for albedo. This PR adds `regridder_type` as an optional argument.


src/standalone/Bucket/Bucket.jl
```diff
@@ -189,7 +195,7 @@ function PrescribedSurfaceAlbedo{FT}(
     )
```
Sbozzolo, line 195: The regridder type should be here

PR #632, opened 2024-05-24
Fix docs apps

AlexisRenchon:
## Purpose 

Some docs apps are broken, this is a result of these apps code not being part of CI... 

to do:

- [x] Fix DAMM model app
- [x] Squash

### DIFF
--- a/docs/src/standalone/Project.toml
+++ b/docs/src/standalone/Project.toml
@@ -1,0 +1,6 @@
+[deps]
+Bonito = "824d6782-a2ef-11e9-3a09-e5662e0c26f8"
+ClimaLand = "08f4d4ce-cf43-44bb-ad95-9d2d5f413532"
+ClimaParams = "5c42b081-d73a-476f-9059-fd94b934656c"
+ParamViz = "9706fb73-1de9-418d-83da-fef8cd99bb4a"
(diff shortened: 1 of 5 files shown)


docs/src/standalone/Project.toml
```diff
@@ -0,0 +1,6 @@
+[deps]
+Bonito = "824d6782-a2ef-11e9-3a09-e5662e0c26f8"
+ClimaLand = "08f4d4ce-cf43-44bb-ad95-9d2d5f413532"
+ClimaParams = "5c42b081-d73a-476f-9059-fd94b934656c"
+ParamViz = "9706fb73-1de9-418d-83da-fef8cd99bb4a"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
```
Sbozzolo, line 6: Why is this needed?

AlexisRenchon: We could meet to discuss this, sometime (not a priority), docs/src/standalone/apps.jl makes some interacting viz in the docs, they need those dependencies. 

I am sure there is a better way to organize this though (same for ClimaLandSimulations - I am just not very experienced with making clean environment and code)

Sbozzolo (approved):

PR #634, opened 2024-05-28
tag v0.12.2

AlexisRenchon:
## Purpose 
tag bi-weekly release


docs/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 ArtifactWrappers = "a14bc488-3040-4b00-9dc1-f6467924858a"
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
```
Sbozzolo, line 4: Is this needed?

juliasloan25: Oops, this is leftover from when we ran `]add CUDA@5.3` to lower the version in the manifest. You can run `]rm CUDA` in this environment so it stays in the manifest but is removed from the project @AlexisRenchon

AlexisRenchon:
ugh not sure what I did wrong but I'll restart this PR

PR #636, opened 2024-05-28
tag v0.12.2

AlexisRenchon:
## Purpose 
tag bi-weekly release


juliasloan25 (approved):
Looks good, thanks Alexis! It's probably a good idea to open an issue in ClimaLand (in addition to the one in ClimaCore), so that we remember to remove the pinned CUDA version

Issue #541, opened 2024-03-07
test GPU compatibility after interface change


juliasloan25:
**Is your feature request related to a problem? Please describe.**
Before updating the implicit solver interface, we see a failure in `_ldiv_serial` when running RichardsModel globally on GPU (see error [here](https://buildkite.com/clima/climaland-ci/builds/3506#018e1573-ab8d-4835-97b7-2cc53554818a/196-226)). This is part of the code that should change when we update to the newer ClimaTimeSteppers interface (#528), so we should test this setup again after that's done. We want to test GPU compatibility both for a single column and on a sphere.

After https://github.com/CliMA/ClimaLand.jl/pull/542, we test the jacobian calculation on GPU, but we should also add the following runs on GPU to our pipeline:
- [ ] richards_comparison.jl (column)
- [ ] richards_runoff.jl (sphere)

PR #638, opened 2024-05-29
run RichardsModel on GPU

juliasloan25:
## Purpose 
closes https://github.com/CliMA/ClimaLand.jl/issues/541

## To-do
- [x] add column run on GPU to buildkite (richards_comparison.jl)
- [x] add sphere run on GPU to buildkite (richards_runoff.jl)

## Current status
- column run works
  - no plots - doesn't verify accuracy, just that it runs
- sphere run works
  - no plots - doesn't verify accuracy, just that it runs
  - out of memory error fixed by increasing cache save frequency from `dt` to `100dt`

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -45,7 +45,16 @@ steps:
 
       - label: "Richards comparison to Bonan"
         command: "julia --color=yes --project=.buildkite experiments/standalone/Soil/richards_comparison.jl"
-        artifact_paths: "experiments/standalone/Soil/comparison*png"
+        artifact_paths: "experiments/standalone/Soil/cpu/comparison*png"
+
+      - label: "Richards comparison to Bonan: GPU"
+        command: "julia --color=yes --project=.buildkite experiments/standalone/Soil/richards_comparison.jl"
+        artifact_paths: "experiments/standalone/Soil/gpu/comparison*png"
+        agents:
+          slurm_ntasks: 1
+          slurm_gres: "gpu:1"
+        env:
+          CLIMACOMMS_DEVICE: "CUDA"
 
       - label: "vaira_test"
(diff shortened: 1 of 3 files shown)


experiments/standalone/Soil/richards_comparison.jl
```diff
@@ -100,8 +113,8 @@ clay_datapath, sand_datapath =
         # Check that simulation still has correct float type
         @assert eltype(sol.u[end].soil) == FT
 
-        # Check results and plot for Float64 simulation
-        if FT == Float64
+        # Check results and plot for Float64 simulation on CPU
+        if FT == Float64 && context.device isa ClimaComms.CPUSingleThreaded
```
juliasloan25, line 117: only plot on CPU (fails on GPU)

.buildkite/pipeline.yml
```diff
@@ -84,10 +93,21 @@ steps:
 
       - label: "Richards Runoff CPU"
         command: "julia --color=yes --project=.buildkite experiments/standalone/Soil/richards_runoff.jl"
-        artifact_paths: "experiments/standalone/Soil/artifacts/*png"
+        artifact_paths: "experiments/standalone/Soil/artifacts/cpu/*png"
```
juliasloan25, line 96: I changed the artifact paths to use the architecture type. Even though we aren't producing artifacts on GPU currently, I think this is a useful separation

kmdeck (commented):
looks good, thank you!

One thought - do we want to instead run richards runoff on GPU as part of our performance pipeline?

kmdeck (approved):

juliasloan25:
> One thought - do we want to instead run richards runoff on GPU as part of our performance pipeline?

Good point! Maybe we can add it in both places - so the regular CI run will make sure no PR breaks this change, even if the benchmarks pipeline isn't explicitly run, and the benchmarks pipeline wll let us monitor performance changes

PR #641, opened 2024-06-05
zero flux if zero lai

kmdeck:
## Purpose 
When LAI = SAI = RAI is zero, we still have a nonzero net LW radiation for the canopy, a SHF for the canopy, and nonzero autotrophic respiration.

## To-do
[X] Go over with Alexis and Renato
[X] Update tests

## Content
1. In the canopy parameters, we define canopy emissivity. We now scale that by a factor following CLM as emissiv = emissiv_0*(1-exp(-(LAI+SAI)). This behaves correctly: as LAI+SAI -> 0, emissivity -> 0 and LW absorbed by the canopy -> 0. @braghiere 
2. Autotrophic respiration now uses the stem area index to compute Ns, which is zero if SAI is zero (check with @AlexisRenchon)
3. Autotrophic respiration was before combining GPP with Rd. But, GPP is computed as An scaled by LAI, while Rd is at level of An. I am guessing GPP is canopy level, while An/Rd are leaf level. A possible fix (implemented) is to use An in computing the respiration instead of GPP, and then scaled to canopy at the end. Then we would satisfy that respiration is zero as LAI -> 0. Im not sure if this should be LAI+SAI, though. @AlexisRenchon 
4. SHF change would would be easier to discuss in person @braghiere 
5. we no longer use root area index, also easier to discuss in person @braghiere. 

Review checklist

I have:

----

### DIFF
--- a/experiments/standalone/Vegetation/no_vegetation.jl
+++ b/experiments/standalone/Vegetation/no_vegetation.jl
@@ -1,0 +1,270 @@
+import SciMLBase
+using CairoMakie
+using Statistics
+using Dates
+using Insolation
+
+# Load CliMA Packages and ClimaLand Modules:
+
+using ClimaCore
+import ClimaParams as CP
+import ClimaTimeSteppers as CTS
+using StaticArrays
+using ClimaLand
+using ClimaLand.Domains: Point
+using ClimaLand.Canopy
+using ClimaLand.Canopy.PlantHydraulics
+import ClimaLand
+import ClimaLand.Parameters as LP
+const FT = Float32;
+earth_param_set = LP.LandParameters(FT);
+f_root_to_shoot = FT(3.5)
+SAI = FT(0.0)
+plant_ν = FT(2.46e-4) # kg/m^2
+n_stem = Int64(0)
+n_leaf = Int64(1)
+h_leaf = FT(9.5)
+compartment_midpoints = [h_leaf / 2]
+compartment_surfaces = [FT(0), h_leaf]
+land_domain = Point(; z_sfc = FT(0.0))
+include(
+    joinpath(pkgdir(ClimaLand), "experiments/integrated/fluxnet/data_tools.jl"),
+);
+time_offset = 7
+lat = FT(38.7441) # degree
+long = FT(-92.2000) # degree
+atmos_h = FT(32)
+site_ID = "US-MOz"
+data_link = "https://caltech.box.com/shared/static/7r0ci9pacsnwyo0o9c25mhhcjhsu6d72.csv"
+
+include(
+    joinpath(
+        pkgdir(ClimaLand),
+        "experiments/integrated/fluxnet/met_drivers_FLUXNET.jl",
+    ),
+);
+
+z0_m = FT(2)
+z0_b = FT(0.2)
+
+shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
+    z0_m,
+    z0_b,
+    earth_param_set,
+);
+ψ_soil0 = FT(0.0)
+
+soil_driver = PrescribedSoil(
+    FT;
+    root_depths = SVector{10, FT}(-(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0),
+    ψ = t -> ψ_soil0,
+    α_PAR = FT(0.2),
+    α_NIR = FT(0.4),
+    T = t -> 298.0,
+    ϵ = FT(0.99),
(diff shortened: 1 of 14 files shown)


src/standalone/Vegetation/autotrophic_respiration.jl
```diff
@@ -57,20 +57,23 @@ function compute_autrophic_respiration(
...
-    Nl, Nr, Ns = nitrogen_content(ne, Vcmax25, LAI, RAI, ηsl, h, σl, μr, μs)
+    Nl, Nr, Ns =
+        nitrogen_content(ne, Vcmax25, LAI, SAI, RAI, ηsl, h, σl, μr, μs)
     Rpm = plant_respiration_maintenance(Rd, β, Nl, Nr, Ns, f1)
-    Rg = plant_respiration_growth(f2, GPP, Rpm)
-    Ra = Rpm + Rg # Should this be a function in canopy_parameterizations.jl or is it ok here?
-    return Ra
+    Rg = plant_respiration_growth(f2, An, Rpm)
+    Ra = Rpm + Rg
+    return Ra * (1 - exp(-K * LAI * Ω)) / (K * Ω) # adjust to canopy level
```
kmdeck, line 76: @AlexisRenchon

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -943,23 +944,21 @@ end
...
+    SAI::FT,
     RAI::FT,
     ηsl::FT, # live stem  wood coefficient (kg C m-3) 
     h::FT, # canopy height (m)
     σl::FT, # Specific leaf density (kg C m-2 [leaf])
     μr::FT, # Ratio root nitrogen to top leaf nitrogen (-), typical value 1.0
     μs::FT, # Ratio stem nitrogen to top leaf nitrogen (-), typical value 0.1 
 ) where {FT}
-    Sc = ηsl * h * LAI
+    Sc = ηsl * h * SAI
```
kmdeck, line 961: @AlexisRenchon

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -998,18 +997,14 @@ end
...
 plant maintenance respiration (Rpm), and a relative contribution factor, f.
 """
-function plant_respiration_growth(
-    f2::FT, # Factor of relative contribution, usually 0.25
-    GPP::FT, # Gross primary productivity
-    Rpm::FT, # Plant maintenance respiration
-) where {FT}
-    Rg = f2 * (GPP - Rpm)
+function plant_respiration_growth(f2::FT, An::FT, Rpm::FT) where {FT}
+    Rg = f2 * (An - Rpm)
```
kmdeck, line 1008: @AlexisRenchon

experiments/standalone/Vegetation/no_vegetation.jl
```diff
@@ -0,0 +1,270 @@
+import SciMLBase
```
AlexisRenchon, line 1: Nothing urgent, but I think in the coming months we should consider and have a discussion about how we write those experiments. 

I think it will be neat to use ClimaLandSimulations - it is not great at the moment and might need a refactor, but the idea is have fewer dependencies and fewers lines of code when writing experiments, making it easier to customize simulations. We will also use ClimaDiagnostics and ClimaAnalysis in those experiments, making them more generalizable and user friendly? 

Having said that, I understand that writing everything explicitly has advantages for debugging and also "no black box" which can be useful. 

Nothing to do in this PR, but I think we should talk about it. (I think end users like Mitra/Renato may have a different perspective than software engineer like Gabriele, and having a discussion with everyone's input would be informative)

AlexisRenchon (approved):
The PR looks good to me! 
I just had one comment for a future discussion, but no change request

braghiere:
All looks good.

braghiere (approved):

experiments/standalone/Vegetation/no_vegetation.jl
kmdeck: I agree we should talk about it! definitely we need ClimaLandSimulations+ClimaDiagnostics+Analaysis for end users (especially for large global simulations), but having these explicit scripts for some items may be useful too. 
- maybe it is also a question of being more clear about what our different tasks/uses are, and renaming "experiments" to be more clear about what they are
- maybe the need for "experiments" as we have them now will go down with time, after we add more core features to the models, and ClimaLandSimulations can become more stable

PR #642, opened 2024-06-06
Add information about benchmark run

Sbozzolo:
Run type, timestep, duration.

### DIFF
--- a/experiments/benchmarks/bucket.jl
+++ b/experiments/benchmarks/bucket.jl
@@ -46,7 +46,7 @@ earth_param_set = ClimaLand.Parameters.LandParameters(FT);
 outdir = "bucket_benchmark_$(device_suffix)"
 !ispath(outdir) && mkpath(outdir)
 
-function setup_prob(t0, tf, Δt)
+function setup_prob(t0, tf, Δt; nelements = (100, 10))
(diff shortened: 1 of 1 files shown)


experiments/benchmarks/bucket.jl
```diff
@@ -143,21 +142,29 @@ function setup_prob(t0, tf, Δt)
...
     t0 = 0.0
     tf = 7 * 86400
     Δt = 3600.0
-    prob, cb = setup_prob(t0, tf, Δt)
+    nelements = (100, 10)
+    if greet
+        @info "Run: Bucket with temporal map"
+        @info "Resolution: $nelements"
+        @info "Timestep: $Δt"
+        @info "Duration: $tf"
```
juliasloan25, line 156: It might be better to use `$(tf - t0)` so we aren't assuming `t0 = 0`, and it would also be nice to include that the units for `dt` and `tf` are seconds

juliasloan25 (approved):
Looks good! Thank you for adding the run info printing for the benchmark run

experiments/benchmarks/bucket.jl
Sbozzolo: Added, thanks!

PR #643, opened 2024-06-06
Preallocate bucket auxs

Sbozzolo:
This PR preallocates some of the Fields that are created when computing the bucket tendency.

New time: 3.6s

https://buildkite.com/clima/climaland-benchmark/builds/355#018fef98-5edd-4f39-9cbd-a11a529eba6f

### DIFF
--- a/src/standalone/Bucket/Bucket.jl
+++ b/src/standalone/Bucket/Bucket.jl
@@ -317,11 +317,41 @@ auxiliary_types(::BucketModel{FT}) where {FT} = (
     FT,
     FT,
     FT,
+    FT,
+    FT,
+    NamedTuple{(:F_melt, :F_into_snow, :G_under_snow), Tuple{FT, FT, FT}},
+    FT,
+    FT,
+    FT,
 )
-auxiliary_vars(::BucketModel) =
-    (:q_sfc, :turbulent_fluxes, :R_n, :T_sfc, :α_sfc, :ρ_sfc)
-auxiliary_domain_names(::BucketModel) =
-    (:surface, :surface, :surface, :surface, :surface, :surface, :surface)
+auxiliary_vars(::BucketModel) = (
+    :q_sfc,
+    :turbulent_fluxes,
+    :R_n,
+    :T_sfc,
+    :α_sfc,
+    :ρ_sfc,
+    :snow_cover_fraction,
+    :F_sfc,
+    :partitioned_fluxes,
+    :G,
+    :snow_melt,
+    :infiltration,
+)
+auxiliary_domain_names(::BucketModel) = (
+    :surface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
+    :surface,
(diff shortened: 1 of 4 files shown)


src/standalone/Bucket/Bucket.jl
```diff
@@ -452,6 +440,59 @@ function make_update_aux(model::BucketModel{FT}) where {FT}
...
+        # In this case, there is just one set of surface fluxes to compute.
+        # Since q_sfc itself has already been modified to account for
+        # snow covered regions, and since the surface temperature is
+        # assumed to be the same for snow and underlying land,
+        # the returned fluxes are computed correctly for the cell
+        # regardless of snow-coverage.
+
+        # The below must be adjusted to compute F_sfc over snow and over soil
+        # if we want the snow cover fraction to be intermediate between 0 and 1.
+        @. p.bucket.F_sfc = (
```
kmdeck, line 457: Should we add a test for this? E.g. here
https://github.com/CliMA/ClimaLand.jl/blob/97f06a17e349b8666b6b73d89657b4bdaae03c4a/test/standalone/Bucket/soil_bucket_tests.jl#L216

src/standalone/Bucket/Bucket.jl
```diff
@@ -452,6 +440,59 @@ function make_update_aux(model::BucketModel{FT}) where {FT}
...
+            p.bucket.turbulent_fluxes.shf .+ p.bucket.turbulent_fluxes.lhf +
+            p.bucket.R_n
+        ) # Eqn (21)
+        _T_freeze = LP.T_freeze(model.parameters.earth_param_set)
+        _LH_f0 = LP.LH_f0(model.parameters.earth_param_set)
+        _ρ_liq = LP.ρ_cloud_liq(model.parameters.earth_param_set)
+        _ρLH_f0 = _ρ_liq * _LH_f0 # Latent heat per unit volume.
+        # partition energy fluxes for snow covered area
+        p.bucket.partitioned_fluxes .=
```
kmdeck, line 466: same here, should we test that these are set properly in unit tests?
https://github.com/CliMA/ClimaLand.jl/blob/97f06a17e349b8666b6b73d89657b4bdaae03c4a/test/standalone/Bucket/snow_bucket_tests.jl#L155

Im not sure.

Sbozzolo: Added

src/standalone/Bucket/Bucket.jl
Sbozzolo: Added

Sbozzolo:
> I think this looks fine! I am just wondering if any unit tests need to change to reflect these additions. How are we determining that the behavior doesnt change with this PR?

This is a good question. The PR only moves code around, so it things were previously properly tested, they will stay properly tested. I added unit tests to check that things should be consistent.

> One other comment is that I think at this point we dont need to improve the bucket performance on its own, but should make changes that improve the performance of all models (e.g. allocations in `turbulent_fluxes`, `net_radiation`, etc)

I agree!

PR #646, opened 2024-06-07
absorbed radiation is not zero if LAI is zero

### LINKED ISSUE #645: Roundoff error in two stream scheme
## Describe the bug
We get incorrect absorbed/reflected/transmitted radiation from `plant_absorbed_pfd` for both of our radiation schemes when the zenith angle is pi/2 or the sun is below the horizon, and when LAI is very small. This is because we have things like:
K = 1/cos(theta), where theta = zenith angle
and exp(-K*LAI)

We should be able to remove the LAI clipping done in the global run once we do this.

For now, I clip LAI in the TimeVaryingInput so that if LAI < 0.05, LAI = 0 exactly. Then exp(-K*LAI) =1.
## To Reproduce
see failing test in kd/radiation_roundoff_bug

kmdeck:
## Purpose 
Fixes #645 
We compute transmitted radiation as proportional to exp(-K*LAI*Omega), where Omega is O(1). The absorbed radiation is proportional to (1- exp(-K*LAI*Omega)). 

When the sun is at the horizon, K ~ 1/eps(FT), and no radiation is transmitted (the extinction coefficient K -> infinity).

However, when LAI -> 0, we need to have exp(-K*LAI*Omega) -> 1, so absorbed radiation -> 0. The issue is that if LAI is O(eps(FT)), K*LAI ~O(1), and absorbed radiative fraction is O(1/e).

Solution: 
Im not sure. I dont understand the behavior of K with zenith angle

## To-do
deal with 1/cos(theta) when theta = pi/2
deal with exp(-K*LAI) when K ~ 1/eps(FT) (from 1/cos(theta)) and LAI ~ eps(FT)

## Content
added failing test

----

### DIFF
--- a/ext/CreateParametersExt.jl
+++ b/ext/CreateParametersExt.jl
@@ -539,6 +539,7 @@ end
         α_NIR_leaf = 0.4,
         τ_NIR_leaf = 0.25,
         Ω = 1,
+        LS_c = 0.05,
         n_layers = UInt64(20),
         kwargs...
     )
@@ -549,6 +550,7 @@ end
         α_NIR_leaf = 0.4,
         τ_NIR_leaf = 0.25,
         Ω = 1,
+        LS_c = 0.05,
         n_layers = UInt64(20),
         kwargs...
     )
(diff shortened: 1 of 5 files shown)


kmdeck:
@Espeer5 in case you have any ideas, but I can implement/try and fix

kmdeck:
Im closing this PR because for now we are clipping LAI

PR #648, opened 2024-06-07
add RichardsModel to benchmarks

### LINKED ISSUE #647: Add global GPU RichardsModel to benchmarks pipeline
We want to monitor the performance of the RichardsModel implicit solver (and later on, of the full soil implicit solver), so we should add this run to the performance benchmarks pipeline

Follows off of https://github.com/CliMA/ClimaLand.jl/pull/638

juliasloan25:
# Purpose
closes #647

### Questions
The resolution here is nelements = (101, 15) (copied from richards_runoff), but the other benchmark (bucket) uses nelements = (100, 10). Do we prefer to be consistent with richards_runoff, or with the bucket benchmark?

### DIFF
--- a/experiments/benchmarks/richards.jl
+++ b/experiments/benchmarks/richards.jl
@@ -1,0 +1,378 @@
+# # Global run of RichardsModel
+
+# The code sets up and runs RichardsModel for 7 days on a spherical domain,
+# using prescribed precipitation from ERA5 data. In this simulation, we have
+# turned lateral flow off because horizontal boundary conditions and the
+# land/sea mask are not yet supported by ClimaCore.
+
+# This code runs the model multiple times and collects statistics for execution time and
+# allocations
+#
+# When run with buildkite on clima, this code also compares with the previous best time
+# saved at the bottom of this file
+
+# Simulation Setup
+# This simulation setup is taken from the richards_runoff.jl experiment.
+# Number of spatial elements: 101 in horizontal, 15 in vertical
+# Soil depth: 50 m
+# Simulation duration: 7 days
+# Timestep: 1800 s (30 min)
+# Timestepper: ARS111
+# Maximum iterations: 2
+# Convergence criterion: 1e-8
+# Jacobian update: Every Newton iteration
+# Precipitation data update: every timestep
+
+import SciMLBase
+using Dates
+using Test
+import NCDatasets
+import Interpolations
+
+import ClimaComms
+@static pkgversion(ClimaComms) >= v"0.6" && ClimaComms.@import_required_backends
+import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput
+import ClimaUtilities.SpaceVaryingInputs: SpaceVaryingInput
+import ClimaUtilities.Regridders: InterpolationsRegridder
+import ClimaTimeSteppers as CTS
+using ClimaCore
+using ClimaParams
+
+import ClimaLand
+import ClimaLand.Parameters
+using ClimaLand:
+    initialize,
+    make_update_aux,
+    make_exp_tendency,
+    make_imp_tendency,
+    make_tendency_jacobian,
+    make_set_initial_cache
+
+import Profile, ProfileCanvas
+
+const FT = Float64;
+
+regridder_type = :InterpolationsRegridder
+context = ClimaComms.context()
+device = ClimaComms.device()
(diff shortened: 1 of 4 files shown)


Sbozzolo:
> - The resolution here is `nelements = (101, 15)` (copied from richards_runoff), but the other benchmark (bucket) uses `nelements = (100, 10)`. Do we prefer to be consistent with richards_runoff, or with the bucket benchmark?

It is not important to be consistent with the bucket benchmark. You can increase the bucket benchmark to (101, 15) if you prefer.

experiments/benchmarks/richards.jl
```diff
@@ -0,0 +1,372 @@
...
+        p,
+    )
+    save_every = 100
+    saveat = Array(t0:(save_every * Δt):tf)
+    sv = (;
+        t = Array{Float64}(undef, length(saveat)),
+        saveval = Array{NamedTuple}(undef, length(saveat)),
+    )
+    saving_cb = ClimaLand.NonInterpSavingCallback(sv, saveat)
+    updateat = Array(t0:Δt:tf)
```
kmdeck, line 275: We may update less frequently but I think this is OK for now, unless updating takes a lot of time

experiments/benchmarks/richards.jl
```diff
@@ -0,0 +1,372 @@
+# # Global run of RichardsModel
```
kmdeck, line 1: Could we add this?

"In this simulation, we have turned lateral flow off because horizontal boundary conditions and the land/sea mask are not yet supported by ClimaCore."

kmdeck: Maybe could we also add the pertinient info re: timestepping and discretization? (n elements, n layers, dt, timestepper, max iters,?)

kmdeck: maybe even the tolerance and how often we update the Jacobian

kmdeck (approved):
looks good! This will be very helpful moving forwards, thank you.

Sbozzolo:
How often is data read from file in the benchmark?

kmdeck:
> How often is data read from file in the benchmark?

I think it should be every "updateat" - every dt in this case. that is how often we do the driver callback update which calls the time varying input update

kmdeck:
> > How often is data read from file in the benchmark?
> 
> I think it is every "updateat" - every dt in this case. that is how often we do the driver callback update which calls the time varying input update

(we could get away with less frequent)

Sbozzolo:
Having very frequent is perfect because it allows us to stress-test IO.

experiments/benchmarks/richards.jl
```diff
@@ -0,0 +1,383 @@
...
+Profile.Allocs.@profile sample_rate = 0.1 setup_and_solve_problem()
+results = Profile.Allocs.fetch()
+profile = ProfileCanvas.view_allocs(results)
+alloc_flame_file = joinpath(outdir, "alloc_flame_$device_suffix.html")
+ProfileCanvas.html_file(alloc_flame_file, profile)
+@info "Saved allocation flame to $alloc_flame_file"
+
+if ClimaComms.device() isa ClimaComms.CUDADevice
+    import CUDA
+    CUDA.@profile external = true setup_and_solve_problem()
```
Sbozzolo, line 370:
```suggestion
    CUDA.@profile setup_and_solve_problem()
```

experiments/benchmarks/richards.jl
```diff
@@ -0,0 +1,383 @@
...
+
+if get(ENV, "BUILDKITE_PIPELINE_SLUG", nothing) == "climaland-benchmark"
+    PREVIOUS_BEST_TIME = 13.0
+    if average_timing_s > 1.2PREVIOUS_BEST_TIME
+        @info "Possible performance regression, previous average time was $(PREVIOUS_BEST_TIME)"
+    elseif average_timing_s < 0.8PREVIOUS_BEST_TIME
+        @info "Possible significant performance improvement, please update PREVIOUS_BEST_TIME in $(@__DIR__)"
+    end
+    @testset "Performance" begin
+        @test 0.8PREVIOUS_BEST_TIME <= average_timing_s ≤ 1.1PREVIOUS_BEST_TIME
```
Sbozzolo, line 381:
```suggestion
        @test 0.8PREVIOUS_BEST_TIME <= average_timing_s ≤ 1.2PREVIOUS_BEST_TIME
```

PR #650, opened 2024-06-07
Turn off dss

kmdeck:
## Purpose 
Turn off dss, but preserve the capabilities/code for future use
-> turn off lateral flow
-> comment out some tests of horizontal tendencies in soil

Plots for the global bucket run without DSS - we see some grid imprinting
T_sfc_raw_cpu
σS_raw_cpu
W_raw_cpu

These are made with CC Makie:
image
image
image


test/standalone/Soil/soil_test_3d.jl
```diff
@@ -11,6 +11,7 @@ import ClimaLand
 import ClimaLand.Parameters as LP
 
 for FT in (Float32, Float64)
+    #=
```
kmdeck, line 14: commented out these tests since they compare the output of exp_tendency! with an analytic expression, but exp_tendency! does not do anything if there is no lateral flow.

experiments/standalone/Bucket/global_bucket_staticmap.jl
```diff
@@ -14,7 +14,7 @@
 # between CPU and GPU runs.
 # Plots of the temporal evolution of water content, snow cover fraction,
 # surface temperature, evaporation, and surface energy flux.
-
+ENV["CLIMACOMMS_CONTEXT"] = "SINGLETON"
```
kmdeck, line 17: not sure why I added this, will revert

experiments/standalone/Bucket/global_bucket_staticmap.jl
```diff
@@ -347,5 +349,15 @@ for (i, (field_ts, field_name)) in enumerate(
     end
 
 end
-outfile = joinpath(outdir, string("ts_$device_suffix.png"))
-CairoMakie.save(outfile, fig_ts)
+
+if device_suffix == "cpu"
```
kmdeck, line 353: I made this to see how gridd-y the field are. I think it's useful to keep?

Artifacts.toml
```diff
@@ -1,9 +1,15 @@
+[bareground_albedo]
```
kmdeck, line 1: will restore

experiments/standalone/Bucket/global_bucket_staticmap.jl
Sbozzolo: I would prefer not having `Plots` in the dependency tree, it often leads to problems. There should be a way to do this with Makie, but I don't know it

kmdeck: it's a long term goal to move our plotting to Makie, but I believe we use plots in all the fluxnet runs also. so perhaps a larger issue.

that said, I can use ClimaCoreMakie here instead!

src/shared_utilities/utils.jl
```diff
@@ -86,10 +86,10 @@ Computes the weighted direct stiffness summation and updates `Y` in place.
...
 """
 function dss!(Y::ClimaCore.Fields.FieldVector, p::NamedTuple, t)
-    for key in propertynames(Y)
-        property = getproperty(Y, key)
-        dss_helper!(property, axes(property), p)
-    end
+    #for key in propertynames(Y)
+    #    property = getproperty(Y, key)
+    #    dss_helper!(property, axes(property), p)
+    #end
```
szy21, line 92: Do we want to remove this, or just comment it out?

szy21 (approved):

src/shared_utilities/utils.jl
kmdeck: I *think* just comment it out, since we want to come back to lateral flow at some point...and maybe if this PR causes issues, we want to be able to turn it back on? but also this stuff is preserved in git history.

kmdeck:
@Sbozzolo Im not sure if this is "grid imprinting" or just coarse resolution images. Either way, I think this is ready to merge after rebasing, unless we want to try a very long coupler run using this clima land branch? I had already triggered a regular Coupler long run and Zhaoyi thought it looked fine.

src/standalone/Soil/rre.jl
```diff
@@ -92,6 +92,7 @@ function RichardsModel{FT}(;
         subsurface_source = subsurface_runoff_source(top_bc.runoff)
         sources = append_source(subsurface_source, sources)
     end
+    @assert ~lateral_flow
```
Sbozzolo, line 95:
```suggestion
    @assert !lateral_flow
```

`!` is for logical operations, `~` is for bitwise operations

experiments/Manifest.toml
Sbozzolo, line 1: I think this Manifest should not be included

Sbozzolo (approved):

experiments/Manifest.toml
kmdeck: removed entirely

PR #651, opened 2024-06-10
initial full soil imp solver

juliasloan25:
## Purpose 
Adds a simple implicit tendency for the `EnergyHydrology` model, which only treats the liquid water content implicitly (using the functions already defined for `RichardsModel`)

### Changes in src
- all prognostic variables must have an entry `(@name(var), @name(var)) => block` in the Jacobian. The block should be the negative identity matrix for variables that are stepped explicitly, and a tridiagonal matrix for variables that are stepped implicitly
- in the explicit tendency, we need to reset `dY.model .= 0` for implicitly-stepped variables 
- in the implicit tendency, we need to reset `dY.model .= 0` for explicitly-stepped variables 
### Changes in driver
- in addition to `exp_tendency!`, create `imp_tendency!`, `tendency_jacobian!`, and `jac_kwargs`
- change the timestepper from an explicit to an imex method (e.g. RK4 -> ARS111 or ARS343)
- change the type of the ODE algorithm from `CTS.ExplicitAlgorithm` to `CTS.IMEXAlgorithm`, and add the `NewtonsMethod` argument
- when constructing the problem, pass an ODE function for both the explicit and implicit tendencies (e.g. add `SciMLBase.ODEFunction(imp_tendency!; jac_kwargs...)` to `CTS.ClimaODEFunction` call)

## Next steps (separate PRs)
- [ ] add implicit tendency for energy `dT/dI`
- [ ] include boundary condition contributions
- [ ] include contribution from `drhoe_int / dtheta_l`

## Content
- [x] move jacobian struct and related implicit timestepping functions into `src/Soil`
- [x] add `imp_tendency` function for `EnergyHydrology`
- [x] update all instances of `EnergyHydrology` to use implicit tendency (docs, tests)
- [x] use implicit solver with `SoilCanopyModel`
  - add jacobian terms for canopy, soilco2 prognostic variables in `ImplicitEquationJacobian`
  - set implicit tendency for canopy, soilco2 prognostic variables to 0 in `imp_tendency!` (replace `do_nothing_imp_tendency` ? --> update `compute_imp_tendency` (and `exp_tendency` defaults to set `dY .= 0`)
  - add `make_jac_tendency` function for `AbstractLandModel` - calling functions of component models, analogous to ALM `update_aux`
- [x] add plots from docs

### docs
- [x] soil/canopy tutorial
ozark_canopy_flux_test
ozark_soil_plant_flux
ozark_soil_test

- [x] coarse sand evaporation
evaporation_lehmann2008_fig8b

- [x] Gilat Loess evaporation
evaporation_lehmann2024_figS6
evaporation_gardner_fig1

- [x] freezing front

- [x] energy hydrology

- [x] sublimation
water fluxes
profiles

### experiments (visually inspected)
- [x] vaira
  - plots (ET, SHF, temperature) look smoother now! (probably from timestepper, dt changes)
  - slower because dt decreased by factor of 2
- [x] ozark
  - plots are nearly identical
- [x] niwot
  - plots are nearly identical
- [x] harvard
  - plots look identical
- [x] ozark PFT
  - plots are nearly identical
- [ ] ozark conservation
  - energy conservation fractional error shows same pattern, but larger by ~10^13 orders of magnitude
  - water conservation fractional error larger by ~10^13 orders of magnitude
- [x] soil/canopy lsm performance
  - solve takes longer now

----


kmdeck:
for the freezing front test, maybe we can try a higher order IMEX stepper? E.g. ARS343

It may be that using explicit euler as the explicit stepper is not sufficient or requires a very small step, while when we were using RK4 we could use a bigger step.

juliasloan25:
> for the freezing front test, maybe we can try a higher order IMEX stepper? E.g. ARS343
> 
> It may be that using explicit euler as the explicit stepper is not sufficient or requires a very small step, while when we were using RK4 we could use a bigger step.

Even using ARS343 with a timestep as small as 0.1s (was 60s when explicit), all prognostic variables are NaN at the second timestep. Maybe we do need to add more terms to our jacobian approximation?

kmdeck:
> > for the freezing front test, maybe we can try a higher order IMEX stepper? E.g. ARS343
> > It may be that using explicit euler as the explicit stepper is not sufficient or requires a very small step, while when we were using RK4 we could use a bigger step.
> 
> Even using ARS343 with a timestep as small as 0.1s (was 60s when explicit), all prognostic variables are NaN at the second timestep. Maybe we do need to add more terms to our jacobian approximation?

Let's think about this a bit more. It seems odd that treating RHS terms all explicitly is OK, but moving one to be treated implicitly (which should only making things more stable) and leaving the others explicit causes errors. Do you want to meet this afternoon to try some debugging?

juliasloan25:
> > > for the freezing front test, maybe we can try a higher order IMEX stepper? E.g. ARS343
> > > It may be that using explicit euler as the explicit stepper is not sufficient or requires a very small step, while when we were using RK4 we could use a bigger step.
> > 
> > 
> > Even using ARS343 with a timestep as small as 0.1s (was 60s when explicit), all prognostic variables are NaN at the second timestep. Maybe we do need to add more terms to our jacobian approximation?
> 
> Let's think about this a bit more. It seems odd that treating RHS terms all explicitly is OK, but moving one to be treated implicitly (which should only making things more stable) and leaving the others explicit causes errors. Do you want to meet this afternoon to try some debugging?

Yeah, good point. Want to meet at 1pm?

kmdeck:
> Yeah, good point. Want to meet at 1pm?

sure!

experiments/integrated/performance/conservation/ozark_conservation_setup.jl
```diff
@@ -192,8 +192,15 @@ land = SoilCanopyModel{FT}(;
     canopy_model_args = canopy_model_args,
 )
 exp_tendency! = make_exp_tendency(land)
```
juliasloan25, line 194: Interestingly, this file was already using ARS111 and an IMEXAlgorithm (lines 18-32 of this file)

kmdeck: oh! that's totally it. We were using explicit euler before, but the only option for that is IMEX ars111 (with imp_tendency! doing nothing). there is no pure explicit euler option in ClimaTimestppers. this matters just because the expected energy and water changes we compute are computed assuming explicit euler timesteps. 

like, if Var(t+dt)= Var(t) + function(fluxes(t)) * dt, we can check conservation by comparing Var(t+dt) - Var(t) = function(fluxes(t)) * dt (but over many steps - and just expect roundoff error)

but if the timestepper changes, this is not the case. if it was implicit, we could do
Var(t+dt) - Var(t) = function(fluxes(t+dt)) * dt

basically im not really sure how to compute the "truth" for higher order imex schemes.

kmdeck: maybe we can ask Dennis. We mostly want to verify that d(energy in land) = energy in - energy out over the whole sim. 

I think the important thing is that this is not growing. so I am inclined to think this is OK, we just are computing what we want

juliasloan25: Ohh I see! Would it be possible to use the equation you gave for the explicit case for explicitly-stepped variables, and the implicit one for implicitly-stepped variables? Or does that not make sense because we have to consider both variables and fluxes?

lib/ClimaLandSimulations/src/utilities/make_timestepper.jl
```diff
@@ -4,20 +4,33 @@ export make_timestepper
...
+        ),
+    )
 
-Define the setup for the simulation timestepper. 
-the default timestepper is 4th order Runge Kutta method,
- other timesteppers from ClimaTimeSteppers.jl can be used. 
+Define the setup for the simulation timestepper.
+The default timestepper (ARS343) is an IMEX ARK algorithm with
+3 implicit stages, 4 implicit stages, and 3rd order accuracy.
+Other timesteppers from ClimaTimeSteppers.jl can be used.
```
kmdeck, line 20: maybe other IMEX

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -230,6 +221,136 @@ function ClimaLand.make_compute_exp_tendency(
...
+
+Using this Jacobian with a backwards Euler timestepper is equivalent
+to using the modified Picard scheme of Celia et al. (1990).
+"""
+function ClimaLand.make_update_jacobian(model::EnergyHydrology{FT}) where {FT}
+    function update_jacobian!(jacobian::ImplicitEquationJacobian, Y, p, dtγ, t)
+        (; matrix) = jacobian
+        (; ν, hydrology_cm, S_s, θ_r) = model.parameters
+
+        # Create divergence operator
```
kmdeck, line 281: Maybe we can open an issue to note partial code duplication between this and rre make_update_jacobian?

src/shared_utilities/models.jl
```diff
@@ -246,11 +249,15 @@ end
...
 Return a `compute_imp_tendency!` function that updates state variables
 that we will be stepped implicitly.
+This fallback sets all tendencies of this model to zero, which is appropriate
+for models that do not have any implicit tendencies to update.
 
 `compute_imp_tendency!` should be compatible with SciMLBase.jl solvers.
 """
 function make_compute_imp_tendency(model::AbstractModel)
-    function compute_imp_tendency!(dY, Y, p, t) end
+    function compute_imp_tendency!(dY, Y, p, t)
```
kmdeck, line 258: perhaps we can leave a comment about why we dont set dY .= 0?

src/shared_utilities/implicit_timestepping.jl
```diff
@@ -0,0 +1,166 @@
...
+variables `p` in place and then updates the entries of the
+Jacobian matrix `W` for the `model` in place.
+
+The default is that no updates are required, no implicit tendency is
+present, and hence the timestepping is entirely explicit.
+
+Note that the returned function `tendency_jacobian!` should be
+used as `Wfact!` in `ClimaTimeSteppers.jl` and `SciMLBase.jl`.
+"""
+function make_tendency_jacobian(model::AbstractModel)
```
kmdeck, line 24: Maybe we can open an issue to unify the terminology between `make_update_jacobian` and `make_compute_imp_tendency`, and between `make_tendency_jacobian` and `make_imp_tendency`? I think they serve the same roles but it might be clearer if we name them the same. What do you think?

kmdeck: (feel free to assign some of these issues to me!)

src/shared_utilities/implicit_timestepping.jl
```diff
@@ -0,0 +1,166 @@
...
+`T!(dY, Y, p, t) = make_implicit_tendency(model)`, the Jacobian
+should be given by `W_{i,j}! = ∂T!_i/∂Y_j`, where `Y_j` is the
+`j-th` state variable
+and `T!_i` is the implicit tendency of the `i-th` state variable.
+
+The default is that no updates are required, no implicit tendency is
+present, and hence the timestepping is entirely explicit.
+"""
+function make_update_jacobian(model::AbstractModel)
+    function update_jacobian!(W, Y, p, dtγ, t) end
```
kmdeck, line 52: Should this do nothing? or should it return minus identity matrix? 

or do we never rely on this default (and never will)? if so, maybe we can note that this is only a stub and must be extended or ? not sure.

src/shared_utilities/implicit_timestepping.jl
```diff
@@ -0,0 +1,166 @@
...
+"""
+    ImplicitEquationJacobian{M, S}
+
+A struct containing the necessary information for constructing a tridiagonal
+Jacobian matrix (`W`) for solving Richards equation, treating only the vertical
+diffusion term implicitly.
+
+`matrix` is a block matrix containing the tri-diagonal matrix `∂ϑres∂ϑ`
+in the RichardsModel case.
+`solver` is a diagonal solver for the RichardsModel case because our matrix is
```
kmdeck, line 82: can update docstring

src/shared_utilities/implicit_timestepping.jl
```diff
@@ -0,0 +1,166 @@
...
+block diagonal.
+
+Note that the diagonal, upper diagonal, and lower diagonal entry values
+are stored in this struct and updated in place.
+"""
+struct ImplicitEquationJacobian{M, S}
+    "Jacobian matrix stored as a MatrixFields.FieldMatrix"
+    matrix::M
+    "Solver to use for solving the tridiagonal system"
```
kmdeck, line 91: maybe just "system"?

src/shared_utilities/implicit_timestepping.jl
```diff
@@ -0,0 +1,166 @@
...
+
+"""
+    ImplicitEquationJacobian(
+        Y::ClimaCore.Fields.FieldVector;
+)
+
+Outer constructor for the ImplicitEquationJacobian Jacobian
+matrix struct.
+
+Initializes all variables to zeros.
```
kmdeck, line 103: to NaN?

src/shared_utilities/implicit_timestepping.jl
```diff
@@ -0,0 +1,166 @@
...
+    tridiag_type = MatrixFields.TridiagonalMatrixRow{FT}
+    # Create a field containing a `TridiagonalMatrixRow` at each point
+    tridiag_field = Fields.Field(tridiag_type, center_space)
+    fill!(parent(tridiag_field), NaN)
+
+    # Only add jacobian blocks for fields that are in Y
+    is_in_Y(name) = MatrixFields.has_field(Y, name)
+
+    implicit_names = (@name(soil.ϑ_l),)
+    explicit_names = (
```
kmdeck, line 119: we mentioned opening an issue to generalize this.

kmdeck (approved):
I think this looks great as a starting point! Mostly comments about keeping track of our to-do list afterwards

experiments/integrated/performance/conservation/ozark_conservation_setup.jl
kmdeck: im not sure! maybe we can discuss more tomorrow? Im not worried about this really for this PR though... we maybe need to rethink that conversation test in general

src/standalone/Soil/energy_hydrology.jl
juliasloan25: good point! I opened an issue here https://github.com/CliMA/ClimaLand.jl/issues/659

src/shared_utilities/implicit_timestepping.jl
juliasloan25: Good point! I was getting confused between these while working on this PR....

juliasloan25: I opened an issue here https://github.com/CliMA/ClimaLand.jl/issues/660

src/shared_utilities/implicit_timestepping.jl
juliasloan25: I don't think it needs to return -I. I know that we had previously set the jacobian blocks for explicitly-stepped variables to -I in `update_jacobian!`, but I realized that we just can initialize these as -I in our `ImplicitEquationJacobian` constructor. Then we just do nothing for these blocks in `update_jacobian!`.

And now that I'm writing this out, I notice that `update_jacobian!` is now identical for `RichardsModel` and `EnergyHydrology` with the jacobian approximation we're using right now. I can move it into `src/Soil` and change the type dispatch to use one method for both, or we could just keep it split up since we'll be changing the `EnergyHydrology` version as we iterate on the jacobian anyway. I'm fine with either.

I'll add a note to the generic `update_jacobian!` to explain that it needs to be extended

src/shared_utilities/implicit_timestepping.jl
juliasloan25: opened here: https://github.com/CliMA/ClimaLand.jl/issues/661

experiments/integrated/performance/conservation/ozark_conservation_setup.jl
juliasloan25: happy to discuss tomorrow! I opened an issue for this too so we can keep track of it https://github.com/CliMA/ClimaLand.jl/issues/662

PR #654, opened 2024-06-12
Bug fix for atmos type

kmdeck:
## Purpose 
if the land model has an atmosphere driver, we check if the params stored in the atmos type agree with the params of the land model itself. The coupled atmos driver does not have these params, so coupled runs break. this updates the check to only be carried out in the prescribed atmos case.

### DIFF
--- a/src/shared_utilities/models.jl
+++ b/src/shared_utilities/models.jl
@@ -388,7 +388,7 @@ function add_drivers_to_cache(p::NamedTuple, model::AbstractModel, coords)
     (atmos, radiation) = get_drivers(model)
     if hasproperty(model, :parameters) &&
        hasproperty(model.parameters, :earth_param_set) &&
-       !isnothing(atmos)
+       atmos isa ClimaLand.PrescribedAtmosphere
         if LP.thermodynamic_parameters(model.parameters.earth_param_set) !=
            atmos.thermo_params
             error(


kmdeck:
@AlexisRenchon I think this should go in before the release, but I am making sure now that this fixes the bug in coupled runs

src/shared_utilities/models.jl
```diff
@@ -388,7 +388,7 @@ function add_drivers_to_cache(p::NamedTuple, model::AbstractModel, coords)
     (atmos, radiation) = get_drivers(model)
     if hasproperty(model, :parameters) &&
        hasproperty(model.parameters, :earth_param_set) &&
-       !isnothing(atmos)
+       typeof(atmos) == ClimaLand.PrescribedAtmosphere
```
Sbozzolo, line 391:
```suggestion
       atmos isa ClimaLand.PrescribedAtmosphere
```
is a little more Julian.

Technically, the type of `atmos` is not `PrescribedAtmosphere`, but
```
PrescribedAtmosphere{
    FT,
    LP <: Union{Nothing, AbstractTimeVaryingInput},
    SP <: Union{Nothing, AbstractTimeVaryingInput},
    TA <: Union{Nothing, AbstractTimeVaryingInput},
    UA <: Union{Nothing, AbstractTimeVaryingInput},
    QA <: Union{Nothing, AbstractTimeVaryingInput},
    RA <: Union{Nothing, AbstractTimeVaryingInput},
    CA <: Union{Nothing, AbstractTimeVaryingInput},
    DT,
    TP,
}
``` 
(with all the parametric types)

Sbozzolo (approved):

PR #658, opened 2024-06-13
Store z etc in domain

kmdeck:
## Purpose 
Every step of the soil model, we create the field of `z` and delta z_top, bottom. This stores them in the cache, along with z_sfc.

There is a lot of other improvements to be made to the boundary condition code in terms of allocations, but this was low hanging fruit.

I also allocated and stored the volumetric heat capacity since we use that in a number of places.

## To-do
CI


src/standalone/Soil/energy_hydrology.jl
```diff
@@ -555,13 +556,31 @@ function ClimaLand.source!(
...
     _ρ_l = FT(LP.ρ_cloud_liq(earth_param_set))
     _ρ_i = FT(LP.ρ_cloud_ice(earth_param_set))
-    ρc = volumetric_heat_capacity.(p.soil.θ_l, Y.soil.θ_i, params)
-    τ = thermal_time.(ρc, src.Δz, p.soil.κ)
-
-    liquid_source =
-        phase_change_source.(p.soil.θ_l, Y.soil.θ_i, p.soil.T, τ, params)
-    @. dY.soil.ϑ_l += -liquid_source
-    @. dY.soil.θ_i += (_ρ_l / _ρ_i) * liquid_source
+    Δz_top = model.domain.fields.Δz_top # center face distance
```
kmdeck, line 559: some rearrangement here to reduce allocations also - basically, putting it all on one line

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -379,11 +381,12 @@ function horizontal_components!(
     hgrad = Operators.Gradient()
     # The flux is already covariant, from hgrad, so no need to convert.
     @. dY.soil.ϑ_l += -hdiv(-p.soil.K * hgrad(p.soil.ψ + z))
-    ρe_int_l = volumetric_internal_energy_liq.(p.soil.T, model.parameters)
     @. dY.soil.ρe_int +=
         -hdiv(
             -p.soil.κ * hgrad(p.soil.T) -
-            p.soil.K * ρe_int_l * hgrad(p.soil.ψ + z),
+            p.soil.K *
+            volumetric_internal_energy_liq(p.soil.T, model.parameters) *
```
kmdeck, line 388: rearrangement to reduce allocating intermediate quantities

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -175,7 +176,8 @@ function ClimaLand.make_compute_exp_tendency(
     model::EnergyHydrology{FT},
 ) where {FT}
     function compute_exp_tendency!(dY, Y, p, t)
-        z = ClimaCore.Fields.coordinate_field(model.domain.space.subsurface).z
+        z = model.domain.fields.z
+        rre_top_flux_bc = p.soil.top_bc.water
```
kmdeck, line 180: Ill remove this line, we dont need it here

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -131,8 +131,9 @@ end
 
 function make_update_boundary_fluxes(model::EnergyHydrology)
     function update_boundary_fluxes!(p, Y, t)
-        z = ClimaCore.Fields.coordinate_field(model.domain.space.subsurface).z
-        Δz_top, Δz_bottom = ClimaLand.Domains.get_Δz(z)
+        z = model.domain.fields.z
```
kmdeck, line 134: we may not need z here, double check

src/shared_utilities/Domains.jl
```diff
@@ -357,6 +372,10 @@ function HybridBox(;
 
     surface_space = obtain_surface_space(subsurface_space)
     space = (; surface = surface_space, subsurface = subsurface_space)
+    z = ClimaCore.Fields.coordinate_field(subsurface_space).z
+    Δz_top, Δz_bottom = get_Δz(z)
+    z_sfc = get_z_sfc(z)
+    fields = (; z = z, Δz_top = Δz_top, Δz_bottom = Δz_bottom, z_sfc = z_sfc)
```
juliasloan25, line 378: Since we do the exact same thing for each domain type, it might be nice to pull this out into a helper function, e.g. `get_domain_fields(subsurface_space)`. Then it will be much easier if we want to change the fields we store here

src/shared_utilities/Domains.jl
```diff
@@ -721,6 +746,27 @@ function get_Δz(z::ClimaCore.Fields.Field)
     return Δz_top ./ 2, Δz_bottom ./ 2
 end
 
+"""
+    get_z_sfc(z::ClimaCore.Fields.Field)
+
+A function to return the coordinates of the surface `z`, projected onto the one-/two-dimensional surface space.
+This space is distinct from the level space of `z`.
```
juliasloan25, line 753: It would be good to note that this function is non-allocating, since it constructs a Field of field_values and a space that both already exist

src/shared_utilities/Domains.jl
```diff
@@ -721,6 +746,27 @@ function get_Δz(z::ClimaCore.Fields.Field)
...
+"""
+function get_z_sfc(z::ClimaCore.Fields.Field)
```
juliasloan25, line 755: Could we rename this to a more general `top_face_to_surface`, analogous to `top_center_to_surface`? It seems like the only part specific to `z` here is `z_face = ClimaCore.Fields.coordinate_field(fs).z`, so maybe we could keep this function and just have it pass that `z_face` to the generalized `top_face_to_surface` function?

src/standalone/Soil/energy_hydrology.jl
juliasloan25: I think we can get rid of it!

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -555,13 +556,31 @@ function ClimaLand.source!(
...
-    @. dY.soil.θ_i += (_ρ_l / _ρ_i) * liquid_source
+    Δz_top = model.domain.fields.Δz_top # center face distance
+    @. dY.soil.ϑ_l +=
+        -phase_change_source(
+            p.soil.θ_l,
+            Y.soil.θ_i,
+            p.soil.T,
+            thermal_time(
+                volumetric_heat_capacity(p.soil.θ_l, Y.soil.θ_i, params),
+                2 * Δz_top,
```
juliasloan25, line 567: maybe add a short comment to explain the factor of 2 here

juliasloan25: (and in the θ_i equation)

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -772,20 +795,7 @@ end
...
-        ClimaCore.Fields.field_values(
-            ClimaCore.Fields.level(
-                ClimaCore.Fields.coordinate_field(face_space).z,
-                ClimaCore.Utilities.PlusHalf(N - 1),
-            ),
-        ),
-        surface_space,
-    )
-    return z_sfc
+    return model.domain.fields.z_sfc
```
juliasloan25, line 798: awesome!!

test/standalone/Soil/climate_drivers.jl
```diff
@@ -115,6 +115,7 @@ for FT in (Float32, Float64)
             )
 
             Y, p, coords = initialize(model)
+            Δz_top, _ = ClimaLand.Domains.get_Δz(coords.subsurface.z)
```
juliasloan25, line 118: should we instead get `Δz_top` from the domain here to test the value we're storing, since that's how we get `Δz_top` in the code now?

juliasloan25 (approved):
This looks great, thank you!! I left some minor comments, but after those I think it's good to merge

src/standalone/Soil/energy_hydrology.jl
kmdeck: will do!

src/shared_utilities/Domains.jl
kmdeck: I think it allocates z_face, but other than that is non-allocating. I left that comment!

src/shared_utilities/Domains.jl
kmdeck: This is a good idea! But I realized there is an easier way to get z_sfc, and since we dont need the `top_face_to_surface` function otherwise, I didnt add it for now. currently we just do:
    `z_sfc = ClimaCore.Fields.coordinate_field(surface_space).z`

src/shared_utilities/Domains.jl
kmdeck: great idea! I did this

src/shared_utilities/Domains.jl
kmdeck: edit- that didnt work. so I added in top_face_to_surface

PR #666, opened 2024-06-17
Reduce allocations in the canopy update aux

kmdeck:
## Purpose 
Reduces allocation in Canopy `update_aux`. Small improvement in time to solution (13%), but most of the solution time is now in updating the drivers, so this is potentially a larger improvement in RHS functions.

## Content
1. Instead of computing esat, eair, and then relative humidity and allocating each, I wrote a function to compute the relative humidity, and we now broadcast that over existing fields when updating `frac_diff` in place.
2. allocate `frac_diff` in cache and update in place.
3. refactor `update_photosynthesis`
4. refactor `compute_PAR`, `compute_NIR`
5. refactor compute_absorbances (change structure of canopy cache)

----


AlexisRenchon (approved):
Nice, the code looks cleaner like that! Thanks! 
Maybe add a line in NEWS.md? 
I don't think there is any breaking change, right? (all internal)

AlexisRenchon:
What labels would be good for this PR? 
`performance` or `code clarity`?

kmdeck:
> What labels would be good for this PR? `performance` or `code clarity`?

I added a "performance" label for now :) and will add in the NEWS md before merging. Thanks!

PR #675, opened 2024-06-24
Add implicit tendency for canopy temperature

kmdeck:
## Purpose 
Add implicit tendency for canopy temperature
Address #696 

Previously, all canopy variables were stepped explicitly. Now temperature
is stepped implicitly in both standalone CanopyModel runs and integrated
SoilCanopyModel runs. Changes include adding a jacobian and implicit 
tendency for the CanopyModel, and updating scripts to use ImEx steppers.
We've also added scripts to test convergence behavior of the standalone
and integrated cases, which are output as plots on buildkite.

This change increases the stability of our global
SoilCanopyModel runs, decreasing the number of NaNs we see, and allows
us to take timesteps of 450s.

### Notes
There are two things we dont understand about this change:
- there seem to be some times during the simultation where the error is always large, and a smaller dt doesnt help. For example, the 99th percentile error curve dips below the mean error for small dt, which indicates there are outliers that are very large in the top 1% error
  - update 9/18: it's possible this may happen during times in the simulation where temperature is changing a lot, e.g. shortly after a driver update, as opposed to when the temperature is in equilibrium. See plots in comments for more information
- the derivative of SHF with respect to temperature is correct to a few percent, but the derivative of LHF with respect to temeperature is ~30% wrong. This is partly due to d_qsat/d_T errors (but Ive double checked that I entered the formula correct), but seems like it is also due to d_LHF/d_qsat. I think I have followed CLM exactly. Not sure why the error is large.

These two questions will be investigated in a later PR - see https://github.com/CliMA/ClimaLand.jl/issues/782

## Results
With an explicit stepper, RK4, a test script at the Ozark site with the canopy alone permits dt = 60 [ac_canopy = 1e3]. Time to solution = 114 seconds - see kd/canopy_explicit_tests branch. Using dt = 180 fails.
With an implicit stepper ARS111, updated jacobian every newton iteration, max_iters = 6, the same simulation can be run with dt=3600. Time to solution = 16 seconds. Using only a single iteration failed.

Accuracy as a function of timestep (ref solution = dt = 6, implicit solver)
Float 32:
errors_f32
Float64:
errors_f64

In the global long run, on this branch: [kd/canopy_explicit_global_nans](https://buildkite.com/clima/climaland-long-runs/builds?branch=kd%2Fcanopy_explicit_global_nans) I just changed the value of ac_canopy but left everything else the same as #main - i.e. using an explicit solver. The resulting map is almost entirely NaN [ look at time 86400]

On the implicit branch (this PR), we have no NaNs after one day. [ look at time 86400]

## To-do
- [X] Merge ClimaCore Branch
- [X] Test timestep improvement
- [X] Unit test Jacobian
- [x] Cleanup canopy timestep_test.jl
- [x] Add script to show soil/canopy convergence (analogous to canopy timestep_test.jl) - add to buildkite pipeline
- [x] Rebase
- [x] Check longruns
  - requires https://github.com/CliMA/ClimaCore.jl/pull/2016 for GPU compatibility
- [x] Review

## Content
- Adds canopy temp to the list of implicitly stepped variables
- Adds the canopy jacobian
- Adds the canopy implicit tendency
- updates scripts and some timesteps
- adds a script testing timestep convergence, adds to buildkite
- update docs Manifest to ClimaCore 0.14.11

----

### DIFF
--- a/experiments/integrated/fluxnet/fluxnet_simulation.jl
+++ b/experiments/integrated/fluxnet/fluxnet_simulation.jl
@@ -6,14 +6,15 @@ N_spinup_days = 30
 N_days = N_spinup_days + 30
 tf = Float64(t0 + 3600 * 24 * N_days)
 t_spinup = Float64(t0 + N_spinup_days * 3600 * 24)
-saveat = Array(t_spinup:(n * dt):tf)
+dt_save = 3600.0
+saveat = Array(t_spinup:dt_save:tf)
 
 # Set up timestepper
-timestepper = CTS.ARS343();
+timestepper = CTS.ARS111();
 ode_algo = CTS.IMEXAlgorithm(
     timestepper,
     CTS.NewtonsMethod(
-        max_iters = 1,
+        max_iters = 6,
         update_j = CTS.UpdateEvery(CTS.NewNewtonIteration),
     ),
 );
--- a/experiments/long_runs/land.jl
+++ b/experiments/long_runs/land.jl
@@ -662,7 +662,7 @@ setup_and_solve_problem(; greet = true);
 # read in diagnostics and make some plots!
 #### ClimaAnalysis ####
 simdir = ClimaAnalysis.SimDir(outdir)
-short_names = ["gpp", "ct", "lai", "swc", "si"]
+short_names = ["gpp", "ct", "lai", "swc", "si", "swo", "lwo", "et", "er", "sr"]
 for short_name in short_names
     var = get(simdir; short_name)
     times = ClimaAnalysis.times(var)
--- a/experiments/long_runs/land.jl
+++ b/experiments/long_runs/land.jl
@@ -9,10 +9,10 @@
 # Number of spatial elements: 101 in horizontal, 15 in vertical
 # Soil depth: 50 m
 # Simulation duration: 365 d
-# Timestep: 900 s
-# Timestepper: ARS343
-# Fixed number of iterations: 1
-# Jacobian update: every new timestep
+# Timestep: 450 s
+# Timestepper: ARS111
+# Fixed number of iterations: 3
+# Jacobian update: every new Newton iteration
 # Atmos forcing update: every 3 hours
 import SciMLBase
 import ClimaComms
@@ -363,7 +363,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
     τ_NIR_leaf = FT(0.25)
 
     # Energy Balance model
-    ac_canopy = FT(2.5e4) # this will likely be 10x smaller!
+    ac_canopy = FT(2.5e3)
 
     #clm_data is used for g1 and vcmax maps
     clm_artifact_path = ClimaLand.Artifacts.clm_data_folder_path(; context)
@@ -634,7 +634,7 @@ function setup_and_solve_problem(; greet = false)
 
     t0 = 0.0
     tf = 60 * 60.0 * 24 * 365
-    Δt = 900.0
+    Δt = 450.0
     nelements = (101, 15)
     if greet
         @info "Run: Global Soil-Canopy Model"
--- a/experiments/standalone/Vegetation/timestep_test.jl
+++ b/experiments/standalone/Vegetation/timestep_test.jl
@@ -1,0 +1,332 @@
+# # Timestep test for standalone CanopyModel
+
+# This experiment runs the standalone canopy model for 20 days at a single point
+# and outputs useful plots related to the timestepping and stability of
+# the simulation using different timesteps. This was initially implemented to
+# test the behavior of stepping canopy temperature (T) implicitly.
+
+# The simulation is run multiple times: first, with a small reference timestep,
+# then with increasing timestep sizes. The goal here is to see which timesteps
+# are small enough to produce a stable simulation, and which timesteps are too
+# large to produce accurate results.
+
+# The runs with larger timesteps are compared with the reference timestep run
+# using multiple metrics, including:
+# - mean, 99%th, and 95th% error over the course of the simulation
+# - T at the end of the simulation (used to test convergence with each dt)
+# - T throughout the entire simulation
+
+# Note that the simulation is run for 20 days + 80 seconds. This somewhat
+# unusual length is necessary to test the convergence behavior of the solver.
+# Convergence behavior must be assessed when T is actively changing, rather
+# than when it is in equilibrium. T changes in the time period after a driver
+# update, which occurs every 3 hours (so exactly at 20 days). Running for
+# 20 days + 80 seconds allows us to look at results over a longer simulation,
+# as well as convergence behavior. See the discussion in
+# https://github.com/CliMA/ClimaLand.jl/pull/675 for more information.
+
+# Simulation Setup
+# Space: single point
+# Simulation duration: 20 days + 80 seconds (= 480.02 hours)
+# Timestep: variable, ranging from 0.6s to 3600s
+# Timestepper: ARS111
+# Fixed number of iterations: 6
+# Jacobian update: Every Newton iteration
+# Atmospheric and radiation driver updates: Every 3 hours
+
+import SciMLBase
+import ClimaComms
+@static pkgversion(ClimaComms) >= v"0.6" && ClimaComms.@import_required_backends
+using CairoMakie
+using Statistics
+import StatsBase: percentile
+using Dates
+using Insolation
+using StaticArrays
+
+# Load CliMA Packages and ClimaLand Modules:
+
+using ClimaCore
+import ClimaParams as CP
+import ClimaTimeSteppers as CTS
+using ClimaLand
+using ClimaLand.Domains: Point
+using ClimaLand.Canopy
+using ClimaLand.Canopy.PlantHydraulics
+import ClimaLand
+import ClimaLand.Parameters as LP
+const FT = Float64;
+earth_param_set = LP.LandParameters(FT);
+f_root_to_shoot = FT(3.5)
+plant_ν = FT(2.46e-4) # kg/m^2
+n_stem = Int64(1)
+n_leaf = Int64(1)
+h_leaf = FT(9.5)
+h_stem = FT(9)
+compartment_midpoints = [h_stem / 2, h_stem + h_leaf / 2]
+compartment_surfaces = [FT(0), h_stem, h_stem + h_leaf]
+land_domain = Point(; z_sfc = FT(0.0))
+include(
+    joinpath(pkgdir(ClimaLand), "experiments/integrated/fluxnet/data_tools.jl"),
+);
+time_offset = 7
+lat = FT(38.7441) # degree
+long = FT(-92.2000) # degree
+atmos_h = FT(32)
+site_ID = "US-MOz"
+data_link = "https://caltech.box.com/shared/static/7r0ci9pacsnwyo0o9c25mhhcjhsu6d72.csv"
+
+include(
+    joinpath(
+        pkgdir(ClimaLand),
+        "experiments/integrated/fluxnet/met_drivers_FLUXNET.jl",
+    ),
+);
+
+z0_m = FT(2)
+z0_b = FT(0.2)
+
+shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
+    z0_m,
+    z0_b,
+    earth_param_set,
+);
+ψ_soil0 = FT(0.0)
+
+soil_driver = PrescribedSoil(
+    FT;
+    root_depths = SVector{10, FT}(-(10:-1:1.0) ./ 10.0 * 2.0 .+ 0.2 / 2.0),
+    ψ = t -> ψ_soil0,
+    α_PAR = FT(0.2),
+    α_NIR = FT(0.4),
+    T = t -> 298.0,
+    ϵ = FT(0.99),
+);
+
+rt_params = TwoStreamParameters(
+    FT;
+    G_Function = ConstantGFunction(FT(0.5)),
+    α_PAR_leaf = FT(0.1),
+    α_NIR_leaf = FT(0.45),
+    τ_PAR_leaf = FT(0.05),
+    τ_NIR_leaf = FT(0.25),
+    Ω = FT(0.69),
+    λ_γ_PAR = FT(5e-7),
+    λ_γ_NIR = FT(1.65e-6),
+)
+rt_model = TwoStreamModel{FT}(rt_params);
+
+cond_params = MedlynConductanceParameters(FT; g1 = FT(141.0))
+stomatal_model = MedlynConductanceModel{FT}(cond_params);
+
+is_c3 = FT(1)
+photo_params = FarquharParameters(FT, is_c3; Vcmax25 = FT(5e-5))
+photosynthesis_model = FarquharModel{FT}(photo_params);
+
+AR_params = AutotrophicRespirationParameters(FT)
+AR_model = AutotrophicRespirationModel{FT}(AR_params);
+
+f_root_to_shoot = FT(3.5)
+SAI = FT(1.0)
+RAI = FT(3f_root_to_shoot)
+ai_parameterization = PrescribedSiteAreaIndex{FT}(LAIfunction, SAI, RAI)
+
+K_sat_plant = FT(1.8e-6)
+ψ63 = FT(-4 / 0.0098)
+Weibull_param = FT(4)
+a = FT(0.05 * 0.0098)
+
+conductivity_model =
+    PlantHydraulics.Weibull{FT}(K_sat_plant, ψ63, Weibull_param)
+
+retention_model = PlantHydraulics.LinearRetentionCurve{FT}(a);
+
+ν = FT(0.7)
+S_s = FT(1e-2 * 0.0098)
+
+plant_hydraulics_ps = PlantHydraulics.PlantHydraulicsParameters(;
+    ai_parameterization = ai_parameterization,
+    ν = ν,
+    S_s = S_s,
+    rooting_depth = FT(0.5),
+    conductivity_model = conductivity_model,
+    retention_model = retention_model,
+);
+
+
+plant_hydraulics = PlantHydraulics.PlantHydraulicsModel{FT}(;
+    parameters = plant_hydraulics_ps,
+    n_stem = n_stem,
+    n_leaf = n_leaf,
+    compartment_surfaces = compartment_surfaces,
+    compartment_midpoints = compartment_midpoints,
+);
+ac_canopy = FT(1e3)
+energy_model = ClimaLand.Canopy.BigLeafEnergyModel{FT}(
+    BigLeafEnergyParameters{FT}(ac_canopy),
+)
+
+canopy = ClimaLand.Canopy.CanopyModel{FT}(;
+    parameters = shared_params,
+    domain = land_domain,
+    autotrophic_respiration = AR_model,
+    radiative_transfer = rt_model,
+    photosynthesis = photosynthesis_model,
+    conductance = stomatal_model,
+    energy = energy_model,
+    hydraulics = plant_hydraulics,
+    soil_driver = soil_driver,
+    atmos = atmos,
+    radiation = radiation,
+);
+
+exp_tendency! = make_exp_tendency(canopy)
+imp_tendency! = make_imp_tendency(canopy)
+jacobian! = make_jacobian(canopy)
+drivers = ClimaLand.get_drivers(canopy)
+updatefunc = ClimaLand.make_update_drivers(drivers)
+
+ψ_leaf_0 = FT(-2e5 / 9800)
+ψ_stem_0 = FT(-1e5 / 9800)
+S_l_ini =
+    inverse_water_retention_curve.(
+        retention_model,
+        [ψ_stem_0, ψ_leaf_0],
+        ν,
+        S_s,
+    )
+
+seconds_per_day = 3600 * 24.0
+t0 = 150seconds_per_day
+N_days = 20.0
+tf = t0 + N_days * seconds_per_day + 80
+sim_time = round((tf - t0) / 3600, digits = 2) # simulation length in hours
+set_initial_cache! = make_set_initial_cache(canopy)
+
+timestepper = CTS.ARS111();
+ode_algo = CTS.IMEXAlgorithm(
+    timestepper,
+    CTS.NewtonsMethod(
+        max_iters = 6,
+        update_j = CTS.UpdateEvery(CTS.NewNewtonIteration),
+    ),
+);
+
+ref_dt = 6.0
+dts = [ref_dt, 12.0, 48.0, 225.0, 450.0, 900.0, 1800.0, 3600.0]
+
+ref_T = []
+mean_err = []
+p95_err = []
+p99_err = []
+sol_err = []
+T_states = []
+times = []
+ΔT = FT(0)
+for dt in dts
+    @info dt
+
+    # Initialize model before each simulation
+    Y, p, coords = ClimaLand.initialize(canopy)
+    jac_kwargs = (;
+        jac_prototype = ClimaLand.ImplicitEquationJacobian(Y),
+        Wfact = jacobian!,
+    )
+
+    # Set initial conditions
+    Y.canopy.hydraulics.ϑ_l.:1 .= augmented_liquid_fraction.(ν, S_l_ini[1])
+    Y.canopy.hydraulics.ϑ_l.:2 .= augmented_liquid_fraction.(ν, S_l_ini[2])
+    evaluate!(Y.canopy.energy.T, atmos.T, t0)
+    set_initial_cache!(p, Y, t0)
+
+    # Create callback for driver updates
+    saveat = vcat(Array(t0:(3 * 3600):tf), tf)
+    updateat = vcat(Array(t0:(3 * 3600):tf), tf)
+    cb = ClimaLand.DriverUpdateCallback(updateat, updatefunc)
+
+    # Solve simulation
+    prob = SciMLBase.ODEProblem(
+        CTS.ClimaODEFunction(
+            T_exp! = exp_tendency!,
+            T_imp! = SciMLBase.ODEFunction(imp_tendency!; jac_kwargs...),
+            dss! = ClimaLand.dss!,
+        ),
+        Y,
+        (t0, tf),
+        p,
+    )
+    @time sol =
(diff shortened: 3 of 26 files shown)


experiments/long_runs/land.jl
```diff
@@ -362,7 +362,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
     τ_NIR_leaf = FT(0.25)
 
     # Energy Balance model
-    ac_canopy = FT(2.5e4) # this will likely be 10x smaller!
+    ac_canopy = FT(2.5e3) # this will likely be 10x smaller!
```
kmdeck, line 365: remove comment and update the value in the benchmark case also

src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -173,7 +173,8 @@ function canopy_boundary_fluxes!(
     shf .= canopy_tf.shf
     lhf .= canopy_tf.lhf
     r_ae .= canopy_tf.r_ae
-
+    p.canopy.energy.∂LHF∂qc .= canopy_tf.∂LHF∂qc
```
kmdeck, line 176: ideally we would set p.canopy.energy.var .= canopy_tf, but this will be a different PR. it would save an allocation

src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -334,15 +335,24 @@ function canopy_turbulent_fluxes_at_a_point(
     T_in::FT = Thermodynamics.air_temperature(thermo_params, ts_in)
     ΔT = T_in - T_sfc
     r_ae::FT = 1 / (conditions.Ch * SurfaceFluxes.windspeed(sc))
-    ρ_air::FT = Thermodynamics.air_density(thermo_params, ts_in)
+    ρ_sfc::FT = Thermodynamics.air_density(thermo_params, ts_sfc)
```
kmdeck, line 338: renaming since we use the surface density, not the atmos density at the first interior point

test/standalone/Vegetation/canopy_model.jl
```diff
@@ -761,6 +766,241 @@ end
...
+    @test parent(
+        abs.(
+            (
+                (
+                    p_2.canopy.radiative_transfer.LW_n .-
+                    p.canopy.radiative_transfer.LW_n
+                ) ./ FT(0.1) .- p.canopy.energy.∂LW_n∂Tc
+            ) ./ p.canopy.energy.∂LW_n∂Tc
+        ),
+    )[1] < 0.05
```
kmdeck, line 980: 5% error in derivative of LW_n

src/standalone/Vegetation/canopy_energy.jl
```diff
@@ -154,3 +170,67 @@ function root_energy_flux_per_ground_area!(
...
+
+"""
+    partial_q_sat_partial_T_liq(P::FT, T::FT) where {FT}
+
+Computes the quantity ∂q_sat∂T at temperature T and pressure P,
+over liquid water. The temperature must be in Celsius.
+
+Uses the polynomial approximation from Flatau et al. (1992).
+"""
+function partial_q_sat_partial_T_liq(P::FT, T::FT) where {FT}
```
Sbozzolo, line 215: I have the feeling that a function like this might not be super accurate with Float32

kmdeck: That's true. we can convert everything to float 64 for the computation and then back to FT?

experiments/standalone/Vegetation/timestep_test.jl
```diff
@@ -0,0 +1,249 @@
...
+photo_params = FarquharParameters(FT, Canopy.C3(); Vcmax25 = FT(5e-5))
+
+photosynthesis_model = FarquharModel{FT}(photo_params);
+
+AR_params = AutotrophicRespirationParameters(FT)
+AR_model = AutotrophicRespirationModel{FT}(AR_params);
+
+f_root_to_shoot = FT(3.5)
+SAI = FT(1.0)
+RAI = FT(3 * f_root_to_shoot)
```
Sbozzolo, line 97: A reminder that you can also write
```suggestion
RAI = FT(3f_root_to_shoot)
```

Sbozzolo: It makes no difference and it can improve readability in certain cases

src/standalone/Vegetation/canopy_energy.jl
```diff
@@ -154,3 +170,67 @@ function root_energy_flux_per_ground_area!(
...
+        ∂LHF∂qc = p.canopy.energy.∂LHF∂qc
+        ∂SHF∂Tc = p.canopy.energy.∂SHF∂Tc
+        ∂LW_n∂Tc = p.canopy.energy.∂LW_n∂Tc
+        ∂qc∂Tc = p.canopy.energy.∂qc∂Tc
+        ϵ_c = p.canopy.radiative_transfer.ϵ
+        area_index = p.canopy.hydraulics.area_index
+        ac_canopy = model.parameters.ac_canopy
+        earth_param_set = canopy.parameters.earth_param_set
+        _T_freeze = FT(LP.T_freeze(earth_param_set))
+        _σ = FT(LP.Stefan(earth_param_set))
```
Sbozzolo, line 192: Aren't parameters already in the correct `FT`?

src/standalone/Vegetation/component_models.jl
```diff
@@ -123,11 +123,48 @@ may be needed and passed in (via the `canopy` model itself).
...
 """
-function ClimaLand.make_compute_exp_tendency(::AbstractCanopyComponent, canopy)
-    function compute_exp_tendency!(dY, Y, p, t) end
+function ClimaLand.make_compute_exp_tendency(
+    component::AbstractCanopyComponent,
+    canopy,
+)
+    function compute_exp_tendency!(dY, Y, p, t)
+        vars = prognostic_vars(component)
+        if vars != ()
```
Sbozzolo, line 132:
```suggestion
        if !isempty(vars)
```

src/standalone/Vegetation/component_models.jl
```diff
@@ -123,11 +123,48 @@ may be needed and passed in (via the `canopy` model itself).
...
+The right hand side for the entire canopy model can make use of
+these functions for the individual components.
+"""
+function ClimaLand.make_compute_imp_tendency(
+    component::AbstractCanopyComponent,
+    canopy,
+)
+    function compute_imp_tendency!(dY, Y, p, t)
+        vars = prognostic_vars(component)
+        if vars != ()
```
Sbozzolo, line 155:
```suggestion
        if !isempty(vars)
```

There's also probably a better way to write this block

experiments/standalone/Vegetation/timestep_test.jl
```diff
@@ -0,0 +1,249 @@
...
+    )
+
+Y.canopy.hydraulics.ϑ_l.:1 .= augmented_liquid_fraction.(ν, S_l_ini[1])
+Y.canopy.hydraulics.ϑ_l.:2 .= augmented_liquid_fraction.(ν, S_l_ini[2])
+
+
+
+t0 = 150 * 3600 * 24.0
+N_days = 100
+tf = t0 + 3600 * 24 * N_days
```
Sbozzolo, line 181:
```suggestion
seconds = second = 1.0
hours = hour = 3600seconds
day = 24hours

t0 = 150days
tf = t0 + 24days
```

src/standalone/Vegetation/canopy_energy.jl
kmdeck: (already doing that, closing)

experiments/standalone/Vegetation/timestep_test.jl
kmdeck: changing

src/standalone/Vegetation/component_models.jl
kmdeck: clarified it a little bit

src/standalone/Vegetation/canopy_energy.jl
kmdeck: I changed it here, but it could be a future PR to change everywhere

experiments/standalone/Vegetation/timestep_test.jl
kmdeck: I went with
```
seconds_per_day = 3600 * 24.0
t0 = 150seconds_per_day
N_days = 100
tf = t0 + N_days*seconds_per_day
```

experiments/standalone/Vegetation/timestep_test.jl
```diff
@@ -0,0 +1,249 @@
...
+    ylabel = "Temperature (K)",
+    xscale = log10,
+    yscale = log10,
+)
+dts = dts ./ 60
+lines!(ax, dts, FT.(mean_err), label = "Mean Error")
+lines!(ax, dts, FT.(p95_err), label = "95th% Error")
+lines!(ax, dts, FT.(p99_err), label = "99th% Error")
+axislegend(ax)
+save(joinpath(savedir, "errors.png"), fig)
```
dennisYatunin, line 249: The mean error plot currently on Buildkite indicates a convergence order much lower than 1, which is the value I would expect to see for IMEX Euler (`ARS111`). This might be evidence of a non-negligible issue with the Jacobian approximation, but it could also be an issue with the test setup. In order to measure reasonable convergence orders in `ClimaTimeSteppers.jl`, I needed to use a reference `dt` that was several orders of magnitude smaller than the tested values of `dt`, even for simple nonlinear problems. It would be good to rerun this test with smaller `dt`s, particularly with a smaller reference `dt`, to see whether you can get closer to the expected convergence order.

dennisYatunin (approved):
These changes look good overall, but I have some concerns about the reliability of the tests. The current test results indicate that there are significant errors in the Jacobian approximation, but these errors might appear bigger than they actually are if there is a problem with the reference solutions. This issue can be dealt with in future PRs, though; the new infrastructure added here is fine to merge in.

experiments/integrated/fluxnet/fluxnet_simulation.jl
```diff
@@ -6,14 +6,15 @@ N_spinup_days = 30
 N_days = N_spinup_days + 30
 tf = Float64(t0 + 3600 * 24 * N_days)
 t_spinup = Float64(t0 + N_spinup_days * 3600 * 24)
-saveat = Array(t_spinup:(n * dt):tf)
+dt_save = 3600.0
```
juliasloan25, line 9: should we keep `dt_save = n * dt` here?

experiments/standalone/Vegetation/timestep_test.jl
```diff
@@ -0,0 +1,249 @@
+import SciMLBase
```
juliasloan25, line 1: It would be nice to have a comment at the top of this file explaining the goal of this experiment - are we just testing running the vegetation model for a single timestep? Are we testing for stability over n days? etc

src/shared_utilities/implicit_timestepping.jl
```diff
@@ -107,53 +108,63 @@ to the `explicit_names` tuple.
...
-            (name, name) => Fields.Field(
-                tridiag_type,
-                axes(MatrixFields.get_field(Y, name)),
-            ),
-        available_implicit_names,
-    )
 
-    matrix = MatrixFields.FieldMatrix(identity_blocks..., tridiagonal_blocks...)
+
+    matrix = MatrixFields.FieldMatrix(implicit_blocks..., explicit_blocks...)
```
juliasloan25, line 167: Before, we were setting up the FieldMatrix with the explicit (identity) blocks first, and then the implicit (tridiagonal). Do we need to switch the argument order here to `(explicit_blocks..., implicit_blocks...)`?

juliasloan25: RichardsModel and EnergyHydrology results look the same between this branch and main, so the order here probably doesn't matter (maybe it would if we had off-diagonal interaction terms)

src/standalone/Vegetation/canopy_energy.jl
```diff
@@ -87,11 +103,11 @@ where the canopy temperature is modeled prognostically.
 canopy_temperature(model::BigLeafEnergyModel, canopy, Y, p, t) =
     Y.canopy.energy.T
 
-function make_compute_exp_tendency(
+function make_compute_imp_tendency(
```
juliasloan25, line 106: do we still need a function `compute_exp_tendency!` that sets `dY.canopy.energy.T .= 0`? In our other ImEx models, we set this to avoid updating variables twice

juliasloan25: e.g. here in EnergyHydrology https://github.com/CliMA/ClimaLand.jl/pull/678/files#diff-7ba5ae758c389ffc507c642329638590c755bbd4a2edf53294e4dc357bf913c6R193

juliasloan25: ah, I just saw the fallback method for `AbstractCanopyComponent`

src/shared_utilities/implicit_timestepping.jl
dennisYatunin: The FieldMatrix is equivalent to a Dict, so order doesn't matter here.

experiments/integrated/fluxnet/fluxnet_simulation.jl
kmdeck: I actually found this more confusing, because every site then has a different n*dt to reach hourly saving.

juliasloan25:
Some insights we gained during explicitly-stepped experiments in #771 - 

In testing the canopy temperature implicit solver implemented in https://github.com/CliMA/ClimaLand.jl/pull/675, we saw unexpected convergence behavior. To investigate if the cause of this was due to inacurracy of the reference solution or flaws in the simulation setup, we take a look here at the convergence behavior of the same setup using an explicit solver.


### Key point 1: `ac_canopy` affects time to steady state
 

Here we can see the evolution of canopy temperature over time using two different `ac_canopy` values. Atmospheric and radiative drivers are updated every 3 hours, so we see the temperature sharply change every 3 hours in both cases. The temperature then tends towards steady state but only reaches steady state with the smaller `ac_canopy=1e3`, creating the flat sections of the plot.

This pattern of temperature changing following a driver update then remaining constant upon reaching steady state produces two distinct regimes.

### Key point 2: convergence behavior should be assessed outside of the steady state regime
 

Here we have two plots showing convergence using ARS111 with only an explicit tendency (nothing stepped implicitly). The left plot shows convergence for simulations run for 12 hours + 80 seconds, while the left was run for 12 hours. The plotted line has slope 1, so we see the expected convergence behavior in the first case but not in the second.

In the first case, the simulations end shortly after a driver update, so outside of the steady state regime. All simulations are converging towards the steady state, but are not yet there, so we see the expected convergence behavior.

In the second case, the simulations end during the steady state regime. Simulations with all timesteps have converged to the same steady state, so the error for each of them is very small and doesn't converge as we would expect. This doesn't mean the convergence behavior of the solver/tendency is wrong; we've just stopped the simulation at an uninformative time.

### Conclusion
When testing the convergence of a solver for canopy temperature (or similar equations that reach steady state), we need to make sure we check convergence at a time outside of the steady state regime. Also note that this could have effects on errors throughout the simulation - e.g. error peaking sharply after a driver update then returning to a small value as steady state is reached.

juliasloan25:
Here is the convergence of the implicit solver when running the simulation for 3 hours and updating the drivers at every timestep. This shows the expected convergence behavior for timesteps up to ~1800s (all dts shown are [12.0, 24.0, 48.0, 100.0, 225.0, 450.0, 900.0, 1800.0, 3600.0].
image

kmdeck:
> Here is the convergence of the implicit solver when running the simulation for 3 hours and updating the drivers at every timestep. This shows the expected convergence behavior for timesteps up to ~1800s (all dts shown are [12.0, 24.0, 48.0, 100.0, 225.0, 450.0, 900.0, 1800.0, 3600.0].
> 

Which convergence checker/max iters combo are we using? this is with ARS111?  Just curious because, as this looks quite good already, I am wondering if we can improve the error

juliasloan25:
> > Here is the convergence of the implicit solver when running the simulation for 3 hours and updating the drivers at every timestep. This shows the expected convergence behavior for timesteps up to ~1800s (all dts shown are [12.0, 24.0, 48.0, 100.0, 225.0, 450.0, 900.0, 1800.0, 3600.0].
> 
> Which convergence checker/max iters combo are we using? this is with ARS111? Just curious because, as this looks quite good already, I am wondering if we can improve the error

This is using ARS111 with `max_iters = 6` and no convergence checker

juliasloan25:
Here are plots showing the difference in T at the end of a 100 day simulation between runs with dt=10s and dt=1800s. The plot on the left uses `max_iters = 6`, and the plot on the right uses `max_iters = 10`. Both have the same features and look nearly identical, except the mean error is slightly larger in the case with `max_iters = 10`.

error_T_100days_maxiters6
error_T_100days_maxiters10

kmdeck:
> Here are plots showing the difference in T at the end of a 100 day simulation between runs with dt=10s and dt=1800s. The plot on the left uses `max_iters = 6`, and the plot on the right uses `max_iters = 10`. Both have the same features and look nearly identical, except the mean error is _slightly_ larger in the case with `max_iters = 10`.


thanks for checking this! I think this is consistent with there being certain times where the Newton iterations do not converge (and not because we dont carry out enough iterations, but because J is so far off)

src/standalone/Vegetation/canopy_boundary_fluxes.jl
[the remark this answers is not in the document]
juliasloan25: Good point - I opened an issue to take care of this in a separate PR: https://github.com/CliMA/ClimaLand.jl/issues/786

src/shared_utilities/implicit_timestepping.jl
```diff
@@ -107,53 +108,63 @@ to the `explicit_names` tuple.
...
-    available_implicit_names =
-        MatrixFields.unrolled_filter(is_in_Y, implicit_names)
-    available_explicit_names =
-        MatrixFields.unrolled_filter(is_in_Y, explicit_names)
+    available_implicit_vars =
+        MatrixFields.unrolled_filter(is_in_Y, implicit_vars)
+    available_explicit_vars =
+        MatrixFields.unrolled_filter(is_in_Y, explicit_vars)
+
+    get_jac_type(
```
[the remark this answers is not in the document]
kmdeck: I think for now this is OK, but if our system becomes more complex we would need to do something like what you said.

src/standalone/Vegetation/component_models.jl
```diff
@@ -123,11 +123,50 @@ may be needed and passed in (via the `canopy` model itself).
...
+            getproperty(dY_canopy, name(component)) .= 0
+        end
+
+    end
+    return compute_imp_tendency!
+end
+
+
+function make_compute_jacobian(component::AbstractCanopyComponent, canopy)
+    function compute_jacobian!(W, Y, p, dtγ, t) end
```
kmdeck, line 167: For the implicit and explicit tendencies, we set dY = 0 by default (above). Here, do we need to set elements of the jacobian to zero? I think not, but Im just noticing a difference in how we handle it.

kmdeck: and so want to double check

test/standalone/Vegetation/canopy_model.jl
```diff
@@ -760,7 +765,276 @@ end
...
+
+        finitediff_SHF = (p_2.canopy.energy.shf .- p.canopy.energy.shf) ./ ΔT
+        estimated_SHF = p.canopy.energy.∂SHF∂Tc
+        @test parent(abs.(finitediff_SHF .- estimated_SHF) ./ finitediff_SHF)[1] <
+              0.05
+
+        finitediff_LHF = (p_2.canopy.energy.lhf .- p.canopy.energy.lhf) ./ ΔT
+        estimated_LHF = p.canopy.energy.∂LHF∂qc .* p.canopy.energy.∂qc∂Tc
+        @test parent(abs.(finitediff_LHF .- estimated_LHF) ./ finitediff_LHF)[1] <
+              0.5
```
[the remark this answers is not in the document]
kmdeck: This will be tackled in another PR! We think they come from the approximation

experiments/long_runs/land_region.jl
```diff
@@ -51,6 +51,7 @@ time_interpolation_method = LinearInterpolation(PeriodicCalendar())
 regridder_type = :InterpolationsRegridder
 context = ClimaComms.context()
 device = ClimaComms.device()
+ClimaComms.init(context)
```
kmdeck, line 54: can remove, duplicate

src/diagnostics/default_diagnostics.jl
```diff
@@ -169,9 +169,12 @@ function default_diagnostics(
             # "pwc", # return a Tuple
             "si",
             "sie",
+            "swo",
+            "lwo",
```
kmdeck, line 173: alos "er", "et"?

src/diagnostics/define_diagnostics.jl
```diff
@@ -726,6 +726,49 @@ function define_diagnostics!(land_model)
...
+
+    # Ecosystem respiration
+    add_diagnostic_variable!(
+        short_name = "er",
+        long_name = "Ecosystem Respiration",
+        standard_name = "ecosystem respiration",
+        units = "mol m^-2 s^-1",
+        comments = "Total respiration flux out of the surface.",
+        compute! = (out, Y, p, t) ->
+            compute_total_respiration!(out, Y, p, t, land_model),
```
kmdeck, line 769: I think we also need a add_diagnostic_variable! for runoff.

and we can make a note that this is Surface runoff only

src/diagnostics/land_compute_methods.jl
```diff
@@ -158,6 +158,51 @@ end # Convert from kg C to mol CO2.
 @diagnostic_compute "soilco2_diffusivity" SoilCanopyModel p.soilco2.D
 @diagnostic_compute "soilco2_source_microbe" SoilCanopyModel p.soilco2.Sm
 
+## Other ##
+
+@diagnostic_compute "lw_out" SoilCanopyModel p.LW_out
+@diagnostic_compute "sw_out" SoilCanopyModel p.SW_out
+@diagnostic_compute "runoff" SoilCanopyModel p.soil.R_s
```
kmdeck, line 165: "surface_runoff"

src/diagnostics/land_compute_methods.jl
```diff
@@ -158,6 +158,51 @@ end # Convert from kg C to mol CO2.
...
+function compute_total_respiration!(
+    out,
+    Y,
+    p,
+    t,
+    land_model::SoilCanopyModel{FT},
+) where {FT}
+    if isnothing(out)
+        return p.soilco2.top_bc .* FT(83.26) .+
+               p.canopy.autotrophic_respiration.Ra
```
kmdeck, line 199: do we also need heterotrophic respiration? Could you note what is 83.26? @AlexisRenchon

src/diagnostics/define_diagnostics.jl
kmdeck: moles of what?

experiments/long_runs/land.jl
```diff
@@ -662,7 +662,7 @@ setup_and_solve_problem(; greet = true);
 # read in diagnostics and make some plots!
 #### ClimaAnalysis ####
 simdir = ClimaAnalysis.SimDir(outdir)
-short_names = ["gpp", "ct", "lai", "swc", "si"]
+short_names = ["gpp", "ct", "lai", "swc", "si", "swo", "lwo", "et", "er"]
```
kmdeck, line 665: and runoff

kmdeck: @AlexisRenchon

src/diagnostics/define_diagnostics.jl
kmdeck: @AlexisRenchon

src/diagnostics/land_compute_methods.jl
kmdeck: @AlexisRenchon

src/diagnostics/land_compute_methods.jl
AlexisRenchon: heterotrophic respiration is `p.soilco2.top_bc .* FT(83.26)`, I'll add a comment for the factor

kmdeck: thank you! I missed that

experiments/long_runs/land.jl
```diff
@@ -10,9 +10,9 @@
 # Soil depth: 50 m
 # Simulation duration: 365 d
 # Timestep: 900 s
```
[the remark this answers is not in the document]
juliasloan25: that sounds good to me! I think 450s is good enough for now and we can improve on it

experiments/long_runs/land_region.jl
juliasloan25: I'm still confused about this - I don't see anywhere in the code where we initialize our comms contexts... but I'll remove this for now

kmdeck: oh im sorry! I mis read this, I was reviewing it too quickly. I was confusing it with `context = ClimaComms.context()` on line 52.

juliasloan25: I'll add them in as part of another PR - see #811

src/standalone/Vegetation/component_models.jl
juliasloan25: I also don't think we need to zero them out, but not 100% sure. @dennisYatunin do you have any insights?

california_longrun_cpu/regional_diagnostics/output_active
```diff
@@ -0,0 +1 @@
+output_0005
```
kmdeck, line 1: (should remove these before merging)

experiments/long_runs/matrix_fields_reproducer.jl
```diff
@@ -0,0 +1,191 @@
+#=
+using Revise; include("matrix")
+=#
+using Test
```
kmdeck, line 4: should probbaly also be removed?

src/standalone/Vegetation/component_models.jl
juliasloan25: We talked about this offline and clarified that the blocks in the jacobian for explicitly-stepped variables will always be `-I`. There's nowhere else these blocks would be updated, so we expect them to remain unchanged and don't need to reset any blocks of the jacobian here

PR #677, opened 2024-06-25
conv checker

kmdeck:
## Purpose 
Remove convergence checker from select simulations, since we set max_iters = 1 or 2 anyways

## Content
I kept the convergence checker in the following:
`./experiments/standalone/Soil/richards_comparison.jl` [more accurate better for comparison?]
`./experiments/standalone/Soil/water_conservation.jl` [testing limit of performance of jacobian approx]
`./experiments/integrated/performance/conservation/ozark_conservation_setup.jl` [perhaps for conservation we want convergence]
`./docs/tutorials/standalone/Soil/layered_soil.jl` [not sure, but this had a very high number of max_iters, so the convergence checker may matter]

I removed it from all global runs, though

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@ main
 --------
 Improve Performance of soil canopy model
 - PR [#666]
+- PR [#677] Remove norm condition
 Add base global soil canopy run to benchmark and experiments
 - PR [#591] adds benchmark run
 - PR [#669] adds topmodel runoff
--- a/experiments/benchmarks/land.jl
+++ b/experiments/benchmarks/land.jl
@@ -17,8 +17,7 @@
 # Simulation duration: 6 hours
 # Timestep: 180 s
 # Timestepper: ARS343
-# Maximum iterations: 1
-# Convergence criterion: 1e-8
+# Fixed number of iterations: 1
 # Jacobian update: Every Newton iteration
 # Precipitation data update: every timestep
 import SciMLBase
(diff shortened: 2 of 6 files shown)


experiments/benchmarks/land.jl
```diff
@@ -622,8 +622,7 @@ function setup_and_solve_problem(; greet = false)
 
     # Define timestepper and ODE algorithm
     stepper = CTS.ARS343()
-    norm_condition = CTS.MaximumError(FT(1e-8))
-    conv_checker = CTS.ConvergenceChecker(; norm_condition = norm_condition)
+    conv_checker = nothing
```
juliasloan25, line 625: the default is nothing, so we could just not pass `conv_checker` at all :)

kmdeck: good call!

NEWS.md
```diff
@@ -5,6 +5,7 @@ main
 --------
 Improve Performance of soil canopy model
 - PR [#666]
+- PR [] Remove norm condition
```
juliasloan25, line 8: can add #677 here

juliasloan25 (approved):
This looks good! There are still some more places we use `convergence_checker` (docs/layered_soil.jl, ozark_conservation_setup.jl, richards_comparison.jl, water_conservation.jl). Do we want to remove it from all of these places for consistency?

kmdeck:
> This looks good! There are still some more places we use `convergence_checker` (docs/layered_soil.jl, ozark_conservation_setup.jl, richards_comparison.jl, water_conservation.jl). Do we want to remove it from all of these places for consistency?

I think in water_conservation.jl, richards_comparison.jl, and ozark_conservation.jl I want to keep it, because there having a more accurate solution (to check conservation, compare to Bonan, etc) is more important. Ill try without it in the layered soil case, I had kept it for now because max_iters=50 for that sim (so we actually probably use the norm convergence checker. But im not sure we need that)

PR #678, opened 2024-06-25
Approximate soil energy Jacobian

kmdeck:
## Purpose 
Add in our first approximation for the energy equation Jacobian, which only includes d\rho e_int/dT.

This does allow us to take much bigger timesteps, more comparable to our target of 15 minutes, except in the cases where the soil is freezing or the canopy is present. If we make the thermal mass of the canopy larger, we can take bigger steps. 
-> need to think about how we handle soil freezing in terms of implicit stepping
-> canopy temp probably does need an implicit solver.

## Content
- Alter imp/exp tendencies for soil energy hydrology
- Alter update jacobian for soil energy hydrology
- Increase timesteps in tutorials and *some* experiments. In cases with a canopy, this involved increasing ac_canopy as well, so those are unchanged
- Add unit test/update unit tests
- Adds a buildkite run showing performance as a function of timestep, similar to what we had in Richards equation. We simulation the exact same system, but in one case with T > T freeze, and in one case with T < Tfreeze. The timestep required is orders of magnitude different.
No Phase change:
energy_conservation_flux_full_soil_no_pc
water_conservation_flux_full_soil_no_pc

With Phase change:
energy_conservation_flux_full_soil_pc
water_conservation_flux_full_soil_pc

Note the differences in x and y axes. The solution at dt = 10, 100 seconds for the phase change one has lots of oscillations and is not good enough, although the relative error is not terrible

----

### DIFF
--- a/docs/tutorials/standalone/Soil/evaporation.jl
+++ b/docs/tutorials/standalone/Soil/evaporation.jl
@@ -172,7 +172,7 @@ init_soil!(Y, z, soil.parameters);
 # Timestepping:
 t0 = Float64(0)
 tf = Float64(24 * 3600 * 13)
-dt = Float64(2.0)
+dt = Float64(900.0)
 
 # We also set the initial conditions of the cache here:
 set_initial_cache! = make_set_initial_cache(soil)
--- a/docs/tutorials/standalone/Soil/freezing_front.jl
+++ b/docs/tutorials/standalone/Soil/freezing_front.jl
@@ -215,7 +215,7 @@ tendency_jacobian! = ClimaLand.make_tendency_jacobian(soil);
 jac_kwargs =
     (; jac_prototype = ImplicitEquationJacobian(Y), Wfact = tendency_jacobian!);
 
-dt = Float64(10)
+dt = Float64(100)
 
 timestepper = CTS.ARS111();
 ode_algo = CTS.IMEXAlgorithm(
--- a/docs/tutorials/standalone/Soil/soil_energy_hydrology.jl
+++ b/docs/tutorials/standalone/Soil/soil_energy_hydrology.jl
@@ -251,7 +251,7 @@ set_initial_cache!(p, Y, t0);
 # We use [ClimaTimesteppers.jl](https://github.com/CliMA/ClimaTimesteppers.jl) for carrying out the time integration.
 
 # Choose a timestepper and set up the ODE problem:
-dt = Float64(60.0);
+dt = Float64(1000.0);
 timestepper = CTS.ARS111();
 ode_algo = CTS.IMEXAlgorithm(
     timestepper,
--- a/experiments/benchmarks/land.jl
+++ b/experiments/benchmarks/land.jl
@@ -15,7 +15,7 @@
 # Number of spatial elements: 101 in horizontal, 15 in vertical
 # Soil depth: 50 m
 # Simulation duration: 6 hours
-# Timestep: 180 s
+# Timestep: 900 s
 # Timestepper: ARS343
 # Fixed number of iterations: 1
 # Jacobian update: Every Newton iteration
@@ -363,7 +363,7 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
     τ_NIR_leaf = FT(0.25)
 
     # Energy Balance model
-    ac_canopy = FT(2.5e5)
+    ac_canopy = FT(2.5e3)
 
     # Conductance Model
     g1 = FT(141) # Wang et al: 141 sqrt(Pa) for Medlyn model; Natan used 300.
@@ -608,7 +608,7 @@ function setup_and_solve_problem(; greet = false)
     # to set up for both CPU/GPU at the same time
     t0 = 0.0
     tf = 60 * 60.0 * 6
-    Δt = 180.0
+    Δt = 900.0
     nelements = (101, 15)
     if greet
         @info "Run: Global RichardsModel"
--- a/experiments/integrated/fluxnet/US-MOz/US-MOz_simulation.jl
+++ b/experiments/integrated/fluxnet/US-MOz/US-MOz_simulation.jl
@@ -20,7 +20,7 @@ h_leaf = FT(9.5) # m
 t0 = Float64(120 * 3600 * 24)# start mid year to avoid snow
 
 # Time step size:
-dt = Float64(120)
+dt = Float64(450)
 
 # Number of timesteps between saving output:
-n = 15
+n = 4
--- a/src/standalone/Soil/energy_hydrology.jl
+++ b/src/standalone/Soil/energy_hydrology.jl
@@ -189,38 +189,15 @@ function ClimaLand.make_compute_exp_tendency(
 ) where {FT}
     function compute_exp_tendency!(dY, Y, p, t)
         z = model.domain.fields.z
-        heat_top_flux_bc = p.soil.top_bc.heat
-        heat_bottom_flux_bc = p.soil.bottom_bc.heat
 
-        interpc2f = Operators.InterpolateC2F()
-        gradc2f = Operators.GradientC2F()
-
-        # Without topography only
-        # In Cartesian coordinates, W (z^) = Cov3 (z^)= Contra3 (n^ = z^)
-        # In spherical coordinates, W (r^) = Cov3 (r^) = Contra3 (n^ = r^)
-        # Passing WVector to gradient BC is passing a normal flux.
-
-        # Reset liquid water and ice content
+        # Don't update the prognostic variables we're stepping implicitly
         dY.soil.ϑ_l .= 0
+        dY.soil.ρe_int .= 0
+
+        # Note that soil ice content is only updated via source terms,
+        # which are plus-added to dY.soil.θ_i, so we must zero it out here first.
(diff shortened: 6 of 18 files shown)


test/standalone/Soil/soiltest.jl
```diff
@@ -169,10 +169,10 @@ for FT in (Float32, Float64)
         set_initial_cache! = make_set_initial_cache(soil_heat_on)
         set_initial_cache!(p, Y, t0)
 
-        # Test explicit tendency - should update internal energy, set others to 0
-        exp_tendency! = make_exp_tendency(soil_heat_on)
+        # Test implicit tendency - should update internal energy only (Ksat = 0 for hydrology)
```
juliasloan25, line 172: It might be nice to add a comment at the top of this test explaining that water and energy are stepped implicitly so we expect `imp_tendency` to update those, and `exp_tendency` to update ice. The only downside is we'd have to remember to update that comment if we change what's stepped implicitly vs explicitly

docs/tutorials/standalone/Soil/evaporation.jl
```diff
@@ -172,7 +172,7 @@ init_soil!(Y, z, soil.parameters);
 # Timestepping:
 t0 = Float64(0)
 tf = Float64(24 * 3600 * 13)
-dt = Float64(2.0)
+dt = Float64(900.0)
```
juliasloan25, line 175: awesome!! did 1000s fail?

docs/tutorials/standalone/Soil/freezing_front.jl
```diff
@@ -215,13 +215,13 @@ tendency_jacobian! = ClimaLand.make_tendency_jacobian(soil);
...
 
-dt = Float64(10)
+dt = Float64(100)
 
 timestepper = CTS.ARS111();
 ode_algo = CTS.IMEXAlgorithm(
     timestepper,
     CTS.NewtonsMethod(
-        max_iters = 3,
+        max_iters = 23,
```
juliasloan25, line 224: why 23? it seems like 20 would be more straightforward, but it doesn't really matter :)

docs/tutorials/standalone/Soil/soil_energy_hydrology.jl
```diff
@@ -251,7 +251,7 @@ set_initial_cache!(p, Y, t0);
 # We use [ClimaTimesteppers.jl](https://github.com/CliMA/ClimaTimesteppers.jl) for carrying out the time integration.
 
 # Choose a timestepper and set up the ODE problem:
-dt = Float64(60.0);
+dt = Float64(1000.0);
```
juliasloan25, line 254: awesome!! 🥳

experiments/benchmarks/land.jl
```diff
@@ -364,7 +364,7 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
     τ_NIR_leaf = FT(0.25)
 
     # Energy Balance model
-    ac_canopy = FT(2.5e5)
+    ac_canopy = FT(2.5e4)
```
juliasloan25, line 367: I think it would be nice to keep the parameters corresponding to the setup we want to be able to run, and adjust the timesteppping as needed. It might help us have a better idea of how close we are to where we want to be

experiments/integrated/fluxnet/US-MOz/US-MOz_simulation.jl
```diff
@@ -20,7 +20,7 @@ h_leaf = FT(9.5) # m
 t0 = Float64(120 * 3600 * 24)# start mid year to avoid snow
 
 # Time step size:
-dt = Float64(120)
+dt = Float64(450)
 
 # Number of timesteps between saving output:
-n = 15
+n = 4
```
juliasloan25, line 26: why decreased?

experiments/standalone/Soil/full_soil_implicit.jl
```diff
@@ -0,0 +1,238 @@
+using ClimaCore
```
juliasloan25, line 1: maybe we could just title this file "full_soil.jl" or "energy_hydrology.jl"? The "implicit" is kind of implied since that's how this model will always be stepped

experiments/standalone/Soil/full_soil_implicit.jl
```diff
@@ -0,0 +1,238 @@
...
+stepper = CTS.ARS111()
+err = (FT == Float64) ? 1e-8 : 1e-4
+convergence_cond = CTS.MaximumError(err)
+conv_checker = CTS.ConvergenceChecker(norm_condition = convergence_cond)
+ode_algo = CTS.IMEXAlgorithm(
+    stepper,
+    CTS.NewtonsMethod(
+        max_iters = 50,
+        update_j = CTS.UpdateEvery(CTS.NewNewtonIteration),
+        convergence_checker = conv_checker,
```
juliasloan25, line 114: do we want to just use small `max_iters` and remove `convergence_checker` here too?

experiments/standalone/Soil/full_soil_implicit.jl
```diff
@@ -0,0 +1,238 @@
...
+        label = "Water mass error",
+        color = "purple",
+        linewidth = 3,
+        ylabel = "Mass error",
+        xlabel = "",
+        legend = :bottomright,
+        background_color_legend = nothing,
+    )
+    Plots.savefig(
+        joinpath(savedir, "water_conservation_flux_full_soil_$(exp_name).png"),
```
juliasloan25, line 235: could we add energy conservation too?

juliasloan25 (commented):
This looks great!! Thank you Kat

experiments/benchmarks/land.jl
kmdeck: I agree! Ill revert back

kmdeck:
@juliasloan25 thank you for the review! I wont be able to address your comments until next week, but if you have time, let me know if you think these are the main things, or if you see any issues with how we defined J. I think it's all OK because it was just adding terms copying what we do for theta_l, but it would be helpful if you took a close look at that part too.

test/standalone/Soil/soiltest.jl
kmdeck: I added this!

docs/tutorials/standalone/Soil/evaporation.jl
kmdeck: yes :(

docs/tutorials/standalone/Soil/freezing_front.jl
kmdeck: I think this was a typo, going back to 3.

experiments/benchmarks/land.jl
kmdeck: ah, our target is actually 2.5e3. I had made it bigger to run in the first place. I may need to lower dt.

experiments/integrated/fluxnet/US-MOz/US-MOz_simulation.jl
kmdeck: we save every n*dt=3600, but since dt increased, n decreases. it would be clearer to just save every 3600 seconds and not have it depend on dt!

experiments/standalone/Soil/full_soil_implicit.jl
kmdeck: I think here I want to have it use the convergence checker, since we are testing conservation, and then this serves as sort of a best case scenario. In reality we'll use a smaller max_iters and not use the convergence checker at all, like you're alluding to.

experiments/standalone/Soil/full_soil_implicit.jl
kmdeck: I added it in!

experiments/benchmarks/land.jl
```diff
@@ -363,7 +363,7 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
     τ_NIR_leaf = FT(0.25)
 
     # Energy Balance model
-    ac_canopy = FT(2.5e5)
+    ac_canopy = FT(2.5e3)
```
kmdeck, line 366: This is the more realistic value we want to use in our simulations.

dennisYatunin (approved):
These changes all look good!

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -274,9 +253,26 @@ function ClimaLand.make_compute_imp_tendency(
         @. dY.soil.ϑ_l =
             -(divf2c_rre(-interpc2f(p.soil.K) * gradc2f(p.soil.ψ + z)))
 
+        # Heat equation RHS
+        divf2c_heat = Operators.DivergenceF2C(
+            top = Operators.SetValue(Geometry.WVector.(heat_top_flux_bc)),
+            bottom = Operators.SetValue(Geometry.WVector.(heat_bottom_flux_bc)),
```
dennisYatunin, line 259: Just a note though that this will allocate two surface Fields on every Jacobian update. It would generally be better to cache these Fields somewhere, and you could also use the same cached Field for the top and bottom BCs.

kmdeck: Thanks, Dennis! good catch!

PR #680, opened 2024-06-26
add ClimaCoupler downstream test

juliasloan25:
## Purpose 
It would be useful to have a driver that runs a simple coupled ClimaAtmos/ClimaLand setup for one or two timesteps. This could be used by these packages to test the effects of changes downstream in ClimaCoupler. It wouldn't test physical stability, but it would test interface changes

see https://github.com/CliMA/ClimaCoupler.jl/issues/877

## To-do
It would be good to have a more streamlined test that doesn't require instantiating the entire `ClimaEarth` environment, but for now this is fine

## Content

----

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,0 +1,38 @@
+name: Downstream
+on:
+  push:
+    branches:
+      - main
+    tags: '*'
+  pull_request:
+
+# Needed to allow julia-actions/cache to delete old caches that it has created
+permissions:
+  actions: write
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: downstream ClimaCoupler.jl
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: '1.10'
+      - uses: julia-actions/cache@v2
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: actions/checkout@v4
+        with:
+          repository: 'CliMA/ClimaCoupler.jl'
+          path: ClimaCoupler.jl
+      - run: |
+          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ -e 'using Pkg; Pkg.instantiate()'
+          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ -e 'using Pkg; Pkg.develop(; path = ".")'
+          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ ClimaCoupler.jl/experiments/ClimaEarth/run_amip.jl --config_file ClimaCoupler.jl/config/ci_configs/target_amip_albedo_function.yml --job_id target_amip_albedo_function


.github/workflows/downstream.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: julia-actions/cache@v2
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: actions/checkout@v4
+        with:
+          repository: 'CliMA/ClimaCoupler.jl'
+          path: ClimaCoupler.jl
+      - run: |
+          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ -e 'using Pkg; Pkg.instantiate()'
+          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ -e 'using Pkg; Pkg.develop(; path = ".")'
+          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ experiments/ClimaEarth/run_amip.jl --config_file ClimaCoupler.jl/config/ci_configs/target_amip_albedo_function.yml --job_id target_amip_albedo_function
```
Sbozzolo, line 33:
```suggestion
          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ ClimaCoupler.jl/experiments/ClimaEarth/run_amip.jl --config_file ClimaCoupler.jl/config/ci_configs/target_amip_albedo_function.yml --job_id target_amip_albedo_function
```

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  test:
+    name: downstream ClimaCoupler.jl
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v1
```
Sbozzolo, line 21:
```suggestion
      - uses: julia-actions/setup-julia@v2
```

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          version: '1.10'
+      - uses: julia-actions/cache@v2
```
Sbozzolo, line 24: You also need this 

```
# Needed to allow julia-actions/cache to delete old caches that it has created
permissions:
  actions: write
  contents: read
```

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: actions/checkout@v4
+        with:
+          repository: 'CliMA/ClimaCoupler.jl'
+          path: ClimaCoupler.jl
+      - run: |
+          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ -e 'using Pkg; Pkg.instantiate()'
+          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ -e 'using Pkg; Pkg.develop(; path = ".")'
+          julia --color=yes --project=ClimaCoupler.jl/experiments/ClimaEarth/ ClimaCoupler.jl/experiments/ClimaEarth/run_amip.jl --config_file ClimaCoupler.jl/config/ci_configs/target_amip_albedo_function.yml --job_id target_amip_albedo_function
```
Sbozzolo, line 33: Ideally, we would want to have a leaner test. For example, the `ClimaEarth` env is already massive and takes a long time to instantiate.

It's fine for the time bein

Sbozzolo (approved):

.github/workflows/downstream.yml
juliasloan25: Agreed

.github/workflows/downstream.yml
juliasloan25: does `julia-actions/julia-buildpkg@v1` need to be updated too?

also, I copied this directly from ClimaUtilities, which is still using v1. Maybe we should update there too

Sbozzolo: Yes, I already opened a PR there.

And no, `julia-buildpkg` is still at version 1.

PR #682, opened 2024-06-28
ClimaLand Diagnostics, update

AlexisRenchon:
## Purpose 
This is a second major PR to introduce diagnostics in ClimaLand. 

To do:
- [ ] union type ::BucketModel, ::SoilCanopyModel
- [ ] more tests
- [ ] better docs
- [ ] macro to reduce code duplication in compute_ and add_diag ( ? )
- [ ] fix names, units, comments of SoilCanopyModel diagnostics variables
- [ ] add more default methods, period of averaging, what variables, etc.


AlexisRenchon:
@Sbozzolo just listing our to do list for now.

PR #685, opened 2024-07-02
renaming

kmdeck:
## Purpose 
Close #660 

We have the function `imp_tendency!`, which calls `update_aux!`, `update_boundary_fluxes!`, and `compute_imp_tendency!` (where the actual tendency computation is done). We also have the explicit analogs.

Similarly, we have `tendency_jacobian!`, which calls `update_aux!`, `update_boundary_fluxes!`, and `update_jacobian!` (where the actual tendency computation is done).

## Content
Renames two functions
`update_jacobian -> compute_jacobian` (analog is `compute_imp_tendency`)
`tendency_jacobian -> jacobian` (analog is `imp_tendency`)

----
- [x] I have read and checked the items on the review checklist.

PR #689, opened 2024-07-05
update version to v0.13.0

AlexisRenchon:
## Purpose 

New major version (0.13.0) following jacobian naming change #685


kmdeck (approved):
once the coupler downstream test passes, looks good to me!

akshaysridhar:
We can follow up with a PR updating the compat restriction here: 
https://github.com/CliMA/ClimaCoupler.jl/blob/6541ec571207ca5ca5d06fbb25004a83b39d92b3/experiments/ClimaEarth/Project.toml#L63C1-L63C19

PR #690, opened 2024-07-05
spatially varying and consistent biogeochemistry

kmdeck:
## Purpose 
The only parameters of the biogeochemsitry model which are spatially varying currently are those related to soil properties.  This PR makes them spatially varying (i.e. supporting parameter fields), and it moves them to the "soil driver struct". 

We then make the soil driver struct in the LandModel constructor using the soil parameters themselves, which enforces consistency

## Content
- Move porosity, theta_a100, and b out of the SoilCO2 Parameters
- Move them into PrescribedMet, PrognosticMet (what we call the "soil driver" for Biogeochemistry)
- access them via the driver in the soil CO2 update aux function, rather than from params struct
- add check in integrated model constructor that they are the same - not needed, the constructor handles this.
- update tests

----

### DIFF
--- a/docs/tutorials/integrated/soil_canopy_tutorial.jl
+++ b/docs/tutorials/integrated/soil_canopy_tutorial.jl
@@ -174,7 +174,7 @@ soil_model_type = Soil.EnergyHydrology{FT}
 # The domain is defined similarly to the soil domain described above.
 soilco2_type = Soil.Biogeochemistry.SoilCO2Model{FT}
 
-soilco2_ps = SoilCO2ModelParameters(FT; ν = soil_ν);
+soilco2_ps = SoilCO2ModelParameters(FT);
 
 # soil microbes args
 Csom = (z, t) -> eltype(z)(5); # kg C m⁻³, this is a guess, not measured at the site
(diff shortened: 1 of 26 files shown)


docs/tutorials/integrated/soil_canopy_tutorial.jl
```diff
@@ -306,7 +299,7 @@ canopy_model_args = (; parameters = shared_params, domain = canopy_domain);
 # atmospheric and radiative flux conditions from the observations at the Ozark
 # site as was done in the previous tutorial.
 
-land_input = (atmos = atmos, radiation = radiation)
+land_input = (atmos = atmos, radiation = radiation, soil_organic_carbon = Csom)
```
kmdeck, line 302: we pass this in, now, and the soil co2 driver is created in the LandModel constructor

ext/CreateParametersExt.jl
```diff
@@ -484,10 +484,6 @@ SoilCO2ModelParameters(::Type{FT}; kwargs...) where {FT <: AbstractFloat} =
     SoilCO2ModelParameters(CP.create_toml_dict(FT); kwargs...)
 
 function SoilCO2ModelParameters(toml_dict::CP.AbstractTOMLDict; kwargs...)
-    # These parameters have defaults that should not go in ClimaParams
-    θ_a100 = 0.1816
```
kmdeck, line 488: These are now computed directly from the soil parameters, in "prognostic soil" mode, or passed in in prescribed mode.

src/integrated/soil_canopy_model.jl
```diff
@@ -165,18 +165,20 @@ function SoilCanopyModel{FT}(;
...
             canopy_model_args...,
         )
     end
 
-    soilco2 = Soil.Biogeochemistry.SoilCO2Model{FT}(; soilco2_args...)
-
-    if !(soilco2_args.drivers.met isa PrognosticMet)
-        throw(AssertionError("Must be of type PrognosticMet."))
-    end
+    co2_prognostic_soil = Soil.Biogeochemistry.PrognosticMet(soil.parameters)
```
kmdeck, line 175: created using the soil parameters directly

src/integrated/soil_energy_hydrology_biogeochemistry.jl
```diff
@@ -38,35 +38,56 @@ Additional arguments, like parameters and driving atmospheric data, can be passe
...
+    b::F
+    function PrognosticMet(
+        soil_params::Soil.EnergyHydrologyParameters{FT},
+    ) where {FT}
+        ν = soil_params.ν
+        θ_r = soil_params.θ_r
+        hcm = soil_params.hydrology_cm
+        F = typeof(ν)
+        if F <: AbstractFloat
+            θ_a100 =
```
kmdeck, line 80: different behavior required if scalars or fields

AlexisRenchon (approved):
I have read through. 
I think this is great! Thank you!
I don't have comments for the PR (ok to merge), but we will talk about it in person

PR #692, opened 2024-07-07
Generalize drivers; make driver for soil organic carbon

kmdeck:
## Purpose 
1. Replace PrescribedSOC (which only accepted analytically prescribed SOC) with PrescribedSoilOrganicCarbon, which now accepts a TimeVaryingInput
2. SOC is now update with a driver callback and NOT every step in update_aux!
3. Removes allocation of a field at every timestep since p.drivers.soc is now preallocated and updated in place.

## Content
1. Our driver functionality was hardcoded to have two drivers: atmos, and radiation. Functions expected and returned two arguments, one for atmos, one for radation, set to `nothing` if not needed. Now we pass in  general tuples of any length (atmos, radiation, SOC, ...etc), and an empty tuple indicates no drivers needed. This required a couple new functions (new methods of make_update_drivers and initialize_drivers)
2. Adds a new driver (PrescribedSoilOrganicCarbon) and driver methods for this
3. Removes PrescribedSOC from Biogeochemistry
4. updates all simulations and tests

----


NEWS.md
```diff
@@ -3,6 +3,7 @@ ClimaLand.jl Release Notes
 
 main
 --------
+- PR[#692] Generalize our forcing ``drivers" to include prescribed soil organic carbon
```
AlexisRenchon, line 6: This PR will have breaking change right? 
So we should add `![][badge-💥breaking]`

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -581,7 +553,7 @@ function ClimaLand.boundary_flux(
 end
 
 function ClimaLand.get_drivers(model::SoilCO2Model)
-    return (model.driver.atmos, nothing)
+    return (model.driver.atmos, model.driver.soc)
```
AlexisRenchon, line 556: Consistency question: 
we have p.drivers (with an s) and model.driver (with no s)
is this on purpose?

kmdeck: I noticed that in the previous PR, and I think it should be "drivers" everywhere, now! thanks for catching!

NEWS.md
kmdeck: yes! let me add that

AlexisRenchon (approved):
This looks great. Thanks! I am looking forward to working more on it.

akshaysridhar:
@kmdeck If you start an issue tracking this PR in `ClimaCoupler` we can fix compatibility issues in the Coupler with @AlexisRenchon. Thanks!

PR #699, opened 2024-07-16
Add SoilCanopy diagnostics

AlexisRenchon:
## Purpose 

This is a second major PR to introduce diagnostics in ClimaLand. 

To do:
- [x] Update SoilCanopyModel weekly long run with :long output_vars - test if it works!
- [x] docs: optional kwarg output_vars = :short or :long --> write it so it is called and written in docs, output_period = :hour or :month or ...
- [x] add diagnostics to experiments/integrated/global/global_soil_canopy.jl, plots like Bucket, make sure it is grabbed by buildkite
- [x] macro to reduce code duplication in compute_ and add_diag ( ? )
- [x] fix names, units, comments of SoilCanopyModel diagnostics variables
- [x] add more default methods, period of averaging, what variables, etc.
- [x] Update where variable are stored if needed, add SIF

next PR 1:
- [ ] Add @has_error macro
- [ ] Somewhere, say which variable is 2D and which one is 3D (i.e., which has depth)
- [ ] 2nd round of names, units, comment etc. review, address the one still WIP in this PR (that need further discussion with Renato and Kat)
- [ ] Kat: we also have some runoff variables, but where or not they are present depends on the type of RunoffModel we choose. maybe this is something we can accomodate? E.g. NoRunoff() has no additional variables, while TopModel has subsurface, surface runoff in p.soil
- [ ] Make sure carbon flux units are consistent, and chose: mol co2 m-2 s-1? Co2 or c? mol or kg?

next PR 2:
- [ ] more tests
- [ ] better docs
- [ ] use diagnostics in all files (need to update experiments, tests, etc.)
- [ ] Add land-sea mask as output by default, might be in the domain? (where to compute it? --> ask Kat)

next PR 3:
- [ ] use actual months https://github.com/CliMA/ClimaDiagnostics.jl/pull/67


AlexisRenchon:
@kmdeck 
@braghiere 

Could you just look at:
`land_compute_methods.jl` - do I have everything we compute in p and Y? Does the code organization look good?

`define_diagnostics.jl` - Are the names, units and comments good?

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,146 @@
...
+@land_compute "near_infrared_radiation_transmitted" SoilCanopyModel p.canopy.radiative_transfer.tnir
+@land_compute "photosynthetically_active_radiation" SoilCanopyModel p.canopy.radiative_transfer.par
+@land_compute "photosynthetically_active_radiation_absorbed" SoilCanopyModel p.canopy.radiative_transfer.apar
+@land_compute "photosynthetically_active_radiation_reflected" SoilCanopyModel p.canopy.radiative_transfer.rpar
+@land_compute "photosynthetically_active_radiation_transmitted" SoilCanopyModel p.canopy.radiative_transfer.tpar
+@land_compute "radiation_longwave_net" SoilCanopyModel p.canopy.radiative_transfer.LW_n
+@land_compute "radiation_shortwave_net" SoilCanopyModel p.canopy.radiative_transfer.SW_n
+
+# Canopy - Transpiration
+@land_compute "canopy_transpiration" SoilCanopyModel p.canopy.transpiration
```
AlexisRenchon, line 105: If this is the only variable not in a canopy namespace, should we move it to e.g., conductance?

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,146 @@
...
+@land_compute "radiation_longwave_down" SoilCanopyModel p.drivers.LW_d
+@land_compute "radiation_shortwave_down" SoilCanopyModel p.drivers.SW_d
+@land_compute "snowfall" SoilCanopyModel p.drivers.P_snow
+@land_compute "solar_zenith_angle" SoilCanopyModel p.drivers.θs
+@land_compute "specific_humidity" SoilCanopyModel p.drivers.q
+@land_compute "wind_speed" SoilCanopyModel p.drivers.u
+
+## Soil Module ##
+
+@land_compute "infiltration" SoilCanopyModel p.soil.infiltration
```
AlexisRenchon, line 121: also here a thought: do we want namespace or keep directly p.soil.X

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,146 @@
...
+@land_compute "soil_latent_heat_flux" SoilCanopyModel p.soil.turbulent_flux.lhf
+@land_compute "soil_sensible_heat_flux" SoilCanopyModel p.soil.turbulent_fluxes.shf
+@land_compute "vapor_flux" SoilCanopyModel p.soil.turbulent_fluxes.vapor_flux
+
+# Soil - SoilCO2
+@land_compute "heterotrophic_respiration" SoilCanopyModel p.soilco2.top_bc
+@land_compute "soilco2_diffusivity" SoilCanopyModel p.soilco2.D
+@land_compute "soilco2_source_microbe" SoilCanopyModel p.soilco2.Sm
+
+# variables stored in Y (prognostic or state variables)
```
AlexisRenchon, line 142: Do we really only have 3? 
Surely not. for example, it seems like soil moisture is missing below.

AlexisRenchon:
@kmdeck @Sbozzolo 

Should we add a test that check if all variables of a model (e.g., SoilCanopyModel) are defined in the diagnostics? 
That would be good to ensure in CI that if we add a new variable, it has to be in diagnostics

Sbozzolo:
> @kmdeck @Sbozzolo
> 
> Should we add a test that check if all variables of a model (e.g., SoilCanopyModel) are defined in the diagnostics? That would be good to ensure in CI that if we add a new variable, it has to be in diagnostics

There is no requirement for variables defined to be in the model to be in the diagnostics too: only those that are scientifically interesting

src/diagnostics/land_compute_methods.jl
kmdeck: For soil, we just have p.soil.Var, and not p.soil.energy.Var or p.soil.water.Var. 

I think this makes sense because we dont currently have different "components" of the soil model which require modularity in the same way that we do for the canopy, but we could have designed that. it might be nice at some point (components could be water, energy, CO2, ...).

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,146 @@
...
+@land_compute "solar_zenith_angle" SoilCanopyModel p.drivers.θs
+@land_compute "specific_humidity" SoilCanopyModel p.drivers.q
+@land_compute "wind_speed" SoilCanopyModel p.drivers.u
+
+## Soil Module ##
+
+@land_compute "infiltration" SoilCanopyModel p.soil.infiltration
+@land_compute "soil_hydraulic_conductivity" SoilCanopyModel p.soil.K
+@land_compute "soil_thermal_conductivity" SoilCanopyModel p.soil.κ
+@land_compute "soil_water_liquid" SoilCanopyModel top_center_to_surface(
```
kmdeck, line 124: just a note that this only returns the top surface value (2D), and not the whole field (3D), while all others return the whole field at each layer. We may want the 3D field as long as there is an easy way to plot just the surface value if we want it.

I had added this in because the diagnostics previously were making plots of 2d fields only

src/diagnostics/land_compute_methods.jl
kmdeck: Y.soil.\rhoe_int
Y.soil.\vartheta_l
Y.soil.\theta_i
Y.soilco2.C
(only four for soil)

canopy T and canopy water content are prognostic

src/diagnostics/land_compute_methods.jl
AlexisRenchon: I will remove it from diagnostics for now and only keep the 3D one!

src/diagnostics/land_compute_methods.jl
AlexisRenchon: Right! 
No need to make any change until it becomes useful - but let's keep it in mind =)

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,146 @@
...
+## Canopy Module ##
+
+# Canopy - Solar Induced Fluorescence
+@land_compute "solar_induced_fluorescence" SoilCanopyModel p.canopy.sif.SIF
+
+# Canopy - Autotrophic respiration
+@land_compute "autotrophic_respiration" SoilCanopyModel p.canopy.autotrophic_respiration.Ra
+
+# Canopy - Conductance
+@land_compute "medlyn_term" SoilCanopyModel p.canopy.conductance.medlyn_term
```
kmdeck, line 71: Im not sure we need this one, we care more about gs and transpiration

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,146 @@
...
+
+# Canopy - Energy
+@land_compute "canopy_aerodynamic_resistance" SoilCanopyModel p.canopy.energy.r_ae
+@land_compute "canopy_latent_heat_flux" SoilCanopyModel p.canopy.energy.lhf
+@land_compute "canopy_sensible_heat_flux" SoilCanopyModel p.canopy.energy.shf
+
+# Canopy - Hydraulics
+@land_compute "area_index" SoilCanopyModel p.canopy.hydraulics.area_index.leaf
+@land_compute "canopy_water_potential" SoilCanopyModel p.canopy.hydraulics.ψ
+@land_compute "cross_section" SoilCanopyModel p.canopy.hydraulics
```
kmdeck, line 82: not sure about this, unfinished?

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,146 @@
...
+@land_compute "cross_section_roots" SoilCanopyModel p.canopy.hydraulics.fa_roots
```
kmdeck, line 83: maybe RAI_scaled_water_flux?

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,146 @@
...
+@land_compute "moisture_stress_factor" SoilCanopyModel p.canopy.hydraulics.β
+
+# Canopy - Photosynthesis
+@land_compute "photosynthesis_net_canopy" SoilCanopyModel p.canopy.photosynthesis.GPP
+@land_compute "photosynthesis_net_leaf" SoilCanopyModel p.canopy.photosynthesis.An
+@land_compute "respiration_leaf" SoilCanopyModel p.canopy.photosynthesis.Rd
+@land_compute "vcmax25" SoilCanopyModel p.canopy.photosynthesis.vcmax25
```
kmdeck, line 90: maybe Vcmax25? (capitalized)

src/diagnostics/land_compute_methods.jl
kmdeck: we can revisit this, but maybe a second PR? or maybe even plant hydraulics

src/diagnostics/land_compute_methods.jl
kmdeck: 💯 !!

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,146 @@
...
+)
+@land_compute "soil_water_potential" SoilCanopyModel p.soil.ψ
+@land_compute "soil_net_radiation" SoilCanopyModel p.soil.R_n
+@land_compute "soil_temperature" SoilCanopyModel p.soil.T
+
+# Soil - Turbulent Fluxes
+@land_compute "soil_aerodynamic_resistance" SoilCanopyModel p.soil.turbulent_fluxes.r_ae
+@land_compute "soil_latent_heat_flux" SoilCanopyModel p.soil.turbulent_flux.lhf
+@land_compute "soil_sensible_heat_flux" SoilCanopyModel p.soil.turbulent_fluxes.shf
+@land_compute "vapor_flux" SoilCanopyModel p.soil.turbulent_fluxes.vapor_flux
```
kmdeck, line 132: we also have some runoff variables, but where or not they are present depends on the type of RunoffModel we choose. maybe this is something we can accomodate? E.g. NoRunoff() has no additional variables, while TopModel has subsurface, surface runoff in p.soil

src/diagnostics/land_compute_methods.jl
AlexisRenchon: I agree... I will remove it for now

Artifacts.toml
```diff
@@ -1,16 +1,19 @@
+[ameriflux_data_US-MOz]
```
kmdeck, line 1: not sure this file should change

src/diagnostics/define_diagnostics.jl
```diff
@@ -189,36 +194,40 @@ function define_diagnostics!(land_model)
...
     )
 
+    # Infiltration
     add_diagnostic_variable!(
         short_name = "infil",
         long_name = "Infiltration",
         standard_name = "infiltration",
-        units = "m s^-1", # double check
+        units = "m s^-1",
         comments = "The flux of liquid water into the soil.",
```
kmdeck, line 203: looks good. might say "volume flux of liquid water"

AlexisRenchon: hmm a flux is the amount of something going through a surface over some period of time. 
It is not always a volume (e.g., flux of photons), but for water, it usually is a volume? 
(it feels weird to me to say "volume flux"). But I agree (m s^-1) can be a bit unintuitive. I'll add (m^3 or water per m^2 of ground per second).

Artifacts.toml
AlexisRenchon: uh I am not sure what happened - I'll remove this

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
+	    # Solar Induced Fluorescence
     add_diagnostic_variable!(
-        short_name = "slw",
-        long_name = "Soil Liquid Water",
-        standard_name = "soil_liquid_water",
-        units = "m^3 m^-3",
-        comments = "Soil moisture, the liquid water volume per soil volume. This liquid water is located in the soil pores.",
+        short_name = "sif",
+        long_name = "Solar Induced Fluorescence",
+        standard_name = "solar_induced_fluorescence",
```
kmdeck, line 193: should we say at 755 nm?

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
     add_diagnostic_variable!(
-        short_name = "scd",
-        long_name = "Soil CO2 Diffusivity",
-        standard_name = "soil_co2_diffusivity",
-        units = "", # need to add
-        comments = "The diffusivity of CO2 in the porous phase of the soil. Depends on soil texture, moisture, and temperature.",
+        short_name = "gs",
+        long_name = "Stomatal Conductance",
+        standard_name = "stomatal_conductance",
+        units = "m s^-1",
```
kmdeck, line 218: mol/m^2/s
https://github.com/CliMA/ClimaLand.jl/blob/main/src/standalone/Vegetation/canopy_parameterizations.jl#L908

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
     add_diagnostic_variable!(
-        short_name = "infil",
-        long_name = "Infiltration",
-        standard_name = "infiltration",
-        units = "m s^-1", # double check
-        comments = "The flux of liquid water into the soil.",
+        short_name = "ra",
+        long_name = "Autotrophic Respiration",
+        standard_name = "autotrophic_respiration",
+        units = "mol CO2 m^-2 s^-1", # note throughout: make sure what we express as kg or mol, CO2 or C
```
kmdeck, line 206: I think this may be kg of C.
I think we have a bug in the units in the code. Maybe worth fixing in this PR?
https://github.com/CliMA/ClimaLand.jl/blob/main/src/standalone/Vegetation/canopy_parameterizations.jl#L1004 [maintenance is in kg/m^2/s]
but here it is used with An: https://github.com/CliMA/ClimaLand.jl/blob/main/src/standalone/Vegetation/canopy_parameterizations.jl#L1040
and An has units of mol CO2.

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
+            compute_stomatal_conductance!(out, Y, p, t, land_model),
+    )
+	
+        # Canopy transpiration
+    add_diagnostic_variable!(
+        short_name = "trans",
+        long_name = "Canopy Transpiration",
+        standard_name = "canopy_transpiration",
+        units = "m s^-1",
+        comments = "The water evaporated from the canopy due to leaf transpiration.",
```
kmdeck, line 230: Also expressed as a volume flux of liquid water

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
+            compute_canopy_transpiration!(out, Y, p, t, land_model),
+    )	
+	
+	### Canopy - Energy
+	    # Canopy aerodynamic resistance
+    add_diagnostic_variable!(
+        short_name = "crae",
+        long_name = "Canopy Aerodynamic Resistance",
+        standard_name = "canopy_aerodynamic_resistance",
+        units = "s m^-1", # ??? not sure
```
kmdeck, line 241: I think so

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
     add_diagnostic_variable!(
-        short_name = "gs",
-        long_name = "Stomatal Conductance",
-        standard_name = "stomatal_conductance",
-        units = "m s^-1",
-        comments = "The conductance of leaves. This depends on stomatal opening. It varies with factors such as soil moisture or atmospheric water demand.",
+        short_name = "cwp",
+        long_name = "Canopy Water Potential",
+        standard_name = "canopy_water_potential",
+        units = "Pa",
```
kmdeck, line 275: this one is measured in `m` actually - same as soil -  but we get Pa when needed by multiplying by rho_l *g

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
     add_diagnostic_variable!(
-        short_name = "mt",
-        long_name = "Medlyn Term",
-        standard_name = "medlyn_term",
-        units = "", # check
-        comments = "",
+        short_name = "fa",
+        long_name = "Cross Section",
+        standard_name = "cross_section",
+        units = "m^2 m^-2",
```
kmdeck, line 286: this is a flux*cross section. flux is in water volume per meter^2 of plant per second, and then we multiply by the area index (plant area/ground area) to get a flux per area ground. net result is m/s

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
     add_diagnostic_variable!(
-        short_name = "trans",
-        long_name = "Canopy Transpiration",
-        standard_name = "canopy_transpiration",
-        units = "", # check
-        comments = "The water evaporated from the canopy due to leaf transpiration.",
+        short_name = "far",
+        long_name = "Root Cross Section",
+        standard_name = "Cross Section",
+        units = "m^2 m^-2",
```
kmdeck, line 297: same comment, but this one is the flux from the ground to the roots

src/diagnostics/define_diagnostics.jl
kmdeck: note that this and canopy.hydraulics.psi, canopy.hydraulics.vartheta are tuples.

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
         compute! = (out, Y, p, t) -> compute_vcmax25!(out, Y, p, t, land_model),
     )
 
+    ### Canopy - Radiative Transfer
+	    # NIR - near infrared radiaton
+    add_diagnostic_variable!(
+        short_name = "nir",
+        long_name = "Near Infrared Radiation",
+        standard_name = "near_infrared_radiation",
+        units = "W m^-2",
```
kmdeck, line 397: I think NIR, ANIR, TNIR, RNIR and same for Par are:
mol photons per m^2 per 
second for a radiation band.

src/diagnostics/define_diagnostics.jl
```diff
@@ -362,177 +513,295 @@ function define_diagnostics!(land_model)
...
     add_diagnostic_variable!(
-        short_name = "rnir",
-        long_name = "Reflected Near Infrared Radiation",
-        standard_name = "reflected_near_infrared_radiation",
-        units = "W m^-2",
-        comments = "The amount of near infrared radiation reaching the canopy.",
+        short_name = "rain",
+        long_name = "Rainfall",
+        standard_name = "rainfall",
+        units = "m s^-1", # do we express rainfall as a flux? Check then delete this comment
```
kmdeck, line 564: yes! a flux of liquid water volume. same for snow

src/diagnostics/define_diagnostics.jl
AlexisRenchon: We mention this before, but to be correct: 
PAR (radiation) should be `W m^-2`
PPFD (photons flux density) should be `mol photons m^2 s^-1`
I don't really mind and I understand we can name something and chose its units, but I've been asked by reviewers to correct this before (so it might happen again).

AlexisRenchon: (I updated to mol photons m^-2 s^-1 as this is what we have)

src/diagnostics/define_diagnostics.jl
AlexisRenchon: this is not `p.canopy.hydraulics.fa` ?

src/diagnostics/define_diagnostics.jl
AlexisRenchon: I think it would be good to decide together what to do with flux of CO2 @braghiere , do we express them is `mol CO2`, `kg C`, or other? 
These are all SI, so we just have to pick one. 
Usually, in the literature, we use mol CO2 for fluxes (umol), and kg C for budgets (change in pool size or pool size). 
I would vote for mol CO2, but let's all agree.

src/diagnostics/define_diagnostics.jl
AlexisRenchon: amended to `mol H2O m^-2 s^-1`

src/diagnostics/define_diagnostics.jl
AlexisRenchon: @braghiere 

I am not sure, if future implementations could be other wavelength as well?

src/diagnostics/land_compute_methods.jl
AlexisRenchon: I think this is fine - user decide which diagnostic variable to get, so we can put runoff here, and then user can grab it or not... 
For the default, I will think about how to handle this, I am not sure

src/diagnostics/land_compute_methods.jl
AlexisRenchon: And for fa? Plant_scaled_water_flux or Stem_scaled_water_flux?

src/diagnostics/land_compute_methods.jl
AlexisRenchon: Is soil temperature prognostic? 
Or is it diagnostic, calculated from rhoe_int?

src/diagnostics/land_compute_methods.jl
AlexisRenchon: it is actually already in conductance. I am not sure if this was a typo or if it has been moved at some point

src/diagnostics/define_diagnostics.jl
braghiere: Yes. This is only for 755nm. SIF can be calculate in multiple wavelengths.

src/diagnostics/define_diagnostics.jl
braghiere: Yep. Carbon fluxes is usually in umolCO2.m-2.s-1 (or kgCm-2-day-1) and pools are in kgC.

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,145 @@
...
+    return quote
+        $(functions...)
+    end
+end
+
+### BucketModel ###
+
+# variables stored in p (diagnostics variables stored in the cache)
+@land_compute "aerodynamic_resistance" BucketModel p.bucket.turbulent_fluxes.r_ae
+@land_compute "albedo" BucketModel p.bucket.α_sfc
```
braghiere, line 43: For this we want to be specific about the band. Is it shortwave (400-2500nm), PAR or NIR?

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,145 @@
...
+@land_compute "latent_heat_flux" BucketModel p.bucket.turbulent_fluxes.lhf
+@land_compute "net_radiation" BucketModel p.bucket.R_n
+@land_compute "sensible_heat_flux" BucketModel p.bucket.turbulent_fluxes.shf
+@land_compute "surface_air_density" BucketModel p.bucket.ρ_sfc
+@land_compute "surface_specific_humidity" BucketModel p.bucket.q_sfc
+@land_compute "surface_temperature" BucketModel p.bucket.T_sfc
+@land_compute "vapor_flux" BucketModel p.bucket.turbulent_fluxes.vapor_flux
```
braghiere, line 50: This is also a bit unusual. Water vapor flux? Can we call it water flux? Is it ET?

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,145 @@
...
+@land_compute "stem_area_index" SoilCanopyModel p.canopy.hydraulics.area_index.stem
+
+# Canopy - Photosynthesis
+@land_compute "photosynthesis_net_canopy" SoilCanopyModel p.canopy.photosynthesis.GPP
+@land_compute "photosynthesis_net_leaf" SoilCanopyModel p.canopy.photosynthesis.An
+@land_compute "respiration_leaf" SoilCanopyModel p.canopy.photosynthesis.Rd
+@land_compute "vcmax25" SoilCanopyModel p.canopy.photosynthesis.Vcmax25
+
+# Canopy - Radiative Transfer
+@land_compute "near_infrared_radiation" SoilCanopyModel p.canopy.radiative_transfer.inc_nir
```
braghiere, line 95: Maybe this could be called down as well? Like SW_down?

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,145 @@
...
+@land_compute "near_infrared_radiation_absorbed" SoilCanopyModel p.canopy.radiative_transfer.nir.abs
+@land_compute "near_infrared_radiation_reflected" SoilCanopyModel p.canopy.radiative_transfer.nir.refl
+@land_compute "near_infrared_radiation_transmitted" SoilCanopyModel p.canopy.radiative_transfer.nir.trans
+@land_compute "photosynthetically_active_radiation" SoilCanopyModel p.canopy.radiative_transfer.inc_par
```
braghiere, line 99: same here. PARdown

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,145 @@
...
+@land_compute "soil_temperature" SoilCanopyModel p.soil.T
+
+# Soil - Turbulent Fluxes
+@land_compute "soil_aerodynamic_resistance" SoilCanopyModel p.soil.turbulent_fluxes.r_ae
+@land_compute "soil_latent_heat_flux" SoilCanopyModel p.soil.turbulent_flux.lhf
+@land_compute "soil_sensible_heat_flux" SoilCanopyModel p.soil.turbulent_fluxes.shf
+@land_compute "vapor_flux" SoilCanopyModel p.soil.turbulent_fluxes.vapor_flux
+
+# Soil - SoilCO2
+@land_compute "heterotrophic_respiration" SoilCanopyModel p.soilco2.top_bc
```
braghiere, line 134: why is it called top_bc?

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
+
+    ### Canopy - Photosynthesis
+    # GPP - Gross Primary Productivity
     add_diagnostic_variable!(
         short_name = "gpp",
         long_name = "Gross Primary Productivity",
         standard_name = "gross_primary_productivity",
-        units = "",
-        comments = "Net photosynthesis of the canopy.",
+        units = "mol CO2 m^-2 s^-1",
```
braghiere, line 353: GPP, HR, AR should all be in the same units

src/diagnostics/define_diagnostics.jl
```diff
@@ -179,130 +181,276 @@ function define_diagnostics!(land_model)
...
+            ),
+    )
+
+    # PAR - photosynthetically active radiation
     add_diagnostic_variable!(
         short_name = "par",
         long_name = "Photosynthetically Active Radiation",
         standard_name = "photosynthetically_active_radiation",
-        units = "",
+        units = "mol photons m^-2 s^-1",
```
braghiere, line 453: For diagnostics, energy is usually in W.m-2. Much easier to compare with observations.

src/diagnostics/define_diagnostics.jl
```diff
@@ -362,284 +513,317 @@ function define_diagnostics!(land_model)
...
         compute! = (out, Y, p, t) ->
             compute_soil_ice!(out, Y, p, t, land_model),
     )
+
+    # Soil internal energy
+    add_diagnostic_variable!(
+        short_name = "sie",
+        long_name = "Soil Internal Energy",
+        standard_name = "soil_internal_energy",
+        units = "", # need units
```
braghiere, line 823: If it's energy should be W.m-2.

src/diagnostics/land_compute_methods.jl
AlexisRenchon: top boundary condition 

In the equations we calculate CO2 production by microbial respiration at each layer of soil, which produces CO2 (gain), CO2 also diffuses through layers (gain and loss), we calculate CO2 concentration in pores at each layers of soil. 

But most observation of "soil respiration" measure CO2 efflux from the soil, not actual CO2 production by microbes, which can be delayed from production to efflux from surface because of diffusion. In our model diffusion is a function of soil texture (clay soil slower diffusion than sandy soils, for example) and soil moisture (moist soil slower diffusion) and soil temperature. 

So we don't directly calculate soil CO2 efflux in the model, but we can get it from the top boundary condition (that we call top_bc. We have that for all prognostic variables that are calculated by layers in the soil domain)

src/diagnostics/land_compute_methods.jl
AlexisRenchon: NIR_down you mean - will do, I agree.

src/diagnostics/land_compute_methods.jl
AlexisRenchon: @kmdeck

src/diagnostics/land_compute_methods.jl
AlexisRenchon: @kmdeck

src/diagnostics/define_diagnostics.jl
AlexisRenchon: Definitely. I'll make sure they are all in the same units

src/diagnostics/define_diagnostics.jl
AlexisRenchon: @kmdeck 
Currently we are computing PAR and NIR and others in mol photons m-2 s-1. 
See for example https://github.com/CliMA/ClimaLand.jl/blob/main/src/standalone/Vegetation/canopy_parameterizations.jl#L623

We could convert it to W m-2. Do you think we should? 
(it's easy to convert in data analysis, and we are probably going to convert most units anyway because we use SI)

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,145 @@
+"""
+    @land_compute name model compute
+
+
```
Sbozzolo, line 4:
```suggestion

```

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,144 @@
...
+            else
+                out .= $compute
+            end
+        end
+    end)
+    push!(functions, function_body)
+    return quote
+        $(functions...)
+    end
+end
```
Sbozzolo, line 36:
```suggestion
macro land_compute(name, model, compute)
   function_name = Symbol("compute_", name, "!")
    return esc(quote
        function $function_name(out, Y, p, t, land_model::$model)
            if isnothing(out)
                return copy($compute)
            else
                out .= $compute
            end
        end
    end)
end
```

.buildkite/Project.toml
```diff
@@ -1,4 +1,5 @@
 [deps]
+AbbreviatedStackTraces = "ac637c84-cc71-43bf-9c33-c1b4316be3d4"
```
Sbozzolo, line 2: Please remove this before when you clean things up

Artifacts.toml
```diff
@@ -4,6 +4,9 @@ git-tree-sha1 = "ec424296df6b60cfe273ac8f981701fbbed0bd8a"
 [land_albedo]
 git-tree-sha1 = "3f07b70ab43a05123e93753a45e094dcf7a66b5b"
 
+["processed_topographic_index 2.5 deg"]
+git-tree-sha1 = "0c252917b24ce3c201fa62d639342ab4d872b388"
+
```
Sbozzolo, line 9: This shouldn't be related to this PR

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,144 @@
...
+@land_compute "stomatal_conductance" SoilCanopyModel p.canopy.conductance.gs
+@land_compute "canopy_transpiration" SoilCanopyModel p.canopy.conductance.transpiration
+
+# Canopy - Energy
+@land_compute "canopy_aerodynamic_resistance" SoilCanopyModel p.canopy.energy.r_ae
+@land_compute "canopy_latent_heat_flux" SoilCanopyModel p.canopy.energy.lhf
+@land_compute "canopy_sensible_heat_flux" SoilCanopyModel p.canopy.energy.shf
+
+# Canopy - Hydraulics
+# @land_compute "canopy_water_potential" SoilCanopyModel p.canopy.hydraulics.ψ # return a Tuple
```
Sbozzolo, line 79: This is the leaf water potential (the last element of the tuple)

```suggestion
# @land_compute "canopy_water_potential" SoilCanopyModel last(p.canopy.hydraulics.ψ) 
```

experiments/integrated/global/global_soil_canopy.jl
```diff
@@ -361,76 +366,90 @@ prob = SciMLBase.ODEProblem(
...
+    "hr",
+    "scd",
+    "scms",
+    "ct",
+    "sco2",
+    "swc",
+    "si",
+    "sie",
+]
+for short_name in short_names
```
Sbozzolo, line 449:
```suggestion
for short_name in ClimaAnalysis.available_vars(simdir)
```

experiments/long_runs/land.jl
```diff
@@ -646,13 +646,64 @@ setup_and_solve_problem(; greet = true);
...
+    "scd",
+    "scms",
+    "ct",
+    "sco2",
+    "swc",
+    "si",
+    "sie",
+]
 times = 0.0:(60.0 * 60.0 * 12):(60.0 * 60.0 * 24 * 7)
 for t in times
```
Sbozzolo, line 702: This is going to plot a lot of stuff. Can we select a subset of variables and times we care about?

src/diagnostics/default_diagnostics.jl
```diff
@@ -75,51 +75,83 @@ function default_diagnostics(
...
-            "mt",
             "trans",
-            "rain", # do we want?
-            "an",
+            "crae",
+            "clhf",
+            "cshf",
+            # "cwp",
+            # "fa",
+            # "far",
```
Sbozzolo, line 94: Let's add a comment to explain why they are commented out

src/diagnostics/define_diagnostics.jl
```diff
@@ -362,284 +507,321 @@ function define_diagnostics!(land_model)
...
     add_diagnostic_variable!(
-        short_name = "msf",
-        long_name = "Moisture Stress Factor",
-        standard_name = "moisture_stress_factor",
-        units = "",
-        comments = "Sensitivity of plants conductance to soil water content.",
+        short_name = "soilrae",
+        long_name = "Soil Aerodynamic Resistance",
+        standard_name = "soil_aerodynamic_resistance",
+        units = "s m^-1", # ??? not sure
```
Sbozzolo, line 694: Let's try to figure out the "not sure"

src/diagnostics/land_compute_methods.jl
kmdeck: Temperature is diagnostic and computed from all three prognostic variables

src/diagnostics/land_compute_methods.jl
kmdeck: For the bucket, it is a bulk shortwave albedo

src/diagnostics/land_compute_methods.jl
kmdeck: we can call it water vapor flux! I guess for the bucket it is ET. for soil, it is evaporation+sublimation, for canopy, transpiration

src/diagnostics/define_diagnostics.jl
kmdeck: I think some photosynthesis computation required APAR in moles of photons, right? 

For now I would leave source code the same, but we can convert units in diagnostics?

docs/src/diagnostics/users_diagnostics.md
```diff
@@ -58,6 +58,10 @@ sol = SciMLBase.solve(prob, ode_algo; dt = Δt, callback = diag_cb)
 
 Your diagnostics have now been written in netcdf files in your output folder.
 
+Note that by default, `default_diagnostics` assign two optional kwargs: `output_vars = :long` and `average_period` = :hourly.
```
Sbozzolo, line 61:
```suggestion
Note that by default, `default_diagnostics` assign two optional kwargs: `output_vars = :long` and `average_period` = :daily.

```

experiments/integrated/global/global_soil_canopy.jl
```diff
@@ -24,6 +24,11 @@ using Statistics
 using Dates
 import NCDatasets
 
+using ClimaDiagnostics
+using ClimaAnalysis
+import ClimaAnalysis.Visualize as viz
+using ClimaUtilities
```
Sbozzolo, line 30: I would recommend using `import` instead of `using` to keep the namespace tidier

Sbozzolo, line 30:
```suggestion
import ClimaDiagnostics
import ClimaAnalysis
import ClimaAnalysis.Visualize as viz
import ClimaUtilities
```

src/diagnostics/land_compute_methods.jl
```diff
@@ -0,0 +1,150 @@
+"""
+    @diagnostic_compute name model compute
+
+Macro generating a function to compute a land diagnostic,
+needed in the ClimaDiagnostics framework.
```
Sbozzolo, line 5:
```suggestion
Macro generating a function to compute a land diagnostic,
needed in the ClimaDiagnostics framework.

See the ClimaDiagnostics docmentation for more information.
```

Sbozzolo (approved):

PR #701, opened 2024-07-17
add boundary fluxes function for snow

kmdeck:
## Purpose 
This PR simply adds a function which computes boundary fluxes in the same way we already compute them, but within a standalone function specifically for that purpose.

We construct our tendencies in the following fashion:
(1) update "auxiliary vars" of model, stored in cache `p`
(2) update boundary fluxes (also in `p`, but use the values updated in (1))
(3) compute tendency using updated cache values, boundary conditions and `Y`

For standalone models (snow by itself, soil by itself, etc), there is no reason to split (1) and (2). However, when we combine these models, the presence of interactions between the components affect the boundary fluxes each experiences. Because of that, we split the tendency of the land model as
1. update all auxiliary vars in the cache for all components
2. update all the boundary conditions for all the components [ taking into account interactions]
3. compute tendencies for all the components (only use precomputed quantities and Y)

in preparation for the snow model being integrated with soil/canopy, create the boundary_flux function and update the boundary conditions there, rather than in aux or in the tendency itself.

## To-do
Review

## Content
move certain quantities (Rn, turbulent fluxes) computed in update_aux! and in the tendency for snow into boundary_fluxes!

Review checklist

I have:

----

### DIFF
--- a/src/standalone/Snow/Snow.jl
+++ b/src/standalone/Snow/Snow.jl
@@ -16,6 +16,7 @@ using ClimaLand:
 import ClimaLand:
     make_update_aux,
     make_compute_exp_tendency,
+    make_update_boundary_fluxes,
     prognostic_vars,
     auxiliary_vars,
     name,
@@ -39,6 +40,7 @@ Currently, the only supported concrete example is called `SnowModel`
 and is used as a bulk snow model.
 """
 abstract type AbstractSnowModel{FT} <: ClimaLand.AbstractExpModel{FT} end
+
 
 """
     SnowParameters{FT <: AbstractFloat, PSE}
@@ -156,10 +158,10 @@ end
 function SnowModel(;
     parameters::SnowParameters{FT, PSE},
     domain::ClimaLand.Domains.AbstractDomain,
-    atmosphere::ATM,
+    atmos::ATM,
     radiation::RAD,
 ) where {FT, PSE, ATM, RAD}
-    args = (parameters, atmosphere, radiation, domain)
+    args = (parameters, atmos, radiation, domain)
     SnowModel{FT, typeof.(args)...}(args...)
 end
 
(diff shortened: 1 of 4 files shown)


src/standalone/Snow/Snow.jl
```diff
@@ -236,11 +249,27 @@ auxiliary_domain_names(::SnowModel) = (
...
+
+In the future we can play around with other forms, this makes it a bit more
+explicit.
+"""
+function scf(x::FT; α = FT(1e-3))::FT where {FT}
+    return heaviside(x - α)
+end
+
 function ClimaLand.make_update_aux(model::SnowModel{FT}) where {FT}
     function update_aux!(p, Y, t)
```
kmdeck, line 274: This function does not care if about other components and is only dependent on the snow state.

src/standalone/Snow/Snow.jl
```diff
@@ -253,20 +282,19 @@ function ClimaLand.make_update_aux(model::SnowModel{FT}) where {FT}
...
             p.snow.water_runoff * volumetric_internal_energy_liq(FT, parameters)
+        @. p.snow.snow_cover_fraction = scf(Y.snow.S)
     end
 end
 
-function ClimaLand.make_compute_exp_tendency(model::SnowModel{FT}) where {FT}
-    function compute_exp_tendency!(dY, Y, p, t)
+function ClimaLand.make_update_boundary_fluxes(model::SnowModel{FT}) where {FT}
+    function update_boundary_fluxes!(p, Y, t)
+        p.snow.turbulent_fluxes .= turbulent_fluxes(model.atmos, model, Y, p, t)
```
kmdeck, line 296: This function could care about other components. For example, the net radiation will change if the snow is under a canopy.

kmdeck: (we would account for that using dispatch in `net_radiation` or similar)

szy21 (approved):

PR #704, opened 2024-07-23
Do not profile setup in benchmarks

Sbozzolo:
Split set up and runtime phases in benchmarks so that we can get a more accurate reading for the runtime costs

### DIFF
--- a/experiments/benchmarks/bucket.jl
+++ b/experiments/benchmarks/bucket.jl
@@ -141,9 +141,7 @@ function setup_prob(t0, tf, Δt; nelements = (100, 10))
     return prob, cb
 end
 
-function setup_and_solve_problem(; greet = false)
-    # We profile the setup phase as well here. This is not intended, but it is the easiest
-    # to set up for both CPU/GPU at the same time
+function setup_simulation(; greet = false)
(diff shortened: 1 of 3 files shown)


experiments/benchmarks/bucket.jl
```diff
@@ -141,9 +141,7 @@ function setup_prob(t0, tf, Δt; nelements = (100, 10))
     return prob, cb
 end
 
-function setup_and_solve_problem(; greet = false)
-    # We profile the setup phase as well here. This is not intended, but it is the easiest
-    # to set up for both CPU/GPU at the same time
+function setup_one(; greet = false)
```
kmdeck, line 144: why this function name?

kmdeck (approved):
looks good. thank you!

experiments/benchmarks/bucket.jl
Sbozzolo: `one`, as one instance of a problem that is fully ready to be solved

kmdeck: perhaps "setup_simulation" is clearer, or "setup_instance"? as you prefer.

PR #705, opened 2024-07-24
SOC can be spatially varying only

kmdeck:
## Purpose 
Our goal is to prescribed soil organic carbon from time varying data. We added this in, but our only use case currently was TimeVaryingInput((t) -> constant). Since SOC is a 3d field, I think the allocations in regridded_snapshot! led to a big slowdown with this change.
Before that PR:
https://buildkite.com/clima/climaland-benchmark/builds/709#0190a29f-b2ff-40e3-9dae-a007eae845c4 (10 seconds)
After.: https://buildkite.com/clima/climaland-benchmark/builds/744#0190b780-87fb-493d-ad30-3ee2aabf1d30 (25 seconds)

For now, I added the option to have SOC be only a spatially varying field and not use TimeVaryingInput (and hence not need regridded_snapshot!). this should speed things up for us, and also be relevant for shorter runs where SOC is not varying much, or for runs where we do not have time-varying data.

The time per simulation is now back to 10 seconds.

## Content
Allow for the prescribed SOC driver to be a field, and not a TimeVaryingInput. In this case, set p.drivers.soc = the field each evaluate step, intead of evaluating the TimeVaryingInput object at a specific time.

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@ ClimaLand.jl Release Notes
 
 main
 --------
+- Allow for simple spatially varying prescribed organic carbon, rather than require temporal variation as well.
+  PR[#705](https://github.com/CliMA/ClimaLand.jl/pull/705)
 
 v0.14.0
 --------
--- a/docs/tutorials/integrated/soil_canopy_tutorial.jl
+++ b/docs/tutorials/integrated/soil_canopy_tutorial.jl
@@ -176,8 +176,12 @@ soilco2_type = Soil.Biogeochemistry.SoilCO2Model{FT}
 
 soilco2_ps = SoilCO2ModelParameters(FT);
 
-# soil microbes args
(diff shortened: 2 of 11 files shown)


Sbozzolo:
Can you check this again once #704 is merged?

`setup_prob` is 86% of the execution, so maybe most of the time is spent in the setup only.

kmdeck:
@Sbozzolo this doesnt seem to affect the time to solution, though the flame graphs look different. I think I will close this for now, unless you want to take a look at the flame graphs for the soil canopy benchmark. weirdly, this PR seems to have more "update driver" allocations than the run associated with the PR you just merged.

Sbozzolo:
The time to solution should not be affected, only the flame graph should. 

My PR essentially removed initialization from the flame graph so that we can focus on what happens at every timestep instead of only the first time.

PR #708, opened 2024-07-30
Kd/rb/ar/basic sif refactor

kmdeck:
Implements a simple SIF model, `Lee2015SIFModel`, in `AbstractSIFModel` within the `Canopy` module. 

@braghiere as you suggested, we made a new AbstractSIFModel, independent from others components of the Canopy module (where we now have photosynthesis, stomatal conductance, radiative transfer, autotrophic respiration, solar induced fluorescence, etc.). We named this first simple sif model "Lee2015SIFModel", if you have a better name let us know. This implementation allows to easily dispatch on other sif models in the future. Can you check the equations and units? Thanks!


braghiere (approved):
Looks good!

PR #710, opened 2024-07-31
Add support for regional simulations

### LINKED ISSUE #709: Add support for regional simulations
The capability to run simulations on a specified path on the globe is generally useful. This feature is also useful in aiding scientific debugging because it allows to track how NaN arise without the entire complexity of the globe.

At this point (with no lateral flow), implementing this is easy.

Sbozzolo:
This PR adds support for regional simulations, box simulations centered around a specific long/lat. Regional simulations are useful for all sorts of reasons, including debugging. Here, I use the fact that there is no later information exchange and set up a box domain. 

Closes #709

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -34,10 +34,6 @@ tutorials = [
         "Integrated soil+canopy modeling" => [
             "Coupled Canopy and Soil" => "integrated/soil_canopy_tutorial.jl",
         ],
-        "Bucket LSM" => [
-            "standalone/Bucket/bucket_tutorial.jl",
-            "standalone/Bucket/coupled_bucket.jl",
-        ],
         "Snow Modeling" => [
             "standalone/Snow/base_tutorial.jl",
             "standalone/Snow/data_tutorial.jl",
--- a/experiments/standalone/Bucket/bucket_era5.jl
+++ b/experiments/standalone/Bucket/bucket_era5.jl
@@ -1,7 +1,7 @@
-# # Global bucket run using spatial map albedo
+# # Global/regional bucket run using spatial map albedo
 
 # The code sets up and runs the bucket for 7 days using albedo read in from
-# a file containing static data over the globe, and analytic atmospheric and
+# a file containing static data over the globe or on a region, and ERA5 atmospheric and
 # radiative forcings.
 # Moving forward, this driver will serve as our more complex global bucket run,
 # eventually running for a longer time period (1+ year) and using temporally
@@ -64,7 +64,6 @@ function compute_clims(v)
 end
 
 anim_plots = true
-
 # Set to true if you want to run a regional simulation. By default, it is false,
 # unless the `CLIMALAND_CI_REGIONAL_BUCKET` environment variable is defined.
 regional_simulation = haskey(ENV, "CLIMALAND_CI_REGIONAL_BUCKET")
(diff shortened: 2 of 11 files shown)


experiments/standalone/Bucket/global_bucket_staticmap.jl
```diff
@@ -1,7 +1,7 @@
 # # Global bucket run using spatial map albedo
```
AlexisRenchon, line 1: Should we make a different file for a regional experiment? 
Since this file is called "global_ ..."

(or rename this file)

AlexisRenchon (approved):
Is this PR ready for merge, or do want to add some docs first?

This is also relevant enough to add content to the README.md I believe. We should mention that ClimaLand can be run: 
- Standalone of integrated
- At single site/pixel or regionally or globally

experiments/standalone/Bucket/global_bucket_staticmap.jl
Sbozzolo: I renamed this to `bucket_era5` because `staticmap` was misleading

Sbozzolo:
Note that PR hardcodes the radius of the Earth in same places, which is not ideal.

src/shared_utilities/Domains.jl
```diff
@@ -196,9 +198,11 @@ struct Plane{FT} <: AbstractDomain{FT}
     xlim::Tuple{FT, FT}
     "Domain interval limits along y axis, in meters"
     ylim::Tuple{FT, FT}
+    "When not `nothing`, a Tuple that contains the center `long` and `lat`."
```
kmdeck, line 201: what units and convention?

docs/make.jl
```diff
@@ -34,10 +34,6 @@ tutorials = [
         "Integrated soil+canopy modeling" => [
             "Coupled Canopy and Soil" => "integrated/soil_canopy_tutorial.jl",
         ],
-        "Bucket LSM" => [
```
kmdeck, line 37: add back in once albedo file works?

experiments/standalone/Bucket/bucket_era5.jl
```diff
@@ -63,26 +63,45 @@ function compute_clims(v)
     return (minimum(means) - maximum(sigmas), maximum(means) + maximum(sigmas))
 end
 
-anim_plots = true
+anim_plots = false
```
kmdeck, line 66: set back to true?

src/shared_utilities/Domains.jl
```diff
@@ -184,7 +184,9 @@ end
 A struct holding the necessary information
 to construct a domain, a mesh, a 2d spectral
 element space, and the resulting coordinate field.
-Note that only periodic domains are currently supported.
```
kmdeck, line 187: I think we need to keep this - because in principle, if we turned on lateral flow, we would be using periodic BC

Alternatively we can state that only independent columns are supported currently (no need for BC in the horizontal)

src/shared_utilities/Domains.jl
```diff
@@ -196,9 +198,11 @@ struct Plane{FT} <: AbstractDomain{FT}
     xlim::Tuple{FT, FT}
     "Domain interval limits along y axis, in meters"
     ylim::Tuple{FT, FT}
+    "When not `nothing`, a Tuple that contains the center `long` and `lat`."
+    longlat::Union{Nothing, Tuple{FT, FT}}
     "Number of elements to discretize interval, (nx, ny)"
     nelements::Tuple{Int, Int}
-    "Flags for periodic boundaries; only true is supported"
```
kmdeck, line 201: likewise here

src/shared_utilities/Domains.jl
kmdeck: (ClimaCore does not support non-periodic BC, I believe, even for boxes)

kmdeck (commented):
To avoid the hardcoding of R_earth, my only thought would be that we could adjust `Plane` as you have it, but instead make a `RegionalDomain` type (akin to `HybridBox`, which we would then leave untouched - except with `longlat=nothing` passed to `Plane` in the constructor) which takes the same arguments as `HybridBox` except additional R_earth and longlat

docs/make.jl
Sbozzolo: This is duplicated. There's another Bucket LSM above.

src/shared_utilities/Domains.jl
Sbozzolo: I added a note saying that only "no lateral flow is supported"

src/shared_utilities/Domains.jl
Sbozzolo: I added a note saying that only "no lateral flow is supported"

Sbozzolo:
> To avoid the hardcoding of R_earth, my only thought would be that we could adjust `Plane` as you have it, but instead make a `RegionalDomain` type (akin to `HybridBox`, which we would then leave untouched - except with `longlat=nothing` passed to `Plane` in the constructor) which takes the same arguments as `HybridBox` except additional R_earth and longlat

I made `R_earth` a parameter of `Plane`. It is used only when the `latlong` is not nothing.

kmdeck (approved):

PR #717, opened 2024-08-05
use different LAI dataset with data only every week

kmdeck:
## Purpose 
Since LAI changes infrequently, use a dataset with only weekly data so that we read data less frequently.

This new data set is created in this PR: https://github.com/CliMA/ClimaArtifacts/pull/5/files and on the caltech cluster

## To-do
benchmark

## Content
use new file

----


AlexisRenchon (approved):

Sbozzolo:
@AlexisRenchon maybe you can cherry pick this commit

PR #721, opened 2024-08-13
Add @with_error, remove @generate_error_functions

Sbozzolo:
`@generate_error_functions` was not a great macro: it required adding new names to a line every time a diagnostic is added, it was dealing with several variables at the same time, and could not be used outside of its scope of definition and/or called twice.

`@with_error` accomplishes a the same goals as
`@generate_error_functions` without the problems. The only difference is that now `compute` functions have to be decorated.


AlexisRenchon (approved):
This is really nice. Thanks.

Sbozzolo:
Merging bypassing requirements because the Buildkite error is due to #725

PR #722, opened 2024-08-14
Fix method definition overwritten warning from Canopy Model tests

imreddyTeja:
## Purpose 
Fix remaining warning during tests

## Content
- Method definition overwritten warning in Canopy module tests fixed

----


Sbozzolo (approved):
Thank you!

PR #729, opened 2024-08-19
Generalize domain depth

kmdeck:
## Purpose 
In global runs (spherical shell domain), the depth is used to compute TOPMODEL runoff. When we carry out regional simulations (hydrid box domain), we also require the depth, then. However, we were accessing the depth of the domain in a manner specific to the spherical shell.

This generalizes and stores the depth in domain.fields. I added a column method for completeness, even though we may not need it.

## Content

----

### DIFF
--- a/src/standalone/Soil/Runoff/Runoff.jl
+++ b/src/standalone/Soil/Runoff/Runoff.jl
@@ -228,7 +228,7 @@ function update_runoff!(
     @. p.soil.R_ss = topmodel_ss_flux(
         runoff.subsurface_source.R_sb,
         runoff.f_over,
-        model.domain.depth - p.soil.h∇,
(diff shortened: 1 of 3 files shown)


src/standalone/Soil/Runoff/Runoff.jl
```diff
@@ -228,7 +228,7 @@ function update_runoff!(
     @. p.soil.R_ss = topmodel_ss_flux(
         runoff.subsurface_source.R_sb,
         runoff.f_over,
-        model.domain.depth - p.soil.h∇,
```
kmdeck, line 231: previously, only the spherical shell had a depth field

Sbozzolo (approved):

PR #205, opened 2023-05-20
Finish implicit solver for soil RRE

juliasloan25:
## Purpose 
Finish implementing the implicit solver for the soil water equation (RRE). This uses the infrastructure added by #190 and #191, and adds in the rest of the functionality from our experimental implicit solver branch, #141.

Part of SDI #135 

Broken up into two smaller pieces: #212 and #224

## Content
- [x] Add implicit solver utility functions (linsolve, ldiv, etc) to SharedUtilities #218
  - Note: These will be moved to ClimaCore soon, as part of the implicit solver interface change ([ClimaCore #1230](https://github.com/CliMA/ClimaCore.jl/issues/1230)).
- [x] Make dss compatible with ClimaTimeSteppers, generalize for all models, and move to SharedUtilities. #218
- [x] Remove dss function call from tendency function(s). #218
- [x] Go through tests that call tendency functions outside of ODEFunction, manually add `dss!` call. #218
- [x] Use ClimaTimeSteppers instead of OrdinaryDiffEq #218
- [x] Update examples/tutorials/tests using RichardsModel to use ClimaTimeSteppers with the implicit/explicit setup.
- [x] Convert `imex_testing_modified_picard.jl` from #141 to a test.
- [ ] Add tests for 3 combined LandModel cases: 2 explicit models, 1 explicit and 1 implicit model, and 2 implicit models.

----
- [ ] I have read and checked the items on the review checklist.

PR #731, opened 2024-08-19
All CO2 fluxes in mol CO2 m-2 s-1

AlexisRenchon:
There was some inconsistencies, as well as some bug in the code, because of sometimes expressing CO2 fluxes in gC m-2 s-1, and sometimes in mol CO2 m-2 s-1.
In this PR, we express all CO2 fluxes in mol CO2 m-2 s-1. Note: the parameter "f1" (:mol_CO2_to_kg_C_factor) from AutotrophicRespirationParameters has been removed, and the parameter "f2" (:relative_contribution_factor) renamed to "Rel". This change is also happening in [ClimaParams.jl PR #205](https://github.com/CliMA/ClimaParams.jl/pull/205)

### DIFF
--- a/ext/CreateParametersExt.jl
+++ b/ext/CreateParametersExt.jl
@@ -96,8 +96,7 @@ function AutotrophicRespirationParameters(
         :live_stem_wood_coeff => :ηsl,
         :specific_leaf_density => :σl,
         :root_leaf_nitrogen_ratio => :μr,
-        :mol_CO2_to_kg_C_factor => :f1,
-        :relative_contribution_factor => :f2,
+        :relative_contribution_factor => :Rel,
         :stem_leaf_nitrogen_ratio => :μs,
     )
     parameters = CP.get_parameter_values(toml_dict, name_map, "Land")
--- a/lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-MOz/US-MOz_parameters.jl
+++ b/lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-MOz/US-MOz_parameters.jl
@@ -60,10 +60,9 @@ function auto_resp_ozark(;
     σl = FT(0.05),
     μr = FT(1.0),
     μs = FT(0.1),
-    f1 = FT(0.012),
-    f2 = FT(0.25),
+    Rel = FT(0.25),
 )
-    return AutotrophicRespirationParameters(ne, ηsl, σl, μr, μs, f1, f2)
+    return AutotrophicRespirationParameters(ne, ηsl, σl, μr, μs, Rel)
 end
 
 function soil_ozark(; # Function that returns the src function, but with ozark default args
--- a/src/diagnostics/define_diagnostics.jl
+++ b/src/diagnostics/define_diagnostics.jl
@@ -167,7 +167,7 @@ function define_diagnostics!(land_model)
         short_name = "ra",
         long_name = "Autotrophic Respiration",
         standard_name = "autotrophic_respiration",
-        units = "kg C m^-2 s^-1",
+        units = "mol CO2 m^-2 s^-1",
         comments = "The canopy autotrophic respiration, the sum of leaves, stems and roots respiration.",
         compute! = (out, Y, p, t) ->
             compute_autotrophic_respiration!(out, Y, p, t, land_model),
@@ -720,7 +720,7 @@ function define_diagnostics!(land_model)
         short_name = "scms",
         long_name = "Soil CO2 Microbial Source",
         standard_name = "soil_co2_microbial_source",
-        units = "kg C m^-2 s^-1",
+        units = "kg C m^-3 s^-1",
         comments = "The production of CO2 by microbes in the soil. Vary by layers of soil depth.",
         compute! = (out, Y, p, t) ->
             compute_soilco2_source_microbe!(out, Y, p, t, land_model),
--- a/src/diagnostics/land_compute_methods.jl
+++ b/src/diagnostics/land_compute_methods.jl
@@ -139,7 +139,22 @@ end
 @diagnostic_compute "vapor_flux" SoilCanopyModel p.soil.turbulent_fluxes.vapor_flux
 
 # Soil - SoilCO2
-@diagnostic_compute "heterotrophic_respiration" SoilCanopyModel p.soilco2.top_bc
+function compute_heterotrophic_respiration!(
+    out,
+    Y,
+    p,
+    t,
+    land_model::SoilCanopyModel{FT},
+) where {FT}
+    if isnothing(out)
+        return p.soilco2.top_bc .* FT(83.26)
+    else
+        out .= p.soilco2.top_bc .* FT(83.26)
+    end
+end # Convert from kg C to mol CO2. 
+# To convert from kg C to mol CO2, we need to multiply by:
+# [3.664 kg CO2/ kg C] x [10^3 g CO2/ kg CO2] x [1 mol CO2/44.009 g CO2] = 83.26 mol CO2/kg C
+
 @diagnostic_compute "soilco2_diffusivity" SoilCanopyModel p.soilco2.D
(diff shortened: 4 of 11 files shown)


src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -565,7 +567,7 @@ function ClimaLand.boundary_flux(
         Fields.field_values(Fields.level(Y.soilco2.C, p_len)),
         axes(Δz),
     )
-    C_bc = p.drivers.c_co2
+    C_bc = p.drivers.c_co2 # Kat, do we need to convert from kg C to mol CO2 here?
```
AlexisRenchon, line 570: @kmdeck 
What do you think is the best approach?

AlexisRenchon:
@kmdeck 
I think this PR is pretty close, but there is a remaining thing: 
CO2 concentration is in kg C m^-3. 
We use it at the top BC for soilco2. 
Now, I express microbial source flux in mol CO2 m^-3 s^-1. 
I imagine we should express CO2 concentration at BC in mol CO2 m^3.
Where is the best place to do this? In drivers.jl, or in Biogeochemistry.jl?

ext/CreateParametersExt.jl
```diff
@@ -96,8 +96,7 @@ function AutotrophicRespirationParameters(
         :live_stem_wood_coeff => :ηsl,
         :specific_leaf_density => :σl,
         :root_leaf_nitrogen_ratio => :μr,
-        :mol_CO2_to_kg_C_factor => :f1,
-        :relative_contribution_factor => :f2,
+        :relative_contribution_factor => :f,
```
kmdeck, line 99: is the relative_contribution_factor the same as kg_C_to_mol_CO2_factor? Using `f` for both

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
kmdeck: we can actually apply a preprocessing function to the driver as well... so each step it is read in and converted like so: https://github.com/CliMA/ClimaLand.jl/blob/394a295577c7fe4a998e6ed61949034074f8fa37/experiments/long_runs/soil.jl#L85

Otherwise we can specify that the c_co2 from the atmosphere must be in a certain set of units, and perform the conversion on this line?

ext/CreateParametersExt.jl
kmdeck: It might be clearer (even if long) to call this `kgC_to_molCO2`

src/standalone/Vegetation/autotrophic_respiration.jl
```diff
@@ -92,13 +90,14 @@ function compute_autrophic_respiration(
     h,
 )
 
-    (; ne, ηsl, σl, μr, μs, f1, f2) = model.parameters
+    (; ne, ηsl, σl, μr, μs, f) = model.parameters
     Nl, Nr, Ns =
         nitrogen_content(ne, Vcmax25, LAI, SAI, RAI, ηsl, h, σl, μr, μs)
-    Rpm = plant_respiration_maintenance(Rd, β, Nl, Nr, Ns, f1)
-    Rg = plant_respiration_growth(f2, An, Rpm)
+    Rpm = plant_respiration_maintenance(Rd, β, Nl, Nr, Ns)
```
kmdeck, line 96: `Rpm = Rd * (β + (Nr + Ns) / max(Nl, eps(FT)))` uses mol CO2
and 
` Rg = f * (An - Rpm)` here `f` does not do a conversion, so this is still in mol CO2

right?

ext/CreateParametersExt.jl
AlexisRenchon: As discussed, this parameter is not a conversion but to say Rg is proportional to (An - Rpm). 
We will give it a different name to avoid confusion with the f in Heterotrophic params

src/standalone/Vegetation/autotrophic_respiration.jl
AlexisRenchon: exact!

kmdeck (commented):
so Y.soilco2.C = units of mol CO2/m^3, microbe source is now in mol CO2/m^3/s, and all respiration rates are in mol CO2/m^2/s?

And then the last ? is regarding how to convert the driver CO2 concentration?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
AlexisRenchon: I did:
`C_bc = p.drivers.c_co2 * 83.26 # convert from kg C to mol CO2`

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-MOz/US-MOz_parameters.jl
```diff
@@ -60,10 +60,9 @@ function auto_resp_ozark(;
     σl = FT(0.05),
     μr = FT(1.0),
     μs = FT(0.1),
-    f1 = FT(0.012),
-    f2 = FT(0.25),
+    f = FT(0.012),
```
kmdeck, line 63: should be Rel?

src/diagnostics/define_diagnostics.jl
```diff
@@ -720,7 +720,7 @@ function define_diagnostics!(land_model)
         short_name = "scms",
         long_name = "Soil CO2 Microbial Source",
         standard_name = "soil_co2_microbial_source",
-        units = "kg C m^-2 s^-1",
+        units = "mol CO2 m^-2 s^-1",
```
kmdeck, line 723: I think this is kg C still, right?

lib/ClimaLandSimulations/src/Fluxnet/fluxnet_sites/US-MOz/US-MOz_parameters.jl
kmdeck: (`f` no longer in struct?)

src/diagnostics/land_compute_methods.jl
```diff
@@ -139,7 +139,20 @@ end
...
-@diagnostic_compute "heterotrophic_respiration" SoilCanopyModel p.soilco2.top_bc
+function compute_heterotrophic_respiration!(
+    out,
+    Y,
+    p,
+    t,
+    land_model::SoilCanopyModel{FT},
+) where {FT}
+    if isnothing(out)
+        return p.soilco2.top_bc .* FT(83.26)
```
kmdeck, line 150: Maybe we could note that:
# To convert from kg C to mol CO2, we need to multiply by: [3.664 kg CO2/ kg C] x [10^3 g CO2/ kg CO2] x [1 mol CO2/44.009 g CO2] = 83.26 mol CO2/kg C

kmdeck (approved):
thanks! looks good.

src/diagnostics/define_diagnostics.jl
AlexisRenchon: correct - and actually kg C m^-3 (volume) - thanks!

src/diagnostics/land_compute_methods.jl
Sbozzolo: Technically 83.26 is a parameter in ClimaParams. Would it be possible to fetch its value from `land_model`?

AlexisRenchon: I wonder if I should remove it from ClimaParams, as it is just a conversion?

AlexisRenchon: that I added but actually we are not using it (except if we do use it here)

Sbozzolo: To make it makes sense that it should not be a parameter (but there are other constants of nature/conversions in there). I'd remove it.

PR #736, opened 2024-08-22
Alexis docs hackathon #1 Aug 23

AlexisRenchon:
Working on ClimaLand docs 1 day a month - Alexis August 23

https://clima.github.io/ClimaLand.jl/previews/PR736/

- [x] index: explain we have docs "for users" and "for developers"
- [x] index: explain that we use SI units
- [x] getting started: discoverability and help of ClimaLand, tree of package structure (submodules, exported functions and type), how to get info
- [x] reorganize a bit: API now last, tutorials after getting-started, etc.

!! Note: `about(ClimaLand)` in getting-started does not show functions, modules and type by colors in julia 1.10 (it will in julia 1.11), so the line below mentioning what the color mean looks a bit weird, but julia 1.11 should be here soon.


docs/src/getting-started.md
```diff
@@ -0,0 +1,178 @@
+# Getting Started
+
+## For Users
+
+### Installation
+
+Installing ClimaLand in Julia is very easy. First, download and install Julia
```
Sbozzolo, line 7: Describing something as "very easy" is typically not the best for pedagogy. It may be very easy for you, but it is certainly not very easy for my grandma. It might also lead to frustration: how would you feel if you couldn't accomplish a task described as "very easy"? I suggest we avoid language like this when possible to make our docs more welcoming.

Sbozzolo:
https://sbozzolo.github.io/kuibit/features.html

docs/src/getting-started.md
AlexisRenchon: I agree! I adjusted the text

docs/make.jl
```diff
@@ -83,16 +51,15 @@ include("list_standalone_models.jl")
 include("list_diagnostics.jl")
 pages = Any[
     "Home" => "index.md",
-    "APIs" => apis,
-    "Contribution guide" => "Contributing.md",
+    "Getting Started" => "getting-started.md",
```
Sbozzolo, line 54: Typically `snake_case` is preferred over `kebab-case` in Julia

docs/src/getting-started.md
```diff
@@ -0,0 +1,133 @@
...
+```julia
+julia> ] # Enter Package REPL mode
+Pkg> add ClimaLand # Install ClimaLand
+Pkg> # Go back to Julia REPL mode
+Julia> using ClimaLand
+```
+
+### Parameterisation
+
+Let's start with an basic example: compute canopy gross photosynthesis (GPP).
```
Sbozzolo, line 19:
```suggestion
Let's start with a basic example: compute canopy gross photosynthesis (GPP).
```

docs/src/getting-started.md
```diff
@@ -0,0 +1,133 @@
...
+
+```@repl
+using ClimaLand
+@doc ClimaLand.Canopy.compute_GPP
+```
+
+As you can see, our parameterisation for GPP is located in the `Canopy` Module, and requires four FT.
+FT are Float (either Float32 or Float64). For example, with An = 5 µmol m⁻² s⁻¹, K = 0.5, LAI = 3 m² m⁻²,
+Ω = 0.7, you can compute GPP like below:
```
Sbozzolo, line 28:
```suggestion
As you can see, our parameterisation for GPP is located in the `Canopy` Module, and requires four FT.
FT means Float Type (typically Float32 or Float64). For example, with An = 5 µmol m⁻² s⁻¹, K = 0.5, LAI = 3 m² m⁻²,
Ω = 0.7, you can compute GPP like below:
```

docs/src/getting-started.md
```diff
@@ -0,0 +1,133 @@
...
+
+```@repl
+import ClimaLand.Canopy as canopy
+FT = Float64
+canopy.compute_GPP(FT(5.0), FT(0.5), FT(3.0), FT(0.7))
+```
+
+Et voilà!
+
+Note that our package [ParamViz](https://github.com/CliMA/ParamViz.jl) allow interactive visualisation of
```
Sbozzolo, line 38:
```suggestion
Note that our package [ParamViz](https://github.com/CliMA/ParamViz.jl) allows interactive visualisation of
```

docs/src/getting-started.md
```diff
@@ -0,0 +1,133 @@
...
+Then, you can install the ClimaLand package by doing:
+
+```julia
+julia> ] # Enter Package REPL mode
+Pkg> add ClimaLand # Install ClimaLand
+Pkg> # Go back to Julia REPL mode
+Julia> using ClimaLand
+```
+
+### Parameterisation
```
Sbozzolo, line 17: I would spell it `Parameterization` (en-US instead en-GB)

docs/src/getting-started.md
```diff
@@ -0,0 +1,133 @@
...
+First, download and install Julia by following the instructions at [https://julialang.org/downloads/](https://julialang.org/downloads/).
+Then, you can install the ClimaLand package by doing:
+
+```julia
+julia> ] # Enter Package REPL mode
+Pkg> add ClimaLand # Install ClimaLand
+Pkg> # Go back to Julia REPL mode
+Julia> using ClimaLand
+```
+
```
Sbozzolo, line 16: The transition from "Installation" to "Parametrisation" feels abrupt. 

Let's add a transition, like:
```
A typical land simulation employs several different parameterizations to model the various land-surface processes. Let's start our journet into ClimaLand by looking at one of those.
```

docs/src/getting-started.md
```diff
@@ -0,0 +1,133 @@
...
+### Parameterisation
+
+Let's start with an basic example: compute canopy gross photosynthesis (GPP).
+
+```@repl
+using ClimaLand
+@doc ClimaLand.Canopy.compute_GPP
+```
+
+As you can see, our parameterisation for GPP is located in the `Canopy` Module, and requires four FT.
```
Sbozzolo, line 26:
```suggestion
As you can see, our parameterisation for GPP is located in the [`Canopy`](@ref) Module, and requires four FT.
```

Sbozzolo, line 26: I think users should not be worried about `FT` at this point. Let's introduce this complication later.

docs/src/getting-started.md
```diff
@@ -0,0 +1,133 @@
...
+
+To explore what modules, functions and types are exported in a particular module, you can use About.jl:
+
+```@repl
+using ClimaLand
+using About
+about(ClimaLand.Soil.Biogeochemistry)
+```
+
+Where names in red are module, names in blue are functions, and name in yellow are Types.
```
Sbozzolo, line 129: Let's not mention this until it is actually true.

docs/src/getting-started.md
```diff
@@ -0,0 +1,133 @@
...
+
+# Build the tree
+root = build_tree(modules)
+```
+
+```@example climatree_md
+print_sorted_tree(stdout, root) # hide
+```
+
+To explore what modules, functions and types are exported in a particular module, you can use About.jl:
```
Sbozzolo, line 121: Please, add a link to `About.jl`.

docs/src/index.md
```diff
@@ -1,4 +1,54 @@
...
+using Pkg
+io = IOBuffer()
+v = Pkg.installed()["ClimaLand"]
+print(io, """
+    # ClimaLand.jl Documentation (v$(v))
+
+    """)
+import Markdown
+Markdown.parse(String(take!(io)))
+```
```
Sbozzolo, line 11: I think that this adds a chunk of code to be maintained but very little value (the version number).

Sbozzolo (approved):
Thank you! I think this is a good starting point and we can iterate over it with time.

docs/src/getting-started.md
AlexisRenchon: Thank you!! I like that

docs/src/index.md
AlexisRenchon: I agree, and removed it

PR #744, opened 2024-08-30
Extend Vcmax to support SpaceVaryingInput

### LINKED ISSUE #723: Extend Vcmax to support `SpaceVaryingInput`

```[tasklist]
### Tasks
- [ ] Obtain Vcmax map, visualize it, possibly deal with ocean values/NaNs
- [ ] Add it to ClimaArtifacts
- [ ] Extend `FarquharParameters` to support `F <: Union{<: AbstractFloat, ClimaCore.Field}`
- [ ] Change `land.jl` in benchmarks to use `SpaceVaryingInput` and ClimaLand functions and tests that use `Vcmax`
```

imreddyTeja:
## Purpose 

FarquharParameters now also supports Vcmax25 of type ClimaCore.Fields.Field. In relevant tests, Vcmax25 is changed from a float to a field that contains the previous Vcmax25 value everywhere. Both longrun and benchmark Land.jl now use the Vcmax map from clm_data.
Closes #723  

PR also contains whitespace removal.

## To-do
- Add VCmax map to ClimaArtifacts

## Content
- `FarquharParameters` takes a `Vcmax25` of type `F <: Union{<: AbstractFloat, ClimaCore.Field}`
- `Land.jl` benchmark and long run uses the vcmax map from clm_data. Map can be seen below. 
vc
- Relevant tests now use field of constant values for `Vcmax25`. 

#### Previously:
Day6Prev
#### After changing Vcmax25 to the clm_data map:
gpp_518400 0

----

### DIFF
--- a/ext/CreateParametersExt.jl
+++ b/ext/CreateParametersExt.jl
@@ -106,19 +106,19 @@ end
 
 """
     function FarquharParameters(
-        FT, 
+        FT,
         mechanism::AbstractPhotosynthesisMechanism;
         Vcmax25 = FT(5e-5),
         kwargs...  # For individual parameter overrides
     )
 
     function FarquharParameters(
-        toml_dict::CP.AbstractTOMLDict, 
+        toml_dict::CP.AbstractTOMLDict,
         mechanism::AbstractPhotosynthesisMechanism;
         Vcmax25 = FT(5e-5),
         kwargs...  # For individual parameter overrides
     )
-    
+
 Constructors for the FarquharParameters struct. Two variants:
 1. Pass in the float-type and retrieve parameter values from the default TOML dict.
 2. Pass in a TOML dictionary to retrieve parameter values.Possible calls:
@@ -183,7 +183,7 @@ end
     )
 
     function OptimalityFarquharParameters(
-        toml_dict::CP.AbstractTOMLDict, 
+        toml_dict::CP.AbstractTOMLDict,
         kwargs...  # For individual parameter overrides
     )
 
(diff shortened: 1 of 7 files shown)


ext/CreateParametersExt.jl
```diff
@@ -166,7 +166,9 @@ function FarquharParameters(
     parameters = CP.get_parameter_values(toml_dict, name_map, "Land")
     FT = CP.float_type(toml_dict)
     MECH = typeof(mechanism)
-    return FarquharParameters{FT, MECH}(;
+    Vcmax25 = FT.(Vcmax25)
```
kmdeck, line 169: do we need this FT? If Vcmax25 is a field, it should already be made on the space of the simulation and be the right type. If it is a scalar, this definition in the FarquharParameter struct `VC <: Union{FT, ClimaCore.Fields.Field}` should ensure it is the same type as the other scalars.

kmdeck (approved):
This looks good to me, thank you! 

We must already broadcast over functions with VCmax25 in Canopy.jl, since that didnt need to change?

Sbozzolo:
Make sure to add an entry to the NEWS file too. When you do so, it is nice if you can showcase the new feature you have added (see, e.g., how we do it in [ClimaAnalysis](https://github.com/CliMA/ClimaAnalysis.jl/blob/main/NEWS.md)).

ext/CreateParametersExt.jl
imreddyTeja: I think Julia does not perform conversion from scalars to unions. If I remove line 169 the tests fail with 
` TypeError: in FarquharParameters, in VC, expected VC<:Union{Float32, ClimaCore.Fields.Field}, got Type{Float64}`. Additionally, if I try converting from a scalar to `Union{FT, ClimaCore.Fields.Field}` I get 
`MethodError: Cannot convert an object of type 
  Float64 to an object of type 
  Union{Float32, ClimaCore.Fields.Field}
`

Sbozzolo (approved):

AlexisRenchon:
beautiful! thanks @imreddyTeja

PR #746, opened 2024-08-30
Refactor to use diagnostics throughout

AlexisRenchon:
- [x] add diagnostics: net radiation, latent heat flux, sensible heat flux, ground heat flux
- [x] add `diagnostics_as_vector` function, to get diagnostics from DictWriters
- [x] use diagnostics in `run_fluxnet`
- [x] add water runoff for single site experiment (fluxnet)

### DIFF
--- a/experiments/integrated/fluxnet/run_fluxnet.jl
+++ b/experiments/integrated/fluxnet/run_fluxnet.jl
@@ -185,7 +185,12 @@ shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
 canopy_model_args = (; parameters = shared_params, domain = canopy_domain)
 
 # Integrated plant hydraulics and soil model
-land_input = (atmos = atmos, radiation = radiation, soil_organic_carbon = Csom)
+land_input = (
+    atmos = atmos,
+    radiation = radiation,
+    soil_organic_carbon = Csom,
+    runoff = ClimaLand.Soil.Runoff.SurfaceRunoff(),
+)
 land = SoilCanopyModel{FT}(;
     soilco2_type = soilco2_type,
     soilco2_args = soilco2_args,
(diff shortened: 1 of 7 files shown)


Sbozzolo:
We can add an ArrayWriter too. It would probably make single site much simpler.

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -265,14 +270,82 @@ prob = SciMLBase.ODEProblem(
...
+    "si", # si_sfc or si_5 or si_10
+]
+
+hourly_diag_name = short_names_1D .* "_1h_average"
+hourly_diag_name_2D = short_names_2D .* "_1h_average"
+
+SIF, AR, g_stomata, GPP, canopy_T, SW_u, LW_u, ER, ET, β, SHF, LHF, G, Rn = [
+    ClimaLand.Diagnostics.diagnostic_as_vectors(d_writer, diag_name, 1)[2][1:721]
+    for diag_name in hourly_diag_name
+] # [1:721] only because summer subset of model output
```
Sbozzolo, line 339: This doesn't scale very well. `ClimaLand.Diagnostics.diagnostic_as_vectors` returns the time too, can you use that to  detect when it is summer instead?

src/diagnostics/diagnostic.jl
```diff
@@ -125,6 +125,28 @@ macro with_error(compute_function_expr)
...
 end
 
+"""
+    diagnostic_as_vectors(writer::ClimaDiagnostics.DictWriter, diagnostic)
+
+Extract `diagnostic` from given `writer` as tuple of vectors (time and value).
+
+`diagnostic` is typically a string with the short name of the diagnostic.
+"""
+function diagnostic_as_vectors(writer::DictWriter, diagnostic, layer)
```
Sbozzolo, line 135: Let's make `layer` a keyword argument that defaults to nothing for diagnostics with no depth.
```suggestion
function diagnostic_as_vectors(writer::DictWriter, diagnostic; layer)
```
Add it to the docstring too

src/diagnostics/diagnostic.jl
```diff
@@ -125,6 +125,28 @@ macro with_error(compute_function_expr)
...
+    # axes(first(values(writer[diagnostic]))) isa ClimaCore.Spaces.PointSpace || error("diagnostic_as_vectors works only on PointSpaces")
+
```
Sbozzolo, line 137:
```suggestion

```

experiments/integrated/fluxnet/run_fluxnet.jl
AlexisRenchon: I agree that this old script had many hardcoded stuff that make it difficult to maintain, and should be refactored properly.

however for instance, we were doing summer only because there was no snow model. once we have snow, we will run the whole year/dataset. so this is temporary.

src/diagnostics/diagnostic.jl
AlexisRenchon: good idea!

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -265,14 +270,84 @@ prob = SciMLBase.ODEProblem(
     (t0, tf),
     p,
 );
+
+# ClimaDiagnostics
+# ClimaDiagnostics
```
Sbozzolo, line 275:
```suggestion
```

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -265,14 +270,84 @@ prob = SciMLBase.ODEProblem(
...
+    for diag_name in hourly_diag_name
+] # [1:721] only because summer subset of model output
+
+swc_5, soil_T_5, si_5 = [
+    ClimaLand.Diagnostics.diagnostic_as_vectors(
+        d_writer,
+        diag_name;
+        layer = 5,
+    )[2][1:721] for diag_name in hourly_diag_name_2D
+]
```
Sbozzolo, line 347: Let's also explain the `[2]`

Sbozzolo (approved):
Looks good to me!

PR #748, opened 2024-09-01
Update compats and add downgrade workflows

Sbozzolo:
ClimaLand was falsely advertising compatibility with packages that are not really compatible. This commit updates the minimum bounds of the compats to reflect that. I also added a GitHub action workflow to check on that. The downgrade action runs the testsuite with the oldest possible compatible packages.


AlexisRenchon (approved):
That's great thanks!

PR #757, opened 2024-09-05
add land region script and update Manifest to branch

kmdeck:
## Purpose 
Create a regional run example which runs when we run long runs. 

In the future, we will move this to a tutorial.

## Content
- long lat swap in Plane constructor
- regional land script

----


src/shared_utilities/Domains.jl
```diff
@@ -267,22 +267,27 @@ function Plane(;
         @assert periodic == (false, false)
         radius_earth = FT(radius_earth)
         long, lat = longlat
+        dxlim = xlim # long
+        dylim = ylim # lat
```
Sbozzolo, line 271:
```suggestion
        dxlim = xlim # lat
        dylim = ylim # long
```
I had to switch things around to work with climacore

experiments/long_runs/land_region.jl
```diff
@@ -0,0 +1,666 @@
+# # Global run of land model
+
+# The code sets up and runs the soil/canopy model for 6 hours on a spherical domain,
```
Sbozzolo, line 3:
```suggestion
# The code sets up and runs the soil/canopy model for 6 hours on a small region of the globe,
```

experiments/long_runs/land_region.jl
```diff
@@ -0,0 +1,666 @@
...
+    τ_NIR_leaf = FT(0.25)
+
+    # Energy Balance model
+    ac_canopy = FT(2.5e4) # this will likely be 10x smaller!
+
+    # Conductance Model
+    g1 = FT(141) # Wang et al: 141 sqrt(Pa) for Medlyn model; Natan used 300.
+
+    #Photosynthesis model
+    Vcmax25 = FT(9e-5) # from Yujie's paper 4.5e-5 , Natan used 9e-5
```
Sbozzolo, line 375: Maybe we can use the spatially varying VCmax here?

experiments/long_runs/land_region.jl
```diff
@@ -0,0 +1,666 @@
...
+    driver_cb = ClimaLand.DriverUpdateCallback(updateat, updatefunc)
+    return prob, SciMLBase.CallbackSet(driver_cb, diag_cb)
+end
+
+function setup_and_solve_problem(; greet = false)
+
+    t0 = 0.0
+    tf = 60 * 60.0 * 24 * 2
+    Δt = 900.0
+    nelements = (2, 2, 15)
```
Sbozzolo, line 628: There's no reason to run so few elements on a GPU. It is pretty much free to run many more if you want.

src/shared_utilities/Domains.jl
kmdeck: I thought though that the user passes in xlim refering to long, and ylim refering to lat, and "longlat". And then here internally we swap them to work with climacore. I updated the comments to how I understand it, but let me know if the user actually passes in xlim refering to lat, ylim refering long.

Sbozzolo: Ah yes. You were and are right!

experiments/long_runs/land_region.jl
```diff
@@ -0,0 +1,676 @@
...
+        dz_tuple = FT.((10.0, 0.5)),
+    )
+    surface_space = domain.space.surface
+    subsurface_space = domain.space.subsurface
+
+    ref_time = DateTime(2021)
+    t_start = t0
+    # Forcing data
+    era5_artifact_path =
+        ClimaLand.Artifacts.era5_land_forcing_data2021_folder_path(; context)    # Precipitation:
```
Sbozzolo, line 82:
```suggestion
        ClimaLand.Artifacts.era5_land_forcing_data2021_folder_path(; context)    
        # Precipitation:
```

experiments/long_runs/land_region.jl
```diff
@@ -0,0 +1,676 @@
...
+simdir = ClimaAnalysis.SimDir(outdir)
+short_names = ["gpp", "ct", "swc", "si"]
+for short_name in short_names
+    var = get(simdir; short_name)
+    times = ClimaAnalysis.times(var)
+    for t in times
+        fig = CairoMakie.Figure(size = (800, 600))
+        kwargs = ClimaAnalysis.has_altitude(var) ? Dict(:z => 1) : Dict()
+        tmp = ClimaAnalysis.slice(var, time = t; kwargs...)
+        if ~all(isnan.(tmp.data))
```
Sbozzolo, line 671:
```suggestion
        if !all(isnan.(tmp.data))
```

`~` is bitwise 'not', `!` is logical 'not'. You are almost never working with bits, so you should pretty much always use `!`

Sbozzolo (approved):

PR #758, opened 2024-09-06
extraction fix

kmdeck:
## Purpose 
Update root extraction so that it tends to zero as the soil dries or freezes.
We have the Darcy flux:
Flux = -K_eff [(psi_canopy - psi_soil)/(z_canopy -z_soil) + 1]

Previously: K_eff = [K_canopy(psi_soil) + K_canopy(psi_canopy)]/2.

Two things have changed: we now use the harmonic mean, and we replace K_canopy(psi_soil) with K_soil(psi_soil). Note that K_soil() and K_canopy() are two different functions. 

Currently: K_eff = K_soil(psi_soil) * K_canopy(psi_canopy)/ [K_soil(psi_soil) + K_canopy(psi_canopy)]

So, if K_soil(psi_soil) -> 0, then K_eff -> 0, and the root extraction goes to zero consistently. Although as psi_soil ->0, K_canopy(psi_soil) ->0, it may not drop to zero quickly enough due to the differing functional forms.

Please see CLM:
Screenshot 2024-09-10 at 15 24 54
which is different in that it uses conductances and has canopy -> root, but see Equation 11.16 (harmonic mean) and Equation 11.18 (K_soil in soil layer, i.e. K_soil(psi_soil))

## Effect on NaNs:
Global run after soil evaporation fix (very similar to current main):
swc_2 88576e7_new
Global run after these changes:
image

## To-do
Review

## Content
1. Change the root extraction to depend on K_soil, and not K_plant(psi_soil). Note that this is only changed in the prognostic soil +canopy mode, and not the standalone canopy mode (prescribed soil). In the latter, we use K_plant(psi_soil) still.
2. Use the geometric mean to compute K_eff, and not arithmetic mean
3. Update tests

----

### DIFF
--- a/experiments/long_runs/land.jl
+++ b/experiments/long_runs/land.jl
@@ -28,7 +28,8 @@ import ClimaAnalysis
 import ClimaAnalysis.Visualize as viz
 import ClimaUtilities
 
-import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput
+import ClimaUtilities.TimeVaryingInputs:
+    TimeVaryingInput, LinearInterpolation, PeriodicCalendar
 import ClimaUtilities.SpaceVaryingInputs: SpaceVaryingInput
 import ClimaUtilities.Regridders: InterpolationsRegridder
 import ClimaUtilities.ClimaArtifacts: @clima_artifact
@@ -47,7 +48,7 @@ using Dates
 import NCDatasets
 
 const FT = Float64;
-
+time_interpolation_method = LinearInterpolation(PeriodicCalendar())
 regridder_type = :InterpolationsRegridder
 context = ClimaComms.context()
 device = ClimaComms.device()
@@ -83,6 +84,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         reference_date = start_date,
         regridder_type,
         file_reader_kwargs = (; preprocess_func = (data) -> -data / 3600,),
+        method = time_interpolation_method,
     )
 
     snow_precip = TimeVaryingInput(
@@ -92,6 +94,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         reference_date = start_date,
         regridder_type,
         file_reader_kwargs = (; preprocess_func = (data) -> -data / 3600,),
+        method = time_interpolation_method,
     )
 
     u_atmos = TimeVaryingInput(
@@ -100,6 +103,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         surface_space;
         reference_date = start_date,
         regridder_type,
+        method = time_interpolation_method,
     )
     q_atmos = TimeVaryingInput(
         joinpath(era5_artifact_path, "era5_2021_0.9x1.25_clima.nc"),
@@ -107,6 +111,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         surface_space;
         reference_date = start_date,
         regridder_type,
+        method = time_interpolation_method,
     )
     P_atmos = TimeVaryingInput(
         joinpath(era5_artifact_path, "era5_2021_0.9x1.25.nc"),
@@ -114,6 +119,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         surface_space;
         reference_date = start_date,
         regridder_type,
+        method = time_interpolation_method,
     )
 
     T_atmos = TimeVaryingInput(
@@ -122,6 +128,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         surface_space;
         reference_date = start_date,
         regridder_type,
+        method = time_interpolation_method,
     )
     h_atmos = FT(10)
 
@@ -145,6 +152,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         reference_date = start_date,
         regridder_type,
         file_reader_kwargs = (; preprocess_func = (data) -> data / 3600,),
+        method = time_interpolation_method,
     )
     LW_d = TimeVaryingInput(
         joinpath(era5_artifact_path, "era5_2021_0.9x1.25.nc"),
@@ -153,6 +161,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         reference_date = start_date,
         regridder_type,
         file_reader_kwargs = (; preprocess_func = (data) -> data / 3600,),
+        method = time_interpolation_method,
     )
 
     function zenith_angle(
@@ -489,6 +498,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         file_reader_kwargs = (;
             preprocess_func = (data) -> data > 0.05 ? data : 0.0,
         ),
+        method = time_interpolation_method,
     )
     ai_parameterization =
         Canopy.PrescribedSiteAreaIndex{FT}(LAIfunction, SAI, RAI)
@@ -613,6 +623,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         start_date;
         output_writer = nc_writer,
         output_vars = :short,
+        average_period = :monthly,
     )
 
     diagnostic_handler =
@@ -627,7 +638,7 @@ end
 function setup_and_solve_problem(; greet = false)
 
     t0 = 0.0
-    tf = 60 * 60.0 * 24 * 14
+    tf = 60 * 60.0 * 24 * 365
     Δt = 900.0
     nelements = (101, 15)
     if greet
--- a/src/standalone/Vegetation/PlantHydraulics.jl
+++ b/src/standalone/Vegetation/PlantHydraulics.jl
@@ -23,7 +23,7 @@ import ClimaLand:
     name
 export PlantHydraulicsModel,
     AbstractPlantHydraulicsModel,
-    flux,
+    water_flux,
     effective_saturation,
     augmented_liquid_fraction,
     water_retention_curve,
@@ -328,7 +328,7 @@ end
 
 
 """
-    flux(
+    water_flux(
         z1,
         z2,
         ψ1,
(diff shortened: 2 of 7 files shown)


.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -37,7 +37,7 @@ steps:
         artifact_paths: "land_longrun_gpu/*png"
         agents:
           slurm_gpus: 1
-          slurm_time: 03:00:00
+          slurm_time: 08:00:00
```
juliasloan25, line 40: do we need this change? it looks like the run is still taking ~2.5 hours to run

juliasloan25:
> Change the root extraction to depend on K_soil, and not K_plant(psi_soil). Note that this is only changed in the prognostic soil +canopy mode, and not the standalone canopy mode (prescribed soil). In the latter, we use K_plant(psi_soil) still.

It seems unusual to me to use slightly different equations for standalone vs integrated modes. Although maybe it doesn't make sense to evaluate K_soil with prescribed soil?

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -342,13 +342,15 @@ Computes the water flux given the absolute potential (pressure/(ρg))
...
 the flow path between these two points.
 
-We currently assuming an arithmetic
-mean for mean K_sat between the two points (Bonan, 2019; Zhu, 2008)
-to take into account the change in K_sat halfway
-between z1 and z2; this is incorrect for compartments of differing sizes.
-"""
-function flux(z1, z2, ψ1, ψ2, K1, K2)
-    flux = -(K1 + K2) / 2 * ((ψ2 - ψ1) / (z2 - z1) + 1)
+We currently assuming an geometric
```
juliasloan25, line 345:
```suggestion
We are currently asuuming a geometric
```

juliasloan25: it would be nice to also cite CLM here

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -342,13 +342,15 @@ Computes the water flux given the absolute potential (pressure/(ρg))
...
+mean for mean K_sat between the two points. Previously,
+we used the arithmetic mean (Bonan, 2019; Zhu, 2008),
+but then when the soil K was very low, root extraction would
+continue. This should be modified for compartments of differing sizes.
```
juliasloan25, line 349: what do you mean by "This should be modified for compartments of differing sizes."? what should be modified?

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -342,13 +342,15 @@ Computes the water flux given the absolute potential (pressure/(ρg))
...
+"""
+function flux(z1::FT, z2::FT, ψ1::FT, ψ2::FT, K1::FT, K2::FT) where {FT}
```
juliasloan25, line 351: I think a more descriptive name than "flux" would be nice - maybe "water_flux"? up to you

test/standalone/Vegetation/plant_hydraulics_test.jl
```diff
@@ -215,7 +215,7 @@ for FT in (Float32, Float64)
             SAI,
             RAI,
         )
-        K_sat_plant = 1.8e-8 # m/s.
+        K_sat_plant = 1.0 # m/s.
```
juliasloan25, line 218: why did you change this value (and T, LAI further down) to 1?

test/standalone/Vegetation/plant_hydraulics_test.jl
```diff
@@ -307,7 +307,7 @@ for FT in (Float32, Float64)
                                     vcat(root_depths, [0.0])[2:end] -
                                     vcat(root_depths, [0.0])[1:(end - 1)]
                                 ),
-                            ) * AI[:root]
+                            ) * AI[:stem]
```
juliasloan25, line 310: why did this change from root to stem?

juliasloan25 (commented):
Awesome!! This looks great, I just had a couple comments :)

kmdeck:
> > Change the root extraction to depend on K_soil, and not K_plant(psi_soil). Note that this is only changed in the prognostic soil +canopy mode, and not the standalone canopy mode (prescribed soil). In the latter, we use K_plant(psi_soil) still.
> 
> It seems unusual to me to use slightly different equations for standalone vs integrated modes. Although maybe it doesn't make sense to evaluate K_soil with prescribed soil?

Yeah, I agree...it would entail changing the PrescribedSoil struct to include a function for K_soil(psi) as well, or a timeseries of Ksoil(t) which is consistent with psi_soil(t), which would be made up at this point. We dont really have a use case for it yet.

.buildkite/longruns_gpu/pipeline.yml
kmdeck: thanks!

experiments/long_runs/land.jl
```diff
@@ -628,7 +629,7 @@ end
 function setup_and_solve_problem(; greet = false)
 
     t0 = 0.0
-    tf = 60 * 60.0 * 24 * 14
+    tf = 60 * 60.0 * 24 * 340
```
braghiere, line 632: Any reason why you stop in 340? We have a full 365 days right?

src/standalone/Vegetation/PlantHydraulics.jl
braghiere: +1

test/standalone/Vegetation/plant_hydraulics_test.jl
```diff
@@ -284,8 +284,8 @@ for FT in (Float32, Float64)
             )
             # Set system to hydrostatic equilibrium state by setting fluxes to zero, and setting LHS of both ODEs to 0
             function initial_compute_exp_tendency!(F, Y)
-                AI = (; leaf = LAI(0.0), root = RAI, stem = SAI)
-                T0A = FT(1e-8) * AI[:leaf]
+                AI = (; leaf = LAI(1.0), root = RAI, stem = SAI)
```
braghiere, line 287: Shouldn't this LAI vary spatially too? Why is it 1.0 everywhere?

test/standalone/Vegetation/plant_hydraulics_test.jl
```diff
@@ -335,7 +332,9 @@ for FT in (Float32, Float64)
             soln = nlsolve(
                 initial_compute_exp_tendency!,
                 Vector{FT}(-0.03:0.01:0.07);
-                ftol = 1e-11,
+                ftol = eps(FT),
+                method = :newton,
+                iterations = 20,
```
braghiere, line 337: I am not sure this is relevant but are 20 iterations enough to converge everywhere? Perhaps some places would take more iterations?

src/standalone/Vegetation/PlantHydraulics.jl
braghiere: +1

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: I added a comment for what CLM does, but will double check.

braghiere:
> > > Change the root extraction to depend on K_soil, and not K_plant(psi_soil). Note that this is only changed in the prognostic soil +canopy mode, and not the standalone canopy mode (prescribed soil). In the latter, we use K_plant(psi_soil) still.
> > 
> > 
> > It seems unusual to me to use slightly different equations for standalone vs integrated modes. Although maybe it doesn't make sense to evaluate K_soil with prescribed soil?
> 
> Yeah, I agree...it would entail changing the PrescribedSoil struct to include a function for K_soil(psi) as well, or a timeseries of Ksoil(t) which is consistent with psi_soil(t), which would be made up at this point. We dont really have a use case for it yet.

Agree. If it is not too difficult, maybe we can change the standalone as well? If it is a bit of work, we can leave it for now since we're focusing on canopy+soil development, which is more aligned to OKRs.

braghiere (commented):
It looks good to me. Left a few comments.

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: I changed this

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: Ok, I fleshed out the comment more. This is something we probably should do. But it would require introducing additional parameters (root length density, root biomass). I believe the PlantHydraulics code will need some other improvements too, and I think we could tackle it then.

test/standalone/Vegetation/plant_hydraulics_test.jl
kmdeck: In this case, it is just a test at a single location, so it is OK to have it be constant. but we vary spatially in our global and regional runs!

test/standalone/Vegetation/plant_hydraulics_test.jl
kmdeck: good question! In this case, we are just running the test at one location/in a column, so it doenst matter, but yes, some places on the globe might take longer to converge. That said, I think many models still cap the number of iterations in places like this so that the entire run is not held up by a single point.

test/standalone/Vegetation/plant_hydraulics_test.jl
kmdeck: In our src code, we use the above ground area index here, not the root area index. This was making sure the two align since this test is testing the src function.

test/standalone/Vegetation/plant_hydraulics_test.jl
kmdeck: I changed it back - thought it turns out the LAI one was a typo! and our test must not have caught it because it was not strict enough

juliasloan25:
> > > > Change the root extraction to depend on K_soil, and not K_plant(psi_soil). Note that this is only changed in the prognostic soil +canopy mode, and not the standalone canopy mode (prescribed soil). In the latter, we use K_plant(psi_soil) still.
> > > 
> > > 
> > > It seems unusual to me to use slightly different equations for standalone vs integrated modes. Although maybe it doesn't make sense to evaluate K_soil with prescribed soil?
> > 
> > 
> > Yeah, I agree...it would entail changing the PrescribedSoil struct to include a function for K_soil(psi) as well, or a timeseries of Ksoil(t) which is consistent with psi_soil(t), which would be made up at this point. We dont really have a use case for it yet.
> 
> Agree. If it is not too difficult, maybe we can change the standalone as well? If it is a bit of work, we can leave it for now since we're focusing on canopy+soil development, which is more aligned to OKRs.

I think it makes sense to keep PrescribedSoil as-is for now, until we do have a use case that necessitates this change

test/standalone/Vegetation/plant_hydraulics_test.jl
juliasloan25: makes sense, thank you!

kmdeck:
> > > > > Change the root extraction to depend on K_soil, and not K_plant(psi_soil). Note that this is only changed in the prognostic soil +canopy mode, and not the standalone canopy mode (prescribed soil). In the latter, we use K_plant(psi_soil) still.
> > > > 
> > > > 
> > > > It seems unusual to me to use slightly different equations for standalone vs integrated modes. Although maybe it doesn't make sense to evaluate K_soil with prescribed soil?
> > > 
> > > 
> > > Yeah, I agree...it would entail changing the PrescribedSoil struct to include a function for K_soil(psi) as well, or a timeseries of Ksoil(t) which is consistent with psi_soil(t), which would be made up at this point. We dont really have a use case for it yet.
> > 
> > 
> > Agree. If it is not too difficult, maybe we can change the standalone as well? If it is a bit of work, we can leave it for now since we're focusing on canopy+soil development, which is more aligned to OKRs.
> 
> I think it makes sense to keep PrescribedSoil as-is for now, until we do have a use case that necessitates this change

Thanks guys! I updated an old issue to have this included: https://github.com/CliMA/ClimaLand.jl/issues/159

kmdeck:
@braghiere @juliasloan25 I will extend the run to 365 days (or document why we cant and open an issue, I forget why it didnt run before).

would this be ready to merge then? If so, if you approve it now Ill merge it after the last fix to set tf=365 days.

I would really like to rethink aspects of our plant hydraulics model thoroughly #159 , but for now want to move forward with what we have

juliasloan25:
> @braghiere @juliasloan25 I will extend the run to 365 days (or document why we cant and open an issue, I forget why it didnt run before).
> 
> would this be ready to merge then? If so, if you approve it now Ill merge it after the last fix to set tf=365 days.
> 
> I would really like to rethink aspects of our plant hydraulics model thoroughly #159 , but for now want to move forward with what we have

Yeah, looks good to me!

juliasloan25 (approved):

experiments/long_runs/land.jl
kmdeck: I extended this in both our long runs

braghiere (approved):
Looks good. Thanks!

PR #759, opened 2024-09-09
Extend the MedlynConductance slope parameter to support SpaceVaryingInput

### LINKED ISSUE #756: Extend the MedlynConductance slope parameter to support `SpaceVaryingInput`

- [x] Visualize the medlynslope map and possibly deal with ocean values/NaNs
- [x] Extend `MedlynConductanceParameters` to support `G1 <: Union{FT, ClimaCore.Field}`
- [x] Change `land.jl` in benchmarks and long runs to use the medlynslope map
- [x] Change tests that use `MedlynConductanceParameters` to test medlynslope as a `SpaceVaryingInput`

imreddyTeja:
## Purpose 
Extend MedlynConductanceParameters to support a slope parameter that is
`<: Union{FT, ClimaCore.Fields.Field}`
Closes #756  -- this will automatically close issue 756 on PR merge

## To-do
- The clm_data file for vegetation properties and the script that generates it should be changed to reflect
the actual units. 

## Content

- MedlynConductanceParameters struct change to support g1 as Field
- Constructor for MedlynConductanceParameters change to reflect struct changes
- Add for loop for tests that check g1 as Float and Field
- unrelated: change way the vcmix kwarg is passed to the FarquharParameters constructor to unify code style
- Change g1 in land.jl to use the map in clm_data (see map below)

### Previous map image (has mask over ocean and no unit conversion)
med_slope
### Re-made map with no mask and converted to sqrt(Pa)

med_slope

----

### DIFF
--- a/experiments/benchmarks/land.jl
+++ b/experiments/benchmarks/land.jl
@@ -365,11 +365,21 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
     # Energy Balance model
     ac_canopy = FT(2.5e3)
 
+    #clm_data is used for g1 and vcmax maps
+    clm_artifact_path = ClimaLand.Artifacts.clm_data_folder_path(; context)
+
     # Conductance Model
-    g1 = FT(141) # Wang et al: 141 sqrt(Pa) for Medlyn model; Natan used 300.
+    # g1 is read in units of sqrt(kPa) and then converted to sqrt(Pa)
+    g1 = SpaceVaryingInput(
+        joinpath(clm_artifact_path, "vegetation_properties_map.nc"),
+        "medlynslope",
+        surface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+        file_reader_kwargs = (; preprocess_func = (data) -> data * 10^(3 / 2),),
+    )
 
     #Photosynthesis model
-    clm_artifact_path = ClimaLand.Artifacts.clm_data_folder_path(; context)
     # vcmax is read in units of umol CO2/m^2/s and then converted to mol CO2/m^2/s
     Vcmax25 = SpaceVaryingInput(
         joinpath(clm_artifact_path, "vegetation_properties_map.nc"),
--- a/experiments/long_runs/land.jl
+++ b/experiments/long_runs/land.jl
@@ -365,11 +365,21 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
     # Energy Balance model
     ac_canopy = FT(2.5e4) # this will likely be 10x smaller!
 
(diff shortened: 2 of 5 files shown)


Sbozzolo (approved):
Thank you! Please squash commits that need to be squashed before merging

kmdeck (approved):
Thank you! This looks great

kmdeck:
Would it be possible to update the map of g1 here to plot in units of sqrt(Pa)? Why is value over the ocean nonzero/non NaN?

imreddyTeja:
> Would it be possible to update the map of g1 here to plot in units of sqrt(Pa)? Why is value over the ocean nonzero/non NaN?

I added an re-made map under the original in the pull request. The original map in the PR had an ocean mask applied to it. `vegetation_properties_map.nc` lists the`_FillValue` as NaN, but there are no NaNs in the data. It also seems like there are no zero values in the data. I believe the value over the ocean is 284.60498 $\sqrt{Pa}$ because the median and maximum of the data is also  284.60498 $\sqrt{Pa}$

kmdeck:
> > Would it be possible to update the map of g1 here to plot in units of sqrt(Pa)? Why is value over the ocean nonzero/non NaN?
> 
> I added an re-made map under the original in the pull request. The original map in the PR had an ocean mask applied to it. `vegetation_properties_map.nc` lists the`_FillValue` as NaN, but there are no NaNs in the data. It also seems like there are no zero values in the data. I believe the value over the ocean is 284.60498 P a because the median and maximum of the data is also 284.60498 P a

@braghiere It is not a big deal, but does it make sense to you why the ocean has a nonzero/nonNan value for medlyn slope?

@imreddyTeja Thanks for remaking the plot!

braghiere:
@kmdeck It was just the way it was constructed. Bare soil and water were assigned values in the map but those should be ignored. 

@imreddyTeja The values look wrong now. Medlyn g1 is in sqrt(kPa) which can be anywhere from ~1.5 to ~6. The values were "right" but unit names were wrong. 

 
Screenshot 2024-09-10 at 10 23 35 AM

imreddyTeja:
Could I have performed the unit conversion incorrectly? I converted to sqrt(Pa) by multiplying by 10^(3/2). This is what the map looks like with no unit conversion or mask: 
med_slope

The data without unit conversion ranges from 1.62 to 9.0

braghiere:
> Could I have performed the unit conversion incorrectly? I converted to sqrt(Pa) by multiplying by 10^(3/2). This is what the map looks like with no unit conversion or mask: med_slope
> 
> The data without unit conversion ranges from 1.62 to 9.0

Got it! I see you converted from sqrt(kPa) ) to sqrt(Pa) to be in the same units as VPD. That makes sense! Thanks.

kmdeck (approved):

PR #769, opened 2024-09-13
Misc fixes to regional domain, long run scripts

kmdeck:
## Purpose 
- We should convert from meters D to degrees as: D/(2piR_earth)*360
- Increase resolution in soil model in land long runs
- g1 spatially varying in land_region

## Content
Update Domains.jl and test
g1 spatially varying in land_region.jl
dz_tuple -> (10, 0.05) in land.jl and land_region.jl

## Results

On the cluster, we have for the long land.jl run with dz_top = 5cm:
swc_3 1536e7

----

### DIFF
--- a/src/shared_utilities/Domains.jl
+++ b/src/shared_utilities/Domains.jl
@@ -235,7 +235,7 @@ Example: `longlat = (FT(-118.14452), FT(34.14778))`. `longlat` is in degrees, wi
 is not accounted for. We compute `xlim` and `ylim` as
 
 ```julia
-xlim = (long - xlim[1] / (2radius_earth), long + xlim[2] / (2radius_earth))
+xlim = (long - xlim[1] / (2π*radius_earth) * 360, long + xlim[2] / (2π*radius_earth) * 360)
 ```
 """
 function Plane(;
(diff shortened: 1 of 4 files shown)


src/shared_utilities/Domains.jl
```diff
@@ -235,7 +235,7 @@ Example: `longlat = (FT(-118.14452), FT(34.14778))`. `longlat` is in degrees, wi
 is not accounted for. We compute `xlim` and `ylim` as
 
 ```julia
-xlim = (long - xlim[1] / (2radius_earth), long + xlim[2] / (2radius_earth))
+xlim = (long - xlim[1] / (2π*radius_earth) * 360, long + xlim[2] / (2π*radius_earth) * 360)
```
Sbozzolo, line 238: There are functions `deg2rad` and `rad2deg`

kmdeck: I might keep it this way for readability unless you feel strongly about it!

Sbozzolo: I personally would find it more readable as
```
long - rad2deg(xlim[1] / 2radius_earth)
```
but I don't feel strongly about it, so have it as you prefer

Sbozzolo (approved):

PR #770, opened 2024-09-13
LAI lower threshold for nonzero LAI

kmdeck:
## Purpose 
We set LAI to be zero if it is below 0.05. This is what CLM does. 

Do we need to do this? Our fluxes should tend to zero as LAI -> 0, and our tendency for the canopy temperature is a sum of fluxes/max(LAI*height of canopy, eps(FT)). Therefore, for small LAI, everything should be well behaved.  That said, there was a roundoff bug for very small LAI: https://github.com/CliMA/ClimaLand.jl/issues/645, and, empirically, small values of LAI leads to NaNs (see below). 

I had thought our clipping of LAI was happening in the TimeVaryingInput preprocessfunction, but that is applied prior to any linear interpolation. This could cause issues near the coastline and also inland in regions where LAI is very small

While it would be good to understand the issues at small LAI better, this PR makes it so we are doing what we already thought we are doing. 
## Content
Previous Main:
image

After fix:
image

----

### DIFF
--- a/experiments/benchmarks/land.jl
+++ b/experiments/benchmarks/land.jl
@@ -473,21 +473,12 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
         )
     )
     # Set up plant hydraulics
-    # Note that we clip all values of LAI below 0.05 to zero.
-    # This is because we currently run into issues when LAI is
-    # of order eps(FT) in the SW radiation code.
-    # Please see Issue #644
-    # or PR #645 for details.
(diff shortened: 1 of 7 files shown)


juliasloan25 (approved):
This is great! Another good catch :) 

Maybe with this change, we should no longer do the clipping in the preprocess function of the TimeVaryingInput. If we clip before the interpolation, we could end up with smaller values after interpolating that will just get clipped again.

juliasloan25 (approved):

PR #771, opened 2024-09-13
Temporary canopy explicit PR

kmdeck:
## Purpose 
In testing the canopy temperature implicit solver implemented in https://github.com/CliMA/ClimaLand.jl/pull/675, we saw unexpected convergence behavior. To investigate if the cause of this was due to inacurracy of the reference solution or flaws in the simulation setup, we take a look here at the convergence behavior of the same setup using an explicit solver.

### Key point 1: `ac_canopy` affects time to steady state
 

Here we can see the evolution of canopy temperature over time using two different `ac_canopy` values. Atmospheric and radiative drivers are updated every 3 hours, so we see the temperature sharply change every 3 hours in both cases. The temperature then tends towards steady state but only reaches steady state with the smaller `ac_canopy=1e3`, creating the flat sections of the plot.

This pattern of temperature changing following a driver update then remaining constant upon reaching steady state produces two distinct regimes.

### Key point 2: convergence behavior should be assessed outside of the steady state regime
 

Here we have two plots showing convergence using ARS111 with only an explicit tendency (nothing stepped implicitly). The left plot shows convergence for simulations run for 12 hours + 80 seconds, while the left was run for 12 hours. The plotted line has slope 1, so we see the expected convergence behavior in the first case but not in the second.

In the first case, the simulations end shortly after a driver update, so outside of the steady state regime. All simulations are converging towards the steady state, but are not yet there, so we see the expected convergence behavior.

In the second case, the simulations end during the steady state regime. Simulations with all timesteps have converged to the same steady state, so the error for each of them is very small and doesn't converge as we would expect. This doesn't mean the convergence behavior of the solver/tendency is wrong; we've just stopped the simulation at an uninformative time.

### Conclusion
When testing the convergence of a solver for canopy temperature (or similar equations that reach steady state), we need to make sure we check convergence at a time outside of the steady state regime. Also note that this could have effects on errors throughout the simulation - e.g. error peaking sharply after a driver update then returning to a small value as steady state is reached.


kmdeck:
@juliasloan25 closing this, but we can reopen if needed

PR #774, opened 2024-09-17
Extend farquhar mechanism to spatially varying

### LINKED ISSUE #762: Extend Photosynthesis mechanism to support `SpaceVaryingInput`

- [x]  obtain or create map of C3/C4 plants
- [ ]  add the map to ClimaArtifacts
- [x] Extend `FarquharParameters` to support `MECH <:Union{AbstractPhotosynthesisMechanism, ClimaCore.Fields.Field}`
- [x]  change tests to use field of mechanisms
- [x] change land.jl to use the C3/C4 plant map
- [ ] add to NEWS.md and mark as breaking change

imreddyTeja:
## Purpose 
Extend FarquharParemeters and OptimalityFarquharParameters to support a mechanism parameter that is
<: Union{FT, ClimaCore.Fields.Field}
Closes #762  

## To-do
- Add to NEWS.md
- Add to ClimaArtifacts repo

## Content
This change is applied to both OptimalityFarquharParameters
and FarquharParameters. Previously, the mechanism parameter
was held as an empty struct of type ::C3 or ::C4 to allow
for dispatch. Because it is not possible to create a
ClimaCore field of different structs, the mechanism param is changed to a
is_c3 param that holds 1s and 0s. This choice was made because
reading a NetCDF file requires regridding, which must be performed over Numbers.
If the input data file must store the mechanism as a float, then it is easiest to
also internally represent the mechanism as a float. The C3/C4 structs
are now removed. The functions that previously used C3/C4 structs for dispatching
are converted to two seperate functions, which are called based on the `is_c3` value.

If you have a better name for the parameter than `is_c3`, please suggest it. I previously considered
`c3_mechanism_flag`. 

The constructor checks that the mechanism value of field is
between 0 and 1, and throws an error otherwise. Regridding a field of
0s and 1s may result in values between 0 and 1, so the constuctor rounds them
to the nearest integer. This implementation allows for future support of
mixed mechanisms in an area represented by proportion c3 without many changes.

All calls to the FarquharParameters constructor are changed to
work with the changes to mechanism representation. Tests are also
changed to work with new constructor and representation of mechanism.

The tests in canopy_model are changed to run twice using a for loop. The first run has all the parameters
as floats, and the second run has all canopy parameters that can be fields as fields. In this
case, the photosynthesis mechanism is tested as a field that alternates between 1.0 and 0.0,
which represents alternating between C3 and C4. The tests were previously all run over a Point
domain. The tests that use FarquharParameters are changed to run on a spherical surface.
Additionally, a test for vcmax as a field is added.

Land.jl and land_region.jl are changed to use the mechanism map from
the clm data.

----

### DIFF
--- a/docs/src/standalone/apps/leaf_an.jl
+++ b/docs/src/standalone/apps/leaf_an.jl
@@ -38,10 +38,11 @@ function ParamViz.parameterisation(
     Γstar = co2_compensation(Γstar25, ΔHΓstar, T, To, R)
     medlynt = 1 + g1 / sqrt(VPD)
     ci = intercellular_co2(ca, Γstar, medlynt)
-    Aj = light_assimilation(Canopy.C3(), J, ci, Γstar)
+    is_c3 = 1.0
+    Aj = light_assimilation(is_c3, J, ci, Γstar)
     Kc = MM_Kc(Kc25, ΔHkc, T, To, R)
     Ko = MM_Ko(Ko25, ΔHko, T, To, R)
-    Ac = rubisco_assimilation(Canopy.C3(), Vcmax, ci, Γstar, Kc, Ko, oi)
+    Ac = rubisco_assimilation(is_c3, Vcmax, ci, Γstar, Kc, Ko, oi)
     Rd = dark_respiration(Vcmax25, β, f, ΔHRd, T, To, R)
     An = net_photosynthesis(Ac, Aj, Rd, β)
     return An
@@ -83,7 +84,7 @@ function An_app_f()
             "Ω", # Clumping index
             "Γstar25", # co2 compensation at 25c
             "ΔHΓstar", # a constant energy of activation
-            "To", # a standard temperature 
+            "To", # a standard temperature
             "R", # the universal gas constant
             "Vcmax25", # the maximum rate of carboxylation of Rubisco
             "ΔHJmax", # a constant
--- a/docs/src/standalone/apps/leaf_an_ci.jl
+++ b/docs/src/standalone/apps/leaf_an_ci.jl
@@ -36,8 +36,9 @@ function ParamViz.parameterisation(
     Γstar = co2_compensation(Γstar25, ΔHΓstar, T, To, R)
     Kc = MM_Kc(Kc25, ΔHkc, T, To, R)
     Ko = MM_Ko(Ko25, ΔHko, T, To, R)
-    Ac = rubisco_assimilation(Canopy.C3(), Vcmax, ci, Γstar, Kc, Ko, oi)
-    Aj = light_assimilation(Canopy.C3(), J, ci, Γstar)
+    is_c3 = 1.0
+    Ac = rubisco_assimilation(is_c3, Vcmax, ci, Γstar, Kc, Ko, oi)
+    Aj = light_assimilation(is_c3, J, ci, Γstar)
     Rd = dark_respiration(Vcmax25, β, f, ΔHRd, T, To, R)
     An = net_photosynthesis(Ac, Aj, Rd, β)
     return An
@@ -79,7 +80,7 @@ function An_ci_app_f()
             "Ω", # Clumping index
             "Γstar25", # co2 compensation at 25c
             "ΔHΓstar", # a constant energy of activation
-            "To", # a standard temperature 
+            "To", # a standard temperature
             "R", # the universal gas constant
             "ΔHJmax", # a constant
             "θj", # an empirical "curvature parameter"
--- a/experiments/benchmarks/land.jl
+++ b/experiments/benchmarks/land.jl
@@ -380,6 +380,14 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
         regridder_kwargs = (; extrapolation_bc,),
         file_reader_kwargs = (; preprocess_func = (data) -> data / 1_000_000,),
     )
+    # photosynthesis mechanism is read as a float, where 1.0 indicates c3 and 0.0 c4
+    is_c3 = SpaceVaryingInput(
+        joinpath(clm_artifact_path, "mechanism_map.nc"),
+        "c3_dominant",
+        surface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+    )
 
     # Plant Hydraulics and general plant parameters
     SAI = FT(0.0) # m2/m2
@@ -465,13 +473,8 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
     conductance_args =
         (; parameters = Canopy.MedlynConductanceParameters(FT; g1))
     # Set up photosynthesis
-    photosynthesis_args = (;
-        parameters = Canopy.FarquharParameters(
-            FT,
-            Canopy.C3();
-            Vcmax25 = Vcmax25,
-        )
-    )
+    photosynthesis_args =
+        (; parameters = Canopy.FarquharParameters(FT, is_c3; Vcmax25 = Vcmax25))
     # Set up plant hydraulics
     # Note that we clip all values of LAI below 0.05 to zero.
     # This is because we currently run into issues when LAI is
--- a/experiments/long_runs/land_region.jl
+++ b/experiments/long_runs/land_region.jl
@@ -368,8 +368,18 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (10, 10, 15))
 
     # Energy Balance model
     ac_canopy = FT(2.5e4) # this will likely be 10x smaller!
-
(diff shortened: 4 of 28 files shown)


Artifacts.toml
Sbozzolo, line 1: This change should not be added to this PR

experiments/benchmarks/land.jl
```diff
@@ -380,6 +380,15 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
...
         file_reader_kwargs = (; preprocess_func = (data) -> data / 1_000_000,),
     )
+    # photosynthesis mechanism is read as a float, where 1.0 indicates c3 and 0.0 c4
+    is_c3 = SpaceVaryingInput(
+        joinpath(clm_artifact_path, "mechanism_map.nc"),
+        "c3_dominant",
+        surface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+        file_reader_kwargs = (;),
```
Sbozzolo, line 390:
```suggestion
```

if you don't need to pass extra arguments, you can remove the line

experiments/long_runs/land_region.jl
```diff
@@ -380,14 +390,14 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (10, 10, 15))
...
-        "medlynslope",
+    # c3 is read in as a float
+    is_c3 = SpaceVaryingInput(
+        joinpath(clm_artifact_path, "mechanism_map.nc"),
+        "c3_dominant",
         surface_space;
         regridder_type,
         regridder_kwargs = (; extrapolation_bc,),
-        file_reader_kwargs = (; preprocess_func = (data) -> data * 10^(3 / 2),),
+        file_reader_kwargs = (;),
```
Sbozzolo, line 400:
```suggestion
```

src/standalone/Vegetation/photosynthesis.jl
```diff
@@ -207,7 +193,6 @@ function update_photosynthesis!(
     )
     Vcmax25field .= Vcmax25
 end
-Base.broadcastable(m::AbstractPhotosynthesisMechanism) = tuple(m)
```
Sbozzolo, line 210: Why was this removed?

Sbozzolo (approved):
Thank you!

src/standalone/Vegetation/photosynthesis.jl
imreddyTeja: The AbstractPhotosynthesisMechanism struct definition is deleted above (line 3)

docs/src/standalone/apps/leaf_an.jl
```diff
@@ -38,10 +38,10 @@ function ParamViz.parameterisation(
     Γstar = co2_compensation(Γstar25, ΔHΓstar, T, To, R)
     medlynt = 1 + g1 / sqrt(VPD)
     ci = intercellular_co2(ca, Γstar, medlynt)
-    Aj = light_assimilation(Canopy.C3(), J, ci, Γstar)
+    Aj = light_assimilation(1.0, J, ci, Γstar)
```
kmdeck, line 41: for clarity here, perhaps we could define is_c3 = 1.0 and then use that in the functions

docs/src/standalone/apps/leaf_an_ci.jl
```diff
@@ -36,8 +36,8 @@ function ParamViz.parameterisation(
     Γstar = co2_compensation(Γstar25, ΔHΓstar, T, To, R)
     Kc = MM_Kc(Kc25, ΔHkc, T, To, R)
     Ko = MM_Ko(Ko25, ΔHko, T, To, R)
-    Ac = rubisco_assimilation(Canopy.C3(), Vcmax, ci, Γstar, Kc, Ko, oi)
-    Aj = light_assimilation(Canopy.C3(), J, ci, Γstar)
+    Ac = rubisco_assimilation(1.0, Vcmax, ci, Γstar, Kc, Ko, oi)
```
kmdeck, line 39: likewise here

ext/CreateParametersExt.jl
```diff
@@ -123,25 +123,25 @@ Constructors for the FarquharParameters struct. Two variants:
...
-ClimaLand.Canopy.FarquharParameters(Float64, ClimaLand.Canopy.C3())
+ClimaLand.Canopy.FarquharParameters(Float64, 1.0)
 # Kwarg overrides
-ClimaLand.Canopy.FarquharParameters(Float64, ClimaLand.Canopy.C3(); Vcmax25 = 99999999, pc = 444444444)
+ClimaLand.Canopy.FarquharParameters(Float64, 1.0; Vcmax25 = 99999999, pc = 444444444)
 # TOML Dictionary:
 import ClimaParams as CP
 toml_dict = CP.create_toml_dict(Float32);
-ClimaLand.Canopy.FarquharParameters(toml_dict, ClimaLand.Canopy.C3(); Vcmax25 = 99999999, pc = 444444444)
+ClimaLand.Canopy.FarquharParameters(toml_dict, 1.0; Vcmax25 = 99999999, pc = 444444444)
```
kmdeck, line 132: for Float32, should this be 1.0f0?

ext/CreateParametersExt.jl
```diff
@@ -165,11 +165,22 @@ function FarquharParameters(
     )
     parameters = CP.get_parameter_values(toml_dict, name_map, "Land")
     FT = CP.float_type(toml_dict)
-    MECH = typeof(mechanism)
+    if maximum(is_c3) > 1
```
kmdeck, line 168: What was the motivation for doing the rounding here, rather than having someone pass in 1 or 0? Since it should represent a bool.

That is, is_c3 should be zero or 1, and not between zero and 1

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -516,8 +516,22 @@ function co2_compensation(
...
-    rubisco_assimilation(::C3,
-                         Vcmax::FT,
+    rubisco_assimilation(is_c3::AbstractFloat, args...)
+
+Calls the correct rubisco assimilation function based on the `is_c3`.
+
+A `is_c3` value of 1.0 corresponds to C3 photosynthesis and calls
+`c3_rubisco_assimilation`, while 0.0 corresponds to C4 photsynthesis and calls
+`c4_rubisco_assimilation`. If a number between 0 and 1 is passed, if it is greater
+than 0.5, c3 will be assumed (essentially this takes the dominant mechanism)
```
kmdeck, line 526: I see. I think this is fine, but Im wondering if we should think of is_c3 as represent a bool and not a number between 0 and 1 (as above).

src/standalone/Vegetation/optimality_farquhar.jl
```diff
@@ -8,9 +8,12 @@ The required parameters for the optimality Farquhar photosynthesis model.
...
 $(DocStringExtensions.FIELDS)
 """
-Base.@kwdef struct OptimalityFarquharParameters{FT <: AbstractFloat}
+Base.@kwdef struct OptimalityFarquharParameters{
+    FT <: AbstractFloat,
+    MECH <: Union{FT, ClimaCore.Fields.Field},
+}
     "Photosynthesis mechanism: C3 only"
-    mechanism::C3
+    is_c3::MECH
```
kmdeck, line 16: since we only allow for C3 right now with the Optimality model, should we @assert all(is_c3 .== 1) in the constructor?

kmdeck (commented):
very nice, thank you! I left some questions in the comments but otherwise this looks good to me.

ext/CreateParametersExt.jl
imreddyTeja: The input map from the NetCDF has 0s and 1s everywhere, but that may not be true after regridding. I thought the constructor should take care of that, but that can also be left to the user.

src/standalone/Vegetation/optimality_farquhar.jl
imreddyTeja: I followed the behavior of the previous constructor, which ignores any kwarg for is_c3 and sets is_c3 to FT(1). Should this be changed?

ext/CreateParametersExt.jl
kmdeck: That makes sense to me, thanks for explaining. Maybe we could leave a comment to that effect?

src/standalone/Vegetation/optimality_farquhar.jl
kmdeck: I see that now, I think that is fine. I guess someone could set up the parameters using the default constructor with is_c3 = 0, but we note that this only supports C3 for now in the doc string.

kmdeck (approved):

PR #776, opened 2024-09-17
Consistent naming for radiation, land properties for radiation

kmdeck:
## Purpose 
Address inconsistencies in naming for radiation: upwelling/downwelling vs in/out

Store the albedo, emissivity, and effective temperature of the land

Partially addresses OKR2.3.8

## Content
We now follow the convention everywhere:
- downwelling "_d" fluxes are positive if down
- upwelling "_u" fluxes are positive if up
- store alpha_sfc, epsilon_sfc, and T_sfc in the cache. These are computed using SW_u, SW_d, LW_u, and the Stefan-Boltzmann constant.

----

### DIFF
--- a/experiments/integrated/fluxnet/ozark_pft.jl
+++ b/experiments/integrated/fluxnet/ozark_pft.jl
@@ -396,12 +396,12 @@ else
     )
 end
 
-# SW_OUT
-SW_out_model = [parent(sv.saveval[k].SW_out)[1] for k in 1:length(sv.saveval)]
+# Upwelling shortwave radiation - referred to as "outgoing" in the data
+SW_u_model = [parent(sv.saveval[k].SW_u)[1] for k in 1:length(sv.saveval)]
 if drivers.SW_OUT.status == absent
     plot_daily_avg(
-        "SW OUT",
-        SW_out_model,
+        "SW_up",
+        SW_u_model,
         dt_save,
         num_days,
         "w/m^2",
@@ -409,14 +409,14 @@ if drivers.SW_OUT.status == absent
         "model",
     )
 else
-    SW_out_data = FT.(drivers.SW_OUT.values)[Int64(t_spinup ÷ DATA_DT):Int64(
+    SW_u_data = FT.(drivers.SW_OUT.values)[Int64(t_spinup ÷ DATA_DT):Int64(
         tf ÷ DATA_DT,
     )]
     plot_avg_comp(
-        "SW OUT",
-        SW_out_model,
+        "SW up",
+        SW_u_model,
         dt_save,
-        SW_out_data,
+        SW_u_data,
         FT(DATA_DT),
         num_days,
         drivers.SW_OUT.units,
@@ -424,12 +424,12 @@ else
     )
 end
 
-# LW_OUT
-LW_out_model = [parent(sv.saveval[k].LW_out)[1] for k in 1:length(sv.saveval)]
+# Upwelling longwave radiation - referred to as "outgoing" in the data
+LW_u_model = [parent(sv.saveval[k].LW_u)[1] for k in 1:length(sv.saveval)]
 if drivers.LW_OUT.status == absent
     plot_daily_avg(
-        "LW OUT",
-        LW_out_model,
+        "LW up",
+        LW_u_model,
         dt_save,
         num_days,
         "w/m^2",
@@ -437,14 +437,14 @@ if drivers.LW_OUT.status == absent
         "model",
     )
 else
-    LW_out_data = FT.(drivers.LW_OUT.values)[Int64(t_spinup ÷ DATA_DT):Int64(
+    LW_u_data = FT.(drivers.LW_OUT.values)[Int64(t_spinup ÷ DATA_DT):Int64(
         tf ÷ DATA_DT,
     )]
     plot_avg_comp(
-        "LW OUT",
-        LW_out_model,
+        "LW up",
+        LW_u_model,
         dt_save,
-        LW_out_data,
+        LW_u_data,
         FT(DATA_DT),
         num_days,
         drivers.LW_OUT.units,
--- a/experiments/integrated/fluxnet/run_fluxnet.jl
+++ b/experiments/integrated/fluxnet/run_fluxnet.jl
@@ -340,12 +340,12 @@ else
     )
 end
 
-# SW_OUT
-SW_out_model = [parent(sv.saveval[k].SW_out)[1] for k in 1:length(sv.saveval)]
+# Upwelling shortwave radiation is referred to as outgoing in the data
+SW_u_model = [parent(sv.saveval[k].SW_u)[1] for k in 1:length(sv.saveval)]
 if drivers.SW_OUT.status == absent
     plot_daily_avg(
-        "SW OUT",
-        SW_out_model,
+        "SW up",
+        SW_u_model,
         dt_save,
         num_days,
         "w/m^2",
@@ -353,14 +353,14 @@ if drivers.SW_OUT.status == absent
         "model",
     )
 else
-    SW_out_data = FT.(drivers.SW_OUT.values)[Int64(t_spinup ÷ DATA_DT):Int64(
+    SW_u_data = FT.(drivers.SW_OUT.values)[Int64(t_spinup ÷ DATA_DT):Int64(
         tf ÷ DATA_DT,
     )]
     plot_avg_comp(
-        "SW OUT",
-        SW_out_model,
+        "SW up",
+        SW_u_model,
         dt_save,
-        SW_out_data,
+        SW_u_data,
         FT(DATA_DT),
         num_days,
         drivers.SW_OUT.units,
@@ -368,12 +368,12 @@ else
     )
 end
 
-# LW_OUT
-LW_out_model = [parent(sv.saveval[k].LW_out)[1] for k in 1:length(sv.saveval)]
+# Upwelling longwave radiation is referred to outgoing in the data
+LW_u_model = [parent(sv.saveval[k].LW_u)[1] for k in 1:length(sv.saveval)]
 if drivers.LW_OUT.status == absent
     plot_daily_avg(
-        "LW OUT",
-        LW_out_model,
+        "LW up",
+        LW_u_model,
         dt_save,
         num_days,
         "w/m^2",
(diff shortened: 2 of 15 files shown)


experiments/standalone/Vegetation/no_vegetation.jl
```diff
@@ -221,11 +221,11 @@ resp = [
     parent(sv.saveval[k].canopy.autotrophic_respiration.Ra)[1] * 1e6 for
     k in 1:length(sol.t)
 ]
-SW_abs = [
+SW_n = [
     parent(sv.saveval[k].canopy.radiative_transfer.SW_n)[1] for
     k in 1:length(sol.t)
 ]
-LW_abs = [
+LW_n = [
```
braghiere, line 228: Now, LW_n can be negative because the atmosphere can be colder than the land. 

Then I guess Rn = SW_d - SW_u + LW_d - LW_u.

experiments/standalone/Vegetation/no_vegetation.jl
```diff
@@ -253,8 +253,8 @@ axislegend(ax)
 save(joinpath(savedir, "no_veg_state.png"), fig)
 fig2 = Figure()
 ax = Axis(fig2[1, 1], xlabel = "Time (days)", ylabel = "Energy Fluxes")
-lines!(ax, sol.t ./ 24 ./ 3600, SW_abs, label = "SW")
-lines!(ax, sol.t ./ 24 ./ 3600, LW_abs, label = "LW")
+lines!(ax, sol.t ./ 24 ./ 3600, SW_n, label = "SW")
```
braghiere, line 256: Usually, we don't calculate SW_n or LW_n. We keep all fluxes separate, i.e., SW_d, SW_u, LW_d, LW_u, or we already compute Rn.

experiments/standalone/Vegetation/no_vegetation.jl
```diff
@@ -253,8 +253,8 @@ axislegend(ax)
...
+lines!(ax, sol.t ./ 24 ./ 3600, LW_n, label = "LW")
 lines!(ax, sol.t ./ 24 ./ 3600, SHF, label = "SHF")
 lines!(ax, sol.t ./ 24 ./ 3600, LHF, label = "LHF")
 lines!(ax, sol.t ./ 24 ./ 3600, RE, label = "RE")
```
braghiere, line 260: What is RE?

experiments/standalone/Vegetation/no_vegetation.jl
```diff
@@ -221,11 +221,11 @@ resp = [
     parent(sv.saveval[k].canopy.autotrophic_respiration.Ra)[1] * 1e6 for
     k in 1:length(sol.t)
 ]
-SW_abs = [
+SW_n = [
```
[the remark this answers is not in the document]
kmdeck: Yes, that will be our convention. It's on the to-do list to go through and make sure this is defined consistently throughout

experiments/standalone/Vegetation/no_vegetation.jl
kmdeck: Agreed!

experiments/standalone/Vegetation/no_vegetation.jl
kmdeck: Here I am checking them individually just because I want to make sure each behaves correctly as LAI -> 0

experiments/standalone/Vegetation/no_vegetation.jl
kmdeck: Root extraction. In the soil model, we include the energy of the water as part of the soil energy budget. so if water leaves the soil due to root extraction, there is some (small - a few W/m^2) energy flux into the canopy also. We just include this in the canopy energy budget to conserve energy

src/integrated/soil_canopy_model.jl
```diff
@@ -383,13 +405,20 @@ function lsm_radiant_energy_fluxes!(
...
     @. LW_u_soil = ϵ_soil * _σ * p.T_ground^4 + (1 - ϵ_soil) * LW_d_canopy # double checked
     @. R_net_soil += ϵ_soil * LW_d_canopy - ϵ_soil * _σ * p.T_ground^4 # double checked
     @. LW_net_canopy =
         ϵ_canopy * LW_d - 2 * ϵ_canopy * _σ * T_canopy^4 + ϵ_canopy * LW_u_soil
-    @. LW_out = (1 - ϵ_canopy) * LW_u_soil + ϵ_canopy * _σ * T_canopy^4 # double checked
+
+    @. LW_u = (1 - ϵ_canopy) * LW_u_soil + ϵ_canopy * _σ * T_canopy^4 # double checked
+
+    # Effective (radiative) land properties 
+    @. p.α_sfc = SW_u / SW_d
```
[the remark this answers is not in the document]
kmdeck: good catch! I will change this. We compute the ratio of SW_d to SW_u already in the canopy radiative transfer (reflected fraction) and I think we can just output that directly.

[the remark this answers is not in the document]
kmdeck: It turns out that is a little more work. We compute the absorbed, transmitted, and reflected fractions, but only store them multipled by SW_d. In a separate PR, I propose to change these to store the fractions, and then we can multiply them by SW_d when needed. Then the reflected fraction is the albedo independent of SW_d, so it wont go to NaN at night.

(for now what is in this PR will set albedo to zero at night)

AlexisRenchon:
There's some change needed in src/diagnostics/land_compute_methods.jl https://github.com/CliMA/ClimaLand.jl/blob/main/src/diagnostics/land_compute_methods.jl#L163

kmdeck:
> There's some change needed in src/diagnostics/land_compute_methods.jl https://github.com/CliMA/ClimaLand.jl/blob/main/src/diagnostics/land_compute_methods.jl#L163

thank you! I added those in

src/diagnostics/define_diagnostics.jl
```diff
@@ -16,7 +16,7 @@ function define_diagnostics!(land_model)
         long_name = "Albedo",
         standard_name = "albedo",
         units = "",
-        comments = "The fraction of incoming radiation reflected by the land surface.",
+        comments = "The fraction of downwelling radiation reflected by the land surface.",
```
braghiere, line 19: Being strict here, we need to define which albedo we're talking about about, given that at the moment we have 3 albedos (SW, PAR, and NIR). This one should be "SW_albedo", or something like that. It is "The fraction of downwelling SHORTWAVE radiation reflected by the land surface."

experiments/integrated/fluxnet/ozark_pft.jl
```diff
@@ -396,55 +396,55 @@ else
     )
 end
 
-# SW_OUT
-SW_out_model = [parent(sv.saveval[k].SW_out)[1] for k in 1:length(sv.saveval)]
+# Upwelling radiation - referred to as "outgoing" in the data
```
braghiere, line 399: Upwelling SHORTWAVE radiation

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -340,55 +340,55 @@ else
     )
 end
 
-# SW_OUT
-SW_out_model = [parent(sv.saveval[k].SW_out)[1] for k in 1:length(sv.saveval)]
+# Upwelling radiation is referred to as outgoing in the data
```
braghiere, line 343: upwelling SHORTWAVE radiation

experiments/integrated/fluxnet/ozark_pft.jl
```diff
@@ -396,55 +396,55 @@ else
...
         FT(DATA_DT),
         num_days,
         drivers.SW_OUT.units,
         savedir,
     )
 end
 
-# LW_OUT
-LW_out_model = [parent(sv.saveval[k].LW_out)[1] for k in 1:length(sv.saveval)]
+# Upwelling radiation - referred to as "outgoing" in the data
```
braghiere, line 427: Upwelling LONGWAVE radiation

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -340,55 +340,55 @@ else
...
         FT(DATA_DT),
         num_days,
         drivers.SW_OUT.units,
         savedir,
     )
 end
 
-# LW_OUT
-LW_out_model = [parent(sv.saveval[k].LW_out)[1] for k in 1:length(sv.saveval)]
+# Upwelling radiation is referred to outgoing in the data
```
braghiere, line 371: Upwelling LONGWAVE radiation

src/integrated/soil_canopy_model.jl
```diff
@@ -376,14 +398,21 @@ function lsm_radiant_energy_fluxes!(
...
     # This is a sign inconsistency. Here Rn is positive if towards soil. X_X
     @. R_net_soil += ϵ_soil * LW_d_canopy - ϵ_soil * _σ * p.T_ground^4 # double checked
     @. LW_net_canopy =
         ϵ_canopy * LW_d - 2 * ϵ_canopy * _σ * T_canopy^4 + ϵ_canopy * LW_u_soil
-    @. LW_out = (1 - ϵ_canopy) * LW_u_soil + ϵ_canopy * _σ * T_canopy^4 # double checked
+
+    @. LW_u = (1 - ϵ_canopy) * LW_u_soil + ϵ_canopy * _σ * T_canopy^4 # double checked
+
+    # Effective (radiative) land properties 
+    @. p.α_sfc = SW_u / max(SW_d, eps(FT)) # TODO: replace with fraction reflected as compute by canopy
```
braghiere, line 413: We need to be careful here because at night, when SW_d goes to zero, the α_sfc will became a very large number. α_sfc should always be a value from zero to one.

src/integrated/soil_canopy_model.jl
```diff
@@ -376,14 +398,21 @@ function lsm_radiant_energy_fluxes!(
...
+    @. p.ϵ_sfc = 1
```
braghiere, line 414: We don't need to do this in this PR, but it would be good to revisit this and make ϵ_sfc a function of canopy water content.

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -453,13 +454,13 @@ function diffuse_fraction(
     T::FT,
     P::FT,
     q::FT,
-    SW_IN::FT,
+    SW_d::FT,
```
braghiere, line 457: Likewise, at the moment we calculate the fraction of diffuse with an empirical function, but when the model is coupled to the atmosphere, we can read in that fraction from CliMA atmos. FYI.

braghiere (approved):

src/diagnostics/define_diagnostics.jl
kmdeck: I fixed this! I also added SW albedo as a diagnostic for the land model (currently it was just an option for the bucket). Now we will save that instead of SW_u.

src/standalone/Vegetation/canopy_parameterizations.jl
kmdeck: thanks for the reminder!

src/integrated/soil_canopy_model.jl
kmdeck: I opened an issue for this.

src/integrated/soil_canopy_model.jl
kmdeck: I opened an issue for this.

PR #779, opened 2024-09-18
Integrated snow and soil modeling - Option 2

kmdeck:
## Content

----


src/integrated/soil_canopy_model.jl
```diff
@@ -92,14 +77,13 @@ function SoilCanopyModel{FT}(;
...
     if :runoff ∈ propertynames(land_args)
-        top_bc = ClimaLand.Soil.AtmosDrivenFluxBC(
+        top_bc = ClimaLand.AtmosDrivenFluxBC(atmos, radiation, land_args.runoff)
+    else #no runoff model
+        top_bc = ClimaLand.AtmosDrivenFluxBC(
             atmos,
-            CanopyRadiativeFluxes{FT}(),
-            land_args.runoff,
+            radiation,
+            ClimaLand.Soil.Runoff.NoRunoff(),
```
kmdeck, line 85: update this to be land_args.runoff!

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -108,6 +108,8 @@ struct EnergyHydrology{FT, PS, D, BCRH, S} <: AbstractSoilModel{FT}
     sources::S
     "A boolean flag which, when false, turns off the horizontal flow of water and heat"
     lateral_flow::Bool
+    "Components"
+    land_components::Tuple
```
juliasloan25, line 112: since EnergyHydrology is a standalone model rather than integrated, does it need land_components? I thought it would be used only for integrated models to keep track of the component models

src/integrated/soil_snow_model.jl
```diff
@@ -0,0 +1,270 @@
...
+            (1 - p.snow.snow_cover_fraction) * (
+                p.soil.turbulent_fluxes.lhf +
+                p.soil.turbulent_fluxes.shf +
+                p.soil.R_n
+            ) +
+            p.snow.snow_cover_fraction * (
+                p.snow.turbulent_fluxes.lhf +
+                p.snow.turbulent_fluxes.shf +
+                p.snow.R_n
+            ) +
```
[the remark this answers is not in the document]
kmdeck: it works because it is (1-frac)*x_soil + frac * x_snow! :)

src/standalone/Soil/energy_hydrology.jl
kmdeck: I think this is a good question!

we somehow need to indicate to `soil_boundary_fluxes!` how it is being run (in a standalone model, or in an integrated model, with which components). we could stick the `components` into the AtmosDrivenFluxBC type, but that has the same philosophical issue in a way (the bc is part of the model, even in standalone runs you would define the land_components). But maybe that is a little preferable? 

What I had before was different BC types ("AtmosDrivenFluxBC", "AtmosDrivenFluxBCWithSnow", etc). I like this the least because it requires us to come up with names.

Another option is to have the soil_boundary_fluxes! update computation be explicitly moved outside of the `soil_boundary_fluxes! and into this one:
`https://github.com/CliMA/ClimaLand.jl/blob/145b001d3678a13876d3fe537f6ec9d755cab78f/src/integrated/soil_canopy_model.jl#L252` because this is made using the type of the land model, which knows what components are present. Then `soil_boundary_fluxes!` for all "IntegratedLandModelBC" types would do nothing, as p.soil.top_bc would already have been set by the land `update_boundary_fluxes!` function. This is what we were talking about Monday with Gabriele, as I understood it.

kmdeck: The way it is right now, the `land_components` default is (:soil,), so we can still run standalone runs in the same way. but youre right that it is a bit odd

src/integrated/soil_snow_model.jl
```diff
@@ -0,0 +1,270 @@
...
+
+A concrete type of land model used for simulating systems with
+snow and soil (and eventually rivers).
+$(DocStringExtensions.FIELDS)
+"""
+struct LandHydrologyModel{
+    FT,
+    SoM <: Snow.SnowModel{FT},
+    SnM <: Soil.EnergyHydrology{FT},
+} <: AbstractLandModel{FT}
```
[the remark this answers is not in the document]
kmdeck: I think it just depends on what combos we want to support! We could also just update the SoilCanopy model to be "LandModel" with soil+canopy+snow, and remove SoilCanopy.

I vaguely thought a LandHydrology (eventually soil+snow+rivers) model might be useful, but we can also only keep it as an internal debugger type model? idk

src/integrated/soil_snow_model.jl
```diff
@@ -0,0 +1,270 @@
...
+
+The names of the additional auxiliary variables that are
+included in the integrated Soil-Snow model.
+"""
+lsm_aux_vars(m::LandHydrologyModel) = (
+    :excess_water_flux,
+    :excess_heat_flux,
+    :atmos_energy_flux,
+    :atmos_water_flux,
+)
```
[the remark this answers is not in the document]
kmdeck: right now "LandHydrology" means only snow/soil. but we can rename it, change it, etc. see comment below about what combos we want to support. Does that answer your question? can discuss in person!

src/integrated/soil_snow_model.jl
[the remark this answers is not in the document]
juliasloan25: oh.... snow and soil look too similar 😆🤦‍♀️

src/standalone/Soil/energy_hydrology.jl
kmdeck: After discussing with @juliasloan25, we added the field `prognostic_land_components` to AtmosDrivenFluxBC. In a future PR we might rename this to something else, and we would plan to introduce analogs for Snow and Canopy

src/integrated/soil_canopy_model.jl
```diff
@@ -386,6 +377,7 @@ function lsm_radiant_energy_fluxes!(
     ϵ_canopy = p.canopy.radiative_transfer.ϵ # this takes into account LAI/SAI
     @. LW_d_canopy = ((1 - ϵ_canopy) * LW_d + ϵ_canopy * _σ * T_canopy^4) # double checked
     @. LW_u_soil = ϵ_soil * _σ * p.T_ground^4 + (1 - ϵ_soil) * LW_d_canopy # double checked
+    # This is a sign inconsistency. Here Rn is positive if towards soil. X_X
```
kmdeck, line 380: see Issue #776

experiments/integrated/fluxnet/snow_soil/simulation.jl
```diff
@@ -0,0 +1,300 @@
...
+    label = "2cm",
+)
+lines!(
+    ax1,
+    daily,
+    [parent(sol.u[k].soil.θ_i)[end] for k in 1:1:length(sol.t)],
+    label = "2cm, ice",
+)
+
+lines!(ax1, seconds ./ 3600 ./ 24, drivers.SWC.values[:], label = "Data, ?cm")
```
Sbozzolo, line 197: `?cm`?

src/integrated/soil_snow_model.jl
```diff
@@ -0,0 +1,276 @@
...
+        ),
+    )
+    soil = soil_model_type(;
+        boundary_conditions = boundary_conditions,
+        sources = sources,
+        soil_args...,
+    )
+    snow = snow_model_type(; atmos = atmos, radiation = radiation, snow_args...)
+
+
```
Sbozzolo, line 92:
```suggestion

```

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -576,27 +577,40 @@ to simulate surface and subsurface runoff and this is accounted
 for when setting boundary conditions. The default is to have no runoff
 accounted for.
 
+Finally, because this same boundary condition type is used for the soil
+in integrated land surface models, we also provide a tuple
+indicating the prognostic land components present, as these affect how the boundary
+conditions are computed. The default is
+a tuple containing only (:soil,), indicating a standalone soil run.
```
Sbozzolo, line 584: Can you add more examples of values? It it a tuple of Symbols? Which symbols are allowed?

Sbozzolo (commented):
Looks reasonable to me.

It would be great if you could also add a test and some documentation

experiments/integrated/fluxnet/snow_soil/simulation.jl
kmdeck: I cant find a reported depth where the sensor is...hence "? cm" :(

kmdeck: I changed label to "Unknown depth"

experiments/integrated/fluxnet/snow_soil/simulation.jl
```diff
@@ -0,0 +1,300 @@
...
+imp_tendency! = make_imp_tendency(land)
+jacobian! = ClimaLand.make_jacobian(land);
+jac_kwargs = (; jac_prototype = ImplicitEquationJacobian(Y), Wfact = jacobian!);
+set_initial_cache! = make_set_initial_cache(land)
+
+
+#Initial conditions
+Y.soil.ϑ_l =
+    drivers.SWC.status != absent ?
+    drivers.SWC.values[1 + Int(round(t0 / DATA_DT))] : soil_ν / 2
```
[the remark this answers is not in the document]
kmdeck: good call! I changed to:
```
# Find data index corresponds to t0
t0_idx = 1 + Int(round(t0 / DATA_DT))
```

src/standalone/Soil/boundary_conditions.jl
kmdeck: I noted here that it is a tuple of symbols, and referred to the documentation page I am adding.

kmdeck:
> Looks reasonable to me.
> 
> It would be great if you could also add a test and some documentation

thanks for the suggestion! I added a set of tests and also a tutorial.

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -743,23 +736,53 @@ Returns the net volumetric water flux (m/s) and net energy
...
-If you wish to compute surface fluxes taking into account the
-presence of a canopy, snow, etc, as in a land surface model,
-this is not the correct method to be using.
-
 This function calls the `turbulent_fluxes` and `net_radiation`
 functions, which use the soil surface conditions as well as
 the atmos and radiation conditions in order to
 compute the surface fluxes using Monin Obukhov Surface Theory.
+It also accounts for the presence of other components, if run as
+part of an integrated land model, and their affect on boundary conditions.
```
juliasloan25, line 744: affect -> effect

juliasloan25 (approved):
This looks great! Thank you Kat :)

kmdeck:
@juliasloan25 thank you for the review! I had set to "auto merge" prior to your review, and so it merged when you approved it even though you still had a comment. I will address that comment in a follow on PR with snow!

PR #783, opened 2024-09-19
Extend plant hydraulics rooting depth to support fields

### LINKED ISSUE #772: Add spatially varying rooting depth to plant hydraulics
- [x]  obtain or create map of rooting depths
- [x]  add the map to ClimaArtifacts
- [x] replace root distribution with rooting depth
- [x] change all calls to `PlantHydraulicsParameters` to support change
- [x] extend rooting depth to support `Union{FT, ClimaCore.Fields.Field}`
- [x]  make relevant additions to tests
- [x]  change land.jl to use the root depth map
- [x] add to NEWS.md and mark as breaking change

imreddyTeja:
## Purpose 
Towards spatially varying canopy parameters
Move the root_distribution function to src
Make rooting_depth, an argument to root_distribution, a parameter of PlantHydraulicsParameters
Make rooting depth support ClimaCore fields.
Closes #772 

PlantHydraulicsParameters previously held a root distribution function
that had a default value for rooting depth. This function is
always the same outside testing, so it was removed as a parameter
and placed into src. rooting_depth, an argument to the distribution function
is now a parameters that is either a Float or a field. The tests are changed to
check passing rooting_depth as a float or a field of floats that are the same everywhere.

## To-do
- Add to ClimaArtifacts

## Content
- Rooting_depth map created
- rooting_depth added as parameter
- rooting_distribution moved to source
- all calls to PlantHydraulicsParameters changed accordingly 
- tests changed to also test rooting_depth as a field
- add message in NEWS.md

PlantHydraulicsParameters previously held a root distribution function
that had a default value for rooting depth. This function is
always the same outside testing, so it was removed as a parameter
and placed into src. rooting_depth, an argument to the distribution function
is now a parameters that is either a Float or a field. The tests are changed to
check passing rooting_depth as a float or a field of floats that are the same everywhere.

The rooting_depth map can be seen here:

rooting_depth

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,8 +9,6 @@ main
 v0.15.0
 --------
 - Add regional simulation example PR [#757](https://github.com/CliMA/ClimaLand.jl/pull/757)
-- ![][badge-💥breaking] Extend photosynthesis mechanism parameter to support fields.
-  PR[#774](https://github.com/CliMA/ClimaLand.jl/pull/774)
 - Reduced number of dependencies, leading to faster instantiation and import time,
 - Improved compatibility of ClimaLand with older versions of packages.
   PR[#749](https://github.com/CliMA/ClimaLand.jl/pull/749)
@@ -19,6 +17,9 @@ v0.15.0
   PR[#759](https://github.com/CliMA/ClimaLand.jl/pull/759)
 - Added a regional run example.
   PR[#757](https://github.com/CliMA/ClimaLand.jl/pull/757)
+- ![breaking change][badge-💥breaking] Extend photosynthesis mechanism parameter to support fields.
+PR[#774](https://github.com/CliMA/ClimaLand.jl/pull/774)
+  - C3/C4 structs are removed. Now C3 is represented by a float of 1.0 and C4 by a float of 0.0
 
 v0.14.3
 --------
--- a/src/standalone/Vegetation/PlantHydraulics.jl
+++ b/src/standalone/Vegetation/PlantHydraulics.jl
@@ -157,24 +157,42 @@ function PlantHydraulicsParameters(;
     ai_parameterization::PrescribedSiteAreaIndex{FT},
     ν::FT,
     S_s::FT,
-    rooting_depth::Union{FT, ClimaCore.Fields.Field},
+    rooting_depth::Union{Nothing, FT, ClimaCore.Fields.Field} = nothing,
+    root_distribution::Union{Nothing, Function} = nothing,
     conductivity_model,
     retention_model,
 ) where {FT}
-    return PlantHydraulicsParameters{
-        FT,
-        typeof(ai_parameterization),
-        typeof(conductivity_model),
-        typeof(retention_model),
-        typeof(rooting_depth),
-    }(
-        ai_parameterization,
-        ν,
-        S_s,
-        conductivity_model,
-        retention_model,
-        rooting_depth,
-    )
+    if root_distribution isa Nothing
+        return PlantHydraulicsParameters{
+            FT,
+            typeof(ai_parameterization),
+            typeof(conductivity_model),
+            typeof(retention_model),
+            typeof(rooting_depth),
+        }(
+            ai_parameterization,
+            ν,
+            S_s,
+            conductivity_model,
+            retention_model,
+            rooting_depth,
+        )
+    else
+        return PlantHydraulicsParameters{
+            FT,
+            typeof(ai_parameterization),
+            typeof(conductivity_model),
+            typeof(retention_model),
+            FT,
+        }(
+            ai_parameterization,
+            ν,
+            S_s,
+            conductivity_model,
+            retention_model,
+            FT(1.0 / root_distribution(0.0)),
+        )
+    end
 end
 
 """
--- a/src/standalone/Vegetation/PlantHydraulics.jl
+++ b/src/standalone/Vegetation/PlantHydraulics.jl
@@ -374,15 +374,15 @@ end
 """
     root_distribution(z::FT, rooting_depth::FT)
 
-Computes value of rooting probabilty density function at `z`.
+Computes value of rooting probability density function at `z`.
 
-The rooting probabilty density function is derived from the
-cumulative distribution function F(z) = 1 - beta^(100z), which is described
+The rooting probability density function is derived from the
+cumulative distribution function F(z) = 1 - β^(100z), which is described
 by Equation 2.23 of
 Bonan, "Climate Change and Terrestrial Ecosystem Modeling", 2019 Cambridge University Press.
 This probability distribution function is equivalent to the derivative of the
 cumulative distribution function with respect to z,
-where `rooting_depth` replaces (-1)/(100ln(beta)).
+where `rooting_depth` replaces (-1)/(100ln(β)) and z is expected to be negative.
 """
 function root_distribution(z::FT, rooting_depth::FT) where {FT <: AbstractFloat}
     return (1 / rooting_depth) * exp(z / rooting_depth) # 1/m
--- a/src/standalone/Vegetation/PlantHydraulics.jl
+++ b/src/standalone/Vegetation/PlantHydraulics.jl
@@ -187,7 +187,7 @@ function PlantHydraulicsParameters(;
             )
         else
             Base.depwarn(
-                "root_distribution keyword argument is deprecated",
+                "root_distribution keyword argument is deprecated. Use rooting_depth instead.",
                 :PlantHydraulicsParameters,
             )
         end
--- a/src/standalone/Vegetation/PlantHydraulics.jl
+++ b/src/standalone/Vegetation/PlantHydraulics.jl
@@ -153,6 +153,18 @@ struct PlantHydraulicsParameters{
     rooting_depth::RD
 end
 
+"""
+    PlantHydraulicsParameters(;
+        ai_parameterization::PrescribedSiteAreaIndex{FT},
+        ν::FT,
+        S_s::FT,
+        rooting_depth::Union{FT, ClimaCore.Fields.Field},
+        conductivity_model,
+        retention_model,
+    )
+
+Constructor for PlantHydraulicsParameters.
+"""
 function PlantHydraulicsParameters(;
     ai_parameterization::PrescribedSiteAreaIndex{FT},
     ν::FT,
--- a/src/standalone/Vegetation/PlantHydraulics.jl
+++ b/src/standalone/Vegetation/PlantHydraulics.jl
@@ -149,7 +149,7 @@ struct PlantHydraulicsParameters{
     conductivity_model::CP
     "Water retention model and parameters"
     retention_model::RP
-    "Rooting depth parameter (m) - used to calculate root_distribution"
+    "Rooting depth parameter (m) - a characteristic depth below which 1/e of the root mass lies"
     rooting_depth::RD
 end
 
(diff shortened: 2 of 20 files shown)


NEWS.md
```diff
@@ -6,6 +6,9 @@ main
  - Add regional simulation example PR [#757](https://github.com/CliMA/ClimaLand.jl/pull/757)
  - ![][badge-💥breaking] Extend photosynthesis mechanism parameter to support fields.
   PR[#774](https://github.com/CliMA/ClimaLand.jl/pull/774)
+ - ![][badge-💥breaking] Remove root_distribution from PlantHydraulicsParameters and add rooting_depth,
+ which supports fields
+  PR[#783](https://github.com/CliMA/ClimaLand.jl/pull/783)
```
Sbozzolo, line 11: If something is breaking, it is always a good idea to describe how it is breaking and what users might need to do to fix it

test/standalone/Vegetation/plant_hydraulics_test.jl
```diff
@@ -303,31 +306,31 @@ for FT in (Float32, Float64)
...
+                                        conductivity_model,
+                                        Y[i - 1],
+                                    ),
+                                    PlantHydraulics.hydraulic_conductivity(
+                                        conductivity_model,
+                                        Y[i],
+                                    ),
+                                ) * AI[plant_hydraulics.compartment_labels[i]]
+                        end
+                        F[i] = mean(fa .- T0A)
```
Sbozzolo, line 332: Let's discuss this. I would have expected `F` to be a `Field` here too

kmdeck: I believe that in this test, we solve for the steady state of a vector-valued quantity using a root solver. We then set the initial condition of the field to the value of this vector. This seemed easier than getting the root solver to play with Fields, though maybe that is easy enough to do

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -370,6 +371,15 @@ function water_flux(z1::FT, z2::FT, ψ1::FT, ψ2::FT, K1::FT, K2::FT) where {FT}
     return flux # (m/s)
 end
 
+"""
+    root_distribution(z::FT, rooting_depth::FT)
+
+Compute value of rooting probabilty density function at `z`
```
kmdeck, line 377: Could you sketch out here the connection between rooting depth and beta?
E.g. give the CDF F(z) = 1-beta^(100z) and cite Equation 2.23 of Bonan, "Climate Change and Terrestrial Ecosystem Modeling", 2019 Cambridge University Press. And say this is equivalent to this probability density function with coordinate `z` replacing depth.

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -370,6 +371,15 @@ function water_flux(z1::FT, z2::FT, ψ1::FT, ψ2::FT, K1::FT, K2::FT) where {FT}
...
+"""
+function root_distribution(z::FT, rooting_depth::FT) where {FT <: AbstractFloat}
+    return FT(1.0 / rooting_depth) * exp(z / FT(rooting_depth)) # 1/m
```
kmdeck, line 380: If `z` and `rooting_depth` are already FT, and we change 1.0 -> 1, do we still need the FTs in the formula?

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -148,15 +149,15 @@ struct PlantHydraulicsParameters{
     conductivity_model::CP
     "Water retention model and parameters"
     retention_model::RP
-    "Root distribution function P(z)"
-    root_distribution::F
+    "Rooting depth parameter (m) - used to calculate root_distribution"
```
kmdeck, line 152: Maybe:
"Rooting depth parameter (m) - a characteristic depth below which `1/e` of the root mass lies."

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -370,6 +371,23 @@ function water_flux(z1::FT, z2::FT, ψ1::FT, ψ2::FT, K1::FT, K2::FT) where {FT}
...
+Computes value of rooting probabilty density function at `z`.
+
+The rooting probabilty density function is derived from the
+cumulative distribution function F(z) = 1 - beta^(100z), which is described
+by Equation 2.23 of
+Bonan, "Climate Change and Terrestrial Ecosystem Modeling", 2019 Cambridge University Press.
+This probability distribution function is equivalent to the derivative of the
+cumulative distribution function with respect to z,
+where `rooting_depth` replaces (-1)/(100ln(beta)) and z is expected to be negative.
+"""
```
imreddyTeja, line 386: @kmdeck Is this clear and accurate? I was not sure if there is a citation convention to follow. 

In your comment you said to "say this is equivalent to this probability density function with coordinate z replacing depth," but the z coordinate in the CDF is still present in the PDF, just negated. 

Could you sketch out here the connection between rooting depth and beta?
E.g. give the CDF F(z) = 1-beta^(100z) and cite Equation 2.23 of Bonan, "Climate Change and Terrestrial Ecosystem Modeling", 2019 Cambridge University Press. And say this is equivalent to this probability density function with coordinate z replacing depth.

kmdeck: I was making a distinction between "depth" z as Bonan uses it, and the coordinate z, what you have is clear! I would just check for typos ("probabilty" twice), replace beta with \beta (the symbol).

thank you!

test/standalone/Vegetation/plant_hydraulics_test.jl
imreddyTeja: @Sbozzolo I changed this test to not take the mean, and instead follow the original behavior but with fields. My implementation of this change is abnormal (see last commit). Could we discuss?

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -148,32 +149,50 @@ struct PlantHydraulicsParameters{
...
+            FT,
+        }(
+            ai_parameterization,
+            ν,
+            S_s,
+            conductivity_model,
+            retention_model,
+            FT(1.0 / root_distribution(0.0)),
+        )
+    end
```
Sbozzolo, line 195: You are doing manual type dispatch here. Instead, you can just define two methods for the function to return the correct value.

test/standalone/Vegetation/plant_hydraulics_test.jl
```diff
@@ -13,6 +13,7 @@ import ClimaLand
 import ClimaLand.Parameters as LP
 import Insolation
 using Dates
+using Statistics
```
Sbozzolo, line 16: This is no longer needed

test/standalone/Vegetation/plant_hydraulics_test.jl
```diff
@@ -283,70 +282,107 @@ for FT in (Float32, Float64)
...
-                                ),
-                            ) * AI[plant_hydraulics.compartment_labels[i]]
+                                ) * AI[plant_hydraulics.compartment_labels[i]]
+                        end
+                        F[i] = fa - T0A
                     end
-                    F[i] = fa - T0A
                 end
             end
+            #=======================
```
Sbozzolo, line 338: Can you add a description of what this test represents/is doing?

test/standalone/Vegetation/plant_hydraulics_test.jl
Sbozzolo: It's a little abnormal, but I think it's fine! Thanks for the comments in the commit message too

src/standalone/Vegetation/PlantHydraulics.jl
imreddyTeja: Julia doesn't use keyword arguments as dispatch. When I tried defining two methods for the function, the second just overwrote the first. It seems like this was the solution reached here:
https://discourse.julialang.org/t/how-can-kwargs-be-used-in-multiple-dispatch/86790

I could make the keyword arguments positional, but that would make the change breaking. Do you see a better way to do this, or should I revert to making this a breaking change?

Sbozzolo: It's fine! I didn't pay attention to `root_distribution` being a kwarg. Thank you!

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -148,32 +149,50 @@ struct PlantHydraulicsParameters{
...
+    else
+        return PlantHydraulicsParameters{
+            FT,
+            typeof(ai_parameterization),
+            typeof(conductivity_model),
+            typeof(retention_model),
+            FT,
+        }(
+            ai_parameterization,
+            ν,
```
charleskawczynski, line 189:
```suggestion
        rooting_depth = if root_distribution isa Nothing
            rooting_depth
        else
            FT(1.0 / root_distribution(0.0))
        end
        return PlantHydraulicsParameters{
            FT,
            typeof(ai_parameterization),
            typeof(conductivity_model),
            typeof(retention_model),
            typeof(rooting_depth),
        }(
            ai_parameterization,
            ν,
            S_s,
            conductivity_model,
            retention_model,
            rooting_depth,
        )
```
Couldn't we write this like this instead?

test/standalone/Vegetation/plant_hydraulics_test.jl
```diff
@@ -283,70 +281,110 @@ for FT in (Float32, Float64)
...
-                        fa =
-                            sum(
-                                water_flux.(
-                                    root_depths,
+            # Create a system of equations for a given rooting_depth for each cell
+            function create_exp_tendency_func(rooting_depth)
+                # Solve for hydrostatic equilibrium
+                return function initial_compute_exp_tendency!(F, Y)
+                    AI = (; leaf = LAI(1.0), root = RAI, stem = SAI)
+                    T0A = FT(1e-8) * AI[:leaf]
```
charleskawczynski, line 289: `FT` here should probably not be grabbed from the outer-scope, as that will result in a captured variable, and that can lead to type instabilities (which can hurt performance).

imreddyTeja: I'm not sure how this test could be rewritten to not use variables from the outer scope. Would passing in the outer scope variables as arguments fix the issue?

`function create_exp_tendency_func(rooting_depth, LAI, FT)
                # Solve for hydrostatic equilibrium
                    function initial_compute_exp_tendency!(F, Y)
                        AI = (; leaf = LAI(1.0), root = RAI, stem = SAI)
                        T0A = FT(1e-8) * AI[:leaf]`

When I check using @code_warntype, both create_exp_tendency_func and the function it returns are type stable.  Doesn't the inner function still capture the variables FT and LAI though?

kmdeck: I think in this case it is fine to use variables from the outer scope, as this function is run 1x, and in a test where performance is not important.

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -148,32 +149,50 @@ struct PlantHydraulicsParameters{
...
+            S_s,
+            conductivity_model,
+            retention_model,
+            FT(1.0 / root_distribution(0.0)),
```
kmdeck, line 193: Not to stir the pot, I know there was just a discussion about breaking changes, but I dont think we should support passing in the root_distribution function. Or, if we do, we need to know for that distribution how to compute the rooting_depth according to our fixed, single, rooting distribution in src (as of this PR). We cant assume that the rooting distribution they provide has the same functional form, from which we know how to compute the depth

kmdeck:
@imreddyTeja this PR looks good to me - I think you are you able to resolve some of the comments above? Are you waiting on us for approval?

imreddyTeja:
I was making it a non-breaking change again. It is ready for approval review now.

kmdeck (approved):
I appreciate the care you took with this, thank you!

Sbozzolo (approved):

PR #793, opened 2024-09-25
Extend two stream parameters to space varying

### LINKED ISSUE #784: Add spatially varying alpha and tau in radiative transfer model parameters
- [x] obtain or create maps for alpha and tau (PAR and NIR)
- [x] add maps to ClimaArtifacts, if not there already
- [x] Decide if Gfunction should be moved to src. If so, move it, and change all calls to constructors as needed
- [x] extend alpha and tau (PAR and NIR) to support `Union{FT, ClimaCore.Fields.Field}` in TwoStreamParameters
- [x] Do the same for BeerLambartParameters
- [x] make relevant changes to tests

imreddyTeja:
## Purpose 
Towards spatially varying canopy parameters
Closes #784 

G_function can now support its parameters as floats or fields of floats.
A variable G was added to the Radiation
model's auxilliary vars. G is now calculated only once during
update_aux!. Previously, G was calculated once in extinction_coeff
and once  in plant_absorbed_pfd, which are both called during
update_aux.

G was added to the auxilliary vars because it is not possible
to broadcast on a struct containing a field. This means G needs
to be calculated before extinction_coeff and plant_absorbed_pfd
are called, and then passed in as an argument.

α_PAR_leaf, τ_PAR_leaf, α_NIR_leaf, and τ_NIR_leaf in TwoStreamParameters
support fields. α_PAR_leaf and α_NIR_leaf in BeerLambertParameters also
support fields. Both constructors changed to match the new structs.
The tests that used TwoStructParameters and BeerLambartParameters
now test functionality when the parameters are a float, and then as a field
of constants. This is a straight forwards change for all tests except
test_two_stream.jl, where a workaround is used to  avoid broadcasting
isapprox over a field of floats. This results in an error because
it tries to convert bools to floats.

## Content
- alphas and taus extended to support fields in both TwoStreamParameters and BeerLambertParameters
- created maps for alpha and added to clima artifacts (see below)
- changes to tests
rholnir_map

rholvis_map

----

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -6,11 +6,11 @@ git-tree-sha1 = "316096635acaa8102477820f480ca02fe66eb828"
     url = "https://caltech.box.com/shared/static/awz2lku3ngoto6t78muktzfk3kjloxsw.gz"
 
 [clm_data]
-git-tree-sha1 = "5fcb868a751f4a2e32181612bb0c123e9a8e7e97"
+git-tree-sha1 = "459bca9591a26a48864240d2a37de8b5eb66530d"
 
     [[clm_data.download]]
-    sha256 = "bd8b53d691507e746e1de6dd2cc97178510c1d8944df1c23e4737bbb7ac7f8dd"
-    url = "https://caltech.box.com/shared/static/hbi0w0hbz8chaab4ym3kdpthc0d1blfc.gz"
+    sha256 = "dc3c0fc1c84935faf0770ba5e2b85ee7cf943d68f14783b29919c8ef44470fe3"
+    url = "https://caltech.box.com/shared/static/6pu2f6c99g29qawvjjh3j56drkonpd3z.gz"
 
 [era5_land_forcing_data2021]
 git-tree-sha1 = "ec424296df6b60cfe273ac8f981701fbbed0bd8a"
--- a/src/standalone/Vegetation/Canopy.jl
+++ b/src/standalone/Vegetation/Canopy.jl
@@ -463,7 +463,8 @@ function ClimaLand.make_update_aux(
         RT = canopy.radiative_transfer
         compute_PAR!(inc_par, RT, canopy.radiation, p, t)
         compute_NIR!(inc_nir, RT, canopy.radiation, p, t)
-        K = extinction_coeff.(p.canopy.radiative_transfer.G, θs)
+        K = p.canopy.radiative_transfer.K
+        @. K = extinction_coeff(p.canopy.radiative_transfer.G, θs)
         DOY = Dates.dayofyear(
             canopy.atmos.start_date + Dates.Second(floor(Int64, t)),
         )
--- a/src/standalone/Vegetation/Canopy.jl
+++ b/src/standalone/Vegetation/Canopy.jl
@@ -459,10 +459,11 @@ function ClimaLand.make_update_aux(
         @. p.canopy.radiative_transfer.ϵ =
             canopy.radiative_transfer.parameters.ϵ_canopy *
             (1 - exp(-(LAI + SAI))) #from CLM 5.0, Tech note 4.20
+        p.canopy.radiative_transfer.G .= compute_G(G_Function, θs)
         RT = canopy.radiative_transfer
         compute_PAR!(inc_par, RT, canopy.radiation, p, t)
         compute_NIR!(inc_nir, RT, canopy.radiation, p, t)
-        K = extinction_coeff.(G_Function, θs)
+        K = extinction_coeff.(p.canopy.radiative_transfer.G, θs)
         DOY = Dates.dayofyear(
             canopy.atmos.start_date + Dates.Second(floor(Int64, t)),
         )
--- a/src/standalone/Vegetation/canopy_parameterizations.jl
+++ b/src/standalone/Vegetation/canopy_parameterizations.jl
@@ -1,5 +1,6 @@
 using ..ClimaLand.Canopy
-export plant_absorbed_pfd,
+export plant_absorbed_pfd_beer_lambert,
+    plant_absorbed_pfd_two_stream,
     extinction_coeff,
     arrhenius_function,
     intercellular_co2,
@@ -95,16 +96,16 @@ function compute_absorbances!(
     _...,
 ) where {FT}
     RTP = RT.parameters
-    @. p.canopy.radiative_transfer.par = plant_absorbed_pfd(
-        RT,
+    @. p.canopy.radiative_transfer.par = plant_absorbed_pfd_beer_lambert(
+        RTP.Ω,
         PAR / (N_a * energy_per_photon_PAR),
         RTP.α_PAR_leaf,
         LAI,
         K,
         α_soil_PAR,
     )
-    @. p.canopy.radiative_transfer.nir = plant_absorbed_pfd(
-        RT,
+    @. p.canopy.radiative_transfer.nir = plant_absorbed_pfd_beer_lambert(
+        RTP.Ω,
         NIR / (N_a * energy_per_photon_NIR),
         RTP.α_NIR_leaf,
         LAI,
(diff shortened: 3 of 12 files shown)


src/standalone/Vegetation/radiation.jl
```diff
@@ -88,15 +90,19 @@ $(DocStringExtensions.FIELDS)
...
     FT <: AbstractFloat,
     G <: AbstractGFunction{FT},
+    APAR <: Union{FT, ClimaCore.Fields.Field},
+    TPAR <: Union{FT, ClimaCore.Fields.Field},
+    ANIR <: Union{FT, ClimaCore.Fields.Field},
+    TNIR <: Union{FT, ClimaCore.Fields.Field},
 }
     "PAR leaf reflectance (unitless)"
-    α_PAR_leaf::FT
+    α_PAR_leaf::APAR
```
kmdeck, line 99: I believe these should all be the same type, e.g. 
```
α_PAR_leaf::F
τ_PAR_leaf::F
α_NIR_leaf::F
τ_NIR_leaf::F
```
where `{F <: Union{FT, ClimaCore.Fields.Field}}`. I dont think we need to support a mix of floats and fields - either the parameters are scalars or they are fields. And forcing them to have the same type also makes sure the fields are defined on the same space.

I think this should be true for all of these parameter structs. so if there are already Fields you introduced to this struct, it should also have type `F`

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -254,8 +263,6 @@ function plant_absorbed_pfd(
     frac_diff::FT,
 ) where {FT}
 
-    (; G_Function, Ω, n_layers) = RT.parameters
-
     # Get the current leaf angular distribution value based on the solar zenith angle
     G = compute_G(G_Function, θs)
```
kmdeck, line 267: To handle this, we would need to make a field `p.canopy.radiative_transfer.G`, and update it prior to calling `plant_absorbed_pfd`. then we could pass in the field and broadcast over it. The update would dispatch off of the type of Gfunction. This could be done in another PR or here. happy to discuss more!

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -157,8 +158,10 @@ function compute_absorbances!(
     frac_diff,
 ) where {FT}
     RTP = RT.parameters
-    @. p.canopy.radiative_transfer.par = plant_absorbed_pfd(
-        RT,
+    @. p.canopy.radiative_transfer.par = plant_absorbed_pfd_two_stream(
```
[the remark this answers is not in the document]
imreddyTeja: That is correct. Previously RT was passed into the function, and the type was used for dispatching to the correct function, before unpacking the params. I was not able to broadcast plant_absorbed_pfd with RT as an argument because RT contains a field. Instead I passed in the parameters used by plant_absorbed_pfd as arguments to avoid using RT or RTP as arguments.

src/standalone/Vegetation/canopy_parameterizations.jl
imreddyTeja: Is G_function something that is space varying? If not, why would we need to make the field p.canopy.radiative_transfer.G?

kmdeck: The G function has spatially varying parameters. The function is always the same function of those parameters and the independent variable theta_s. So the dependent variable (which we call G) is spatially varying. we can store that, not the function

imreddyTeja: I thought there are two different G functions: ConstantGFunction and CLMGFunction.

If we do not have access to theta_s at the time of constructing the Radiative transfer model parameters struct, then don't we still need to store the function type or at least its parameters, so G can be calculated when theta_s is known?

src/standalone/Vegetation/radiation.jl
```diff
@@ -32,9 +33,10 @@ Base.broadcastable(G::ConstantGFunction) = Ref(G)
 A type for a G function that is parameterized by the solar zenith angle,
 following the CLM approach to parameterizing the leaf angle distribution function.
 """
-struct CLMGFunction{FT} <: AbstractGFunction{FT}
+struct CLMGFunction{F <: Union{AbstractFloat, ClimaCore.Fields.Field}} <:
+       AbstractGFunction{F}
     "Leaf orientation index (unitless)"
-    χl::FT
+    χl::F
 end
```
imreddyTeja, line 40: @kmdeck Is χl a spatially varying parameter? If not, should I remove the option to represent it as a field?

kmdeck: It is - you can search for the phrase "leaf angle" here and see: https://www2.cesm.ucar.edu/models/cesm2/land/CLM50_Tech_Note.pdf

since it is a function of functional type, it is spatially varying
Screenshot 2024-09-30 at 17 04 43

do you have a copy of Bonan's textbook? that might be helpful.

Artifacts.toml
```diff
@@ -1,9 +1,9 @@
 [clm_data]
-git-tree-sha1 = "5fcb868a751f4a2e32181612bb0c123e9a8e7e97"
+git-tree-sha1 = "459bca9591a26a48864240d2a37de8b5eb66530d"
```
kmdeck, line 2: Was this intentionally changed?

src/standalone/Vegetation/Canopy.jl
```diff
@@ -455,10 +455,11 @@ function ClimaLand.make_update_aux(
         @. p.canopy.radiative_transfer.ϵ =
             canopy.radiative_transfer.parameters.ϵ_canopy *
             (1 - exp(-(LAI + SAI))) #from CLM 5.0, Tech note 4.20
+        p.canopy.radiative_transfer.G .= compute_G(G_Function, θs)
         RT = canopy.radiative_transfer
         compute_PAR!(inc_par, RT, canopy.radiation, p, t)
         compute_NIR!(inc_nir, RT, canopy.radiation, p, t)
-        K = extinction_coeff.(G_Function, θs)
+        K = extinction_coeff.(p.canopy.radiative_transfer.G, θs)
```
kmdeck, line 462: I think this was already there previously, but this line allocates. Would you mind also adding K to the radiation aux vars? then at the top:
`K = p.canopy.radiative_transfer.K`

and here: ` @. K = extinction_coeff(p.canopy.radiative_transfer.G, \theta_s)`

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -35,7 +37,7 @@ export plant_absorbed_pfd,
 Returns the constant leaf angle distribution value for the given G function.
 Takes in an arbitrary value for the solar zenith angle, which is not used.
 """
-function compute_G(G::ConstantGFunction{FT}, _::FT) where {FT}
+function compute_G(G::ConstantGFunction, _::FT) where {FT}
```
kmdeck, line 40: Im not sure this is correct (also below). When we call this, we pass it theta_s as a Field, not as type FT. Also, FT is not needed here, in typing the arguments or in the where statement

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -49,8 +51,22 @@ Returns the leaf angle distribution value for CLM G function as a function of th
...
+    return compute_G_CLMG.(G.χl, θs)
+end
+
+"""
+    compute_G_CLMG(
+        χl::FT,
+        θs::FT,
+    )
+
+Returns the leaf angle distribution value for CLM G function as a function of the
```
kmdeck, line 64: I think it is helpful to note that this function applies pointwise (it is broadcasted over fields)

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -465,12 +485,8 @@ end
 Computes the vegetation extinction coefficient (`K`), as a function
 of the sun zenith angle (`θs`), and the leaf angle distribution (`ld`).
```
kmdeck, line 486: can update the doc string here. I think the leaf angle distribution is now `G`

kmdeck (commented):
Looks ready to go after a couple minor changes! Thanks!

kmdeck (approved):

Sbozzolo (approved):

PR #802, opened 2024-09-27
Compute SHF using SF, do not re-compute offline

juliasloan25:
## Purpose 
Since evaporation is no longer included in the SH computation in [SF.jl](https://github.com/CliMA/SurfaceFluxes.jl/pull/176/files#diff-6059b3ed786f8df9a4697af825ca0c680479fec22c617e328302dc4e2aaece66R687-R688), we now use the output of `sensible_heat_flux` provided directly by SF.jl's call to `surface_conditions`. This will make fluxes calculated in atmos and in land more similar, which is a good to prepare for coupled runs.

Our fluxes differ from those calculated in SurfaceFluxes still in these ways:
- land components usually require `r_sfc`, an additional resistance term, in the evaporation calculation.
- for the bucket and snow models, this is the only change
- for the soil model, we do not allow for condensation or frost. Again we modify the evaporation using a surface resistance. We also compute SF twice, once for liquid water, and once for ice. This requires us to combine the SH_ice and SH_water into a single number also.
- for the canopy, the surface resistance affects SH and E

## Content
Note that this change required an increase in the threshold for passing in one of our tests (for the canopy temperature jacobian contribution from SHF). Previously, we used only cp_m of the atmos in the SH flux, which is indep of T_canopy. now, the coefficient in the Jacobian depends on cp_m_sfc. This does depend on T_canopy, because it depends on q_canopy, which we approximate as q_sat(T_canopy).  We hypothesize that this is why the error in the SH approximation grew with this PR. @trontrytel 

Previous: SHF = -rho_sfc cp_m_atmos (T_atmos - T_canopy)/resistances
Now: SHF = -rho_sfc/resistances *  [stuff indep of T_canopy ..... - cp_m_sfc(T_canopy)*T_canopy ....]

Since we already have an issue open to improve the jacobian approximation, we are accepting this for now. @juliasloan25 

For good measure, though, I added the derivative of rho_sfc with respect to T_sfc

## Comparisons
This PR followed by implicit canopy PR long run:
Long Runs at 1/2 year:
Canopy Temp: 
image
image
ET:
image
image
SWC:
image
image

Canopy Alone Convergence:
this PR vs current main
image
image
Errors:
image
image

SoilCanopy Convergence:
this PR vs current main
image
image
Errors:
image
image

----

### DIFF
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -304,6 +304,11 @@ Computes turbulent surface fluxes at a point on a surface given
 
 This returns an energy flux and a liquid water volume flux, stored in
 a tuple with self explanatory keys.
+
+Please note that this function, if r_sfc is set to zero, makes no alteration
+to the output of the `SurfaceFluxes.surface_conditions` call, aside
+from unit conversion of evaporation from a mass to a liquid water volume
+flux.
 """
 function turbulent_fluxes_at_a_point(
     T_sfc::FT,
@@ -341,7 +346,7 @@ function turbulent_fluxes_at_a_point(
     )
 
     # State containers
-    sc = SurfaceFluxes.ValuesOnly(
+    states = SurfaceFluxes.ValuesOnly(
         state_in,
         state_sfc,
         z_0m,
@@ -350,26 +355,33 @@ function turbulent_fluxes_at_a_point(
         gustiness = gustiness,
     )
     surface_flux_params = LP.surface_fluxes_parameters(earth_param_set)
-    conditions = SurfaceFluxes.surface_conditions(
-        surface_flux_params,
-        sc;
-        tol_neutral = SFP.cp_d(surface_flux_params) / 100000,
-    )
+    scheme = SurfaceFluxes.PointValueScheme()
+    conditions =
+        SurfaceFluxes.surface_conditions(surface_flux_params, states, scheme)
     _LH_v0::FT = LP.LH_v0(earth_param_set)
     _ρ_liq::FT = LP.ρ_cloud_liq(earth_param_set)
 
-    cp_m::FT = Thermodynamics.cp_m(thermo_params, ts_in)
-    T_in::FT = Thermodynamics.air_temperature(thermo_params, ts_in)
-    ΔT = T_in - T_sfc
-    r_ae::FT = 1 / (conditions.Ch * SurfaceFluxes.windspeed(sc))
-    ρ_air::FT = Thermodynamics.air_density(thermo_params, ts_in)
+    # aerodynamic resistance
+    r_ae::FT = 1 / (conditions.Ch * SurfaceFluxes.windspeed(states))
 
-    E0::FT = SurfaceFluxes.evaporation(surface_flux_params, sc, conditions.Ch)
-    E = E0 * r_ae / (r_sfc + r_ae)
+    # latent heat flux
+    E0::FT =
+        SurfaceFluxes.evaporation(surface_flux_params, states, conditions.Ch) # mass flux at potential evaporation rate
+    E = E0 * r_ae / (r_sfc + r_ae) # mass flux accounting for additional surface resistance
+    LH = _LH_v0 * E # Latent heat flux
+
+    # sensible heat flux
+    SH = SurfaceFluxes.sensible_heat_flux(
+        surface_flux_params,
+        conditions.Ch,
+        states,
+        scheme,
+    )
+
+    # vapor flux in volume of liquid water with density 1000kg/m^3
     Ẽ = E / _ρ_liq
-    H = -ρ_air * cp_m * ΔT / r_ae
-    LH = _LH_v0 * E
-    return (lhf = LH, shf = H, vapor_flux = Ẽ, r_ae = r_ae)
+
+    return (lhf = LH, shf = SH, vapor_flux = Ẽ, r_ae = r_ae)
 end
 
 """
--- a/src/standalone/Soil/energy_hydrology.jl
+++ b/src/standalone/Soil/energy_hydrology.jl
@@ -859,6 +859,17 @@ Computes turbulent surface fluxes for soil at a point on a surface given
 
 This returns an energy flux and a liquid water volume flux, stored in
 a tuple with self explanatory keys.
+
+Notes:
+In this function, we call SF twice - once for ice, and once for water. 
+We then combine the fluxes from them to get the total SH, LH. 
+The vapor fluxes are applied to ice and water differently - as either a source term
+or a boundary condition-  so they are kept separate.
+
+For ice, we supply a beta factor and do not need an additional surface resistance, 
+and so we use the output of surface_conditions directly. 
+For water, we do, and it's a little complicated how we handle it. 
+But once we correct E, we compute SH and LH the same as SurfaceFluxes.jl does.
 """
 function soil_turbulent_fluxes_at_a_point(
     T_sfc::FT,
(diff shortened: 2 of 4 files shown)


land_longrun_cpu/global_diagnostics/output_active
Sbozzolo, line 1: This file was added by accident, I assumed

Project.toml
```diff
@@ -57,7 +57,7 @@ NCDatasets = "0.13.1, 0.14"
 SciMLBase = "2"
 StaticArrays = "1"
 StatsBase = "0.34"
-SurfaceFluxes = "0.11"
+SurfaceFluxes = "0.11, 0.12"
```
Sbozzolo, line 60: Do we know if have to apply any changes?

juliasloan25: It was hard to tell from the release notes, but I just found this comment by @kmdeck:

> @trontrytel In simulations where land is forced by a prescribed atmosphere, land needs to compute its own fluxes. In principle that means we just call SF functions, but instead right now we do that and then adjust the SF fluxes as follows: https://github.com/CliMA/ClimaLand.jl/blob/a746febe976c4dcab555c2afb8dbbe0483e169b4/src/shared_utilities/drivers.jl#L367-L372
> 
> This is actually more similar to the SF computation in this PR, so that is good. We had switched to this form because it was easier to debug/understand than the previous SF computation (e.g. SHF doesnt depend on E). But note that we also need to account for surface resistance in the fluxes, and SF does not handle that right now, so we do it "offline".
> 
> To couple the full land model to atmos, we need to adjust SF to account for resistance. We should also adjust our standalone land model to use the output of SF directly. This will make coupled runs easier to debug.

kmdeck: This should not be breaking change for us in terms of interface, and it wont change ClimaLand offline simulations since we compute surface fluxes on our own (can explain more), but it may change coupled behavior

Sbozzolo (approved):
Let's regenerate the Manifests before merging.

src/shared_utilities/drivers.jl
```diff
@@ -358,18 +358,27 @@ function turbulent_fluxes_at_a_point(
...
+    LH = _LH_v0 * E
+
+    # sensible heat flux
     cp_m::FT = Thermodynamics.cp_m(thermo_params, ts_in)
     T_in::FT = Thermodynamics.air_temperature(thermo_params, ts_in)
     ΔT = T_in - T_sfc
-    r_ae::FT = 1 / (conditions.Ch * SurfaceFluxes.windspeed(sc))
     ρ_air::FT = Thermodynamics.air_density(thermo_params, ts_in)
+    grav = SFP.grav(surface_flux_params)
+    ΔΦ = grav * Δz(sc)
```
juliasloan25, line 375: I rearranged the code here so the diff is larger than it needs to be, but the only physical change is adding this `\Delta \Phi` term to the SHF calculation

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -970,9 +962,22 @@ function soil_turbulent_fluxes_at_a_point(
     else
         Ẽ *= 0 # condensation, set to zero
     end
-    H_liq::FT = -ρ_air * cp_m * ΔT / r_ae_liq
 
-    # Sublimation:
+    # sensible heat flux
+    SH_liq = SurfaceFluxes.sensible_heat_flux(
```
kmdeck, line 967: For the soil model, we call SF twice - once for ice, and once for water. We then combine the fluxes from them to get the total SH, LH. the vapor fluxes are applied to ice and water differently, so they are kept separate.

For ice, we supply a beta factor and do not need an additional surface resistance, and so we use the output of `surface_conditions` directly. For water, we do, and it's a little complicated how we handle it. But once we correct E, we compute SH and LH the same as SF.jl does.

kmdeck: Note that for both we do not allow condensation/frost. We can/should revisit this.

src/shared_utilities/drivers.jl
```diff
@@ -350,26 +350,32 @@ function turbulent_fluxes_at_a_point(
...
-    ρ_air::FT = Thermodynamics.air_density(thermo_params, ts_in)
+    # aerodynamic resistance
+    r_ae::FT = 1 / (conditions.Ch * SurfaceFluxes.windspeed(states))
 
-    E0::FT = SurfaceFluxes.evaporation(surface_flux_params, sc, conditions.Ch)
-    E = E0 * r_ae / (r_sfc + r_ae)
+    # latent heat flux
+    E0::FT =
+        SurfaceFluxes.evaporation(surface_flux_params, states, conditions.Ch) # mass flux at potential evaporation rate
+    E = E0 * r_ae / (r_sfc + r_ae) # mass flux accounting for additional surface resistance
```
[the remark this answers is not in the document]
kmdeck: i.e. we only modify evaporation vs the output of `surface_conditions`. SH should be the same, and LH is computed from E in the same way,

src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -323,27 +323,30 @@ function canopy_turbulent_fluxes_at_a_point(
...
+        SurfaceFluxes.sensible_heat_flux(
+            surface_flux_params,
+            conditions.Ch,
+            states,
+            scheme,
+        ) * r_ae / (canopy_r_b + r_ae)
+    # The above follows from CLM 5, tech note Equation 5.88, setting H_v = SH and solving to remove T_s, ignoring difference between cp in atmos and above canopy. 
     LH = _LH_v0 * E
-    return (lhf = LH, shf = H, vapor_flux = Ẽ, r_ae = r_ae)
+    return (lhf = LH, shf = SH, vapor_flux = Ẽ, r_ae = r_ae)
```
kmdeck, line 351: For the canopy, additional resistances are also needed for SH.

src/standalone/Soil/energy_hydrology.jl
juliasloan25: it would be nice to add this (or part of it) as comments in the code

src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -323,27 +323,27 @@ function canopy_turbulent_fluxes_at_a_point(
...
     canopy_r_b = r_b / (LAI + SAI)
     E0::FT = SurfaceFluxes.evaporation(surface_flux_params, sc, conditions.Ch)
     E = E0 * r_ae / (leaf_r_b + leaf_r_stomata + r_ae) # CLM 5, tech note Equation 5.101, and fig 5.2b, assuming all sunlit, f_wet = 0
     Ẽ = E / _ρ_liq
-    H = -ρ_air * cp_m * ΔT / (canopy_r_b + r_ae) # CLM 5, tech note Equation 5.88, setting H_v = H and solving to remove T_s
+    SH =
+        SurfaceFluxes.sensible_heat_flux(
+            surface_flux_params,
+            conditions.Ch,
+            states,
```
akshaysridhar, line 343: In L317 this is defined as `sc` ; it seems you've modified this elsewhere to be `states` so perhaps we should use this term! It's clearer than the abbreviated `sc` for `surface conditions` I think.

akshaysridhar (commented):
Two small issues
(1) Variable naming in `canopy_boundary_layer.jl` 
(2) Neutral stability tolerance `kwarg` in `energy_hydrology.jl`

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -996,25 +998,28 @@ function soil_turbulent_fluxes_at_a_point(
         beta = β_i,
         gustiness = gustiness,
     )
-    conditions = SurfaceFluxes.surface_conditions(
-        surface_flux_params,
-        states;
-        tol_neutral = SFP.cp_d(surface_flux_params) / 100000,
```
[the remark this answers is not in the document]
kmdeck: I will remove this! good call

src/standalone/Soil/energy_hydrology.jl
kmdeck: Ill addit!

akshaysridhar:
Thanks @kmdeck ! Another note: `ts_cpu.png` and `ts_gpu.png` plots in the buildkite output (_static map_ albedo tests) are blank; since these are visualisations of time-dependent vars maybe we can remove them entirely from static map cases.

kmdeck:
> Thanks @kmdeck ! Another note: `ts_cpu.png` and `ts_gpu.png` plots in the buildkite output (_static map_ albedo tests) are blank; since these are visualisations of time-dependent vars maybe we can remove them entirely from static map cases.

We have some cleanup to do in our buildkite runs which will be tackled in another PR. I will add this to the list! Thank you!

src/shared_utilities/drivers.jl
```diff
@@ -304,6 +304,11 @@ Computes turbulent surface fluxes at a point on a surface given
 
 This returns an energy flux and a liquid water volume flux, stored in
 a tuple with self explanatory keys.
+
+Please note that this function, if r_sfc is set to zero, makes no alteration
```
juliasloan25, line 308: Maybe we could explain what it does if `r_sfc !=0` too

kmdeck: I am trying to work too quickly today and being a little sloppy - merged before I saw this. I'll fix this in my next PR.

PR #804, opened 2024-09-30
Change outdir and savedir to not use pkgdir

### LINKED ISSUE #545: ClimaLand should not use pkgdir for output
The experiments in ClimaLand use `pkgdir` to determine their output path. This is not a good idea. In geenral, `pkgdir` will be a subfolder in `.julia` and there is no reason to save output files in such a hard-to-find place. Second, `pkgdir` depends on how the package is installed, so it will lead to different results for the same simulation.

imreddyTeja:
All uses of pkgdir to specify save or output paths now use generate_output_path.

## Purpose 
Closes #545 

## Content

- output paths previously determined by pkgdir now use generate_output_path
- only remaining pkdir calls are used for includes

----


Sbozzolo (commented):
Can you please add some docs somewhere about the expected output structure (so that people know where they should look for files)?

It is described here: https://clima.github.io/ClimaUtilities.jl/dev/outputpathgenerator/

We use `ActiveLinkStyle`

experiments/README.md
```diff
@@ -9,6 +9,34 @@ tutorials.
...
+## Experiment Output Paths
+
+Many of the outputs from the experiments are saved using a non-destructive approach called
+`ActiveLinkStyle`. Each experiment has some unique save directory. For example, an experiment might have an
+output directory of `example_outdir`. When ran, if the directory does
+not exist relative to the working directory, then `example_outdir` is created. The results of the latest experiment
+run will be saved into `example_outdir/output_xxx`, where xxxx is the greatest
+index. The lower indices store the outputs of previous runs. The latest results are also linked to
+in `example_outdir/output_active`. More details on this style of output directory handeling
+can be found [here](https://clima.github.io/ClimaUtilities.jl/dev/outputpathgenerator/#ActiveLinkStyle-(Non-Destructive))
```
Sbozzolo, line 25:
```suggestion
Many of the outputs from the experiments are saved using a non-destructive approach called
`ActiveLinkStyle`. Each experiment has some unique save directory. For example, an experiment might have an
output directory of `example_outdir`. When run, if the directory does
not exist relative to the working directory, then `example_outdir` is created. The results of the latest experiment
run will be saved into `example_outdir/output_xxx`, where xxxx is an increasing counter. The lower indices store the outputs of previous runs. The latest results are also linked to
in `example_outdir/output_active`, which can be assumed to always contain the most recent ouput. More details on this style of output directory handeling
can be found [here](https://clima.github.io/ClimaUtilities.jl/dev/outputpathgenerator/#ActiveLinkStyle-(Non-Destructive))
```

experiments/README.md
```diff
@@ -9,6 +9,34 @@ tutorials.
...
+
+The expirements in experiments/long_runs save its image outputs to (experiment_name)\_longrun\_(device_suffix).
+All other outputs are saved using `ActiveLinkStyle` to (experiment_name)\_longrun\_(device_suffix)/global_diagnostics.
+
+### integrated and standalone
+
+Th experiments in experiments/long_runs and experiments/standalone have save paths that use `ActiveLinkStyle` and assume the experiments are being run from
+the ClimaLand.jl directory. If run from the ClimaLand.jl directory, then the outputs of an experiment will be saved into
+the same directory the experiment script is found in. Otherwise the outputs will be saved to
+the path of the script relative to the ClimaLand.jl directory.
```
Sbozzolo, line 42: Why do they have to be run from the ClimaLand.jl directory?

Sbozzolo:
You also have to update the artifact paths in the buildkite pipelines

experiments/README.md
```diff
@@ -9,6 +9,33 @@ tutorials.
...
+When a benchmark is run, its outputs are saved using `ActiveLinkStyle` into (benchmark_name)\_benchmark\_(device_suffix).
+
+### long_runs
+
+The expirements in experiments/long_runs save its image outputs to (experiment_name)\_longrun\_(device_suffix).
+All other outputs are saved using `ActiveLinkStyle` to (experiment_name)\_longrun\_(device_suffix)/global_diagnostics.
+
+### integrated and standalone
+
+Th experiments in experiments/long_runs and experiments/standalone have save paths that use `ActiveLinkStyle` and assume the experiments are being run from
```
Sbozzolo, line 38:
```suggestion
The experiments in experiments/long_runs and experiments/standalone have save paths that use `ActiveLinkStyle` and assume the experiments are being run from
```

Sbozzolo (approved):
Thank you!

Note that https://github.com/CliMA/ClimaLand.jl/pull/866 will be merged in minutes and you will need to rebase your PR because the two have some conflicts.

PR #805, opened 2024-09-30
qsat depends on T not snow presence

kmdeck:
## Purpose 
We had set q_sfc = q_sat(T, ice) if snow was on the ground, and q_sfc = q_sat(T, liq) if not. If the snow variable is incorrect (due to incorrect forcing or otherwise), this could mean very cold land would have a higher q_sfc (over liq). 

## To-do
Run in AMIP, Coupler PR https://github.com/CliMA/ClimaCoupler.jl/pull/976
Update unit tests to reflect change as needed

## Content
change q_sfc function

----


kmdeck:
Added in PR #807

PR #808, opened 2024-10-02
Update dependencies and add ocean mask

kmdeck:
## Purpose 
- add ocean mask (for plotting only!)
- up ClimaAnalysis to 0.5.9
- this updates the other dependencies also

## To-do
Future PR: move long_run/bucket.jl to experiments/standalone/Bucket/ in place of bucket_era5.jl
At the same time, use ClimaDiagnostics for all standalone bucket runs and compare diagnostics, not output state to CSV.
Replace plotting with ClimaAnalysis

## Content

----

### DIFF
--- a/experiments/standalone/Bucket/bucket_era5.jl
+++ b/experiments/standalone/Bucket/bucket_era5.jl
@@ -400,9 +400,9 @@ if device_suffix == "cpu"
     titles = ["W", "σS", "T_sfc"]
     for (f, n) in zip(fields, titles)
         fig = Figure(size = (1000, 1000))
-        ax = Axis(fig[1, 1])
+        ax = Axis(fig[1, 1], title = n)
         clims = extrema(f)
-        ClimaCoreMakie.fieldheatmap!(ax, f, title = n)
+        ClimaCoreMakie.fieldheatmap!(ax, f)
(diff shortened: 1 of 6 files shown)


experiments/standalone/Bucket/bucket_era5.jl
```diff
@@ -400,9 +400,9 @@ if device_suffix == "cpu"
     titles = ["W", "σS", "T_sfc"]
     for (f, n) in zip(fields, titles)
         fig = Figure(size = (1000, 1000))
-        ax = Axis(fig[1, 1])
+        ax = Axis(fig[1, 1], title = n)
```
kmdeck, line 403: I had to make this change to get this to pass, Gabriele.

kmdeck:
@juliasloan25 the long_run/bucket.jl script runs here, so maybe we can merge this first, and then your PR?

juliasloan25 (approved):
This looks good to me! The longrun plots still look good. We might want to make a note (in the commit or PR description) that the ocean mask is just used for plotting.

We'll also need to update to ClimaCore v0.14.18 after it's released later today, but I'm happy to do that as part of https://github.com/CliMA/ClimaLand.jl/pull/675

AlexisRenchon (approved):

PR #810, opened 2024-10-03
Refactor Canopy BC

kmdeck:
## Purpose 
We have recently refactored the boundary conditions for snow and soil (AtmosDriven...) to include a prognostic_land_component field, and we dispatch off of the Value of that field in order to correctly account for fluxes in the presence of other components in integrated models. This adds this to the canopy model.

## Content

----


src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -167,7 +235,7 @@ function canopy_boundary_fluxes!(
     lhf = p.canopy.energy.lhf
     r_ae = p.canopy.energy.r_ae
     i_end = canopy.hydraulics.n_stem + canopy.hydraulics.n_leaf
-
+    bc = canopy.boundary_conditions
```
kmdeck, line 238: we can remove this, because we pass `bc` in directly. Alternative, since there is only one type of bc for the canopy, we dont need to pass it in as an argument, and we can just access it from `canopy` which is passed in instead.

src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -184,29 +252,35 @@ function canopy_boundary_fluxes!(
         t,
     )
     # Update the root flux of water per unit ground area in place
+    # Note that here we dispatch off of the ground conditions `bc.ground`
+    # to handle standalone canopy simulations vs integrated ones
```
juliasloan25, line 256: maybe we could just make this comment once and explain that it applies to the following 3 flux calculations

src/standalone/Vegetation/ground_drivers.jl
```diff
@@ -0,0 +1,107 @@
...
+
+Returns the ground albedo in the PAR for a PrescribedGroundConditions driver. In this case,
+the prognostic_land_components only contain `:canopy`, because the canopy is being run in standalone
+mode.
+"""
+function ground_albedo_PAR(
+    prognostic_land_components::Val{(:canopy,)},
+    ground::PrescribedGroundConditions,
+    _...,
+)
```
juliasloan25, line 90: why not have this function just take in the two arguments?

juliasloan25 (commented):
This looks great! I like the "soil" -> "ground" rename. I just left a few comments but overall looks good to merge

src/standalone/Vegetation/canopy_boundary_fluxes.jl
kmdeck: will do!

src/standalone/Vegetation/ground_drivers.jl
kmdeck: in the integrated case, it also takes in `p`, and can access albedo from there (instead of storing albedo in `ground`)

src/integrated/soil_canopy_model.jl
```diff
@@ -421,36 +441,76 @@ function soil_boundary_fluxes!(
...
-Concrete type of AbstractSoilDriver used for dispatch in cases where both
-a canopy model and soil model are run.
-$(DocStringExtensions.FIELDS)
+A type of Canopy.AbstractGroundConditions to use when the soil model is prognostic and
+of type `EnergyHydrology`. This is required because the canopy model needs albedo of the ground
+in order to compute its update_aux! function, and that function must only depend on the canopy model.
+
+In the future, we will allocate space for albedo in the cache. In that case, we would *not*
+store them here, twice. `PrognosticSoilConditions` would
+then just be a flag, essentially.
```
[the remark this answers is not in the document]
kmdeck: this is a good question. RIght now I am kind of stuck because our "ground_albedo" functions are used in the canopy update_aux! function, which, because of our "standalone" design for models, can only access things stored in `canopy`. In the prescribed case this causes no issues, because we have the `ground` object that stores the albedo. But for the prognostic ground case, Im not sure how to dispatch off of the soil model (or soil + snow models..) since they arent in `canopy`. so the solution was to make a new`ground` with `typeof(ground) <: PrognosticSoilConditions` that instructs the canopy model where to look for the albedo (currently in `ground`, later in `p`.)

A separate issue is that in the future the albedo will be time varying/depend on soil and snow aux vars. so I actually think we may need to move some of the functions in canopy `update_aux` to later in the tendency function, since they depend on things that only get updated in the soil/snow `update_aux` function, and we dont really want to assume an ordering between these.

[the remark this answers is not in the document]
kmdeck: I sort of think if we solve the latter, it will help us with the former. because the land model `update_boundary_fluxes!` function can handle some of the interactions (e.g. update p.alpha_ground, and *then* call radiative transfer, and then update boundary fluxes of the component models).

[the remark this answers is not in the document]
juliasloan25: that makes sense to me! I think the current solution is good given the constraints we have, and we can revisit it down the line if/when things change

juliasloan25 (approved):

kmdeck:
I noticed a comment somewhere also about p.T_ground, but cant find it now. I got rid of it just because when we were using it, we were in a function that is specific to SoilCanopyModel, so I figured we could just use ClimaLand.Domains.top_center_to_surface(p.soil.T) directly (since this does not allocate either)

juliasloan25:
> I noticed a comment somewhere also about p.T_ground, but cant find it now. I got rid of it just because when we were using it, we were in a function that is specific to SoilCanopyModel, so I figured we could just use ClimaLand.Domains.top_center_to_surface(p.soil.T) directly (since this does not allocate either)

Yeah, this was a question I had at first but as I continued looking through the PR I figured it out :) nice change!

PR #739, opened 2024-08-27
fix buildkite run so that unit tests run on GPU

kmdeck:
## Purpose 
Some of our buildkite runs were not running on GPU. Update buildkite so that they are.

closes #860 

## Content
- Run tests on GPU (we thought we were doing this before but were not), and make some GPU-compatibility fixes
- Use Array(parent(...)) in tests for GPU compatibility.
This is required to enable GPU compatibility. Generally the pattern
`Array(parent(...))` is discouraged, but using it to check analytic
values in tests is okay.
- Add workaround to avoid type inference failure (initial bug described below)
This is required as a workaround to the type inference failure detailed 
in https://github.com/CliMA/ClimaCore.jl/issues/2065. The failure appears
when a function is broadcast over one field and two structs; the types
are too complex and inference fails. This workaround introduces a wrapper
struct to contain the two struct inputs, so that the function is broadcast
over one field and only one struct, and the types are easier to infer.

## Initial bug reported
1. Fix bug in `experiments/integrated/performance/profile_allocations.jl`. Note that the same code runs on the Clima cluster on GPU. This is the error message:
```
(base) [kdeck@hpc-22-32 ClimaLand.jl]$ julia --project=.buildkite  experiments/integrated/performance/profile_allocations.jl
[ Info: ProfileCanvas found, running with profiler
[...]

PR #812, opened 2024-10-03
Remove root_distribution as param

imreddyTeja:
Previously, rooting_depth was added as a PlayHydraulics param to replace rootdistribution. Support for root_distribution continued, but this caused an isbits error with broadcasting. This commit removes support for the deprecated root_distribution param

One of the tests I changed in the previous PR fails when run with CUDA. Is this something that I should fix, and if so, should it be in a separate pr?

## Content

----

### DIFF
--- a/deprecated_features.md
+++ b/deprecated_features.md
@@ -2,8 +2,3 @@
 
 This file lists deprecated features and the recommended alternative practice
 
-- ## `root_distribution`
-
-    The `root_distribution` in `PlantHydraulicsParameters` is replaced by `rooting_depth`.
-    If using a `root_distribution` function of the form P(x) = 1/d e^(z/d), then `rooting_depth`
-    is equivalent to d.
--- a/src/integrated/soil_canopy_model.jl
+++ b/src/integrated/soil_canopy_model.jl
@@ -261,11 +261,6 @@ function make_update_boundary_fluxes(
         # Note that in `PrescribedSoil` mode, we compute the flux using K_soil = K_plant(ψ_soil)
         # and K_canopy = K_plant(ψ_canopy). In `PrognosticSoil` mode here, we compute the flux using
         # K_soil = K_soil(ψ_soil) and K_canopy = K_plant(ψ_canopy).
-        # if rooting_depth param is not nothing, use root_distribution from source
-        # otherwise use root_distribution from params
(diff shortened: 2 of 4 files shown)


juliasloan25 (approved):
This looks good to me. Thanks for the quick turnaround!

Which test fails on GPU?

imreddyTeja:
"Plant hydraulics model integration tests" in plant_hydraulics_test. I'm fairly certain the cause was changes I made to make the tests work with fields. 

https://github.com/CliMA/ClimaLand.jl/blob/f780e3644df832ab5e7905d44a6a484d2b443cdd/test/standalone/Vegetation/plant_hydraulics_test.jl#L337-L387

juliasloan25:
> "Plant hydraulics model integration tests" in plant_hydraulics_test. I'm fairly certain the cause was changes I made to make the tests work with fields.
> 
> https://github.com/CliMA/ClimaLand.jl/blob/f780e3644df832ab5e7905d44a6a484d2b443cdd/test/standalone/Vegetation/plant_hydraulics_test.jl#L337-L387

Hmm, it looks like all the tests pass on GPU in this PR (and main). Do you mean that the test was failing on GPU with the root distribution changes, but you updated the test with a workaround? If that's the case I think it's a good idea to revert it, but maybe in a separate PR depending on how much work you think it will be

imreddyTeja:
If I run "test/runtests.jl", or specifically "test/standalone/Vegetation/plant_hydraulics_test.jl", with the ClimaComms device set to CUDA, then there is a bits type error. Everything runs fine if using the cpu.

juliasloan25:
> If I run "test/runtests.jl", or specifically "test/standalone/Vegetation/plant_hydraulics_test.jl", with the ClimaComms device set to CUDA, then there is a bits type error. Everything runs fine if using the cpu.

That's weird - we have a run on buildkite that runs `run_tests.jl` on GPU (e.g. [here](https://buildkite.com/clima/climaland-ci/builds/5410#01925395-7729-41e8-a772-3f853bc56fb9)). It could be that we aren't setting up the environment correctly for that run

juliasloan25:
Ah, I remembered that we aren't actually running our tests on GPU at the moment - see https://github.com/CliMA/ClimaLand.jl/pull/739

PR #816, opened 2024-10-04
Fix calibration perfect model, add observations tutorial, fix ClimaLandSimulations

AlexisRenchon:
- edits to perfect model calibration
- fix ClimaLandSimulations `run_fluxnet()`
- add latent heat flux calibration tutorial

https://clima.github.io/ClimaLand.jl/previews/PR816/

### DIFF
--- a/docs/list_tutorials.jl
+++ b/docs/list_tutorials.jl
@@ -31,9 +31,10 @@ tutorials = [
             "standalone/Snow/data_tutorial.jl",
         ],
     ],
-    #"Calibrating your ClimaLand model" => [
-    #    "Recover parameters of a previous simulation" => "calibration/minimal_working_example.jl",
-    #],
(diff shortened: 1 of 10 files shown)


AlexisRenchon:
@kmdeck @Sbozzolo 

I set nensemble and niterations to 3 each, and it gives nice results still, but documentation build went from 18 min to 37 min... is it okay?

Sbozzolo:
Can't we just have a small use case and paste the picture of the completed calibration, as you suggested?

AlexisRenchon:
> Can't we just have a small use case and paste the picture of the completed calibration, as you suggested?

it is already the small use case. otherwise it takes 2 hours. 
it will be faster with the suggestions - almost done!

Sbozzolo (approved):

PR #683, opened 2024-06-28
tag v0.12.5

AlexisRenchon:
## Purpose 
Update release to v0.12.5

PR #817, opened 2024-10-04
Tag v0.15.1

trontrytel:
I went through the NEWS and did not see any breaking changes. So I'm thinking a patch release should be fine?

Are there any Manifest files I should modify to make sure Thermodynamics v.12.8 is used when testing?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaLand"
 uuid = "08f4d4ce-cf43-44bb-ad95-9d2d5f413532"
 authors = ["Clima Land Team"]
-version = "0.15.0"
+version = "0.15.1"
 
 [deps]
 ArtifactWrappers = "a14bc488-3040-4b00-9dc1-f6467924858a"
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,10 @@ ClimaLand.jl Release Notes
 
 main
 --------
+
+
+v0.15.1
+--------
 - Add a tutorial of a single site calibration of a perfect model PR[#621](https://github.com/CliMA/ClimaLand.jl/pull/621)
 - Add a longrun simulation for the bucket PR[#807](https://github.com/CliMA/ClimaLand.jl/pull/807)
 - Add ground heat flux to snow-soil model PR[#796](https://github.com/CliMA/ClimaLand.jl/pull/796)
--- a/docs/list_tutorials.jl
+++ b/docs/list_tutorials.jl
@@ -31,9 +31,9 @@ tutorials = [
             "standalone/Snow/data_tutorial.jl",
         ],
     ],
-    "Calibrating your ClimaLand model" => [
-        "Recover parameters of a previous simulation" => "calibration/minimal_working_example.jl",
-    ],
+    #"Calibrating your ClimaLand model" => [
+    #    "Recover parameters of a previous simulation" => "calibration/minimal_working_example.jl",
+    #],
     "For model developers" => [
         "Intro to standalone models" => "standalone/Usage/model_tutorial.jl",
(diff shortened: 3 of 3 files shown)


juliasloan25:
I think a patch release is good.

We have two checked-in Manifests: docs and .buildkite, so these will need to be updated for the new Thermodynamics version. I usually update all dependencies in these two Manifests whenever I make a release, to make sure we're using all the newest versions periodically

Project.toml
```diff
@@ -58,6 +58,6 @@ SciMLBase = "2"
 StaticArrays = "1"
 StatsBase = "0.34"
 SurfaceFluxes = "0.11, 0.12"
-Thermodynamics = "0.12.4"
+Thermodynamics = "0.12.8"
```
juliasloan25, line 61: why does land require v0.12.8 now?

trontrytel: I wanted to make sure that in the Coupler all packages can run with Thermodynamics v 0.12.8. Will it not cause problems that we set the patch version here?

trontrytel:
> I think a patch release is good.
> 
> We have two checked-in Manifests: docs and .buildkite, so these will need to be updated for the new Thermodynamics version. I usually update all dependencies in these two Manifests whenever I make a release, to make sure we're using all the newest versions periodically

How do you typically update the buildkite Manifest.toml?

Project.toml
juliasloan25: If we specify 0.12.4 here, land will be compatible with anything in the range [0.12.4, 0.13.0). Since our `test/` Manifest isn't checked in, our tests run with the most recent allowed version of each package (so 0.12.8 for Thermodynamics), so that will tell us if land is able to run with Thermodynamics v0.12.8. We only want to update the patch compat in land if land requires something introduced in 0.12.8 (e.g. a bugfix) to be able to run

juliasloan25:
> > I think a patch release is good.
> > We have two checked-in Manifests: docs and .buildkite, so these will need to be updated for the new Thermodynamics version. I usually update all dependencies in these two Manifests whenever I make a release, to make sure we're using all the newest versions periodically
> 
> How do you typically update the buildkite Manifest.toml?

`julia --project=.buildkite` and then `]up` (same for `docs` environment)

> 
> Also, it seems that the buildkite CI did run with Thermodynamics v0.12.8 even though I did not touch the Manifest?: https://buildkite.com/clima/climaland-ci/builds/5449#01925884-8a82-4cdd-a30c-0a51140f0007

I think this is because you set the `Thermodynamics = "0.12.8"` compat in the top-level project, and the manifests are all resolved at the beginning of the pipeline so they get updated to comply with any compat

trontrytel:
Thank you @juliasloan25 

So I should keep the 0.12.4 in the Project.toml, but make sure the two Manifest files use 0.12.8?

juliasloan25:
> Thank you @juliasloan25
> 
> So I should keep the 0.12.4 in the Project.toml, but make sure the two Manifest files use 0.12.8?

Yes :) you should do this:
- keep 0.12.4 in the Project.toml
- run `]up` in the `docs` and `.buildkite` environments
  - this will update all dependencies (including Thermodynamics to 0.12.8) - you can double check this on the buildkite runs too
- also just update the land version to 0.15.1 in the Project.toml and NEWS.md (which you've already done)

juliasloan25:
The files changed after all that should look something like this [previous release](https://github.com/CliMA/ClimaLand.jl/pull/683/files). I just realized that Kat updated our dependencies yesterday, so the Manifests may not have many (any?) changes

trontrytel:
One more try @juliasloan25  :)

juliasloan25 (approved):

PR #819, opened 2024-10-04
Reduce allocations I

kmdeck:
## Purpose 
Reduce allocations by 
(1) making boundary_flux! update in place rather than boundary_flux() creating and returning a field
(2) Allocate space in cache for canopy stomatal resistance instead of allocating each step
(3) compute rho_sfc and q_sfc in canopy_turbulent_fluxes_at_a_point instead of allocating the fields each step and passing to this function

## Content
As above

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,8 @@ main
 
 v0.15.2
 --------
-
+- Boundary fluxes are non-allocating
+PR[#819](https://github.com/CliMA/ClimaLand.jl/pull/819)
 - Artifacts for the bucket model are now automatically downloaded. PR [#820](https://github.com/CliMA/ClimaLand.jl/pull/820)
 
 v0.15.1
--- a/docs/src/APIs/shared_utilities.md
+++ b/docs/src/APIs/shared_utilities.md
@@ -55,7 +55,7 @@ ClimaLand.source!
 ClimaLand.AbstractBoundary
 ClimaLand.TopBoundary
 ClimaLand.BottomBoundary
-ClimaLand.boundary_flux
+ClimaLand.boundary_flux!
 ClimaLand.diffusive_flux
 ClimaLand.boundary_vars
 ClimaLand.boundary_var_domain_names
--- a/src/standalone/Soil/boundary_conditions.jl
+++ b/src/standalone/Soil/boundary_conditions.jl
@@ -1,7 +1,7 @@
 import ClimaLand:
     AbstractBoundary,
     AbstractBC,
-    boundary_flux,
+    boundary_flux!,
     boundary_vars,
     boundary_var_domain_names,
     boundary_var_types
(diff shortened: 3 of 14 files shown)


docs/src/APIs/shared_utilities.md
```diff
@@ -55,7 +55,7 @@ ClimaLand.source!
 ClimaLand.AbstractBoundary
 ClimaLand.TopBoundary
 ClimaLand.BottomBoundary
-ClimaLand.boundary_flux
+ClimaLand.boundary_flux!
```
Sbozzolo, line 58: Just pointing out that you are modifying the declared public API of the package, so this would be a breaking change if we wanted to be pedantic.

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -156,36 +156,39 @@ boundary_var_types(
...
     Δz::ClimaCore.Fields.Field,
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
     t,
-)::ClimaCore.Fields.Field
+)
     FT = eltype(Δz)
-    return FT.(bc.bc(p, t)) .+ FT.(ClimaCore.Fields.zeros(axes(Δz)))
+    bc_val = FT(bc.bc(p, t))
+    @. bc_field = bc_field * 0
```
Sbozzolo, line 178:
```suggestion
    @. fill!(bc_field, 0)
```
or 
```suggestion
    bc_field .= 0
```

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -439,89 +454,94 @@ end
...
     Δz::ClimaCore.Fields.Field,
     Y::ClimaCore.Fields.FieldVector,
     p::NamedTuple,
     t,
-)::ClimaCore.Fields.Field
+)
     FT = eltype(Δz)
-    return FT.(bc.bc(p, t)) .+ FT.(ClimaCore.Fields.zeros(axes(Δz)))
+    bc_val = FT.(bc.bc(p, t))
+    @. bc_field = bc_field * 0
```
Sbozzolo, line 476: Same as above

Sbozzolo: Actually, why is this needed? You are writing a different value in the line below?

kmdeck: youre right! I think we can just directly do:
`bc_field .= bc_val`

docs/src/APIs/shared_utilities.md
kmdeck: I added a breaking change to NEWS.md

src/standalone/Soil/boundary_conditions.jl
kmdeck: thanks!

Sbozzolo (approved):

NEWS.md
```diff
@@ -6,7 +6,8 @@ main
 
 v0.15.2
 --------
-
+- ![][badge-💥breaking] Boundary fluxes are non-allocating
+PR[#819](https://github.com/CliMA/ClimaLand.jl/pull/819)
```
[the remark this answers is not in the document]
kmdeck: I made this non-breaking, thank you!

Issue #580, opened 2024-04-08
Remove ArtifactsWrappers and document artifacts


Sbozzolo:
Following [ClimaArtifacts](https://github.com/CliMA/ClimaArtifacts/)

Artifacts downloaded
```[tasklist]
### to be moved to ClimaArtifacts
- [x] cesm2_albedo, bareground_albedo [Teja] https://github.com/CliMA/ClimaLand.jl/pull/820
- [x] richards_eqn_bonan_data (clay, sand) [Julia] https://github.com/CliMA/ClimaLand.jl/pull/900
- [x] lehmann_assouline_or2008_evaporation_data [Julia] -> (rename) lehmann2008_evaporation https://github.com/CliMA/ClimaLand.jl/pull/929
- [x] experiment_fluxnet_data (for each site) [Eddie] https://github.com/CliMA/ClimaLand.jl/pull/914
- [x] 2_str_test_data (only used in test) [Eddie] https://github.com/CliMA/ClimaLand.jl/pull/910
- [x] topmodel_data [Katherine] https://github.com/CliMA/ClimaLand.jl/pull/928
- [x] water_conservation_test_data (flux_dataset, dirichlet_dataset) [Julia] https://github.com/CliMA/ClimaLand.jl/pull/1005
- [X] huang_et_al2011_soil_van_genuchten_data [Katherine]
- [ ] snotel data [Gabriele]
- [X] mizoguchi1990_soil_freezing_data - at least move out of src [?]
```

PR #820, opened 2024-10-04
Change albedo artifacts to use ClimaArtifacts

imreddyTeja:
Both cesm2_albedo_dataset_path and bareground_albedo_dataset_path now use the new ClimaArtifact cesm2_albedo. This commit also deletes src/standalone/Bucket/artifacts, which was in an include, but never used.

Previously, the there were to functions in Artifacts.jl, bareground_albedo_dataset_path and cesm2_albedo_dataset_path. These now call a  single function cesm2_albedo_dataset_folder and join path with the correct file name.  To completely conform to the way other ClimaArtifacts are used, the two functions that return the specific file path should be removed. However, this would make the places the files are used messy, so I left the functions.   

Note that esw_albedo_Amon_CESM2_historical_r1i1p1f1_gn_185001-201412_v2.nc is replaced with sw_albedo_Amon_CESM2_historical_r1i1p1f1_gn_185001-201412_v2_no-nans.nc, which is exactly the same. 

For more details see related ClimaArtifacts PR [here](https://github.com/CliMA/ClimaArtifacts/pull/48)

## Purpose 
This PR progressed towards #580 
It gets rid of ArtifactWrappers for land albedo and average bareground albedo. 

## Content

----


juliasloan25 (approved):
This looks great! I think it's a good idea to keep the path functions - it's nice to abstract the specific filenames away from the user

PR #821, opened 2024-10-04
recreate ClimaUtilities read! bug

juliasloan25:
## Purpose 
bug in ClimaUtilities v0.1.11, fixed in 0.1.12

## Content

----


juliasloan25:
we have recreated and fixed the bug in #119, so we don't need this anymore

Issue #760, opened 2024-09-10
streamlined PrescribedAtmos constructor


juliasloan25:
**Is your feature request related to a problem? Please describe.**
Following https://github.com/CliMA/ClimaUtilities.jl/pull/105, we're able to do transformations on multiple data variables to create our forcing variables (and can read them all from one file). It would be nice to use this to make a `PresribedAtmos` constructor that avoids the user having to do this:
```julia
  u_atmos = TimeVaryingInput(
        joinpath(era5_artifact_path, "era5_2021_0.9x1.25_clima.nc"),
        "ws",
        surface_space;
        reference_date = ref_time,
        t_start,
        regridder_type,
    )
    q_atmos = TimeVaryingInput(
        joinpath(era5_artifact_path, "era5_2021_0.9x1.25_clima.nc"),
        "q",
        surface_space;
        reference_date = ref_time,
        t_start,
        regridder_type,
    )
    P_atmos = TimeVaryingInput(
        joinpath(era5_artifact_path, "era5_2021_0.9x1.25.nc"),
        "sp",
        surface_space;
        reference_date = ref_time,
        t_start,
        regridder_type,
    )
```
etc for each variable and then
```julia
    atmos = PrescribedAtmosphere(
        precip,
        snow_precip,
        T_atmos,
        u_atmos,
        q_atmos,
        P_atmos,
        ref_time,
        h_atmos,
        earth_param_set,
    )
```
and instead just takes the file, earth_param_set, ref_time... etc and makes atmos directly. We may still have to create an object for each individual variable internally.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #823, opened 2024-10-07
composition of era5 data and helper functions

kmdeck:
## Purpose 
Closes https://github.com/CliMA/ClimaLand.jl/issues/760

## To-do
[X] verify runs in CI

## Content
1. Uses built in composition capability of ClimaUtilities, instead of the pre-composed vars in the _clima.nc files
2. Writes a helper function to make the prescribed atmos, radiation, and lai objects for ERA5 data. I kept LAI separate since we may want to use MODIS data for that instead.

Note that: We cant test the helper function with github unit tests because we dont have era5 data accessible there (only on caltech cluster). But we do use the function on buildkite CI, so it seems OK/it is tested

----


src/shared_utilities/drivers.jl
```diff
@@ -713,6 +722,50 @@ function relative_humidity(
...
+We first compute the relative humidity using the Magnus formula, then the saturated vapor pressure, and then
+we compute q from vapor pressure and saturated vapor pressure.
+
+For more information on the Magnus formula, see e.g. 
+Lawrence, Mark G. (1 February 2005). 
+"The Relationship between Relative Humidity and the Dewpoint Temperature in Moist Air: 
+A Simple Conversion and Applications". 
+Bulletin of the American Meteorological Society. 86 (2): 225–234.
+"""
+function specific_humidity_from_dewpoint(
```
kmdeck, line 741: @trontrytel

docs/tutorials/shared_utilities/driver_tutorial.jl
```diff
@@ -139,3 +139,31 @@ updatefunc = update_radiation!;
 cb = ClimaLand.DriverUpdateCallback(updateat, updatefunc);
 
 # This callback must then be provided to the simulation [`solve`](https://docs.sciml.ai/DiffEqCallbacks/stable/) function.
+
+# # Using ERA5 data
+# If you wish to force your ClimaLand simulation with ERA5 reanalysis data, there is a helper function
+# which can make this easier than specifying each atmospheric variable as individual `TimeVaryingInput`
+# objects, and then making the `PrescribedAtmosphere` struct.
+# You first need a local path to the netcdf file with the ERA5 data. This file must have the following variables:
+# - "tp" Total precipitation as a mass/m^2, accumulated over an hour
```
AlexisRenchon, line 148: it is okay as is but this is equivalent to mass/m^2/hour, right? 
I know it is common to report precip like that, but is it not clearer to just give the full unit? just a thought.

src/shared_utilities/drivers.jl
```diff
@@ -713,6 +723,69 @@ function relative_humidity(
     return es / ea
 end
 
+"""
+    specific_humidity_from_dewpoint(dewpoint_temperature, temperature, air_pressure, earth_param_set)
+
+Estimates the specific humidity given the dewpoint temperature, temperature of the air
+in Kelvin, and air pressure in Pa,  along with the ClimaLand earth_param_set. This is useful
```
AlexisRenchon, line 730: typo: many spaces between Pa, and along with

src/shared_utilities/drivers.jl
```diff
@@ -713,6 +723,69 @@ function relative_humidity(
...
+function specific_humidity_from_dewpoint(
+    T_dew_air::data_FT,
+    T_air::data_FT,
+    P_air::data_FT,
+    earth_param_set,
+) where {data_FT <: Real}
+    thermo_params = LP.thermodynamic_parameters(earth_param_set)
+    _T_freeze = LP.T_freeze(earth_param_set)
+    sim_FT = typeof(_T_freeze)
+    # Obtain the relative humidity. This function requires temperatures in Celcius
```
AlexisRenchon, line 751: typo, celsius

src/shared_utilities/drivers.jl
```diff
@@ -713,6 +723,69 @@ function relative_humidity(
...
+    e = rh * esat
+    q = data_FT(0.622 * e / (P_air - 0.378 * e))
+    return q
+end
+
+"""
+    rh_from_dewpoint(Td_C, T_C)
+
+Returns the relative humidity given the dewpoint temperature in Celsius and the 
+air temperature in Celcius, using the Magnus formula.
```
AlexisRenchon, line 773: Celsius =)

AlexisRenchon: I know we're in the land of Fahrenheit but...

AlexisRenchon: "Celcius" is a common typo for several reasons:

Phonetic Similarity: The "s" sound in "Celsius" can lead people to mistakenly insert an "i" after the "l" when spelling it, making it look like "Celcius."

Familiarity with Other Words: Words like "delicious" or "vicious" have a similar ending, so people may unconsciously apply that pattern when spelling "Celsius."

Lack of Awareness: Since "Celsius" is derived from the name of a person (Anders Celsius), the correct spelling might not be as well-known, leading to more frequent errors.

Overall, it's an understandable mistake, but the correct form is "Celsius."

AlexisRenchon (approved):
just some typo and some thoughts, but the PR looks great! Thank you!

src/shared_utilities/drivers.jl
kmdeck: > I know we're in the land of Fahrenheit but...

lol, Im slightly embarassed!!

AlexisRenchon: yes! I asked ChatGPT if this was a common typo

docs/tutorials/shared_utilities/driver_tutorial.jl
kmdeck: I updated this! Thanks

Issue #777, opened 2024-09-17
Consistency in fluxes and setting radiative properties of land


kmdeck:
## OKR2.3.8:
"Prepare for coupled runs: ensure that we follow the flux and unit conventions set by the coupler. Compute bulk albedo and temperature of surface, set emissivity to 1."

## Units
- all energy fluxes are in J/m^2/s
- we store evaporation as a liquid water volume flux. The coupler converts the mass flux computed by atmos for us: https://github.com/CliMA/ClimaCoupler.jl/blob/59abc31b66ccb6d62a06fb86be9f0340fae4d4c0/experiments/ClimaEarth/components/land/climaland_bucket.jl#L211
- we store precipitation as a liquid water volume flux. The coupler converts a mass flux to volume flux for us right now: https://github.com/CliMA/ClimaCoupler.jl/blob/59abc31b66ccb6d62a06fb86be9f0340fae4d4c0/experiments/ClimaEarth/components/land/climaland_bucket.jl#L197
**Land to do: convert internally to volume fluxes**
- What units will carbon fluxes be in? (mass flux of C)

## Flux sign convention
We have followed the convention for fluxes that positive = towards atmos and negative = towards land. This applies for precipitation, SHF, LHF, evap, and everywhere in the land model, *except radiation*. 

**To do: change precip to be defined to be positive**

## Radiation internals for Land
For radiation, we have "LW_d" and "SW_d" in prescribed atmosphere runs. These are scalars, and indicate the magnitude of the downwelling radiation. They are always positive. We then compute "LW_u" and SW_u", which are the magnitude of the upwelling radiation, and hence are always positive. 

**To do: Change all instances of in/out to down/up.**

2. Internally, we also compute the net radiation per land component "R_n = downwelling - upwelling".  Hence this is positive if the land component is absorbing energy. Note that this is *opposite* the flux convention we use everywhere else, with negative indicating "towards land".
- ChosencConvention: R_n > 0 implies land absorbs energy. The net flux at the surface would then be SHF + LHF - R_n.

 **To do: make sure this convention is followed throughout**
## Coupled atmosphere runs - radiation:
- we will need SW_d in PAR and NIR wavelength bands currently, eventually we may have a more spectrally resolved model. Can the coupler carry out the conversion between the bands of the atmosphere model and of the land model?
- What sign convention will the downwelling SW fluxes follow?  **Answer: same as ours**
- we will need LW_d. What sign convention will this follow? **Answer: same as ours**
[...]

PR #834, opened 2024-10-09
Snowy land

### LINKED ISSUE #630: O2.1.3 Integrate snow model with soil and canopy model
Now that we have a basic standalone snow model, we need to integrate it with the other components. We will do this in two stages
- [x] integrate with soil #629 #775 #773 #779 
- [ ] integrate with soi/canopy model https://github.com/CliMA/ClimaLand.jl/pull/834

kmdeck:
## Purpose 
Snowy land. Closes #630 
Note: I will turn back on the long runs before merging, after incorporating review comments, to save running a lot on clima

ClimaLand with snow (note that the IC is zero snow) from global run:
water
bio
other

## To-do
This PR is already so big!
Tutorial?
Tests? 
Assess conservation in future PR, also addressing #847. Branch is kd/assess_conservation

## Content
1. This adds the snow/soil/canopy model, called "LandModel".
2. It adds diagnostics for this model (this has to be done in this PR to see output in run_fluxnet or the global runs, since those use diagnostics)
3. Creates a global run
4. Uses the snowy land model in our fluxnet runs and in the conservation run

----

### DIFF
--- a/.buildkite/longruns_gpu/pipeline.yml
+++ b/.buildkite/longruns_gpu/pipeline.yml
@@ -31,6 +31,16 @@ steps:
   - group: "Global Land Models"
     steps:
 
+      - label: ":snow_capped_mountain: Snowy Land"
+        command:
+          - julia --color=yes --project=.buildkite experiments/long_runs/snowy_land.jl
+        artifact_paths: "snowy_land_longrun_gpu/*png"
+        agents:
+          slurm_gpus: 1
+          slurm_time: 03:30:00
+        env:
+          CLIMACOMMS_DEVICE: "CUDA"
+
       - label: ":seedling: Soil-Canopy"
         command:
           - julia --color=yes --project=.buildkite experiments/long_runs/land.jl
@@ -41,7 +51,7 @@ steps:
         env:
           CLIMACOMMS_DEVICE: "CUDA"
 
-      - label: ":sun: California regional simulation"
+      - label: ":sunglasses: California regional simulation"
         command:
           - julia --color=yes --project=.buildkite experiments/long_runs/land_region.jl
         artifact_paths: "california_longrun_gpu/*png"
--- a/experiments/integrated/fluxnet/US-Ha1/US-Ha1_simulation.jl
+++ b/experiments/integrated/fluxnet/US-Ha1/US-Ha1_simulation.jl
@@ -17,7 +17,7 @@ h_stem = FT(14) # m
 # TIME STEPPING:
 
 # Starting time:
-t0 = Float64(120 * 3600 * 24)# start mid year to avoid snow
+t0 = Float64(0)
 
 # Time step size:
 dt = Float64(450)
--- a/experiments/integrated/fluxnet/US-NR1/US-NR1_simulation.jl
+++ b/experiments/integrated/fluxnet/US-NR1/US-NR1_simulation.jl
@@ -17,7 +17,7 @@ h_stem = FT(7.5) # m
 # TIME STEPPING:
 
 # Starting time:
-t0 = Float64(120 * 3600 * 24)# start mid year to avoid snow
+t0 = Float64(0)
 
 # Time step size:
-dt = Float64(450)
+dt = Float64(200)
--- a/experiments/integrated/fluxnet/fluxnet_simulation.jl
+++ b/experiments/integrated/fluxnet/fluxnet_simulation.jl
@@ -2,8 +2,9 @@
 fluxtower sites. These work in tandem with the site-specific timing parameters
 found in the {site-ID}_simulation.jl files in each site directory."""
 
-N_spinup_days = 30
-N_days = N_spinup_days + 30
+N_spinup_days = 15
+N_days = N_spinup_days + 340
+
 tf = Float64(t0 + 3600 * 24 * N_days)
 t_spinup = Float64(t0 + N_spinup_days * 3600 * 24)
 
--- a/experiments/integrated/fluxnet/run_fluxnet.jl
+++ b/experiments/integrated/fluxnet/run_fluxnet.jl
@@ -12,6 +12,7 @@ using StatsBase
 
 using ClimaLand
 using ClimaLand.Domains: Column
+using ClimaLand.Snow
 using ClimaLand.Soil
 using ClimaLand.Soil.Biogeochemistry
 using ClimaLand.Canopy
@@ -186,6 +187,11 @@ shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
 )
 
 canopy_model_args = (; parameters = shared_params, domain = canopy_domain)
+
+# Snow model
+snow_parameters = SnowParameters{FT}(dt; earth_param_set = earth_param_set);
+snow_args = (; parameters = snow_parameters, domain = canopy_domain);
+snow_model_type = Snow.SnowModel
 # Integrated plant hydraulics and soil model
 land_input = (
     atmos = atmos,
@@ -193,7 +199,7 @@ land_input = (
     soil_organic_carbon = Csom,
     runoff = ClimaLand.Soil.Runoff.SurfaceRunoff(),
 )
-land = SoilCanopyModel{FT}(;
+land = LandModel{FT}(;
     soilco2_type = soilco2_type,
     soilco2_args = soilco2_args,
     land_args = land_input,
@@ -202,6 +208,8 @@ land = SoilCanopyModel{FT}(;
     canopy_component_types = canopy_component_types,
     canopy_component_args = canopy_component_args,
     canopy_model_args = canopy_model_args,
+    snow_args = snow_args,
+    snow_model_type = snow_model_type,
 )
 
 Y, p, cds = initialize(land)
@@ -240,6 +248,10 @@ for i in 1:(n_stem + n_leaf)
 end
 
 Y.canopy.energy.T = drivers.TA.values[1 + Int(round(t0 / DATA_DT))] # Get atmos temperature at t0
+
+Y.snow.S .= 0.0
+Y.snow.U .= 0.0
+
 set_initial_cache! = make_set_initial_cache(land)
 set_initial_cache!(p, Y, t0);
 
@@ -309,6 +321,7 @@ short_names_1D = [
     "lhf", # LHF
     "ghf", # G
     "rn", # Rn
+    "swe",
 ]
 short_names_2D = [
     "swc", # swc_sfc or swc_5 or swc_10
@@ -323,11 +336,11 @@ hourly_diag_name_2D = short_names_2D .* "_1h_average"
 # diagnostic_as_vectors()[2] is a vector of a variable,
 # whereas diagnostic_as_vectors()[1] is a vector or time associated with that variable.
 # We index to only extract the period post-spinup.
-SIF, AR, g_stomata, GPP, canopy_T, SW_u, LW_u, ER, ET, β, SHF, LHF, G, Rn = [
-    ClimaLand.Diagnostics.diagnostic_as_vectors(d_writer, diag_name)[2][(N_spinup_days * 24):end]
-    for diag_name in hourly_diag_name
-]
-
+SIF, AR, g_stomata, GPP, canopy_T, SW_u, LW_u, ER, ET, β, SHF, LHF, G, Rn, SWE =
+    [
+        ClimaLand.Diagnostics.diagnostic_as_vectors(d_writer, diag_name)[2][(N_spinup_days * 24):end]
+        for diag_name in hourly_diag_name
+    ]
 
 swc, soil_T, si = [
     ClimaLand.Diagnostics.diagnostic_as_vectors(
@@ -510,7 +523,7 @@ if isfile(
     )
 end
 
-# Water content in soil
+# Water content in soil and snow
 # Soil water content
 # Current resolution has the first layer at 0.1 cm, the second at 5cm.
 plt1 = Plots.plot(size = (1500, 800))
@@ -546,7 +559,23 @@ if drivers.SWC.status != absent
         label = "Data",
     )
 end
-plt2 = Plots.plot(
+plt2 = Plots.plot(size = (1500, 800))
+Plots.plot!(
+    plt2,
+    model_times ./ 3600 ./ 24,
+    SWE,
+    label = "Model",
+    xlim = [
+        minimum(model_times ./ 3600 ./ 24),
+        maximum(model_times ./ 3600 ./ 24),
+    ],
+    ylim = [0.0, 0.15],
+    xlabel = "Days",
+    ylabel = "SWE [m]",
+    color = "blue",
+    margin = 10Plots.mm,
+)
+plt3 = Plots.plot(
     data_times ./ 3600 ./ 24,
     (drivers.P.values .* (-1e3 * 24 * 3600) .* (1 .- snow_frac))[data_id_post_spinup],
     label = "Rain (data)",
@@ -560,13 +589,13 @@ plt2 = Plots.plot(
     size = (1500, 400),
 )
 Plots.plot!(
-    plt2,
+    plt3,
     data_times ./ 3600 ./ 24,
     (drivers.P.values .* (-1e3 * 24 * 3600) .* snow_frac)[data_id_post_spinup],
     label = "Snow (data)",
     ylabel = "Precipitation [mm/day]",
 )
-Plots.plot(plt2, plt1, layout = grid(2, 1, heights = [0.3, 0.7]))
+Plots.plot(plt3, plt2, plt1, layout = grid(3, 1, heights = [0.2, 0.4, 0.4]))
 Plots.savefig(joinpath(savedir, "ground_water_content.png"))
 
 plt1 = Plots.plot(size = (1500, 800))
--- a/experiments/long_runs/snowy_land.jl
+++ b/experiments/long_runs/snowy_land.jl
@@ -82,7 +82,6 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         earth_param_set,
         FT;
         time_interpolation_method = time_interpolation_method,
-        regridder_type = regridder_type,
     )
 
     spatially_varying_soil_params =
@@ -235,7 +234,6 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         surface_space,
         start_date;
         time_interpolation_method = time_interpolation_method,
-        regridder_type = regridder_type,
     )
     ai_parameterization =
         Canopy.PrescribedSiteAreaIndex{FT}(LAIfunction, SAI, RAI)
--- a/experiments/long_runs/snowy_land.jl
+++ b/experiments/long_runs/snowy_land.jl
@@ -1,0 +1,458 @@
+# # Global run of land model
+
+# The code sets up and runs ClimaLand v1, which
+# includes soil, canopy, and snow, for 365 days on a spherical domain,
+# using ERA5 data as forcing. In this simulation, we have
+# turned lateral flow off because horizontal boundary conditions and the
+# land/sea mask are not yet supported by ClimaCore.
+
+# Simulation Setup
+# Number of spatial elements: 101 in horizontal, 15 in vertical
+# Soil depth: 50 m
+# Simulation duration: 365 d
+# Timestep: 450 s
+# Timestepper: ARS111
+# Fixed number of iterations: 3
+# Jacobian update: every new Newton iteration
+# Atmos forcing update: every 3 hours
+import SciMLBase
+import ClimaComms
+ClimaComms.@import_required_backends
+import ClimaTimeSteppers as CTS
+using ClimaCore
+using ClimaUtilities.ClimaArtifacts
+
+import ClimaDiagnostics
+import ClimaAnalysis
+import ClimaAnalysis.Visualize as viz
+import ClimaUtilities
+
+import ClimaUtilities.TimeVaryingInputs:
+    TimeVaryingInput, LinearInterpolation, PeriodicCalendar
+import ClimaUtilities.SpaceVaryingInputs: SpaceVaryingInput
+import ClimaUtilities.ClimaArtifacts: @clima_artifact
+import ClimaParams as CP
+
+using ClimaLand
+using ClimaLand.Snow
+using ClimaLand.Soil
+using ClimaLand.Canopy
+import ClimaLand
+import ClimaLand.Parameters as LP
+
+using Statistics
+using CairoMakie
+import GeoMakie
+using Dates
+import NCDatasets
+
+const FT = Float64;
+time_interpolation_method = LinearInterpolation(PeriodicCalendar())
+context = ClimaComms.context()
+device = ClimaComms.device()
+device_suffix = device isa ClimaComms.CPUSingleThreaded ? "cpu" : "gpu"
+root_path = "snowy_land_longrun_$(device_suffix)"
+diagnostics_outdir = joinpath(root_path, "global_diagnostics")
+outdir =
+    ClimaUtilities.OutputPathGenerator.generate_output_path(diagnostics_outdir)
+
+function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
+    earth_param_set = LP.LandParameters(FT)
+    radius = FT(6378.1e3)
+    depth = FT(50)
+    domain = ClimaLand.Domains.SphericalShell(;
+        radius = radius,
+        depth = depth,
+        nelements = nelements,
+        npolynomial = 1,
+        dz_tuple = FT.((10.0, 0.05)),
+    )
+    surface_space = domain.space.surface
+    subsurface_space = domain.space.subsurface
+
+    start_date = DateTime(2021)
+    # Forcing data
+    era5_artifact_path =
+        ClimaLand.Artifacts.era5_land_forcing_data2021_folder_path(; context)
+    era5_ncdata_path = joinpath(era5_artifact_path, "era5_2021_0.9x1.25.nc")
+    atmos, radiation = ClimaLand.prescribed_forcing_era5(
+        era5_ncdata_path,
+        surface_space,
+        start_date,
+        earth_param_set,
+        FT;
+        time_interpolation_method = time_interpolation_method,
+        regridder_type = regridder_type,
+    )
+
+    spatially_varying_soil_params =
+        ClimaLand.default_spatially_varying_soil_parameters(
+            subsurface_space,
+            surface_space,
+            FT,
+        )
+    (;
+        ν,
+        ν_ss_om,
+        ν_ss_quartz,
+        ν_ss_gravel,
+        hydrology_cm,
+        K_sat,
+        S_s,
+        θ_r,
+        PAR_albedo,
+        NIR_albedo,
+        f_max,
+    ) = spatially_varying_soil_params
+    soil_params = Soil.EnergyHydrologyParameters(
+        FT;
+        ν,
+        ν_ss_om,
+        ν_ss_quartz,
+        ν_ss_gravel,
+        hydrology_cm,
+        K_sat,
+        S_s,
+        θ_r,
+        PAR_albedo = PAR_albedo,
+        NIR_albedo = NIR_albedo,
+    )
+    f_over = FT(3.28) # 1/m
+    R_sb = FT(1.484e-4 / 1000) # m/s
+    runoff_model = ClimaLand.Soil.Runoff.TOPMODELRunoff{FT}(;
+        f_over = f_over,
+        f_max = f_max,
+        R_sb = R_sb,
+    )
+
+    # Spatially varying canopy parameters from CLM
+    clm_parameters = ClimaLand.clm_canopy_parameters(surface_space)
+    (;
+        Ω,
+        rooting_depth,
+        is_c3,
+        Vcmax25,
+        g1,
+        G_Function,
+        α_PAR_leaf,
+        τ_PAR_leaf,
+        α_NIR_leaf,
+        τ_NIR_leaf,
+    ) = clm_parameters
+
+    # Energy Balance model
+    ac_canopy = FT(2.5e3)
+    # Plant Hydraulics and general plant parameters
+    SAI = FT(0.0) # m2/m2
+    f_root_to_shoot = FT(3.5)
+    RAI = FT(1.0)
+    K_sat_plant = FT(5e-9) # m/s # seems much too small?
+    ψ63 = FT(-4 / 0.0098) # / MPa to m, Holtzman's original parameter value is -4 MPa
+    Weibull_param = FT(4) # unitless, Holtzman's original c param value
+    a = FT(0.05 * 0.0098) # Holtzman's original parameter for the bulk modulus of elasticity
+    conductivity_model =
+        Canopy.PlantHydraulics.Weibull{FT}(K_sat_plant, ψ63, Weibull_param)
+    retention_model = Canopy.PlantHydraulics.LinearRetentionCurve{FT}(a)
+    plant_ν = FT(1.44e-4)
+    plant_S_s = FT(1e-2 * 0.0098) # m3/m3/MPa to m3/m3/m
+    n_stem = 0
+    n_leaf = 1
+    h_stem = FT(0.0)
+    h_leaf = FT(1.0)
+    zmax = FT(0.0)
+    h_canopy = h_stem + h_leaf
+    compartment_midpoints =
+        n_stem > 0 ? [h_stem / 2, h_stem + h_leaf / 2] : [h_leaf / 2]
+    compartment_surfaces =
+        n_stem > 0 ? [zmax, h_stem, h_canopy] : [zmax, h_leaf]
+
+    z0_m = FT(0.13) * h_canopy
+    z0_b = FT(0.1) * z0_m
+
+
+    soilco2_ps = Soil.Biogeochemistry.SoilCO2ModelParameters(FT)
+
+    soil_args = (domain = domain, parameters = soil_params)
+    soil_model_type = Soil.EnergyHydrology{FT}
+
+    # Soil microbes model
+    soilco2_type = Soil.Biogeochemistry.SoilCO2Model{FT}
+
+    # soil microbes args
+    Csom = ClimaLand.PrescribedSoilOrganicCarbon{FT}(TimeVaryingInput((t) -> 5))
+
+    # Set the soil CO2 BC to being atmospheric CO2
+    soilco2_top_bc = Soil.Biogeochemistry.AtmosCO2StateBC()
+    soilco2_bot_bc = Soil.Biogeochemistry.SoilCO2FluxBC((p, t) -> 0.0) # no flux
+    soilco2_sources = (Soil.Biogeochemistry.MicrobeProduction{FT}(),)
+
+    soilco2_boundary_conditions =
+        (; top = soilco2_top_bc, bottom = soilco2_bot_bc)
+
+    soilco2_args = (;
+        boundary_conditions = soilco2_boundary_conditions,
+        sources = soilco2_sources,
+        domain = domain,
+        parameters = soilco2_ps,
+    )
+
+    # Now we set up the canopy model, which we set up by component:
+    # Component Types
+    canopy_component_types = (;
+        autotrophic_respiration = Canopy.AutotrophicRespirationModel{FT},
+        radiative_transfer = Canopy.TwoStreamModel{FT},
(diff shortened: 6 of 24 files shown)


src/integrated/soil_canopy_model.jl
```diff
@@ -285,50 +285,6 @@ function make_update_boundary_fluxes(
     return update_boundary_fluxes!
 end
 
-"""
```
kmdeck, line 288: moved to separate file since this is used by multiple integrated models now, does not "belong" to this one.

src/integrated/soil_snow_model.jl
```diff
@@ -280,16 +280,16 @@ function update_soil_snow_ground_heat_flux!(
 
     H = p.subsfc_scratch
     @. H = ClimaLand.heaviside(
-        soil_domain.fields.z_sfc - soil_domain.fields.z,
         Δz_soil,
+        soil_domain.fields.z_sfc - soil_domain.fields.z,
```
kmdeck, line 284: this was a bug!!

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -33,8 +33,7 @@ end
 Returns the surface height of the `Snow` model.
 """
 function ClimaLand.surface_height(model::SnowModel{FT}, Y, p) where {FT}
-    z_sfc = ClimaCore.Fields.coordinate_field(model.domain.space.surface).z
-    return z_sfc
+    return p.snow.z
```
kmdeck, line 36: I had to change this to get the snow model to run globally, because if it was defined a point space, there was a `z` coordinate, but if it was a spectral element space, there was not. 

Problems will arise in the surface flux computation, however, if the snow depth is higher than the atmosphere lowest level that we use for the forcing (I think currently a few m). But in principle, this shouldnt happen!

test/integrated/soil_snow.jl
```diff
@@ -25,7 +25,7 @@ precip_snow = (t) -> 0
 T_atmos = (t) -> 285
 u_atmos = (t) -> 3
 q_atmos = (t) -> 0.005
-h_atmos = FT(3)
+h_atmos = FT(10)
```
kmdeck, line 28: here was where I had to fix that!

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -244,37 +255,20 @@ end
 
 Y.canopy.energy.T = drivers.TA.values[1 + Int(round(t0 / DATA_DT))] # Get atmos temperature at t0
 
+Y.snow.S .= 0.0
+Y.snow.U .= 0.0
+
 set_initial_cache! = make_set_initial_cache(land)
 set_initial_cache!(p, Y, t0);
 
-# Simulation
```
kmdeck, line 250: no longer using the SavingCallback here (using Diagnostics). This was leftover from a previous PR so I just removed it here!

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -332,31 +338,32 @@ short_names_2D = [
...
+    ]
 
-swc_5, soil_T_5, si_5 = [
+swc, soil_T, si = [
     ClimaLand.Diagnostics.diagnostic_as_vectors(
         d_writer,
         diag_name;
-        layer = 5,
-    )[2][1:721] for diag_name in hourly_diag_name_2D
+        layer = 20,#surface
```
kmdeck, line 354: layer 20 = surface layer for plotting

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -593,3 +587,87 @@ if isfile(
         ),
     )
 end
+
+# Water content in soil and snow
```
kmdeck, line 591: Added in some plots that were accidentally removed at one point, and added SWE to it

src/diagnostics/land_compute_methods.jl
```diff
@@ -61,88 +61,130 @@ end
 @diagnostic_compute "subsurface_water_storage" BucketModel Y.bucket.W
 @diagnostic_compute "surface_water_content" BucketModel Y.bucket.Ws
 
-### SoilCanopyModel ###
+### Union{SoilCanopyModel, LandModel} ###
```
juliasloan25, line 64: Seeing that we have so many instances of `Union{SoilCanopyModel, LandModel}` here, it might be good to update our type hierarchy with a new abstract type - something like `AbstractSoilCanopyModel`, where we would have both `SoilCanopyModel <: AbstractSoilCanopyModel` and  `LandModel <: AbstractSoilCanopyModel`. Maybe even `AbstractIntegratedModel` or `AbstractComplexModel`, but these values are specific to having canopy + soil so we might want to be more specific in the naming.

We don't have to do this in this PR, just something to think about

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -186,14 +187,26 @@ shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
 )
 
 canopy_model_args = (; parameters = shared_params, domain = canopy_domain)
+
+# Snow model
+ρ_snow = FT(300.0)
+α_snow = FT(0.8)
```
juliasloan25, line 193: This is just the default albedo value, so I don't think we need to pass it in.

Also, why not just use the default density?

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -240,6 +255,10 @@ for i in 1:(n_stem + n_leaf)
 end
 
 Y.canopy.energy.T = drivers.TA.values[1 + Int(round(t0 / DATA_DT))] # Get atmos temperature at t0
+
+Y.snow.S .= 0.0
```
juliasloan25, line 259: I'm thinking about how we can clean up the user experience in drivers, and I think it would be nice to have a `set_initial_values` function where we have default values for each prognostic + diagnostic variable, with the option for the user to pass in alternative values that overwrite the defaults. What do you think?

Again, not for this PR, just an idea :)

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+A constructor for the `LandModel`, which takes in the concrete model
+type and required arguments for each component, constructs those models,
+and constructs the `LandModel` from them.
+
+Each component model is constructed with everything it needs to be stepped
+forward in time, including boundary conditions, source terms, and interaction
+terms.
+"""
+function LandModel{FT}(;
+    soilco2_type::Type{MM},
```
juliasloan25, line 71: This is more of a style choice, but I think it would be nice to unify the order of component models across the `LandModel` struct, the arguments to this function, and the model creation inside this function.

e.g. move the `land_args` to the first position in this function, then within the function extract land args first, then set up soil co2 model, then soil model, then canopy, then snow

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+    )
+
+    transpiration = Canopy.PlantHydraulics.DiagnosticTranspiration{FT}()
+    ground_conditions =
+        PrognosticGroundConditions{FT, typeof(soil.parameters.PAR_albedo)}(
+            snow.parameters.α_snow,
+            soil.parameters.PAR_albedo,
+            soil.parameters.NIR_albedo,
+        )
+    if :energy in propertynames(canopy_component_args)
```
juliasloan25, line 139:
```suggestion
    if :energy in propertynames(canopy_component_args)
        energy = canopy_component_types.energy(
                canopy_component_args.energy.parameters,
            )
    else
        energy = PrescribedCanopyTempModel{FT}()
    end
```

Then we can construct the `CanopyModel` the same way for both cases, setting `energy = energy` for the energy model. This will minimize the code in branches and make the code more readable

juliasloan25: we could also do this for runoff when we construct the `ClimaLand.AtmosDrivenFluxBC` if you like this design pattern

src/integrated/land.jl
juliasloan25: the order could be alphabetical if that's easier to keep track of

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+        #Now update snow boundary conditions, which rely on the ground heat flux
+        update_snow_bf!(p, Y, t)
+
+        # Now we have access to the actual applied and initially computed fluxes for snow
+        @. p.excess_water_flux =
+            (p.snow.total_water_flux - p.snow.applied_water_flux)
+        @. p.excess_heat_flux =
+            (p.snow.total_energy_flux - p.snow.applied_energy_flux)
+
+        # Now we can update the soil BC, and use the excess fluxes there in order to conserv
```
juliasloan25, line 339: unfinished sentence?

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+        # Update soil CO2
+        update_soilco2_bf!(p, Y, t)
+
+        # compute net flux with atmosphere, this is useful for monitoring conservation
+        _LH_f0 = FT(LP.LH_f0(earth_param_set))
+        _ρ_liq = FT(LP.ρ_cloud_liq(earth_param_set))
+        ρe_falling_snow = -_LH_f0 * _ρ_liq # per unit vol of liquid water
+        @. p.atmos_energy_flux =
+            (1 - p.snow.snow_cover_fraction) * (
+                p.soil.turbulent_fluxes.lhf + p.soil.turbulent_fluxes.shf -
```
juliasloan25, line 352: Here we have `lhf + shf - R_n`, but in [soil_snow_model.jl](https://github.com/CliMA/ClimaLand.jl/blob/e71238f2ef61af48789e65021fcef8a58582d5eb/src/integrated/soil_snow_model.jl#L205-L214) we have `lhf + shf + R_n`. It seems like these should be the same - is there a typo?

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+                p.soil.R_n
+            ) +
+            p.snow.snow_cover_fraction * (
+                p.snow.turbulent_fluxes.lhf + p.snow.turbulent_fluxes.shf -
+                p.snow.R_n
+            ) +
+            p.drivers.P_snow * ρe_falling_snow +
+            p.canopy.energy.turbulent_fluxes.shf +
+            p.canopy.energy.turbulent_fluxes.lhf -
```
juliasloan25, line 361: same sign comment here

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+    c = LP.light_speed(earth_param_set)
+    h = LP.planck_constant(earth_param_set)
+    N_a = LP.avogadro_constant(earth_param_set)
+    (; λ_γ_PAR, λ_γ_NIR) = canopy_radiation.parameters
+    energy_per_photon_PAR = h * c / λ_γ_PAR
+    energy_per_photon_NIR = h * c / λ_γ_NIR
+    T_canopy =
+        ClimaLand.Canopy.canopy_temperature(canopy.energy, canopy, Y, p, t)
+
+    α_soil_PAR = land.soil.parameters.PAR_albedo
```
juliasloan25, line 417: can we get this from soil.parameters in `soil_canopy_model.jl` too?

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+        energy_per_photon_PAR * N_a * p.canopy.radiative_transfer.par.refl
+
+    # net canopy
+    @. SW_net_canopy =
+        energy_per_photon_NIR * N_a * p.canopy.radiative_transfer.nir.abs +
+        energy_per_photon_PAR * N_a * p.canopy.radiative_transfer.par.abs
+
+
+    # net soil = (1-α)*trans for par and nir
+    @. R_net_soil .=
```
juliasloan25, line 448: It looks like a lot of the radiation update for soil is the same as the update in `src/integrated/soil_canopy_model.jl` - maybe we can pull this out into a shared location

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+        p.canopy.radiative_transfer.nir.trans *
+        (1 - α_snow_NIR) +
+        energy_per_photon_PAR *
+        N_a *
+        p.canopy.radiative_transfer.par.trans *
+        (1 - α_snow_PAR)
+    ϵ_canopy = p.canopy.radiative_transfer.ϵ # this takes into account LAI/SAI
+    @. LW_d_canopy = ((1 - ϵ_canopy) * LW_d + ϵ_canopy * _σ * T_canopy^4) # double checked
+    @. LW_u_soil = ϵ_soil * _σ * T_soil^4 + (1 - ϵ_soil) * LW_d_canopy # double checked
+    @. LW_u_snow = ϵ_snow * _σ * T_snow^4 + (1 - ϵ_snow) * LW_d_canopy # check
```
juliasloan25, line 470: It looks like some of the equations in this section still need to be checked

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+    Y,
+    p,
+    t,
+)
+    return @. (1 - p.snow.snow_cover_fraction) * ground.α_soil_NIR +
+              p.snow.snow_cover_fraction * ground.α_snow
+end
+
+
+function ClimaLand.get_drivers(model::LandModel)
```
juliasloan25, line 650: docstring

src/integrated/land.jl
```diff
@@ -0,0 +1,669 @@
...
+    return (
+        model.canopy.boundary_conditions.atmos,
+        model.canopy.boundary_conditions.radiation,
+        model.soilco2.drivers.soc,
+    )
+end
+
+
+function Canopy.canopy_radiant_energy_fluxes!(
```
juliasloan25, line 659: docstring

src/integrated/soil_canopy_root_interactions.jl
```diff
@@ -0,0 +1,153 @@
...
+    Y::ClimaCore.Fields.FieldVector,
+    p::NamedTuple,
+    t,
+)
+    ClimaCore.Operators.column_integral_definite!(fa, p.root_extraction)
+end
+
+
+
+
```
juliasloan25, line 79:
```suggestion
```

experiments/long_runs/snowy_land.jl
```diff
@@ -0,0 +1,662 @@
+# # Global run of land model
+
+# The code sets up and runs the soil/canopy model for 6 hours on a spherical domain,
```
juliasloan25, line 3:
```suggestion
# The code sets up and runs the full land model, which includes soil, canopy, and snow, for 365 days on a spherical domain,
```

experiments/long_runs/snowy_land.jl
```diff
@@ -0,0 +1,662 @@
...
+        joinpath(
+            soil_params_artifact_path,
+            "vGn_map_gupta_etal2020_1.0x1.0x4.nc",
+        ),
+        "n",
+        subsurface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc,),
+    )
+    x = parent(vg_α)
```
juliasloan25, line 132: This might need to be `x = Array(parent(vg_α))` for GPU compatibility. On GPU, wrapping in `Array()` would force the `CuArray` returned by `parent` to be converted to an `Array`, so that operations on it happen on the CPU instead of GPU. Without this, I think we would get a scalar indexing error in line 133

experiments/long_runs/snowy_land.jl
```diff
@@ -0,0 +1,662 @@
...
+        ρ_snow = ρ_snow,
+        earth_param_set = earth_param_set,
+    )
+    snow_args = (;
+        parameters = snow_parameters,
+        domain = ClimaLand.obtain_surface_domain(domain),
+    )
+    snow_model_type = Snow.SnowModel
+
+    # Integrated plant hydraulics and soil model
```
juliasloan25, line 493: comment out of date?

experiments/long_runs/snowy_land.jl
```diff
@@ -0,0 +1,662 @@
...
+        canopy_component_args = canopy_component_args,
+        canopy_model_args = canopy_model_args,
+        snow_args = snow_args,
+        snow_model_type = snow_model_type,
+    )
+
+    Y, p, cds = initialize(land)
+
+    init_soil(ν, θ_r) = θ_r + (ν - θ_r) / 2
+    Y.soil.ϑ_l .= init_soil.(ν, θ_r)
```
juliasloan25, line 516:
```suggestion
    @. Y.soil.ϑ_l = θ_r + (ν - θ_r) / 2
```

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -31,6 +31,16 @@ steps:
   - group: "Global Land Models"
     steps:
 
+      - label: "Snowy Land"
```
juliasloan25, line 34: we could add the :mountain_snow: emoji for this run :)

juliasloan25:
It looks like the Niwot Fluxnet experiment is unstable :( - it failed with
```
1m[1mERROR: [22m[39mLoadError: DomainError with -7.839898692779407e9:
Exponentiation yielding a complex result requires a complex argument.
Replace x^y with (x+0im)^y, Complex(x)^y, or similar.
...
```

experiments/integrated/fluxnet/run_fluxnet.jl
AlexisRenchon: I think that's a good idea, maybe for the new `src/simulation` folder (which will eventually replace `ClimaLandSimulations`) - we should have a design meeting at some point, I think maybe in January. 
I think right now the priorities are snow model and calibrations, as well as spatially varying params and stable simulations... once all this is done, we should think of the simulations redesign? (happy to discuss)

src/diagnostics/default_diagnostics.jl
```diff
@@ -258,3 +259,124 @@ function default_diagnostics(
...
+        snowyland_diagnostics = [
+            "swa",
+            "sif",
+            "ra",
+            "gs",
+            "trans",
+            "clhf",
+            "cshf",
+            # "lwp", # last(p.canopy.hydraulics.ψ) errors
+            # "fa", # return a Tuple
```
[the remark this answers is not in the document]
AlexisRenchon: good call - I am not sure what the solution is, but we should indeed have an issue (do we need those diagnostics? if not, we can remove these comments)

src/diagnostics/define_diagnostics.jl
```diff
@@ -843,4 +843,24 @@ function define_diagnostics!(land_model)
...
+        short_name = "swe",
+        long_name = "Snow water equivalent",
+        standard_name = "snow_water_equivalent",
+        units = "m",
+        comments = "The height of liquid water if all snow melted",
+        compute! = (out, Y, p, t) ->
+            compute_snow_water_equivalent!(out, Y, p, t, land_model),
+    )
+
+    add_diagnostic_variable!(
```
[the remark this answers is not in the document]
AlexisRenchon: I think it is okay to not say anything for land diagnostics only (but comment if they are also in bucket)?

src/diagnostics/land_compute_methods.jl
AlexisRenchon: Why not `AbstractLandModel` for the parent type? (we can discuss for another PR)

src/standalone/Snow/snow_parameterizations.jl
AlexisRenchon: The snow level in e.g., Greenland is huge, no? (up to 3km), or is this considered as ice?
(compacted snow becoming ice...) - maybe not a concern for this PR but to keep in mind? What do you think?

AlexisRenchon (approved):
I am very excited for this PR! Amazing work, thank you Kat!

src/standalone/Snow/snow_parameterizations.jl
kmdeck: Great question! I will leave note in the docstring.

I also added an error message so that if we encounter this, we wont just get NaNs, but the simulation will quit on us.

experiments/integrated/fluxnet/US-Ha1/US-Ha1_simulation.jl
```diff
@@ -17,7 +17,7 @@ h_stem = FT(14) # m
 # TIME STEPPING:
 
 # Starting time:
-t0 = Float64(120 * 3600 * 24)# start mid year to avoid snow
+t0 = Float64(0 * 3600 * 24)
```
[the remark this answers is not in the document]
kmdeck: nice! fixed here and in the other files

src/diagnostics/default_diagnostics.jl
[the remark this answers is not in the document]
kmdeck: I will open an issue to add leaf water potential and remove these comments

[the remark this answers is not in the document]
kmdeck: https://github.com/CliMA/ClimaLand.jl/issues/892

src/diagnostics/define_diagnostics.jl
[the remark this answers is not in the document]
kmdeck: The bucket only has one possible defined surface temperature, so that is what the bucket one returns. For effective surface temperature, I used this definition: https://en.wikipedia.org/wiki/Effective_temperature but I think I implemented it incorrectly.

I am going to remove this new diagnostic from this PR.

experiments/integrated/fluxnet/run_fluxnet.jl
kmdeck: fair enough! I will use the defaults.

src/standalone/Snow/snow_parameterizations.jl
kmdeck: I asked Akshay, and he said that the atmos model uses the land ice level as topography.

src/diagnostics/land_compute_methods.jl
kmdeck: ^^ I like this direction of thinking! Right now "AbstractLandModel <: AbstractModel" is taken (it is the abstract type of SoilCanopyModel, LandModel, LandHydrologyModel, etc). Like Julia wrote, some of these methods only make sense if we have a canopy, so we cant use it as is. Ill open an issue!

kmdeck: Issue #895

experiments/integrated/fluxnet/run_fluxnet.jl
kmdeck: I agree! This might tie into @Sbozzolo's recent addition of setting IC from a checkpoint file. It might be that global runs start from a setup similar to what we have by default, whereas fluxnet sites set IC based on the data. but either way we need to run for a while to spinup

src/integrated/land.jl
kmdeck: this is much cleaner! Ill adopt it.

kmdeck: also for runoff

experiments/integrated/fluxnet/US-NR1/US-NR1_simulation.jl
```diff
@@ -17,7 +17,7 @@ h_stem = FT(7.5) # m
 # TIME STEPPING:
 
 # Starting time:
-t0 = Float64(120 * 3600 * 24)# start mid year to avoid snow
+t0 = Float64(0 * 3600 * 24)
 
 # Time step size:
-dt = Float64(450)
+dt = Float64(225)
```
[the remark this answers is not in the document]
kmdeck: IM not sure, but it is failing with a domain error. Ill keep debugging it once I address the other review comments!

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -186,14 +187,26 @@ shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
...
+snow_model_type = Snow.SnowModel
 # Integrated plant hydraulics and soil model
 land_input = (
     atmos = atmos,
     radiation = radiation,
     soil_organic_carbon = Csom,
     runoff = ClimaLand.Soil.Runoff.SurfaceRunoff(),
 )
-land = SoilCanopyModel{FT}(;
+land = LandModel{FT}(;
```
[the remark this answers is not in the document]
kmdeck: I think we should just add it to the LandModel, but only add it when we are running global simulations?

.buildkite/longruns_gpu/pipeline.yml
kmdeck: nice! will do :)

experiments/long_runs/snowy_land.jl
kmdeck: made this change

experiments/long_runs/snowy_land.jl
kmdeck: removed

experiments/long_runs/snowy_land.jl
kmdeck: I dont think we actually need the parent...let me check

experiments/integrated/fluxnet/fluxnet_simulation.jl
```diff
@@ -2,8 +2,9 @@
 fluxtower sites. These work in tandem with the site-specific timing parameters
 found in the {site-ID}_simulation.jl files in each site directory."""
 
-N_spinup_days = 30
-N_days = N_spinup_days + 30
+N_spinup_days = 15
+N_days = N_spinup_days + 340
```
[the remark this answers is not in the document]
AlexisRenchon: so the goal is to reach some steady state of soil moisture, particularly at deeper layer? I think we could discuss that tomorrow (wednesday) a bit

experiments/integrated/fluxnet/US-NR1/US-NR1_simulation.jl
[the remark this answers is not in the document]
AlexisRenchon: not important for this PR maybe?

src/integrated/land.jl
kmdeck: I think so. I made this change.

src/integrated/land.jl
kmdeck: it is a sign inconsistency 😭 for some reason, I had defined R_n positive towards soil in the soil/canopy case. It'll be fixed in https://github.com/CliMA/ClimaLand.jl/issues/777.

Here, I will make sure the convention is the same as soil_snow and soil by itself: R_n is a flux which is negative if towards soil. then total flux = lhf+shf+R_n

src/integrated/land.jl
kmdeck: The signs are changed now so that the only outlier is in soil_canopy_model. This can get fixed in a future PR (will require a change in the conservation runs)

src/integrated/land.jl
kmdeck: I did move one part out into a shared function, and started a file called land_radiation.jl which can contain these shared functions.

I am finding it hard to unify the rest of `lsm_radiant_energy_fluxes!` though, until we fix the sign inconsistency, and I think we should also split it into SW and LW functions. But this feels like a bigger change. I opened this issue: https://github.com/CliMA/ClimaLand.jl/issues/898

src/integrated/land.jl
juliasloan25: Sounds good, thank you!

src/integrated/land.jl
juliasloan25: I don't think this has been changed

experiments/long_runs/snowy_land.jl
juliasloan25: not using parent would be ideal :)

src/shared_utilities/drivers.jl
```diff
@@ -355,7 +355,9 @@ function turbulent_fluxes_at_a_point(
         SVector{2, FT}(u, 0),
         ts_in,
     )
-
+    h - d_sfc - h_sfc < 0 ?
+    @error("Surface height is larger than atmos height in surface fluxes") :
+    nothing
```
juliasloan25, line 360:
```suggestion
    h - d_sfc - h_sfc < 0 && @error("Surface height is larger than atmos height in surface fluxes")
```

juliasloan25:
I think the GPU bucket failure might be the same issue we saw in https://github.com/CliMA/ClimaLand.jl/pull/739. I added a reproducer in https://github.com/CliMA/ClimaCore.jl/pull/2062, but still in the process of debugging

The only change I see in src that might affect the bucket is [here](https://github.com/CliMA/ClimaLand.jl/pull/834/files#diff-a97492469e0d60da6ffa468219144b6c7187de8f93bdb8a8b7e7c5bfb3654379R358). I suggested a change - maybe that will help?

src/integrated/land.jl
kmdeck: ah, I only changed it in soil_canopy_model. thanks for catching this!

src/integrated/land.jl
```diff
@@ -0,0 +1,644 @@
...
+            atmos,
+            radiation,
+            prognostic_land_components,
+        ),
+        snow_args...,
+    )
+    # soil
+    sources = (RootExtraction{FT}(), Soil.PhaseChange{FT}())
+
+    if :runoff ∈ propertynames(land_args)
```
kmdeck, line 104: I changed this here though

src/integrated/land.jl
kmdeck: oh wait no, it is changed here too. right?

juliasloan25: oh yes it is! Sorry about that

experiments/long_runs/snowy_land.jl
kmdeck: We do.... I think because if x is a Field, and we do x .> 0, it seems like it is trying to convert a field of booleans into a field of floats:

```
julia> typeof(x)
ClimaCore.Fields.Field{ClimaCore.DataLayouts.VIJFH{Float64, 15, 2, CUDA.CuArray{Float64, 5, CUDA.DeviceMemory}}, ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace{ClimaCore.Grids.ExtrudedFiniteDifferenceGrid{ClimaCore.Grids.SpectralElementGrid2D{ClimaCore.Topologies.Topology2D{ClimaComms.SingletonCommsContext{ClimaComms.CUDADevice}, ClimaCore.Meshes.EquiangularCubedSphere{ClimaCore.Domains.SphereDomain{Float64}, ClimaCore.Meshes.NormalizedBilinearMap}, CartesianIndices{3, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}}, LinearIndices{3, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}}, CUDA.CuArray{Tuple{Int64, Int64, Int64, Int64, Bool}, 1, CUDA.DeviceMemory}, Vector{Tuple{Int64, Int64, Int64, Int64, Bool}}, CUDA.CuArray{Tuple{Int64, Int64}, 1, CUDA.DeviceMemory}, CUDA.CuArray{Int64, 1, CUDA.DeviceMemory}, CUDA.CuArray{Tuple{Bool, Int64, Int64}, 1, CUDA.DeviceMemory}, CUDA.CuArray{Int64, 1, CUDA.DeviceMemory}, CUDA.CuArray{Int64, 1, CUDA.DeviceMemory}, @NamedTuple{}, CUDA.CuArray{Tuple{Int64, Int64}, 1, CUDA.DeviceMemory}}, ClimaCore.Quadratures.GLL{2}, ClimaCore.Geometry.SphericalGlobalGeometry{Float64}, ClimaCore.DataLayouts.IJFH{ClimaCore.Geometry.LocalGeometry{(1, 2), ClimaCore.Geometry.LatLongPoint{Float64}, Float64, StaticArraysCore.SMatrix{2, 2, Float64, 4}}, 2, CUDA.CuArray{Float64, 4, CUDA.DeviceMemory}}, ClimaCore.DataLayouts.IJFH{Float64, 2, CUDA.CuArray{Float64, 4, CUDA.DeviceMemory}}, ClimaCore.DataLayouts.IFH{ClimaCore.Geometry.SurfaceGeometry{Float64, ClimaCore.Geometry.UVVector{Float64}}, 2, CUDA.CuArray{Float64, 3, CUDA.DeviceMemory}}, @NamedTuple{}}, FiniteDifferenceGrid{IntervalTopology{ClimaComms.SingletonCommsContext{ClimaComms.CUDADevice}, ClimaCore.Meshes.IntervalMesh{ClimaCore.Domains.IntervalDomain{ClimaCore.Geometry.ZPoint{Float64}, Tuple{Symbol, Symbol}}, Vector{ClimaCore.Geometry.ZPoint{Float64}}}, @NamedTuple{bottom::Int64, top::Int64}}, ClimaCore.Geometry.CartesianGlobalGeometry, ClimaCore.DataLayouts.VF{ClimaCore.Geometry.LocalGeometry{(3,), ClimaCore.Geometry.ZPoint{Float64}, Float64, StaticArraysCore.SMatrix{1, 1, Float64, 1}}, 15, CUDA.CuArray{Float64, 2, CUDA.DeviceMemory}}, ClimaCore.DataLayouts.VF{ClimaCore.Geometry.LocalGeometry{(3,), ClimaCore.Geometry.ZPoint{Float64}, Float64, StaticArraysCore.SMatrix{1, 1, Float64, 1}}, 16, CUDA.CuArray{Float64, 2, CUDA.DeviceMemory}}}, ClimaCore.Grids.Flat, ClimaCore.Geometry.ShallowSphericalGlobalGeometry{Float64}, ClimaCore.DataLayouts.VIJFH{ClimaCore.Geometry.LocalGeometry{(1, 2, 3), ClimaCore.Geometry.LatLongZPoint{Float64}, Float64, StaticArraysCore.SMatrix{3, 3, Float64, 9}}, 15, 2, CUDA.CuArray{Float64, 5, CUDA.DeviceMemory}}, ClimaCore.DataLayouts.VIJFH{ClimaCore.Geometry.LocalGeometry{(1, 2, 3), ClimaCore.Geometry.LatLongZPoint{Float64}, Float64, StaticArraysCore.SMatrix{3, 3, Float64, 9}}, 16, 2, CUDA.CuArray{Float64, 5, CUDA.DeviceMemory}}}, ClimaCore.Grids.CellCenter}}

julia> μ = mean(x[x .> 0]);
ERROR: Struct type Bool cannot be represented using base type Float64
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] macro expansion
    @ ~/.julia/packages/ClimaCore/SvZXu/src/DataLayouts/struct.jl:34 [inlined]
  [3] check_basetype(::Type{Float64}, ::Type{Bool})
    @ ClimaCore.DataLayouts ~/.julia/packages/ClimaCore/SvZXu/src/DataLayouts/struct.jl:33
  [4] (ClimaCore.DataLayouts.VIJFH{Bool, 15, 2})(array::CUDA.CuArray{Float64, 5, CUDA.DeviceMemory})
    @ ClimaCore.DataLayouts ~/.julia/packages/ClimaCore/SvZXu/src/DataLayouts/DataLayouts.jl:918
  [5] similar(bc::Base.Broadcast.Broadcasted{ClimaCore.DataLayouts.VIJFHStyle{…}, Nothing, typeof(>), Tuple{…}}, ::Type{Bool}, ::Val{15})
    @ ClimaCore.DataLayouts ~/.julia/packages/ClimaCore/SvZXu/src/DataLayouts/broadcast.jl:377
  [6] similar(bc::Base.Broadcast.Broadcasted{ClimaCore.DataLayouts.VIJFHStyle{…}, Nothing, typeof(>), Tuple{…}}, ::Type{Bool})
    @ ClimaCore.DataLayouts ~/.julia/packages/ClimaCore/SvZXu/src/DataLayouts/broadcast.jl:364
  [7] similar(bc::Base.Broadcast.Broadcasted{…}, ::Type{…})
    @ ClimaCore.Fields ~/.julia/packages/ClimaCore/SvZXu/src/Fields/broadcast.jl:142
  [8] copy(bc::Base.Broadcast.Broadcasted{ClimaCore.Fields.FieldStyle{…}, ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace{…}, typeof(>), Tuple{…}})
    @ ClimaCore.Fields ~/.julia/packages/ClimaCore/SvZXu/src/Fields/broadcast.jl:95
  [9] materialize(bc::Base.Broadcast.Broadcasted{ClimaCore.Fields.FieldStyle{…}, Nothing, typeof(>), Tuple{…}})
    @ Base.Broadcast ./broadcast.jl:867
 [10] top-level scope
    @ REPL[54]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

but, that said, these scripts are running on GPU, so for some reason the parent call is OK:
```
julia> x = parent(vg_α);

julia> μ = mean(x[x .> 0]);

julia> typeof(x)
CUDA.CuArray{Float64, 5, CUDA.DeviceMemory}
```


so I will leave this as is

PR #835, opened 2024-10-09
Calibrate global bucket LHF and SHF

AlexisRenchon:
Calibrate global bucket LHF and SHF

### DIFF
--- a/experiments/long_runs/calibrate_bucket_function.jl
+++ b/experiments/long_runs/calibrate_bucket_function.jl
@@ -187,12 +187,12 @@ function setup_prob(t0, tf, Δt,params, outdir; nelements = (101, 7))
         PrescribedRadiativeFluxes(FT, SW_d, LW_d, start_date; θs = zenith_angle)
 
     # Set up parameters
+    (; κ_soil, ρc_soil, f_bucket, W_f, p, z_0m) = params
     z_0b = z_0m
     τc = FT(Δt)
     α_snow = FT(0.8)
-    (; κ_soil, ρc_soil, f_bucket, W_f, p, z_0m) = params
     albedo = PrescribedBaregroundAlbedo{FT}(α_snow, surface_space)
-    bucket_parameters = BucketModelParameters(FT; albedo, z_0m, z_0b, τc, f_bucket, σS_c, p, W_f, κ_soil, ρc_soil)
+    bucket_parameters = BucketModelParameters(FT; albedo, z_0m, z_0b, τc, f_bucket, p, W_f, κ_soil, ρc_soil)
     bucket = BucketModel(
         parameters = bucket_parameters,
         domain = domain,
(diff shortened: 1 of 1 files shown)


src/Artifacts.jl
```diff
@@ -14,6 +14,18 @@ function era5_land_forcing_data2021_folder_path(; context = nothing)
...
+"""
+    era5_surface_data2008_path(; context)
+
+Return the path to the folder that contains the ERA5 monthly surface data
+
+NOTE: We should be able to use @clima_artifact here, but the folders on central
+seem to be named differently than the artifact name
+"""
+function era5_surface_data2008_path(; context = nothing)
+    return "/groups/esm/ClimaArtifacts/artifacts/era5_surface_fluxes_2008"
```
Espeer5, line 26: We should be able to use ClimaArtifacts here, but the folder on central has the wrong name and doesn't match the ClimaArtifact. I didn't want to mess with that in case there are other downstreams of this data that shouldn't be interfered with.

experiments/calibration/global_bucket/calibrate_bucket_function.jl
```diff
@@ -0,0 +1,432 @@
...
+    # end
+
+    close = (x, y) -> abs(x - y) < 5e-1
+    for (lon, lat) in training_locations
+        i += 1
+        # Fetch slices of lhf and shf era5 data from the era5 dataset
+        lat_ind, lon_ind = findall((x) -> close(x, lat), era5_ds["latitude"][:])[1],
+                           findall((x) -> close(x, lon + 180), era5_ds["longitude"][:])[1]
+        lhf_loc = vec(era5_ds["mslhf"][lon_ind, lat_ind, :][:, end, :])
+        shf_loc = vec(era5_ds["msshf"][lon_ind, lat_ind, :][:, end, :])
```
Espeer5, line 379: Using this version of the loop I was able to successfully run the job_script on central/GPU with era5 lhf/shf data.

Espeer5: (Using 5 randomly sample (lon, lat))

Unfortunately, we don't output the lat/lon that we use anywhere that I can tell, so I'm, not actually sure which locations it was running on.

AlexisRenchon: @Espeer5 

I believe when running the job you were still calibrating the perfect model which was the target?

I now changed the target to ERA5, but I get the error you mention:

```
era5_surface_data2008_path not defined in ClimaLand.Artifacts
```

src/Artifacts.jl
AlexisRenchon: [arenchon@login3 ~]$ cd /groups/esm/ClimaArtifacts/artifacts/era5_surface_fluxes_2008
[arenchon@login3 era5_surface_fluxes_2008]$ ls
era5_monthly_surface_fluxes_200801-200812.nc

what is the problem?
the filename should be the same as the folder name? (I have no idea but I do get an error)

Espeer5: To use ClimaArtifacts the name of the directory should be the artifact name - the artifact name is era5_monthly_averages_2008 so if you copy our CliMA Artifacts output to the toml - 
```
[era5_monthly_averages_2008]
git-tree-sha1 = "c34bbd18a9fd59fd0cd48d1737705255805ab233"
```
 It will look for a directory named era5_monthly_averages_2008 - the folder on the cluster is named era5_surface_fluxes_2008 so it wouldn't find it.

experiments/calibration/global_bucket/calibrate_bucket_function.jl
Espeer5: Hmmm this seems strange to me - this would imply it's not able to find the function definition within the Artifacts module? But the diff for this PR shows it should be there?

AlexisRenchon: I rebased and changed the Manifests, maybe that's why?

src/Artifacts.jl
AlexisRenchon: https://github.com/CliMA/ClimaLand.jl/blob/kd/ar/calibrate_global_bucket/experiments/calibration/global_bucket/calibrate_bucket_function.jl#L110

@kmdeck it seems we don't have `era5_land_forcing_data2021_folder_path` in ClimaLand.Artifacts anymore? Should I change this to something else?

Espeer5: I think #920 removed the 2021 data path function and replaced it with 2008 data

PR #836, opened 2024-10-09
nan fix

kmdeck:
## Purpose 
In areas where there is no canopy, albedo = transmissivity = 0. Add a fix so that we dont divide by zero in this case.

## To-do
[X] Test long run - should have NaNs in canopy temperature where this is no canopy, but the other variables should have no NaNs (compared with previous last long run, 2 days ago: https://buildkite.com/clima/climaland-long-runs/builds/1461)

## Content
Clip something to eps(FT) instead of zero

----

### DIFF
--- a/src/diagnostics/land_compute_methods.jl
+++ b/src/diagnostics/land_compute_methods.jl
@@ -204,8 +204,24 @@ function compute_total_respiration!(
 end
 
 # variables stored in Y (prognostic or state variables)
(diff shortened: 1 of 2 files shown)


imreddyTeja (approved):
Looks good! I'll be more careful in checking my PRs' effects on the long runs from now on.

kmdeck:
> Looks good! I'll be more careful in checking my PRs' effects on the long runs from now on.

no worries!

PR #846, opened 2024-10-10
allocations in canopy surface fluxes I

kmdeck:
## Purpose 
Allocations in canopy surface fluxes. Previously, we returned a NamedTuple from a function, with SHF, LHF, etc. This function was broadcasted, so we got a field of NamedTuples. We then unpacked the named tuple to set cache (scalar fields). like:
```
canopy_sf = turbulent_fluxes.(....)
p.canopy.shf = canopy_sf.shf
p.canopy.lhf = canopy_sf.lhf
```
This allocates canopy_shf (field of NamedTuples) each time it is called.

Now we do:
```
@. p.canopy.sf =turbulent_fluxes(...)
shf = p.canopy.sf.shf
lhf = p.canopy.sf.lhf
```
And do not allocate.

Lots of files change since we make use these quantities in experiments, docs, diagnostics.

## Content
We also stored the transpiration in 3 places. Now we store it only in 2: p.canopy.energy.turbulent_fluxes.transpiration, and note that this is also stored in the BC field of the plant hydraulics cache.

----


juliasloan25 (approved):
Thanks Kat!

PR #851, opened 2024-10-14
Compute surface albedo correctly

### LINKED ISSUE #845: Land surface albedo
We currently define land surface albedo to be SW_u/SW_d. At night, SW_d is zero, so albedo is not be defined.  

Since SW_u is proportional to SW_d,  our  solution is SW_u/max(SW_d, eps(FT)). This would give zero albedo at night and prevent dividing by zero, but it is still not physical.

We can instead save the fraction of light absorbed by the land surface and fraction reflected, rather than absolute amount absorbed and reflected. Then the fraction of light reflect = albedo, and we avoid dividing by SW_d. However, this is not the correct value according to Renato.

Changes in the linked PR clean up how we compute absorbed radiation (no longer in photons per m^2 per s, but in W/m^2) and also stores fractional abs/refl/trans, but keeps the albedo computation the same. This would let us change the albedo definitonif we want, in the future.

FOr diagnostics, we must compute albedo using SW_u and SW_d averages, and not the average of the albedo, because at night the albedo is zero and monthly averages would be skewed to small values. But SW_d and SW_u are affected in the exact same way by night values, so they cancel out. That is, the ratio of monthly means = ratio of monthly means only accounting for daytime values

cc @braghiere

kmdeck:
## Purpose 
Close #845 

Because this involves name changing, a *lot* of files changed in experiments, test.

For review, I think it is enough to focus on the changes to files in src/. @braghiere 
Tagging @Espeer5 since you implemented the two stream scheme and this impacts those functions - only review as your time permits :)

## Content
Previously, we computed the absorbed, reflected, and transmitted fluxes in moles of photons per m^2/s in both NIR and PAR. This required:
- converting PAR and NIR downwelling energy flux -> moles of photon fluxes, which is then used to get moles photon/m^2/s absorbed, transmitted, reflected.  We do this because PAR is needed in moles of photon for photosynthesis and SIF computations. But, it required then converting absorbed molar flux etc back to get total absorbed energy flux, transmitted flux, etc
- NIR is not needed in molar flux, but we still computed it this way since we did it for PAR

Content:
- compute fraction absorbed, reflected, transmitted
- when a molar flux of absorbed PAR is needed, convert to it, but otherwise work in energy fluxes
- remove lambda_gamma_NIR (typical energy per photon in the NIR) from parameters and all scripts since we do not need to convert the NIR energy flux into a moles of photon flux.
- renamed inc_* -> *_d (incoming -> downwelling by our convention)
- Changed names: `compute_absorbances` -> `compute_fractional_absorbances`, `plant_absorbed_pfd` -> `canopy_sw_rt` (shortwave radiative transfer)

I compared multiple buildkite plots (ClimaLandSims, fluxnet sites, canopy implicit timestepping, canopy without lai/sai, and all plots were identical between this and the latest PR that merged into main).

----


src/integrated/soil_canopy_model.jl
```diff
@@ -428,8 +421,11 @@ function lsm_radiant_energy_fluxes!(
 
     @. LW_u = (1 - ϵ_canopy) * LW_u_soil + ϵ_canopy * _σ * T_canopy^4 # double checked
 
-    # Effective (radiative) land properties 
-    @. p.α_sfc = SW_u / max(SW_d, eps(FT)) # TODO: replace with fraction reflected as compute by canopy
+    # Effective (radiative) land properties
+    # α = 1/(λ_max - λ_min) * ∫_λ_min ^λ_max α(λ) dλ
+    # We assume that the PAR and NIR bands are each exactly 50% of the incoming SW radiation
+    # Therefore the integral reduces to:
+    @. p.α_sfc = (f_abs_par + f_abs_nir) / 2
```
braghiere, line 428: This cannot be calculated like this. First, I think what you tried to do is: 
@. p.α_sfc = (f_refl_par + f_refl_nir) / 2

but this is not right. This is better: 
@. p.α_sfc = SW_u / max(SW_d, eps(FT))

braghiere (changes requested):
We need to change how we calculate SW albedo.

src/integrated/soil_canopy_model.jl
kmdeck: youre right! I meant f_refl_par, etc.

For the second comment - since SW_u is proportional to SW_d, if SW_d = 0, SW_u = 0, and albedo would not be defined (as you pointed out before). Because of that, in our current main branch, we have what you suggested. 

I thought this was not ok because then we would have albedo =0 at night, which would  mess up our monthly/daily averages that we save as output, and which also isnt exactly physical, since albedo should be a well defined physical property, and not depend on insolation. For example, the albedo of soil is always defined, even if light is not shining on it.

braghiere: So, albedo at night doesn't make sense anyway. So, it definitely shouldn't be included in the calculations of monthly, daily values. A quick workaround would be to set it to NaN when SW_d = 0. But if we need a number for the calculations, we can set it to zero when SW_d is zero.
Because while we do have this general assumption that PAR and NIR are roughy half of the energy, the fraction doesn't hold like this. So, it is best to keep track of SW_u and SW_d
Also, soil albedo and leaf albedo are intrinsic properties of the material, but canopy albedo is different, because it depends on the incident angle of radiation, the proportion of direct and diffuse, canopy structure, LAI, etc. So that is why we call leaf/soil albedo, single scattering albedo. 

That should be fine for the atmos coupling, because if there's no incident radiation SW 0 times 0 = 0. But for the diagnostic (daily, monthly averages), we will need to add NaNs instead. It is the same issue with the canopy temperature before. Otherwise we will have unrealistic low values of canopy albedo, driven by the zeros at night.

Espeer5 (approved):
Changes to the TwoStream scheme look good and make sense to me!

src/integrated/soil_canopy_model.jl
kmdeck: This has been revert to what Renato suggested

PR #853, opened 2024-10-15
Add checkpoint callbacks

### LINKED ISSUE #850: Add restarting capabilities
Currently, ClimaLand has no capabilities to produce checkpoints and restart a simulation from a saved data. Such capabilities are important for ClimaLand running as a standalone package and as part of a coupled simulation. Among the things we need to support this is:
- Integration with `ClimaCore.InputOutput`
- Capability to read in a HDF5 file as state
- Tests to verify that we can reproduce simulations with restarts
- We need to be careful about callbacks and drivers, which have some internal information about when they last updated/called

Sbozzolo:
This commit adds support for checkpoint and restarting a ClimaLand simulation. The functionality is tested in a bucket experiment, where I verify that saving a simulation to disk and restarting it leads to the same state as running the simulation in one go.

In the process, I had to bump some packages (because I am using the latest version of ClimaUtilities). I could not get ClimaLand to stay compatible with 1.9, so I bumped the minimum version to 1.10.

We should try to keep 1.10 as minimum version because it is LTS. 

Closes #850


docs/src/restarts.md
```diff
@@ -0,0 +1,101 @@
...
+it to the solve function along with your other callbacks.
+
+Example:
+
+```julia
+
+# ... your ClimaLand simulation setup ...
+
+# Create a CheckpointCallback to save checkpoints every 6 hours
+checkpoint_cb = CheckpointCallback(Dates.Hour(6), output_dir, start_date, t_start; model)
```
kmdeck, line 39: Do we need to give any guidance re: the timestep dividing the checkpoint interval with no remainder?

Sbozzolo: Yes, I will add those checks in my next PR

Sbozzolo: I added it here.

docs/src/restarts.md
```diff
@@ -0,0 +1,132 @@
...
+    checkpoint and diagnostic frequencies.
+
+
+Similarly, `ClimaLand` does not support updating drivers across restarts and
+drivers might be incorrect upon restart, unless the frequencies of checkpointing
+and updating the drivers are compatible.
+
+!!! note
+
+    The reason why `ClimaLand` does not support these features (at the momenet), is that
```
kmdeck, line 129: moment

docs/src/restarts.md
```diff
@@ -0,0 +1,132 @@
...
+day 75, and not day 60. In other words, the counter starts from 0 with every
+restart.
+
+!!! note
+
+    If you care about accurate accumulated diagnostics, make sure to line up your
+    checkpoint and diagnostic frequencies.
+
+
+Similarly, `ClimaLand` does not support updating drivers across restarts and
```
kmdeck, line 123: possible reword? "Similarly, `ClimaLand` does not guarantee drivers values will be set consistently across restarts. That is, the values of the forcing variables from the last step of the original run will not coincide with the values of the forcing variables upon the initial step of the restart run, unless the frequencies of checkpointing
and updating the drivers are compatible"

src/shared_utilities/models.jl
```diff
@@ -447,6 +447,26 @@ function initialize(model::AbstractModel{FT}) where {FT}
...
 
+
+"""
+    initialize_from_checkpoint(restart_file; model::AbstractModel)
+
+Creates the prognostic and auxiliary states structures, but with unset
+values; constructs and returns the coordinates for the `model` domain.
+We may need to consider this default more as we add diverse components and
+`Simulations`.
+
```
kmdeck, line 458: Maybe we could add to the docstring why model is optional. 

and, what happens if the user saves the checkpoint with the model hash, but does not pass the mode in here? (maybe youve handled this already, I havent gotten there yet! if so, ignore)

src/shared_utilities/models.jl
```diff
@@ -447,6 +447,26 @@ function initialize(model::AbstractModel{FT}) where {FT}
...
+TODO: Combine this function with initialize. We don't really need two.
+"""
+function initialize_from_checkpoint(restart_file; model)
```
kmdeck, line 461: Does this make sense instead?
`initialize(model) = does what it does now`
`initialize(model, checkpoint) = reads from checkpoint, if model hash saved, also checks that

src/shared_utilities/checkpoints.jl
```diff
@@ -0,0 +1,128 @@
...
+ .bucket
+        .T
+        .W
+        .Ws
+```
+`_context_from_Y` tries obtaining the context from the first Field in the
+hierarchy.
+"""
+function _context_from_Y(Y)
+    a_model_type = getproperty(Y, first(propertynames(Y)))
```
kmdeck, line 44: ok! This works because we always do "Y.model.var"

src/shared_utilities/checkpoints.jl
```diff
@@ -0,0 +1,128 @@
...
+  to check for consistency.
+- `comms_ctx` (Optional): The ClimaComms context. This is used for distributed I/O
+  operations. Defaults to the context extracted from the state vector `Y` or the `model`.
+"""
+function save_checkpoint(
+    Y,
+    t,
+    output_dir;
+    model = nothing,
+    comms_ctx = isnothing(model) ? _context_from_Y(Y) :
```
kmdeck, line 73: can we always use context_from_Y?

src/shared_utilities/models.jl
kmdeck: ok - nothing is done in that case. that's fine!

src/shared_utilities/utils.jl
```diff
@@ -265,6 +267,75 @@ function DriverUpdateCallback(updateat::Vector{FT}, updatefunc) where {FT}
...
+    )
+
+    if !isnothing(dt)
+        dt_period = Dates.Millisecond(1000dt)
+        if !isivisible(checkpoint_frequency_period / dt_period)
+            @warn "Checkpoint frequency ($(checkpoint_frequency_period)) is not an integer multiple of dt $(dt_period)"
+        end
+    end
+
+    cond = let schedule = schedule
```
kmdeck, line 328: Where is the schedule function defined? Im confused because on line 315, it is a type. I guess all types are functions, but what is the default behavior?

src/shared_utilities/utils.jl
```diff
@@ -265,6 +267,75 @@ function DriverUpdateCallback(updateat::Vector{FT}, updatefunc) where {FT}
...
+    schedule = EveryCalendarDtSchedule(
+        checkpoint_frequency_period;
+        reference_date = start_date,
+        date_last = start_date + Dates.Millisecond(1000t_start),
+    )
+
+    if !isnothing(dt)
+        dt_period = Dates.Millisecond(1000dt)
+        if !isivisible(checkpoint_frequency_period / dt_period)
+            @warn "Checkpoint frequency ($(checkpoint_frequency_period)) is not an integer multiple of dt $(dt_period)"
```
kmdeck, line 324: What happens practically in this case? Are checkpoints not saved? (it depends on the checkpoint condition, but i dont understant that). 

Is a warning enough or should this be an error?

src/shared_utilities/utils.jl
```diff
@@ -265,6 +267,75 @@ function DriverUpdateCallback(updateat::Vector{FT}, updatefunc) where {FT}
...
+
+    schedule = EveryCalendarDtSchedule(
+        checkpoint_frequency_period;
+        reference_date = start_date,
+        date_last = start_date + Dates.Millisecond(1000t_start),
+    )
+
+    if !isnothing(dt)
+        dt_period = Dates.Millisecond(1000dt)
+        if !isivisible(checkpoint_frequency_period / dt_period)
```
kmdeck, line 323: isivisible -> isdivisible

test/standalone/Bucket/restart.jl
```diff
@@ -0,0 +1,102 @@
...
+import ClimaComms
+ClimaComms.@import_required_backends
+import ClimaLand
+import ClimaParams
+import SciMLBase
+import ClimaTimeSteppers
+import Dates
+import ClimaUtilities
+
+# First, run a simulation for 1 hour
```
kmdeck, line 11: right now it runs for 3hr

src/shared_utilities/models.jl
kmdeck: oh! One other idea is to think of the checkpoint not as an initialize step, but as a way to set initial conditions. This would be more consistent with what `initialize` does currently (i.e. just makes the structures Y,p). 

In that case, we call initialize like usual, but the checkpoint is handled as a way to set initial conditions. e.g.
```
Y, p, cds = initialize(model)
initial_conditions_from_checkpoint!(Y,restart_file; model)
t0 = initial_time_from_checkpoint(restart_file; model)
```
and then we call set_initial_cache! like usual

kmdeck: ^^this might be better, because this PR currently uses `initialize` in two different ways (allocating space vs doing that AND setting IC)

test/standalone/Bucket/restart.jl
```diff
@@ -0,0 +1,102 @@
...
+    (t_restart, tf),
+    p_restart,
+)
+updateat_restarted = collect(t_restart:Δt:tf)
+driver_cb_restarted =
+    ClimaLand.DriverUpdateCallback(updateat_restarted, updatefunc)
+cb_restarted = SciMLBase.CallbackSet(driver_cb_restarted)
+sol_restarted =
+    SciMLBase.solve(prob_restart, ode_algo; dt = Δt, callback = cb_restarted)
+for p in propertynames(Y.bucket)
```
kmdeck, line 100: while `p` is updated by `solve` in place, so the `p` you check here is `p` at tf,  I am not sure if the same is true for `Y`. You might double check (and if not, can use sol.u[end] as the final Y)

kmdeck: nevermind! I checked.
julia> Y == sol.u[end]
true

src/shared_utilities/checkpoints.jl
Sbozzolo: There's no guarantee that `Y.component.var` is a `Field`. One could have a scalar or a matrix there (we never do so, but also never forbid this)

src/shared_utilities/utils.jl
Sbozzolo: Thank you!

test/standalone/Bucket/restart.jl
Sbozzolo: Thank you, I also updated the describe the test.

src/shared_utilities/checkpoints.jl
Sbozzolo: A case where we cannot use `context_from_Y` is in the tests, which have dummy models with no context information attached.

src/shared_utilities/checkpoints.jl
kmdeck: That's true. We make the state using coordinates, but the coordinates just need to be something we can iterate over. 

In that case, what would ClimaComms.context return? probably an error. Do we need a warning then?

kmdeck: I was referring more to the possibility of us doing something like Y.canopy.energy.T (which actually, we do), in which case, this function would not work, right? Or maybe Y.canopy.energy is a field vector, and then it might still work

Sbozzolo: Yes, I think `Y.canopy.energy` would not work if `energy` is a FieldVector.

Sbozzolo: I updated this to be recursive so that it should find a context if there's any Field in the hierarchy.

kmdeck: nice!!

PR #854, opened 2024-10-15
Backup branch

kmdeck:
## Content

----


src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -49,10 +54,206 @@ end
...
 end
 
+"""
+    partial_q_sat_partial_T_ice(P::FT, T::FT) where {FT}
+Computes the quantity ∂q_sat∂T at temperature T and pressure P,
+over ice. The temperature must be in Celsius.
+Uses the polynomial approximation from Flatau et al. (1992).
+"""
+function partial_q_sat_partial_T_ice(P::FT, T::FT) where {FT}
+    T_celsius = T
```
kmdeck, line 69: We need to pass in T-273.15 or else do it here, because our T is in K

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -49,10 +54,206 @@ end
...
+    f_prime(T_s) = (2 * κ / h_sfc) + F_h_prime(T_s)
+
+    fns(T_s) = [f(T_s), f_prime(T_s)]
+    T_s::FT = solve_Ts(fns, T_bulk)
+
+    # adjust T_s and internal energy accordingly if T_s > freezing temperature
+    if T_s > 273.15
+        T_s_new = FT(273.15)
+        _T0::FT = FT(LP.T_0(earth_param_set))
+        U_Ts::FT = calc_U(_ρ_liq, d, cp_m, T_s, _T0)
```
kmdeck, line 248: we should pass c_snow here, not cp_m (that is for air)

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -49,10 +54,206 @@ end
...
+        U_Tf::FT = calc_U(_ρ_liq, d, cp_m, T_s_new, _T0)
+        ΔU::FT = U_Ts - U_Tf
+        ΔF::FT = ΔU / parameters.Δt
```
kmdeck, line 251: Should this be the opposite sign? If T_sfc > T_freeze, energy went into melting the top layer. If we say that instead T_sfc maxed out at T_freeze and the additional energy that would have increased the temperature went into the bulk snow, shouldnt delta F be negative? (negative = towards snow)

I also wonder if we should bring the top layer to q_l of the bulk, not to q_1 = 1

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -49,10 +54,206 @@ end
...
+
+    ρ_air = Thermodynamics.air_density(thermo_params, thermal_state_air)
+    SH(T_s) = cp_m * -ρ_air * (T_air - T_s) / r_ae(T_s)
+
+    _σ = LP.Stefan(earth_param_set)
+    F_R(T_s) = -(1 - α_sfc) * SW_d - ϵ_sfc * (LW_d - _σ * T_s^4)
+
+    F_h(T_s) = SH(T_s) + LH(T_s) + F_R(T_s) + energy_runoff
+    κ = parameters.κ_ice
+    # κ = snow_thermal_conductivity(ρ_snow, parameters)
```
kmdeck, line 226: Let's check how different these are

src/standalone/Snow/snow_parameterizations.jl
kmdeck: On this branch, we have the sign convention where this is correct. After rebasing, we need to change it

PR #857, opened 2024-10-16
new version of formatter plus formatting changes

kmdeck:
## Purpose 
JuliaFormatter made a major release to v2, and so in CI we also use v2 by default (since, prior to this PR, the .dev Project.toml is instantiated without any compat requirement). Hence anyone who is working locally needs to also use v2 (e.g., previously @AlexisRenchon  and I were on v1). If not, your local formatting changes wont match the CI formatting changes, and CI will fail.

This PR 
- formats all files with JuliaFormatter 2.0.0
- adds a compat entry to our .dev Project.toml to prevent the issue happening again.

### DIFF
--- a/.dev/Project.toml
+++ b/.dev/Project.toml
@@ -1,2 +1,5 @@
 [deps]
 JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
+
+[compat]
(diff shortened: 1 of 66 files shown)


juliasloan25 (approved):
Wow! Big change :o

PR #862, opened 2024-10-17
Change soil albedo to depend of soil moisture

### LINKED ISSUE #828: Make soil albedo vary with soil water content
- [x] move PAR and NIR albedos to aux vars
- [x] add parameters for dry/wet albedo
- [ ] obtain default values for above params
- [x] update tests
- [ ] update any experiments, tutorials, simulations as needed
- [x] add support for PAR_albedo and NIR_albedo params and deprecate it

imreddyTeja:
PAR and NIR albedo are moved from the EnergyHydrologyParameters to the auxiliary vars via AtmosDrivenFluxBC. The EnergyHydrologyParameters now hold PAR/NIR wet and dry as a field or float.  It also holds top_depth, which is the thickness of the top of the soil used in the albedo calculation.

The constructor for EnergyHydrologyParameters accepts both (PAR/NIR_albedo_dry/wet) and a general PAR/NIR_albedo. Supplying PAR/NIR_albedo is marked as deprecated, and the constructor sets both the dry and wet albedos to the same value. 

Previosuly, PAR and NIR albedo were held in a PrognosticSoilConditions
struct, which acted as an intermediary between the soil and canopy models.

Now, PAR and NIR albedo are stored in the cache, and the struct is only
used for dispatch.

Soil albedo is only updated when the top boundary condition is AtmosDrivenFluxBC. Using dispatch for the update_albedo function required moving is_saturated to Runoff.jl and reordering some of the includes in Soil.jl.

The integrated soil-canopy test created a PrognosticSoilConditions
struct and filled in with a PAR and NIR albedo. Then it called
ground_albedo_PAR and ground_albedo_NIR to check the values.

Both of those functions now read from the cache. In order to test them, a
SoilCanopy model would need to be created, initialized, and then updated.
This seems to complex for a test, so this test is deleted.
The long runs and benchmarks now use the wet/dry albedo maps from clm_data. See [here](https://github.com/CliMA/ClimaArtifacts/pull/53)

## Purpose 
Closes #828

## To-do
- Make a new soil_canopy test
- Possibly change calls to EnergyHydrologyParameters constructor

## Content
- PAR/NIR albedos moved to aux vars via boundry condition
- albedo only calculated when AtmosDrivenFluxBC
- PAR/NIR albedo depends of top soil moisture volume
- EnergyHydrologyParameters holds dry/wet NIR and PAR as float or field
- soil albedo added to diagnostics

----


ext/CreateParametersExt.jl
```diff
@@ -319,12 +321,21 @@ function EnergyHydrologyParameters(
...
-    PAR_albedo::SF = 0.2,
-    NIR_albedo::SF = 0.4,
+    PAR_albedo_dry::SF = 0.2,
+    NIR_albedo_dry::SF = 0.4,
+    PAR_albedo_wet::SF = 0.2,
+    NIR_albedo_wet::SF = 0.4,
+    ###########DEPRECATED################
+    PAR_albedo::SFD = nothing,
+    NIR_albedo::SFD = nothing,
+    ####################################
```
imreddyTeja, line 331: Should this be marked as deprecated, or is support going to be continued?

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -62,10 +62,16 @@ Base.@kwdef struct EnergyHydrologyParameters{
...
+    "Soil PAR Albedo dry"
+    PAR_albedo_dry::SF
+    "Soil NIR Albedo dry"
+    NIR_albedo_dry::SF
+    "Soil PAR Albedo wet"
+    PAR_albedo_wet::SF
+    "Soil NIR Albedo wet"
+    NIR_albedo_wet::SF
+    "Thickness of top of soil used in albedo calculations (m)"
+    top_depth::FT
```
imreddyTeja, line 74: Do we want to package these into a struct? Maybe just the albedos? The only advantage I can see to that is possibly a cleaner way of supporting a mix of floats and fields.

ext/CreateParametersExt.jl
imreddyTeja: When the formula recomputes albedo, if both wet and dry are the same, then the calculated value is also that same value. For example if a user passes in `PAR_albedo = 0.2`, then the constructor sets `PAR_albedo_dry = 0.2` and `PAR_albedo_wet = 0.2`. Then when calculating albedo, the `p.soil.PAR_albedo` is set to 
`PAR_albedo_wet * \theta + PAR_albedo_dry * (1-\theta) = 0.2`. So isn't the input albedo the same as the one in the cache?

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,108 @@ export volumetric_liquid_fraction,
...
+            soil_domain.fields.z_sfc - soil_domain.fields.z
+        # zeros all θ_l other than surface
+        @. p.soil.sub_sfc_scratch =
+            ClimaLand.heaviside.(p.soil.sfc_scratch, p.soil.sub_sfc_scratch) *
+            effective_saturation(ν, p.soil.θ_l, θ_r)
+        ClimaCore.Operators.column_integral_definite!(
+            ∫H_S_e_dz,
+            p.soil.sub_sfc_scratch,
+        )
+        # here ∫H_S_e_dz is equivalent to the effective_saturation of the top layer
```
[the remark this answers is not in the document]
imreddyTeja: I tried using the `obtain_surface_space` and `top_face_to_surface` functions to do this. Both errored with certain domains because `level` did not work with them. Do you have any suggestions on how to get the values of the first layer?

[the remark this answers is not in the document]
kmdeck: We dont want to use the value of the first layer because it is resolution dependent. Taking the average value over a predefined thickness is better defined (even though for coarse resolution sims, would amount to just the top layer, maybe).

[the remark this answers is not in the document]
kmdeck: Of course, the albedo is probably only determined by the top ~cm or two, which does coincide with the top layer. But what if someone chooses a much coarser or finer resolution? It's good to put in some protection for that to make our simulation less resolution dependence.

If we do not want to support this now, top_center_to_surface should work, e.g.: https://github.com/CliMA/ClimaLand.jl/blob/e2050188bfffa9fe2ac8b581165d5fb2927ffb1a/src/standalone/Soil/boundary_conditions.jl#L235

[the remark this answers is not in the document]
imreddyTeja: When the resolution is fine the function takes the average of all layers centered above the boundary. When the resolution is so coarse that no layer is centered above the boundary, the first layer is used. Isn't that what we want?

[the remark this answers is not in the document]
kmdeck: That is what we want. I thought Gabriele was suggesting using the top layer value.

ext/CreateParametersExt.jl
kmdeck: One idea (would have to be a follow-on PR) is to make an AbstractSoilAlbedoModel with two concrete types, one with wet/dry and the other with the bulk (constant) values. Each struct could store their respective values. Then we could also move the albedo parameters outside of EnergyHydrologyParameters and into the AtmosDrivenBC struct.

kmdeck: the top thickness parameter could also live in the struct

src/diagnostics/land_compute_methods.jl
```diff
@@ -131,6 +131,20 @@ end
...
+    out,
+    Y,
+    p,
+    t,
+    land_model::SoilCanopyModel{FT},
+) where {FT}
+    if isnothing(out)
+        return FT(0.5) .* p.soil.PAR_albedo .+ FT(0.5) .* p.soil.NIR_albedo
+    else
+        @. out = FT(0.5) .* p.soil.PAR_albedo .+ FT(0.5) .* p.soil.NIR_albedo
```
kmdeck, line 144: with the @. in front, can remove the broadcasts in the rest of the expression

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,108 @@ export volumetric_liquid_fraction,
...
     viscosity_factor,
     dψdϑ,
-    is_saturated
+    update_albedo!
+
+"""
+    albedo_from_moisture(surface_liquid_fraction::FT, albedo_dry::FT, albedo_wet::FT)
+
+Calculates pointwise albedo for any band as a function of volumetric soil water content given
+the dry and wet albedo values for that band. Implements eq.1 from Braghiere, R. K et al.
```
kmdeck, line 16: Perhaps can give the full citation. We should also note that we use soil color and albedo maps from CLM

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,108 @@ export volumetric_liquid_fraction,
...
+            ClimaLand.heaviside.(
+                FT(0.5) * alebdo_calc_top_thickness + sqrt(eps(FT)),
+                p.soil.sub_sfc_scratch,
+            ) * effective_saturation(ν, p.soil.θ_l, θ_r)
+        ClimaCore.Operators.column_integral_definite!(
+            ∫H_S_e_dz,
+            p.soil.sub_sfc_scratch,
+        )
+    else
+        # in the case where no layer is centered above boundry, use the top layer
```
kmdeck, line 79: boundary

kmdeck: Here we should be able to use only top_center_to_surface to extract the top center value, do not need integrals below, right? What error do you see?

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,108 @@ export volumetric_liquid_fraction,
...
+    FT = eltype(soil_domain.fields.Δz_top)
+    # checks if there is at least 1 layer centered within the top soil depth
+    if maximum(soil_domain.fields.Δz_top) < FT(0.5) * alebdo_calc_top_thickness
+        # ∫H_dz is integral of 1 from (surface-alebdo_calc_top_thickness) to surface
+        ∫H_dz = p.soil.sfc_scratch
+        # calculate depths of layer centers
+        @. p.soil.sub_sfc_scratch =
+            soil_domain.fields.z_sfc - soil_domain.fields.z
+        # zero all centers lower than boundry, set everything above to one
+        @. p.soil.sub_sfc_scratch =
```
kmdeck, line 56: I think here it would be clearer to do:
```
        @. p.soil.sub_sfc_scratch =
            ClimaLand.heaviside.(
                FT(0.5) * alebdo_calc_top_thickness + sqrt(eps(FT)),
                soil_domain.fields.z_sfc - soil_domain.fields.z,
            )
```
vs:
```
        @. p.soil.sub_sfc_scratch =
            soil_domain.fields.z_sfc - soil_domain.fields.z
        # zero all centers lower than boundry, set everything above to one
        @. p.soil.sub_sfc_scratch =
            ClimaLand.heaviside.(
                FT(0.5) * alebdo_calc_top_thickness + sqrt(eps(FT)),
                p.soil.sub_sfc_scratch,
            )
```

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,108 @@ export volumetric_liquid_fraction,
...
+            ClimaLand.heaviside.(
+                FT(0.5) * alebdo_calc_top_thickness + sqrt(eps(FT)),
+                p.soil.sub_sfc_scratch,
+            )
+        ClimaCore.Operators.column_integral_definite!(
+            ∫H_dz,
+            p.soil.sub_sfc_scratch,
+        )
+        # calculate depths again
+        @. p.soil.sub_sfc_scratch =
```
kmdeck, line 66: we could remove this in the same way

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,108 @@ export volumetric_liquid_fraction,
...
+            ClimaLand.heaviside.(p.soil.sfc_scratch, p.soil.sub_sfc_scratch) *
+            effective_saturation(ν, p.soil.θ_l, θ_r)
+        ClimaCore.Operators.column_integral_definite!(
+            ∫H_S_e_dz,
+            p.soil.sub_sfc_scratch,
+        )
+        # here ∫H_S_e_dz is equivalent to the effective_saturation of the top layer
+    end
+    @. p.soil.sfc_scratch = ∫H_S_e_dz / ∫H_dz
+    @. p.soil.PAR_albedo =
```
kmdeck, line 95: It also might be more readable to say
```
S_sfc = p.soil.sfc_scratch
@. S_sfc = ∫H_S_e_dz / ∫H_dz
```
and use S_sfc in the below

test/integrated/soil_canopy_lsm.jl
```diff
@@ -1,30 +0,0 @@
-using Test
```
kmdeck, line 1: Can we update this test instead of removing it entirely?

kmdeck:
CLM reference: Lawrence, P.J., and Chase, T.N. 2007. Representing a MODIS consistent land surface in the Community Land Model
(CLM 3.0). J. Geophys. Res. 112:G01023. DOI:10.1029/2006JG000168.

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,108 @@ export volumetric_liquid_fraction,
...
+        NIR_albedo_dry,
+        PAR_albedo_wet,
+        NIR_albedo_wet,
+        alebdo_calc_top_thickness,
+    ) = model_parameters
+    # ∫H_S_e_dz is the integral of effective saturation from (surface-alebdo_calc_top_thickness) to surface
+    ∫H_S_e_dz = p.soil.sfc_θ_l
+    FT = eltype(soil_domain.fields.Δz_top)
+    # checks if there is at least 1 layer centered within the top soil depth
+    if maximum(soil_domain.fields.Δz_top) < FT(0.5) * alebdo_calc_top_thickness
```
kmdeck, line 49: If you have a spare sfc_scratch floating around, you can remove the if statement and instead do:
```
    @. albedo_thickness = max(FT(0.5) * alebdo_calc_top_thickness, soil_domain.fields.Δz_top) + sqrt(eps(FT))
```

kmdeck, line 49: Note that we also have precomputed and stored the layer thickness (face-face): 
https://github.com/CliMA/ClimaLand.jl/blob/e2050188bfffa9fe2ac8b581165d5fb2927ffb1a/src/shared_utilities/Domains.jl#L921

under the name Delta z. You would need to get the top layer thickness with top_center_to_surface.

It doesnt change the computation you do below, but it would allow you to remove the 1/2 everywhere, which might improve readability

ext/CreateParametersExt.jl
kmdeck: the update_albedo function could then dispatch off of BC type and internally on albedo model type.

src/standalone/Soil/soil_hydrology_parameterizations.jl
imreddyTeja: I had something similar earlier. @Sbozzolo suggested using an if statement because the case when there is no layer centered within the top boundary can be calculated more easily without taking integrals. Should I use the if statement, or just set the albedo_thickness to adjust the lower boundary when needed, and integrate?

kmdeck: as you prefer! That makes sense too. 

If so, in principle I think it should be the minimum, not the maximum, right? Because if the maximum is > albedo_calc_top_thickness, but the minimum is not, the integral over the some columns will be zero.

src/diagnostics/define_diagnostics.jl
```diff
@@ -605,6 +605,18 @@ function define_diagnostics!(land_model)
             compute_infiltration!(out, Y, p, t, land_model),
     )
 
+    # Soil albedo
+    add_diagnostic_variable!(
+        short_name = "salb",
+        long_name = "Soil Albedo",
+        standard_name = "surface albedo",
+        units = "",
+        comments = "Dependent on soil moisture",
```
Sbozzolo, line 614: Maybe you can add a reference/equation here so that people know how it is calculated? The comment will end up in the NetCDF output

ext/CreateParametersExt.jl
Sbozzolo: > When the formula recomputes albedo, if both wet and dry are the same, then the calculated value is also that same value. For example if a user passes in PAR_albedo = 0.2, then the constructor sets PAR_albedo_dry = 0.2 and PAR_albedo_wet = 0.2. Then when calculating albedo, the p.soil.PAR_albedo is set to
> PAR_albedo_wet * \theta + PAR_albedo_dry * (1-\theta) = 0.2. So isn't the input albedo the same as the one in the cache?

Okay! I thought the integral would change things.

test/integrated/soil_canopy_lsm.jl
imreddyTeja, line 1: This test was deleted because it would not function with previous changes.
As requested, this test was recreated to work with the changes. The previous
test simply set an albedo it the top bc struct and then checked it using
the ground_albedo functions. With the changes, this would only work with an
initialized model. Because this is an integrated model test, I thought it
made sense to test the SoilCanopy model instead of just creating a
Soil model and checking if the function works. This test, like the
old version, calls the ground_albedo functions and checks their value.
This does seem very involved for a test, so I also made an alternative.
The alternative is just creating the Soil model and initializing it. Then, when the ground_albedo function
is called, it manually creates and passes the prognostic land components and boundary condition. See the alt below:

```
# Only care about PAR and NIR albedo values
        NIR_albedo = FT(0.4)
        PAR_albedo = FT(0.2)
        # setup EnergyHydrology with dummy params
        earth_param_set = LP.LandParameters(FT)
        radius = FT(6378.1e3)
        depth = FT(50)
        nelements = (50, 10)
        domain = ClimaLand.Domains.SphericalShell(;
            radius = radius,
            depth = depth,
            nelements = nelements,
            npolynomial = 1,
            dz_tuple = FT.((10.0, 0.1)),
        )
        model_params = Soil.EnergyHydrologyParameters(
            FT;
            ν = FT(0.1),
            ν_ss_om = FT(0.01),
            ν_ss_quartz = FT(0.01),
            ν_ss_gravel = FT(0.01),
            hydrology_cm = vanGenuchten{FT}(; α = FT(0), n = FT(0)),
            K_sat = FT(0),
            S_s = FT(0),
            θ_r = FT(0),
            NIR_albedo,
            PAR_albedo,
            emissivity = FT(0),
            z_0m = FT(0),
            z_0b = FT(0),
        )
        start_date = DateTime(2005)
        atmos = PrescribedAtmosphere(
            TimeVaryingInput(identity),
            TimeVaryingInput(identity),
            TimeVaryingInput(identity),
            TimeVaryingInput(identity),
            TimeVaryingInput(identity),
            TimeVaryingInput(identity),
            start_date,
            FT(0),
            earth_param_set,
        )
        radiation = PrescribedRadiativeFluxes(
            FT,
            TimeVaryingInput(identity),
            TimeVaryingInput(identity),
            start_date,
        )
        top_bc = ClimaLand.Soil.AtmosDrivenFluxBC(atmos, radiation)
        zero_water_flux = WaterFluxBC((p, t) -> 0.0)
        zero_heat_flux = HeatFluxBC((p, t) -> 0.0)
        boundary_fluxes = (;
            top = top_bc,
            bottom = WaterHeatBC(;
                water = zero_water_flux,
                heat = zero_heat_flux,
            ),
        )
        model = EnergyHydrology{FT}(;
            parameters = model_params,
            domain,
            boundary_conditions = boundary_fluxes,
            sources = (),
        )
        # initialize model
        # check that albedos have been added to cache
        @test haskey(p.soil, :PAR_albedo)
        @test haskey(p.soil, :NIR_albedo)
        # initialize cache, then check that albedos are set to the correct values
        set_initial_cache! = make_set_initial_cache(model)
        set_initial_cache!(p, Y, 0.0)
        prognostic_land_components = (:canopy, :soil, :soilco2)
        @test all(
            Array(
                parent(
                    Canopy.ground_albedo_PAR(
                        Val(prognostic_land_components),
                        ClimaLand.PrognosticSoilConditions(),
                        Y,
                        p,
                        0.0,
                    ),
                ),
            ) .== PAR_albedo,
        )
        @test all(
            Array(
                parent(
                    Canopy.ground_albedo_NIR(
                        Val(prognostic_land_components),
                        ClimaLand.PrognosticSoilConditions(),
                        Y,
                        p,
                        0.0,
                    ),
                ),
            ) .== NIR_albedo,
        )
```

src/diagnostics/define_diagnostics.jl
imreddyTeja: done

src/standalone/Soil/soil_hydrology_parameterizations.jl
imreddyTeja: done

src/diagnostics/land_compute_methods.jl
imreddyTeja: done

src/standalone/Soil/soil_hydrology_parameterizations.jl
imreddyTeja: citation moved to function that calls this one

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,108 @@ export volumetric_liquid_fraction,
...
+           albedo_wet * surface_liquid_fraction
+end
+
+
+"""
+    update_albedo!(bc::AtmosDrivenFluxBC, p, soil_domain, model_parameters)
+
+Calculates and updates PAR and NIR albedo as a function of volumetric soil water content at
+the top of the soil. If the soil layers are very large, the water content of the top layer
+is used in the calclulation
```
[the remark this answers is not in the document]
imreddyTeja: Done, but this docstring should be reviewed and edited

src/standalone/Soil/soil_hydrology_parameterizations.jl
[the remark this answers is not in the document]
imreddyTeja: done

src/standalone/Soil/soil_hydrology_parameterizations.jl
imreddyTeja: done

src/standalone/Soil/soil_hydrology_parameterizations.jl
imreddyTeja: done

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,108 @@ export volumetric_liquid_fraction,
...
+    @. p.soil.PAR_albedo =
+        albedo_from_moisture(p.soil.sfc_scratch, PAR_albedo_dry, PAR_albedo_wet)
+    @. p.soil.NIR_albedo =
+        albedo_from_moisture(p.soil.sfc_scratch, NIR_albedo_dry, NIR_albedo_wet)
+end
+
+"""
+update_albedo!(bc::AbstractWaterBC, p, soil_domain, model_parameters)
+
+    Does nothing for boundry conditions where albedo is not used
```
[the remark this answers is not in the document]
imreddyTeja: done

experiments/benchmarks/land.jl
```diff
@@ -206,8 +206,16 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
             soil_params_mask,
             0,
         )
-    PAR_albedo = ClimaCore.Fields.zeros(surface_space) .+ FT(0.2)
-    NIR_albedo = ClimaCore.Fields.zeros(surface_space) .+ FT(0.2)
+    # clm_data is used for g1, vcmax, rooting, soil albedo,  and two_stream param maps
+    clm_artifact_path = ClimaLand.Artifacts.clm_data_folder_path(; context)
+
+    PAR_albedo_dry, NIR_albedo_dry, PAR_albedo_wet, NIR_albedo_wet =
+        Soil.create_soil_albedo_vars(
```
Sbozzolo, line 213: I think we can take this one extra step and return direclty the clm data. This function could be `clm_soil_albedo_properties` and just take the `surface_space`.

src/diagnostics/land_compute_methods.jl
```diff
@@ -131,6 +131,20 @@ end
...
+    p,
+    t,
+    land_model::SoilCanopyModel{FT},
+) where {FT}
+    if isnothing(out)
+        return FT(0.5) .* p.soil.PAR_albedo .+ FT(0.5) .* p.soil.NIR_albedo
+    else
+        @. out = FT(0.5) * p.soil.PAR_albedo + FT(0.5) * p.soil.NIR_albedo
+    end
+end
```
Sbozzolo, line 146: Can this be writen as 
```
@diagnostic_compute  "soil_albedo" SoilCanopyModel (p.soil.PAR_albedo .+ p.soil.NIR_albedo) ./ 2
```

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -23,8 +23,8 @@ $(DocStringExtensions.FIELDS)
 """
 Base.@kwdef struct EnergyHydrologyParameters{
     FT <: AbstractFloat,
-    F <: Union{<:AbstractFloat, ClimaCore.Fields.Field},
-    SF <: Union{<:AbstractFloat, ClimaCore.Fields.Field},
+    F <: Union{<:FT, ClimaCore.Fields.Field},
+    SF <: Union{<:FT, ClimaCore.Fields.Field},
```
Sbozzolo, line 27: There's only one subtype of `FT` ... `FT` itself, so maybe you want to write `F <: Union{FT, ClimaCore.Fields.Field},`

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -745,9 +762,7 @@ Returns the surface albedo field of the
 `EnergyHydrology` soil model.
 """
 function ClimaLand.surface_albedo(model::EnergyHydrology{FT}, Y, p) where {FT}
-    return @. FT(
-        0.5 * model.parameters.PAR_albedo + 0.5 * model.parameters.NIR_albedo,
-    )
+    return @. FT(0.5 * p.soil.PAR_albedo + 0.5 * p.soil.NIR_albedo)
```
Sbozzolo, line 765:
```suggestion
    return @.  (p.soil.PAR_albedo + p.soil.NIR_albedo) / 2
```
so that you don't have to worry about the FT

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -1042,6 +1057,66 @@ function soil_turbulent_fluxes_at_a_point(
...
+        path,
+        "NIR_albedo_wet",
+        surface_space;
+        regridder_type,
+        regridder_kwargs,
+        file_reader_kwargs,
+        compose_function,
+    )
+
+    return PAR_albedo_dry, NIR_albedo_dry, PAR_albedo_wet, NIR_albedo_wet
```
Sbozzolo, line 1118: You could write this as
```suggestion
    return map(
    s -> SpaceVaryingInput(
        path,
       s,
        surface_space;
        regridder_type,
        regridder_kwargs,
        file_reader_kwargs,
        compose_function,
    ), ("PAR_albedo_dry", "NIR_albedo_dry", "PAR_albedo_wet", "NIT_albedo_dry"))
```

I also think that we don't need to expose `compose_function` here. It's hard to see how it would be used.

test/integrated/soil_canopy_lsm.jl
```diff
@@ -5,26 +5,219 @@ using ClimaCore
...
+        atmos = PrescribedAtmosphere(
+            TimeVaryingInput(identity),
+            TimeVaryingInput(identity),
+            TimeVaryingInput(identity),
+            TimeVaryingInput(identity),
+            TimeVaryingInput(identity),
+            TimeVaryingInput(identity),
+            start_date,
+            FT(0),
+            earth_param_set,
```
Sbozzolo, line 48: You could use `prescribed_analytic_forcing` to set up radiation and atmosphere

src/diagnostics/land_compute_methods.jl
imreddyTeja: Yes, this works. The docs for diagnostics, however, say 
"Or, for convenience, you can use the @diagnostic_compute macro which generates the same function. However, it is better to use that macro only if you are getting a defined variable, such as latent heat flux. (without an operation like the bowen ratio above). "

see [docs](https://clima.github.io/ClimaLand.jl/dev/diagnostics/users_diagnostics/)

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,107 @@ export volumetric_liquid_fraction,
...
+    update_albedo!
+
+"""
+    albedo_from_moisture(surface_liquid_fraction::FT, albedo_dry::FT, albedo_wet::FT)
+
+Calculates pointwise albedo for any band as a function of soil effective saturation given
+the dry and wet albedo values for that band.
+"""
+function albedo_from_moisture(
+    surface_liquid_fraction::FT,
```
kmdeck, line 19: surface_eff_sat::FT

src/diagnostics/define_diagnostics.jl
```diff
@@ -605,6 +605,18 @@ function define_diagnostics!(land_model)
             compute_infiltration!(out, Y, p, t, land_model),
     )
 
+    # Soil albedo
+    add_diagnostic_variable!(
+        short_name = "salb",
+        long_name = "Soil Albedo",
+        standard_name = "surface albedo",
+        units = "",
+        comments = "The mean of PAR and NIR albedo, which are calculated as α_soil_band = α_band_dry * (1 - S_e) + α_band_wet * S_e.",
```
kmdeck, line 614: if you like, you can add this as a default also to the :short diagnostics, and then switch the diagnosticsin the long run back to :short.

experiments/long_runs/land.jl
```diff
@@ -549,7 +555,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
         land,
         start_date;
         output_writer = nc_writer,
-        output_vars = :short,
+        output_vars = :long,
```
kmdeck, line 558: I think you switched to :long in order to get salb saved - but then it is not plotted below (can add it to list of short_names). If you dont want to plot it, we can go back to :short here. Or if you do, we can also add "salb" to the :short diagnostics.

test/integrated/soil_canopy_lsm.jl
kmdeck: I think what you have is good. It is involved setup for a test, but I think that is OK. We have to do the same thing in the soil/snow integrated test also.

src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -7,7 +7,107 @@ export volumetric_liquid_fraction,
...
+    return albedo_dry * (1 - surface_liquid_fraction) +
+           albedo_wet * surface_liquid_fraction
+end
+
+
+"""
+    update_albedo!(bc::AtmosDrivenFluxBC, p, soil_domain, model_parameters)
+
+Calculates and updates PAR and NIR albedo as a function of volumetric soil water content at
+the top of the soil. If the soil layers are very large, the water content of the top layer
```
kmdeck, line 32: If the soil layers are larger than a specified parameter `alb...`, the water content of the top layer...

src/standalone/Soil/soil_hydrology_parameterizations.jl
kmdeck: hey! I noticed that you changed the code to remove the factor of 1/2, but kept Delta z_top. If we remove the 1/2, we should use Delta z.

kmdeck (commented):
I think this is ready to merge, but i would like to see the long run output of monthly salb. it would be good to see it has reasonable values first.

kmdeck:
> I think this is ready to merge, but i would like to see the long run output of monthly salb. it would be good to see it has reasonable values first.

(this is easy to do with our existing plotting - so if you are unsure - just ask!)

src/standalone/Soil/energy_hydrology.jl
kmdeck: I am opening an issue for this, we dont need to do it here.

src/standalone/Soil/soil_hydrology_parameterizations.jl
kmdeck: or S_l_sfc

src/diagnostics/land_compute_methods.jl
Sbozzolo: Let's stick with you what you have here. I forgot why I added that comment, but I was problably wiser back then.

experiments/benchmarks/bucket.jl
```diff
@@ -166,7 +166,7 @@ SciMLBase.solve(prob, ode_algo; dt = Δt, callback = cb)
 @info "Starting profiling"
 # Stop when we profile for MAX_PROFILING_TIME_SECONDS or MAX_PROFILING_SAMPLES
 MAX_PROFILING_TIME_SECONDS = 500
-MAX_PROFILING_SAMPLES = 100
+MAX_PROFILING_SAMPLES = 80
```
kmdeck, line 169: any reason why this change was made?

kmdeck (approved):
thank you!

PR #864, opened 2024-10-18
Revert "new version of formatter plus formatting changes"

juliasloan25:
This reverts commit 9c2b29f4f90ef6725177e519fad02ee28468202d. That commit updated to use JuliaFormatter v2, but that release has been yanked, so here we return to v1.


AlexisRenchon (approved):

AlexisRenchon:
Thank you!

PR #831, opened 2024-10-08
run CI on julia 1.11

juliasloan25:
## Purpose 
Run CI on julia 1.11 in addition to 1.10

Closes #830

PR #865, opened 2024-10-18
finish Julia v1.11 updates

juliasloan25:
## Purpose 
I missed a few updated in #831; this PR finishes updating to use Julia v1.11 everywhere

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -25,7 +25,9 @@ jobs:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
         with:
-          version: '1.10'
+          version:
+            - '1.10'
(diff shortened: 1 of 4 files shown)


.github/workflows/downstream.yml
```diff
@@ -25,7 +25,8 @@ jobs:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
         with:
-          version: '1.10'
+          version:
+            - '1.11'
```
Sbozzolo, line 29: I think we should check both 1.10 and 1.11

Sbozzolo: This has to be with a job matrix, see downgrade job as an example

AlexisRenchon:
ready to go? or need rebase?

PR #870, opened 2024-10-21
Add prescribed_analytic_forcing

Sbozzolo:
This commit adds a convenience function to set up analytic forcing.

### DIFF
--- a/.github/workflows/ClimaLandSimulations.yml
+++ b/.github/workflows/ClimaLandSimulations.yml
@@ -13,16 +13,18 @@ jobs:
   lib-climalandsimulations:
     runs-on: ubuntu-20.04
     timeout-minutes: 45
+    strategy:
+      matrix:
+        version:
+          - '1.10'
+          - '1.11'
     steps:
       - name: Checkout
         uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@latest
         with:
-          version:
-            - '1.10'
-            - '1.11'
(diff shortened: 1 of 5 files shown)


kmdeck (commented):
do we need a more general constructor? Sometimes the values of the atmos vars are arbitrary, but sometimes they are chosen for the test

Sbozzolo:
> do we need a more general constructor? Sometimes the values of the atmos vars are arbitrary, but sometimes they are chosen for the test

I think this is the most general constructor we can have. You can just pass one of the arguments, and it will still work. For example, if you want to change precipitation:
```julia
prescribed_analytic_forcing(precip = TimeVaryingInput(....))
```

Sbozzolo:
> > > do we need a more general constructor? Sometimes the values of the atmos vars are arbitrary, but sometimes they are chosen for the test
> > 
> > 
> > I think this is the most general constructor we can have. You can just pass one of the arguments, and it will still work. For example, if you want to change precipitation:
> > ```julia
> > prescribed_analytic_forcing(precip = TimeVaryingInput(....))
> > ```
> 
> thanks! I hadnt appreciated that, I misread it as hardcoded to those values rather than being defaults.
> 
> Did you want to use this elsewhere in this PR or leave as is for now?

It is used in one case (to make sure that it works). I was not planning on changing it to more cases right now.

kmdeck (approved):

PR #873, opened 2024-10-21
ETOPO2022 dataset artifact

akshaysridhar:
## Purpose 
Adds ETOPO2022 https://www.ncei.noaa.gov/products/etopo-global-relief-model datasets to ClimaLand (ice-surface and bedrock). 60arc-second information resolution is lazily downloaded by default. 

----

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -120,3 +120,24 @@ git-tree-sha1 = "1e77fc84a027da43f1bf6ebe0a53849595a2be22"
     [[era5_lai_covers.download]]
     sha256 = "0331e3a2fc2cb53b0f1d9c5a84f49989da0b506258edc229da161e9ca4a61457"
     url = "https://caltech.box.com/shared/static/uzuzk5wh8r9q477jbrusfg1sr0ytzem9.gz"
+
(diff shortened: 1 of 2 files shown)


experiments/long_runs/land.jl
```diff
@@ -57,6 +57,10 @@ diagnostics_outdir = joinpath(root_path, "global_diagnostics")
 outdir =
     ClimaUtilities.OutputPathGenerator.generate_output_path(diagnostics_outdir)
 
+function mask_land_tendencies(Yₜ, land_mask)
```
kmdeck, line 60: just a note that we should not need this, since @sriharshakandala is adding the option of a mask directly to the ClimaCore space.

akshaysridhar: @kmdeck Indeed! This is simply to test your code with the best available dataset from the ETOPO2022 archives. You already have the required function in the code actually - the `ocean_mask_values` blocks

akshaysridhar:
@kmdeck This just updates the `Artifacts` helpers; we can trigger a custom longrun test with the new dataset following this PR

kmdeck (approved):

PR #863, opened 2024-10-18
Spatially varying clumping index

Espeer5:
## Purpose 

This PR makes the foliage clumping index of the radiative transfer model vary spatially, and uses clumping index data derived from MODIS to drive the global run.

The related PR from ClimaArtifacts is found [here](https://github.com/CliMA/ClimaArtifacts/pull/52)

## To-do

 - [x] Link related PR from ClimaArtifacts here
 - [ ] Pass all tests and check output plots

## Content

- Clumping index parameter of radiative transfer model can take in a field
- The various global runs load clumping index from a new ClimaArtifact containing MODIS clumping index data at the correct resolution

----
- [x] I have read and checked the items on the review checklist.

PR #881, opened 2024-10-23
Basic clumping index patch to fix NANs

Espeer5:
## Purpose 

This PR will patch the NANs introduced into the long runs by #863 

## To-do

- [ ] Verify cause and solution of NANs
- [ ] Change CI to 1 based on LAI rather than across globe

## Content

Fixes clumping index values to 1.0 wherever they were previously NAN to avoid carrying NANs into global runs.

----


Espeer5:
(whoops -this was a duplicate)

PR #886, opened 2024-10-24
Fixes a couple issues in run_fluxnet, ozark_pft uses diagnostics

kmdeck:
## Purpose 
Adds back in soil water content and temperature plots
Makes ozark_pft.jl use diagnostics (since it should produce the same plots/etc as run_fluxnet at US-MOz, but just with different parameters)
Removes saving callback completely from run_fluxnet

## Content

----

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 *#
 *html
 *~
+experiments/integrated/fluxnet/out/*
+experiments/integrated/fluxnet/US-*/out/*
 *.nc
 *.png
 *.mp4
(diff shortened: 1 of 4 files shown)


AlexisRenchon (approved):
It looks great, thank you for fixing the plots and adding diagnostics to the pft experiments! 

I think currently there is no plots saved in buildkite artifacts for the pft experiments, I am not sure if it is intended? 

I think it is because the file path is different from before https://github.com/CliMA/ClimaLand.jl/blob/main/.buildkite/pipeline.yml#L99

PR #888, opened 2024-10-25
Tr/fix logo i

### LINKED ISSUE #803: Fix ClimaLand logo
ClimaLand logo from the readme doesn't always render properly, could be browser or OS or... not sure. 

image

imreddyTeja:
## Purpose 
The logo previously had a little bit of white showing over the "I" in Clima and was quite large in the readme.
I do not know why, but the svg has a rectangle there. This removes that rectangle. 
Old: 
Screenshot 2024-10-25 at 10 32 10 AM
New: 
Screenshot 2024-10-25 at 10 37 24 AM

The rectangle was only visible when rendered in certain ways. For example, viewing the svg directly in Firefox does not show it and viewing it in an html file does not show it. When viewed through a markdown file, using html or the normal method of inserting images, the white rectangle is visible. 

closes #803

## Content
- remove rectangle from both .svg files
- scale down logo (please suggest a better looking size if you have one)

----


AlexisRenchon (approved):

Sbozzolo (approved):
Amazing, thanks!

PR #889, opened 2024-10-25
Tr/fix no docs found and undefined bindings

### LINKED ISSUE #848: Docs have many missing docstrings
The documentation has many of the yellow alert boxes for missing docstrings.

When generating the docs, there are multiple undefined binding warnings and no docs found warnings.

There are also other warnings, including  invalid links and duplicate docs

imreddyTeja:
## Purpose 
The documentation has many of the yellow alert boxes for missing docstrings.
When generating the docs, there were multiple warnings for "no docs found", "undefined binding", "invalid links", and "duplicate docs".

Now only two warnings remain. The first is for docstrings not included in the documentation, and the second is
APIs/shared_utilities.md generated html is too large
closes #848

## To-do

- There are 43 docstrings that are not included in the documentation. This should probably be another pr because it requires some thinking about how the documentation should be organized.

## Content
- replaced deprecated Pkg.installed() call
- removed docs for functions that no longer exist
- removed duplicate docs
- fix invalid links
- added basic docstring for PrognosticMet struct

----


docs/src/folderstructure.md
```diff
@@ -2,7 +2,7 @@
 
 [ClimaLand](https://github.com/CliMA/ClimaLand.jl) home directory has 6 main folders:
 
-- docs: contains files to generate [the documentation website](clima.github.io/ClimaLand.jl/).
+- docs: contains files to generate [the documentation website](https://clima.github.io/ClimaLand.jl/dev/).
```
AlexisRenchon, line 5: I think now that we fixed stable docs, that's what we should use.
https://clima.github.io/ClimaLand.jl/stable/

can you edit the README.md in this PR to update the badge from dev to stable as well?

AlexisRenchon (approved):

PR #883, opened 2024-10-23
helper functions for spatially varying params

kmdeck:
## Purpose 
Close #882 

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #890, opened 2024-10-28
Add leaderboard component to ClimaLand's long runs

### LINKED ISSUE #872: Add leaderboard component to ClimaLand's long runs
Similar to leaderboard component in ClimaCoupler, a leaderboard component can be added for ClimaLand.

The leaderboard component would plot the bias of the simulation and observational data for monthly data. See the plot below for what it could potentially look like. This will be done for `et`, `gpp`, and `lwu`. More variables can be added later.

et

ph-kev:
closes #872 - This PR adds a leaderboard component for ClimaLand's long runs. See the plots below.

gpp_bias_plot
lwu_bias_plot
et_bias_plot
global_rmse_and_bias_graphs

The values near the coastlines are `Nan`s because the observational data is resampled so that it matches the same grid as the simulation data.

I added a new file `land_leaderboard.jl` since no simulation being done for long runs are longer are two years or more. I am not sure if this should be done since `land_leaderboard.jl` is exactly the same as `land.jl` beside the length of the simulation. The leaderboard code still works even if the length of the simulation is only a year, it would just not throw away the first year of the simulation and use it instead for comparing against the observational data.

The other thing is that the simulation is not correct because the era5 land forcing data is from 2021.

### DIFF
--- a/docs/src/leaderboard/leaderboard.md
+++ b/docs/src/leaderboard/leaderboard.md
@@ -1,0 +1,122 @@
+# Leaderboard
+
+## Long run
+
+### Add a new variable to compare against observations
+The infrastructure to compute errors against observations is in the `leaderboard` folder.
+This folder contains two files: `data_sources.jl`, responsible for loading and preprocessing
+variables of interest, and `leaderboard.jl`, which computes error and draw plots. To add a
+new variable to the comparison, you modify the `data_sources.jl`.
+
+### Computation
+As of now, the leaderboard produces bias plots with the global bias and global root mean
+squared error (RMSE). These quantities are computed for each month with the first year of
+the simulation not considered as that is the spinup time. The start date of the simulation
+is 2008 which means that only the year 2009 is used to compare against observational data.
+
+### Add a new variable to the bias plots
+There are four functions that you need to modify to add a new variable which are
+`get_sim_var_dict`, `get_obs_var_dict`, `get_mask_dict`, and
+`get_compare_vars_biases_plot_extrema`. Each function returns a dictionary that must be
+modified to add a new variable to the leaderboard. The dictionaries are `sim_var_dict`,
+`obs_var_dict`, `mask_dict`, and `compare_vars_biases_plot_extrema`.
+
+To add a variable for the leaderboard, add a key-value pair to the dictionary `sim_var_dict`
+whose key is the short name of the variable and the value is a function that returns a
+[`OutputVar`](https://clima.github.io/ClimaAnalysis.jl/dev/var/). Any preprocessing is done
+in the function which includes unit conversion and shifting the dates.
+
+```julia
+sim_var_dict["et"] =
+        () -> begin
+            # Load in variable
+            sim_var = get(
+                ClimaAnalysis.SimDir(diagnostics_folder_path),
+                short_name = "et",
+            )
+            # Shift to the first day and subtract one month as preprocessing
+            sim_var =
+                ClimaAnalysis.shift_to_start_of_previous_month(sim_var)
+            return sim_var
+        end
+```
+
+Then, add a key-value pair to the dictionary `obs_var_dict` whose key is the same short name
+as before and the value is a function that takes in a start date and returns a `OutputVar`.
+Any preprocessing is done in the function.
+
+```julia
+obs_var_dict["et"] =
+    (start_date) -> begin
+    # We use ClimaArtifacts to use a dataset from ILAMB
+        obs_var = ClimaAnalysis.OutputVar(
+            ClimaLand.Artifacts.ilamb_dataset_path(;
+                context = "evspsbl_MODIS_et_0.5x0.5.nc",
+            ),
+            "et",
+            # start_date is used to align the dates in the observational data
+            # with the simulation data
+            new_start_date = start_date,
+            # Shift dates to the first day of the month before aligning the dates
+            shift_by = Dates.firstdayofmonth,
+        )
+        # More preprocessing to match the units with the simulation data
+        ClimaAnalysis.units(obs_var) == "kg/m2/s" &&
+            (obs_var = ClimaAnalysis.set_units(obs_var, "kg m^-2 s^-1"))
+        # ClimaAnalysis cannot handle `missing` values, but does support handling NaNs
+        obs_var = ClimaAnalysis.replace(obs_var, missing => NaN)
+        return obs_var
+    end
+```
+
+!!! tip "Preprocessing"
+    Observational and simulational data should be preprocessed for dates and units. When
+    using ClimaDiagnostics to report monthly averages from a simulation, monthly averages
+    are output on the first day following the month when the average was computed. For
+    instance, the monthly average corresponding to January 2010 is on the date 1 Feb 2010.
+    Preprocessing is done to shift this date to 1 Jan 2010. When preprocessing data, we
+    follow the convention that the first day corresponds to the monthly average for that
+    month. For observational data, you should check the convention being followed and
+    preprocess the dates if necessary.
+
+    For `obs_var_dict`, the anonymous function must take in a start date. The start date is
+    used in `leaderboard.jl` to adjust the seconds in the `OutputVar` to match between start
+    date in the simulation data.
+
+    Units should be the same between the simulation and observational data.
+
+Next, add a key-value pair to the dictionary `mask_dict` whose key is the same short name
+as before and the value is a function that takes in a `OutputVar` representing simulation
+data and a `OutputVar` representing observational data and returns a masking function or
+`nothing` if no masking function is needed. The masking function is used to correctly
+normalize the global bias and global RMSE. See the example below where a mask is made using
+the observational data.
+
+```julia
+mask_dict["et"] =
+    (sim_var, obs_var) -> begin
+        return ClimaAnalysis.make_lonlat_mask(
+            # We do this to get a `OutputVar` with only two dimensions:
+            # longitude and latitude
+            ClimaAnalysis.slice(
+                obs_var,
+                time = ClimaAnalysis.times(obs_var) |> first,
+            );
+            # Any values that are NaN should be 0.0
+            set_to_val = isnan,
+            true_val = 0.0
+        )
+    end
+```
+
+Finally, add a key-value pair to the dictionary `compare_vars_biases_plot_extrema` whose
+key is the same short name as before and the value is a tuple of floats which determine
+the range of the bias plots.
+
+```julia
+compare_vars_biases_plot_extrema = Dict(
+    "et" => (-0.00001, 0.00001),
+    "gpp" => (-8.0, 8.0),
+    "lwu" => (-40.0, 40.0),
+)
+```
--- a/experiments/long_runs/leaderboard/data_sources.jl
+++ b/experiments/long_runs/leaderboard/data_sources.jl
@@ -1,0 +1,209 @@
+import ClimaAnalysis
+
+"""
+    get_sim_var_dict(diagnostics_folder_path)
+
+Return a dictionary mapping short names to `OutputVar` containing preprocessed
+simulation data. This is used by the function `compute_leaderboard`.
+
+To add a variable for the leaderboard, add a key-value pair to the dictionary
+`sim_var_dict` whose key is the short name of the variable and the value is an
+anonymous function that returns a `OutputVar`. For each variable, any
+preprocessing should be done in the corresponding anonymous function which
+includes unit conversion and shifting the dates.
+
+The variable should have only three dimensions: time, longitude, and latitude.
+"""
+function get_sim_var_dict(diagnostics_folder_path)
+    # Dict for loading in simulation data
+    sim_var_dict = Dict{String, Any}()
+
+    sim_var_dict["lwu"] =
+        () -> begin
+            sim_var = get(
+                ClimaAnalysis.SimDir(diagnostics_folder_path),
+                short_name = "lwu",
+            )
+            sim_var =
+                ClimaAnalysis.shift_to_start_of_previous_month(sim_var)
+            return sim_var
+        end
+
+    sim_var_dict["et"] =
+        () -> begin
+            sim_var = get(
+                ClimaAnalysis.SimDir(diagnostics_folder_path),
+                short_name = "et",
+            )
+            sim_var =
+                ClimaAnalysis.shift_to_start_of_previous_month(sim_var)
+            (ClimaAnalysis.units(sim_var) == "kg m^-2 s^-1") && (
+                sim_var = ClimaAnalysis.convert_units(
+                    sim_var,
+                    "mm / day",
+                    conversion_function = units -> units * 86400.0,
+                )
+            )
+            return sim_var
+        end
+
+
+    sim_var_dict["gpp"] =
+        () -> begin
+            sim_var = get(
+                ClimaAnalysis.SimDir(diagnostics_folder_path),
+                short_name = "gpp",
+            )
+            sim_var =
+                ClimaAnalysis.shift_to_start_of_previous_month(sim_var)
+            # converting from to `mol CO2 m^-2 s^-1` in sim to `g C m-2 day-1` in obs
+            (ClimaAnalysis.units(sim_var) == "mol CO2 m^-2 s^-1") && (
+                sim_var = ClimaAnalysis.convert_units(
+                    sim_var,
+                    "g m-2 day-1",
+                    conversion_function = units -> units * 86400.0 * 12.011,
+                )
+            )
+            return sim_var
+        end
+    return sim_var_dict
+end
+
+"""
+    get_obs_var_dict()
+
+Return a dictionary mapping short names to `OutputVar` containing preprocessed
+observational data. This is used by the function `compute_leaderboard`.
+
+To add a variable for the leaderboard, add a key-value pair to the dictionary
+`obs_var_dict` whose key is the short name of the variable and the value is an
+anonymous function that returns a `OutputVar`. The function must take in a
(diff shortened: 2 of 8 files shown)


experiments/long_runs/leaderboard/data_sources.jl
```diff
@@ -0,0 +1,151 @@
...
+    sim_var_dict[short_name] =
+        () -> begin
+            sim_var = get(
+                ClimaAnalysis.SimDir(diagnostics_folder_path),
+                short_name = short_name,
+            )
+            # Remove the line below later when start_date is added to the diagnostics (see
+            # issue #879 in ClimaLand.jl)
+            haskey(sim_var.attributes, "start_date") ||
+                (sim_var.attributes["start_date"] = "2012-01-01T00:00:00")
```
Sbozzolo, line 42: It's added as of ClimaDiagnostics 0.2.10

experiments/long_runs/leaderboard/data_sources.jl
```diff
@@ -0,0 +1,151 @@
...
+            sim_var =
+                ClimaAnalysis.shift_to_start_of_previous_month(sim_var)
+            sim_var = ClimaAnalysis.convert_units(
+                sim_var,
+                "g m-2 day-1",
+                conversion_function = units -> units * 86400.0 * 12.011,
```
Sbozzolo, line 65: What is 12.011?

Sbozzolo (commented):
Can you add a page in the documentation describing the bias plots/leaderboard, how to add them? You can start from the comments you already have in the file

experiments/long_runs/leaderboard/data_sources.jl
ph-kev: The units in the simulation is `mol CO2 m^-2 s^-1` which is converted to match the units in the observational data which is `g C m^-2 day^-1`. I multiplied by `(86400 s / 1 day)`, `44.009 g CO2 / 1 mol CO2`, and `12.011 g C / 44.009 g CO2` to convert the units. I am not sure if the unit conversion is right.

kmdeck: This is correct

experiments/long_runs/leaderboard/data_sources.jl
```diff
@@ -0,0 +1,142 @@
...
+                ClimaAnalysis.SimDir(diagnostics_folder_path),
+                short_name = short_name,
+            )
+            sim_var =
+                ClimaAnalysis.shift_to_start_of_previous_month(sim_var)
+            return sim_var
+        end
+end
+
+for short_name in ["gpp"]
```
kmdeck, line 45: Do we need this loop?

experiments/long_runs/leaderboard/data_sources.jl
```diff
@@ -0,0 +1,142 @@
...
+
+# Dict for loading in observational data
+obs_var_dict = Dict{String, Any}()
+obs_var_dict["et"] =
+    (start_date) -> begin
+        obs_var = ClimaAnalysis.OutputVar(
+            ClimaLand.Artifacts.ilamb_dataset_path(;
+                context = "evspsbl_MODIS_et_0.5x0.5.nc",
+            ),
+            "et",
```
kmdeck, line 72: Do we rely on using the same short_name in the data and the diagnostic output from the simulation anywhere?

kmdeck: or it seems like this line could be anything, as long as we set `obs_var_dict["et"]`

docs/src/leaderboard/leaderboard.md
```diff
@@ -0,0 +1,123 @@
...
+            ),
+            "et",
+            # start_date is used to align the dates in the observational data
+            # with the simulation data
+            new_start_date = start_date,
+            # Shift dates to the first day of the month before aligning the dates
+            shift_by = Dates.firstdayofmonth,
+        )
+        # More preprocessing to match the units with the simulation data
+        ClimaAnalysis.units(obs_var) == "kg/m2/s" &&
```
kmdeck, line 66: Maybe we could link to ClimaAnalysis documentation explaining the formatting for units? If I understand correctly, this will carry out unit conversion, but that means the string must be in the format expected by ClimaAnalysis

experiments/long_runs/leaderboard/leaderboard.jl
```diff
@@ -0,0 +1,165 @@
...
+import Dates
+
+include("data_sources.jl")
+
+"""
+    compute_leaderboard(leaderboard_base_path, diagnostics_folder_path)
+
+Plot the biases against observations and simulation data.
+
+The paramter `leaderboard_base_path` is the path to save the leaderboards and bias plots and
```
kmdeck, line 13: parameter

kmdeck: though it should be called an argument :)

experiments/long_runs/leaderboard/leaderboard.jl
```diff
@@ -0,0 +1,165 @@
...
+import GeoMakie
+import CairoMakie
+import Dates
+
+include("data_sources.jl")
+
+"""
+    compute_leaderboard(leaderboard_base_path, diagnostics_folder_path)
+
+Plot the biases against observations and simulation data.
```
kmdeck, line 11: I see that both obs_var_dict and sim_var_dict are used in this function but not passed in as arguments. For clarity and to make it easy on future users, could we pass in those as arguments?

experiments/long_runs/leaderboard/leaderboard.jl
```diff
@@ -0,0 +1,165 @@
...
+
+The paramter `leaderboard_base_path` is the path to save the leaderboards and bias plots and
+`diagnostics_folder_path` is the path to the simulation data.
+"""
+function compute_leaderboard(leaderboard_base_path, diagnostics_folder_path)
+    @info "Error against observations"
+
+    # Set up dict for storing simulation and observational data after processing
+    sim_obs_comparsion_dict = Dict()
```
kmdeck, line 20: comparison is spelled incorrectly

experiments/long_runs/leaderboard/leaderboard.jl
kmdeck: Is this argument used? diagnostics_folder_path? Why pass this if we also have already created the sim_var_dict?

experiments/long_runs/leaderboard/leaderboard.jl
```diff
@@ -0,0 +1,165 @@
...
+
+        # Remove first spin_up_months months from simulation
+        spin_up_months = 12
+        spinup_cutoff = spin_up_months * 30 * 86400.0
+        ClimaAnalysis.times(sim_var)[end] >= spinup_cutoff && (
+            sim_var =
+                ClimaAnalysis.window(sim_var, "time", left = spinup_cutoff)
+        )
+
+        # Get 12 or less months of data
```
kmdeck, line 48: Why is this here? What if we had run for many years, and only used the first year as spinup?

experiments/long_runs/leaderboard/leaderboard.jl
```diff
@@ -0,0 +1,165 @@
...
+
+        sim_obs_comparsion_dict[short_name] = (sim_var, obs_var)
+    end
+
+    # Get the right number of months for plotting
+    _, sim_obs_tuple = first(sim_obs_comparsion_dict)
+    num_times = length(ClimaAnalysis.times(sim_obs_tuple[1]))
+    months = Dates.monthname.(1:num_times)
+
+    # Plot monthly comparsions
```
kmdeck, line 68: comparison

experiments/long_runs/land_leaderboard.jl
```diff
@@ -0,0 +1,596 @@
...
+            update_j = CTS.UpdateEvery(CTS.NewNewtonIteration),
+        ),
+    )
+    SciMLBase.solve(prob, ode_algo; dt = Δt, callback = cb, adaptive = false)
+    return nothing
+end
+
+setup_and_solve_problem(; greet = true);
+
+# read in diagnostics and make some plots!
```
kmdeck, line 591: can remove this line?

experiments/long_runs/leaderboard/data_sources.jl
```diff
@@ -0,0 +1,142 @@
+import ClimaAnalysis
```
kmdeck, line 1: I noticed that we change the units of the simulation output in some cases (GPP), and in the observations (LWU) in others. it would be good to make this consistent (simulation unit changes to match observations)

experiments/long_runs/leaderboard/leaderboard.jl
```diff
@@ -0,0 +1,165 @@
...
+        mask = mask_dict[short_name](sim_var, obs_var)
+        # Compute globas bias and global rmse
+        rmse_vec = []
+        bias_vec = []
+        times = ClimaAnalysis.times(sim_var) |> copy
+        for t in times
+            g_rmse = ClimaAnalysis.global_rmse(
+                ClimaAnalysis.slice(sim_var, time = t),
+                ClimaAnalysis.slice(obs_var, time = t),
+                mask = mask,
```
kmdeck, line 114: ClimaAnalysis uses the mask to change the denominator in RMSE and mean bias?

experiments/long_runs/land_leaderboard.jl
```diff
@@ -0,0 +1,596 @@
...
+    )
+
+    soil_params_artifact_path =
+        ClimaLand.Artifacts.soil_params_artifact_folder_path(; context)
+    extrapolation_bc = (
+        Interpolations.Periodic(),
+        Interpolations.Flat(),
+        Interpolations.Flat(),
+    )
+    soil_params_mask = SpaceVaryingInput(
```
kmdeck, line 93: Once this PR merges: https://github.com/CliMA/ClimaLand.jl/pull/883, could you change this script to use the helper functions for canopy and soil spatially varying parameters?

experiments/long_runs/leaderboard/data_sources.jl
ph-kev: We don't rely on the same short_name in the (observational?) data and the diagnostic output from the simulation.

It is like that because ClimaAnalysis sometimes can't correctly infers what variable is or if there is more than one variable in the .nc file.

The line can be anything as long as `obs_var_dict["et"]` is a function that takes in a start date and return a `OutputVar`.

docs/src/leaderboard/leaderboard.md
ph-kev: For that part of the code, it is just checking that the string of the units is "kg/m2/s" and if it is, then set the units to "kg m^-2 s^-1". The check is just there to make sure the units are set correctly.

There is no convention for the formatting of units unless one want to use automatic unit conversion. We only do this because ClimaAnalysis can't tell that "kg/m2/s" (in the observational data) is the same as "kg m^-2 s^-1" (in the simulation data).

experiments/long_runs/leaderboard/leaderboard.jl
ph-kev: I added that there in case if the simulation run for longer than one year, but less than two years. For instance, if one were to run the simulation for 1.5 years, then only six months will be used after throwing away the first year. 

The code will always use the second year regardless if the simulation run for multiple years. It assumed that we start a year 2012 and end at year 2013 (although, it could be any year as long as data exists in the observations) The only reason is because some datasets are limited in time. For example, for `et`, the last point in time is `2013-12-16`.

experiments/long_runs/leaderboard/leaderboard.jl
ph-kev: I am going to make the entire `data_sources.jl` file a function instead since I realized it is confusing how things are being initialized before.

experiments/long_runs/leaderboard/leaderboard.jl
ph-kev: Yes, the mask is used to normalize the global RMSE and global bias.

experiments/long_runs/leaderboard/data_sources.jl
ph-kev: I agree that this should be consistent, but I think we should change all observation units to match simulation units. This is because there is some consistency with how the units are written for the simulation data, but not for the observational data.

ph-kev:
The plots now look like this. The last month is missing because ClimaDiagnostics.jl did not save the last month.
global_rmse_and_bias_graphs
gpp_bias_plot
lwu_bias_plot
et_bias_plot

docs/src/leaderboard/leaderboard.md
```diff
@@ -0,0 +1,126 @@
+# Leaderboard
+
+## Long run
+
+### Add a new variable to compare against observations
+Computing errors against observations are all contained in the `leaderboard` folder. The
+files in the leaderboard folder are `data_sources.jl` and `leaderboard.jl`. Loading and
+preprocessing variables of interest are done in `data_sources.jl` and computing the errors
+and plotting are done in `leaderboard.jl`. To add a new variable, you ideally only need to
+modify `data_sources.jl`.
```
Sbozzolo, line 10:
```suggestion
The infrastructure to compute errors against observations is in the `leaderboard` folder. This
folder contains two files: `data_sources.jl`, responsible for loading and preprocessing
variables of interest, and `leaderboard.jl`, which computes error and draw plots. To add a new
variable to the comparison, you modify the `data_sources.jl`.
```

docs/src/leaderboard/leaderboard.md
```diff
@@ -0,0 +1,126 @@
...
+        obs_var = ClimaAnalysis.replace(obs_var, missing => NaN)
+        return obs_var
+    end
+```
+
+!!! tip "Preprocessing"
+    Observational and simulational data should be preprocessed for dates and units. For
+    simulation data, monthly averages correspond to the first day following the month.
+    For instance, the monthly average corresponding to January 2010 is on the date
+    2/1/2010. Preprocessing is done to shift this date to 1/1/2010. When preprocessing
```
Sbozzolo, line 81:
```suggestion
    1 Feb 2010. Preprocessing is done to shift this date to 1 Jan 2010. When preprocessing
```

experiments/long_runs/leaderboard/data_sources.jl
Sbozzolo: For some variables, it is more natural to use observation units instead of simulation units, so we should convert towards what makes the most sense

Sbozzolo (approved):

docs/src/leaderboard/leaderboard.md
```diff
@@ -0,0 +1,126 @@
...
+files in the leaderboard folder are `data_sources.jl` and `leaderboard.jl`. Loading and
+preprocessing variables of interest are done in `data_sources.jl` and computing the errors
+and plotting are done in `leaderboard.jl`. To add a new variable, you ideally only need to
+modify `data_sources.jl`.
+
+### Computation
+As of now, the leaderboard produces bias plots with the global bias and global root mean
+squared error (RMSE). These quantities are computed for each month with the first year of
+the simulation not considered as that is the spinup time. The start date of the simulation
+is 2012 which means that only the year 2013 is used to compare against observational data.
```
kmdeck, line 16: I dont know that this is up to date?

kmdeck: I believe we are simulating on year (2008)

docs/src/leaderboard/leaderboard.md
```diff
@@ -0,0 +1,126 @@
...
+            (obs_var = ClimaAnalysis.set_units(obs_var, "kg m^-2 s^-1"))
+        # ClimaAnalysis cannot handle `missing` values, but does support handling NaNs
+        obs_var = ClimaAnalysis.replace(obs_var, missing => NaN)
+        return obs_var
+    end
+```
+
+!!! tip "Preprocessing"
+    Observational and simulational data should be preprocessed for dates and units. For
+    simulation data, monthly averages correspond to the first day following the month.
```
kmdeck, line 79: "When using ClimaDiagnostics to report monthly averages from a simulation, monthly averages are output on the first day following the month when the average was computed."

experiments/long_runs/land.jl
```diff
@@ -433,3 +433,8 @@ mktempdir(root_path) do tmpdir
         run(Cmd([unite, figures..., joinpath(root_path, "figures.pdf")]))
     end
 end
+
```
kmdeck, line 436: I think we should add this instead to snowy_land.jl, and run for two years.

experiments/long_runs/leaderboard/data_sources.jl
kmdeck: fair enough!

experiments/long_runs/leaderboard/data_sources.jl
```diff
@@ -0,0 +1,191 @@
...
+Return a dictionary mapping short names to `OutputVar` containing preprocessed
+simulation data. This is used by the function `compute_leaderboard`.
+
+To add a variable for the leaderboard, add a key-value pair to the dictionary
+`sim_var_dict` whose key is the short name of the variable and the value is an
+anonymous function that returns a `OutputVar`. For each variable, any
+preprocessing should be done in the corresponding anonymous function which
+includes unit conversion and shifting the dates.
+
+The variable should have only three dimensions: latitude, longitude, and time.
```
kmdeck, line 15: I believe the ordering is lon, lat, time? or time lon lat? let's update the docstring to reflect the ordering in the data

experiments/long_runs/land.jl
ph-kev: Is it fine to change it to two years even if the CI will take longer?

PR #905, opened 2024-11-01
integrate neural network into Snow.jl architecture

a-charbon:
## Purpose 

Integrate the neural snow depth model into the Snow model architecture. This involves small changes to Snow.jl and SnowParameterizations.jl as well as expanding the NeuralSnow extension to allow the neural model as a type of AbstractDensityModel. "CSV" was also added to the buildkite dependencies to permit testing of the NeuralSnow extension, and BSON has been appropriately added to the ClimaLand weak dependencies to enable the expanded functionality of the extension.

## To-do

Questions are commented in the changed files about different possible approaches. With this PR merged, I can also add a slightly faster and more abstract/streamlined form of the neural model itself (changes to ModelTools, etc).

## Content

Expanded NeuralSnow extension and changed some of the syntax of the Snow model.

----

### DIFF
--- a/experiments/standalone/Snow/snowmip_simulation.jl
+++ b/experiments/standalone/Snow/snowmip_simulation.jl
@@ -20,6 +20,10 @@ using Statistics
 using Insolation
 using DelimitedFiles
 
+using CSV, HTTP, Flux, cuDNN, BSON, StatsBase
+ModelTools = Base.get_extension(ClimaLand, :NeuralSnowExt).ModelTools;
+NeuralSnow = Base.get_extension(ClimaLand, :NeuralSnowExt).NeuralSnow;
+
 # Site-specific quantities
 # Error if no site argument is provided
 if length(ARGS) < 1
@@ -46,8 +50,16 @@ ndays = (tf - t0) / 3600 / 24
 
 domain = ClimaLand.Domains.Point(; z_sfc = FT(0))
 
-parameters =
-    SnowParameters{FT}(Δt; α_snow = α, ρ_snow = ρ, earth_param_set = param_set)
+#dens_neural = NeuralSnow.NeuralDepthModel(FT)
+dens_constant = Snow.ConstantDensityModel(ρ)
+depths = z[snow_data_avail]
+
+parameters = SnowParameters{FT}(
+    Δt;
+    α_snow = α,
+    density = dens_constant,
+    earth_param_set = param_set,
+)
 model = ClimaLand.Snow.SnowModel(
     parameters = parameters,
     domain = domain,
@@ -57,6 +69,8 @@ Y, p, coords = ClimaLand.initialize(model)
 
 # Set initial conditions
 Y.snow.S .= FT(SWE[1]) # first data point
+#Y.snow.Z .= FT(depths[1]) #uncomment if using dens_neural, not dens_constant
+p.snow.ρ_snow .= FT(SWE[1] / depths[1] * 1000.0)
 Y.snow.U .=
     ClimaLand.Snow.energy_from_q_l_and_swe(FT(SWE[1]), FT(0), parameters) # with q_l = 0
 
--- a/ext/neural_snow/NeuralSnow.jl
+++ b/ext/neural_snow/NeuralSnow.jl
@@ -1,0 +1,240 @@
+module NeuralSnow
+
+using Flux
+using ClimaLand
+using ClimaLand.Snow: AbstractDensityModel, SnowModel, SnowParameters
+import ClimaLand.Snow:
+    density_prog_vars,
+    density_prog_types,
+    density_prog_names,
+    density_aux_vars,
+    density_aux_types,
+    density_aux_names,
+    update_density!,
+    update_density_prog!
+import ClimaLand.Parameters as LP
+using Thermodynamics
+
+using DataFrames, Dates, CSV, HTTP, Flux, cuDNN, StatsBase, BSON
+ModelTools = Base.get_extension(ClimaLand, :NeuralSnowExt).ModelTools
+#^^is this correct format since this is within the extension itself, or do i just use include(./ModelTools); using ModelTools; here?
+#the only reason I am using this is for the make_model() (once) and settimescale!() (twice) functions.
+
+export NeuralDepthModel, update_density!, update_density_prog!
+
+"""
+    NeuralDepthModel{FT <: AbstractFloat} <: AbstractDensityModel{FT}
+Establishes the density parameterization where snow density
+is calculated from a neural network determining the rate of change of snow depth, dzdt.
+"""
+struct NeuralDepthModel{FT} <: AbstractDensityModel{FT}
+    z_model::Flux.Chain
+    α::FT
+end
+
+"""
+    get_znetwork(; download_link = "https://caltech.box.com/shared/static/ay7cv0rhuiytrqbongpeq2y7m3cimhm4.bson")
+A default function for returning the snow-depth neural network from the paper.
+"""
+#This is the only place that creates the necessity of BSON for the neural extension and the make_model() from ModelTools. Keep or get rid of?
+#Can I change this to something CliMA already includes? is BSON vs JLD2 or anything else better? An artifact (can they store BSON/.jld2 files for direct usage without those packages?)?
+function get_znetwork(;
+    download_link = "https://caltech.box.com/shared/static/ay7cv0rhuiytrqbongpeq2y7m3cimhm4.bson",
+)
+    z_idx = 1
+    p_idx = 7
+    nfeatures = 7
+    zmodel = ModelTools.make_model(nfeatures, 4, z_idx, p_idx)
+    zmodel_state = BSON.load(IOBuffer(HTTP.get(download_link).body))[:zstate]
+    Flux.loadmodel!(zmodel, zmodel_state)
+    ModelTools.settimescale!(zmodel, 86400.0)
+    return zmodel
+
+    #should I write this function to give the ability to also load from local file too?
+    #BSON.@load "../SnowResearch/cleancode/testhrmodel.bson" zstate
+    #Flux.loadmodel!(zmodel, zstate)
+end
+
+"""
+    converted_model_type!(model::Flux.Chain, FT::Type)
+A wrapper function to aid conversion of the utilized Flux network weights to the type determined by the simulation.
+Currently supports `Float32` and `Float64` types.
+"""
+function converted_model_type(model::Flux.Chain, FT::DataType)
+    if FT == Float32
+        return Flux.f32(model)
+    elseif FT == Float64
+        return Flux.f64(model)
+    else
+        error(
+            "Conversion of Flux Chain weights to the desired float-type is not supported. Please implement the desired conversion
+      in NeuralSnow.convert_model_type!() or a custom constructor for the NeuralDepthModel with the desired type.",
+        )
+    end
+end
+
+"""
+    NeuralDepthModel(FT::DataType; Δt::Union{Nothing, FT}=nothing; model::Flux.Chain, α::Union{FT, Nothing})
+An outer constructor for the `NeuralDepthModel` density parameterization for usage in a snow model.
+Can choose to use custom models via the `model` argument, or a custom exponential moving average weight for the network inputs.
+The Δt argument can be used to set the lower boundary value on the network (when not using a custom model) in accordance with the paper upon initialization,
+equivalent to ModelTools.settimescale!(model, Δt, dtype = FT).
+"""
+function NeuralDepthModel(
+    FT::DataType;
+    Δt = nothing,
+    model::Flux.Chain = get_znetwork(),
+    α = nothing,
+)
+    usemodel = converted_model_type(model, FT)
+    weight =
+        !isnothing(α) ? FT(α) :
+        (isnothing(Δt) ? FT(1 / 12) : FT(maximum([2 * Δt / 86400, 1])))
+    # should we set the timescale of the model to Δt? getting rid of get_znetwork() and below line mitigates the need to import ModelTools entirely.
+    if !isnothing(Δt)
+        ModelTools.settimescale!(usemodel, Δt, dtype = FT) #assumes Δt is provided in seconds
+    end
+    return NeuralDepthModel{FT}(usemodel, weight)
+end
+
+#Define the additional prognostic variables needed for using this parameterization:
+#do I need to export these for them to be utilized in the simulation?
+ClimaLand.Snow.density_prog_vars(m::NeuralDepthModel) =
+    (:Z, :P_avg, :T_avg, :R_avg, :Qrel_avg, :u_avg)
+ClimaLand.Snow.density_prog_types(m::NeuralDepthModel{FT}) where {FT} =
+    (FT, FT, FT, FT, FT, FT)
+ClimaLand.Snow.density_prog_names(m::NeuralDepthModel) =
+    (:surface, :surface, :surface, :surface, :surface, :surface)
+
+#so far, every type is just blank here - is it even worth providing this feature or just get rid of this feature in Snow.jl entirely?
+ClimaLand.Snow.density_aux_vars(m::NeuralDepthModel) = ()
+ClimaLand.Snow.density_aux_types(m::NeuralDepthModel{FT}) where {FT} = ()
+ClimaLand.Snow.density_aux_names(m::NeuralDepthModel) = ()
+
+"""
+    snow_density(density::NeuralDepthModel, SWE::FT, z::FT, parameters::SnowParameters{FT}) where {FT}
+Returns the snow density given the current model state and the `NeuralDepthModel`
+density paramterization.
+"""
+function snow_density(
+    density::NeuralDepthModel{FT},
+    SWE::FT,
+    z::FT,
+    parameters::SnowParameters{FT},
+)::FT where {FT}
+    ρ_l = FT(LP.ρ_cloud_liq(parameters.earth_param_set))
+    if SWE <= eps(FT) #if there is no snowpack, aka avoid NaN
+        return FT(ρ_l)
+    end
+    ρ_new = SWE / z * ρ_l
+    return ρ_new
+end
+
+"""
+    update_density!(density::NeuralDepthModel, params::SnowParameters, Y, p)
+Updates the snow density given the current model state and the `NeuralDepthModel`
+density paramterization.
+"""
+function update_density!(
+    density::NeuralDepthModel,
+    params::SnowParameters,
+    Y,
+    p,
+)
+    p.snow.ρ_snow .= snow_density.(Ref(density), Y.snow.S, Y.snow.Z, params)
+end
+
+"""
+    eval_nn(vmodel, z::FT, swe::FT, P::FT, T::FT, R::FT, qrel::FT, u::FT)::FT where {FT}
+Helper function for evaluating the neural network in a pointwise manner over a `ClimaCore.Field`
+and returning the output in a broadcastable way.
+"""
+function eval_nn(
+    model,
+    z::FT,
+    swe::FT,
+    P::FT,
+    T::FT,
+    R::FT,
+    qrel::FT,
+    u::FT,
+)::FT where {FT}
+    input = FT.([z, swe, qrel, R, u, T, P])
+    return model(input)[1]
+end
+
+"""
+    dzdt(density::NeuralDepthModel, model::SnowModel{FT}, Y, p, t) where {FT}
+Returns the change in snow depth (rate) given the current model state and the `NeuralDepthModel`
+density paramterization, passing the approximate average of the forcings over the last 24 hours instead of
+the instantaneous value.
+"""
+function dzdt(density::NeuralDepthModel, Y)
+    return eval_nn.(
+        Ref(density.z_model),
+        Y.snow.Z,
+        Y.snow.S,
+        Y.snow.P_avg,
+        Y.snow.T_avg,
+        Y.snow.R_avg,
+        Y.snow.Qrel_avg,
+        Y.snow.u_avg,
+    )
+end
+
+"""
+    clip_dZdt(S::FT, Z::FT, dSdt::FT, dZdt::FT, Δt::FT)::FT
+A helper function which clips the tendency of Z such that
+its behavior is consistent with that of S.
+"""
+function clip_dZdt(S::FT, Z::FT, dSdt::FT, dZdt::FT, Δt::FT)::FT where {FT}
+    if (S + dSdt * Δt) <= eps(FT)
+        return -Z / Δt
+    elseif (Z + dZdt * Δt) < (S + dSdt * Δt)
+        #do we need to do something about setting q_l = 1 here?
+        return (S - Z) / Δt + dSdt
+    else
+        return dZdt
+    end
+end
+
+"""
+    update_density_prog!(density::NeuralDepthModel, model::SnowModel, Y, p)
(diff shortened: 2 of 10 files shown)


src/standalone/Snow/Snow.jl
```diff
@@ -42,6 +42,81 @@ and is used as a bulk snow model.
...
+    ConstantDensityModel{FT <: AbstractFloat} <: AbstractDensityModel{FT}
+Establishes the density parameterization where snow density
+is always treated as a constant (type FT).
+"""
+struct ConstantDensityModel{FT} <: AbstractDensityModel{FT}
+    ρ_snow::FT
+end
+
+"""
+    ConstantDensityModel{FT}(ρ::FT)
```
kmdeck, line 63: We may not need this outer constructor....the default would be:

ConstantDensityModel(rho) and it would use the type of rho as FT

is this ok to use? Or is your intent to make sure the type of rho matches the FT set in the sim script?

src/standalone/Snow/Snow.jl
```diff
@@ -42,6 +42,81 @@ and is used as a bulk snow model.
...
+    cx::FT
+    ρ_d::FT
+end
+
+"""
+    Anderson1976(FT; c1, c2, c3, c4, c5, ρ_d, cx)
+An outer constructor for the `Anderson1976` density parameterization for usage in a snow model.
+Uses the default constants defined by Anderson but allows for redefinition of the model constants.
+These include:
+- c1: fractional increase in density (0.026 cm/hr)
```
kmdeck, line 91: Maybe note what physically causes this increase

src/standalone/Snow/Snow.jl
```diff
@@ -42,6 +42,81 @@ and is used as a bulk snow model.
...
+- c2: compression constant estimated by Kojima in 1967 (21 cm^3/g)
+- c3: fractional settling rate at 0 degrees C for densities less than the critical density ρ_d (0.005 1/hr)
+- c4: constant (0.1 1/degrees C)
+- c5: scaling of the settling rate when no water is present in the snowpack (0, it is 2 when water is present)
+- cx: destructive metamorphism decay factor for densities greater than the critical density ρ_d (23)
+- ρ_d: the critical density (unitless as a fraction of the density of liquid water, 0.15)
+"""
+function Anderson1976(
+    FT::DataType;
```
kmdeck, line 100: I would suggest using https://docs.julialang.org/en/v1.11/base/base/#Base.@kwdef

```

julia> @kwdef struct Anderson1976{FT}
       a::FT = FT(1.0)
       b::FT = FT(2.0)
       end

julia> Anderson1976{Float32}
Anderson1976{Float32}

julia> Anderson1976{Float32}()
Anderson1976{Float32}(1.0f0, 2.0f0)

julia> Anderson1976{Float32}().a
1.0f0
````

File: src/standalone/Snow/Snow.jl [hunk 4]
```diff
@@ -42,6 +42,81 @@ and is used as a bulk snow model.
...
+    c1 = 0.026,
+    c2 = 21.0,
+    c3 = 0.005,
+    c4 = 0.10,
+    c5 = 0.0, #Snow17 code says 1.0, paper says 0.0?
```
kmdeck, line 105: maybe can move comment to the docstring

kmdeck: and elaborate a bit more

src/standalone/Snow/Snow.jl
```diff
@@ -272,8 +368,13 @@ function ClimaLand.make_update_aux(model::SnowModel{FT}) where {FT}
...
 
-        @. p.snow.water_runoff =
-            compute_water_runoff(Y.snow.S, p.snow.q_l, p.snow.T, parameters)
+        @. p.snow.water_runoff = compute_water_runoff(
+            Y.snow.S,
+            p.snow.q_l,
+            p.snow.T,
+            p.snow.ρ_snow,
+            parameters,
+        ) #do we want to dispatch this off the density model type so that it can take Y.snow.Z insetad of p.snow.ρ_snow when it exists to avoid a redundant computation?
```
kmdeck, line 377: what would be better than making this function have multiple methods is to make a helper function:

snow_depth(density_model, Y, p) which returns either Y.snow.Z or computes depth using density, and use this internally to compute_water_runoff. 

Although depth is easy to compute from density, it may be less confusing to do this ^^ than to compute density from Y.snow.z, then compute depth from this...

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -42,7 +42,8 @@ In order to compute surface fluxes, this cannot be large than the
 height of the atmosphere measurement location (z_atmos > z_land always). 
 """
 function ClimaLand.surface_height(model::SnowModel{FT}, Y, p) where {FT}
-    return p.snow.z
+    _ρ_l = FT(LP.ρ_cloud_liq(model.parameters.earth_param_set))
+    return _ρ_l .* Y.snow.S ./ p.snow.ρ_snow
```
kmdeck, line 46: if we had the helper `snow_depth` function, could use that here.

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -194,12 +195,12 @@ function snow_bulk_temperature(
     q_l::FT,
     parameters::SnowParameters{FT},
 ) where {FT}
-    _ρ_i = FT(LP.ρ_cloud_ice(parameters.earth_param_set))
+    _ρ_i = FT(LP.ρ_cloud_ice(parameters.earth_param_set)) #why do we define this if we never use it?
```
kmdeck, line 198: if we never use it, we can remove it from this function. and same for _T_freeze.

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -217,27 +218,28 @@ function snow_liquid_mass_fraction(
...
-    ΔT = _T_freeze - _T_ref
+    #ΔT = _T_freeze - _T_ref #why do we define this if we never use it?
 
     _ρcD_g = parameters.ρcD_g
     Uminus =
         (_ρ_l * SWE * _cp_i + _ρcD_g) * (_T_freeze - _T_ref) -
         _ρ_l * SWE * _LH_f0
     Uplus = (_ρ_l * SWE * _cp_l + _ρcD_g) * (_T_freeze - _T_ref)
     if U < Uminus
-        FT(0)
```
kmdeck, line 236: this returns it, but maybe you think it is clearer to have the return statement?

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -217,27 +218,28 @@ function snow_liquid_mass_fraction(
...
+    #_ρ_i = FT(LP.ρ_cloud_ice(parameters.earth_param_set)) #why do we define this if we never use it?
     _ρ_l = FT(LP.ρ_cloud_liq(parameters.earth_param_set))
     _T_ref = FT(LP.T_0(parameters.earth_param_set))
     _T_freeze = FT(LP.T_freeze(parameters.earth_param_set))
     _LH_f0 = FT(LP.LH_f0(parameters.earth_param_set))
     _cp_i = FT(LP.cp_i(parameters.earth_param_set))
     _cp_l = FT(LP.cp_l(parameters.earth_param_set))
 
-    ΔT = _T_freeze - _T_ref
+    #ΔT = _T_freeze - _T_ref #why do we define this if we never use it?
```
kmdeck, line 230: OK to remove if not used

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -302,12 +304,17 @@ liquid water (runoff) from the snowpack.
...
+    T::FT,
+    ρ_snow::FT,
+    parameters,
+) where {FT}
     _ρ_l = FT(LP.ρ_cloud_liq(parameters.earth_param_set))
-    ρ_snow::FT = parameters.ρ_snow
     Ksat::FT = parameters.Ksat
     Δt::FT = parameters.Δt
-    depth = snow_depth(S, ρ_snow, _ρ_l)
+    depth = snow_depth(S, ρ_snow, _ρ_l) #is it fine to calculate depth like this every time even if Y.snow.Z exists, or should we dispatch this function off the parameterization types to not calculate this if already calculated?
```
kmdeck, line 317: I think the dispatch idea is a little clearer.

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -363,3 +370,202 @@ function energy_from_T_and_swe(S::FT, T::FT, parameters) where {FT}
     end
 
 end
+
+"""
+    compat_density(W_i::FT, sliq::FT, ρ::FT, Δt_t::FT, Tsnow::FT, density::Anderson1976) where {FT}
+Returns the new compressed density ρ_new from the input ρ under the compaction model defined
+by Snow17 for a given `Anderson1976` parameterization configuration.
+"""
+function compat_density(
```
kmdeck, line 379: what does "compat_density" mean?

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -363,3 +370,202 @@ function energy_from_T_and_swe(S::FT, T::FT, parameters) where {FT}
...
+    ρ::FT,
+    Δt_t::FT,
+    Tsnow::FT,
+    density::Anderson1976,
+)::FT where {FT}
+    c5 = (sliq > FT(0)) ? FT(2) : density.c5
+    cx = (ρ > density.ρ_d) ? density.cx : FT(0)
+
+    B =
+        FT(100) *
```
kmdeck, line 391: 100 * a has the same type as a, so no need for FT(100) here. 

Also no need to do sliq > FT(0), can compare directly to zero

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -363,3 +370,202 @@ function energy_from_T_and_swe(S::FT, T::FT, parameters) where {FT}
...
+function newsnow_dens(air_temp::FT)::FT where {FT}
+    if air_temp < FT(-15.0)
+        return FT(0.05)
+    else
+        return FT(0.05) + FT(0.0017) * FT((air_temp + FT(15.0))^1.5)
+    end
+end
+
+"""
+    newsnow_dens(air_temp::FT)::FT where {FT}
```
kmdeck, line 426: can update doc string for the function below

kmdeck: Typically the doc string should first describe what the function does, and then provide details:

"Estimates the temperature of newly fallen snow as a function of air temperature, according to the Snow17 model.

Used in computing the time derivative of snow depth under the `Anderson1976` density parameterization.
"

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -363,3 +370,202 @@ function energy_from_T_and_swe(S::FT, T::FT, parameters) where {FT}
...
+    )
+    factor = factor * A
+
+    ρ_new = minimum([ρ * factor, FT(0.45)])
+    ρ_new = (ρ_new < FT(0.05)) ? ρ : ρ_new
+    return ρ_new
+end
+
+"""
+    newsnow_dens(air_temp::FT)::FT where {FT}
```
kmdeck, line 411: I think here "newsnow_density" is clearer and almost as concise.

src/standalone/Snow/snow_parameterizations.jl
kmdeck: (can propagate to other functions)

note that 0.01*a, if a is Float32, will be Float64

src/standalone/Snow/snow_parameterizations.jl
kmdeck: can change the name to be clearer

experiments/standalone/Snow/snowmip_simulation.jl
```diff
@@ -57,6 +69,8 @@ Y, p, coords = ClimaLand.initialize(model)
 
 # Set initial conditions
 Y.snow.S .= FT(SWE[1]) # first data point
+#Y.snow.Z .= FT(depths[1]) #uncomment if using dens_neural, not dens_constant
+p.snow.ρ_snow .= FT(SWE[1] / depths[1] * 1000.0)
```
kmdeck, line 73: you dont need to set this, it is set in `set_initial_cache!`

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+    z_idx = 1
+    p_idx = 7
+    nfeatures = 7
+    zmodel = ModelTools.make_model(nfeatures, 4, z_idx, p_idx)
+    zmodel_state = BSON.load(IOBuffer(HTTP.get(download_link).body))[:zstate]
+    Flux.loadmodel!(zmodel, zmodel_state)
+    ModelTools.settimescale!(zmodel, 86400.0)
+    return zmodel
+
+    #should I write this function to give the ability to also load from local file too?
```
kmdeck, line 53: our style is more to have the download link download the file, managed by ClimaArtifacts, and then to load from file.

Perhaps you could switch to that style (read from local file), and open an issue in ClimaArtifacts to add this? we can resolve that later, though

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+function NeuralDepthModel(
+    FT::DataType;
+    Δt = nothing,
+    model::Flux.Chain = get_znetwork(),
+    α = nothing,
+)
+    usemodel = converted_model_type(model, FT)
+    weight =
+        !isnothing(α) ? FT(α) :
+        (isnothing(Δt) ? FT(1 / 12) : FT(maximum([2 * Δt / 86400, 1])))
```
kmdeck, line 92: Can you describe more how this is set?

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+    # should we set the timescale of the model to Δt? getting rid of get_znetwork() and below line mitigates the need to import ModelTools entirely.
+    if !isnothing(Δt)
+        ModelTools.settimescale!(usemodel, Δt, dtype = FT) #assumes Δt is provided in seconds
+    end
+    return NeuralDepthModel{FT}(usemodel, weight)
+end
+
+#Define the additional prognostic variables needed for using this parameterization:
+#do I need to export these for them to be utilized in the simulation?
```
kmdeck, line 101: nope! all set

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+ClimaLand.Snow.density_prog_vars(m::NeuralDepthModel) =
+    (:Z, :P_avg, :T_avg, :R_avg, :Qrel_avg, :u_avg)
+ClimaLand.Snow.density_prog_types(m::NeuralDepthModel{FT}) where {FT} =
+    (FT, FT, FT, FT, FT, FT)
+ClimaLand.Snow.density_prog_names(m::NeuralDepthModel) =
+    (:surface, :surface, :surface, :surface, :surface, :surface)
+
+#so far, every type is just blank here - is it even worth providing this feature or just get rid of this feature in Snow.jl entirely?
+ClimaLand.Snow.density_aux_vars(m::NeuralDepthModel) = ()
```
kmdeck, line 110: you can define a single default method in Snow.jl, e.g.
ClimaLand.Snow.density_aux_vars(m::AbstractDensityModel) = ()

and then not define it here - the NeuralDepthModel would fall back on the default if it does not have a more specific method

That said, it *none* of the density models need additional vars, you can just not make this function at all

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+ClimaLand.Snow.density_aux_types(m::NeuralDepthModel{FT}) where {FT} = ()
+ClimaLand.Snow.density_aux_names(m::NeuralDepthModel) = ()
+
+"""
+    snow_density(density::NeuralDepthModel, SWE::FT, z::FT, parameters::SnowParameters{FT}) where {FT}
+Returns the snow density given the current model state and the `NeuralDepthModel`
+density paramterization.
+"""
+function snow_density(
+    density::NeuralDepthModel{FT},
```
kmdeck, line 120: Could you explain why this function is needed? It seems general to any density model.

kmdeck: OK - I see now how it used in Snow.jl...

but that function is called with different arguments in `update_aux!`:
```update_density!(parameters.density, parameters, Y, p)```

so this function has to take the same ones.

Comment (kmdeck) [reply ext/neural_snow/NeuralSnow.jl:120]: i.e. could we replace with:

snow_depth(SWE, z, rho_l) = SWE/z*rho_l

in snow_parameterizations.jl?

File: src/standalone/Snow/snow_parameterizations.jl [hunk 21]
```diff
@@ -363,3 +370,202 @@ function energy_from_T_and_swe(S::FT, T::FT, parameters) where {FT}
...
+        return dZdt
+    end
+end
+
+"""
+    snow_density(density::Anderson1976, SWE::FT, z::FT, parameters::SnowParameters{FT}) where {FT}
+Returns the snow density given the current model state and the `Anderson1976`
+density paramterization.
+"""
+function snow_density(
```
kmdeck, line 494: This is the same function essentially as the snow_density method for NeuralSnowDepth. I dont think we need both?

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+Updates the snow density given the current model state and the `NeuralDepthModel`
+density paramterization.
+"""
+function update_density!(
+    density::NeuralDepthModel,
+    params::SnowParameters,
+    Y,
+    p,
+)
+    p.snow.ρ_snow .= snow_density.(Ref(density), Y.snow.S, Y.snow.Z, params)
```
kmdeck, line 144: no need to dispatch off of density type here

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+    model,
+    z::FT,
+    swe::FT,
+    P::FT,
+    T::FT,
+    R::FT,
+    qrel::FT,
+    u::FT,
+)::FT where {FT}
+    input = FT.([z, swe, qrel, R, u, T, P])
```
kmdeck, line 162: all the inputs are already type FT, so no need to convert them here

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+    sol_rad = abs.(p.drivers.SW_d) #need to change downward direction to scalar
+    rel_hum =
+        Thermodynamics.relative_humidity.(
+            Ref(model.boundary_conditions.atmos.thermo_params),
+            p.drivers.thermal_state,
+        )
+    wind_speed = p.drivers.u
+    dt = model.parameters.Δt
+    β = density.α
+
```
kmdeck, line 229: You can do this:
@. dY.var = beta * (a - b)

to fuse all of the broadcasts into a single broadcast. It is slightly more efficient than doing it for each function individually

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+    air_temp = p.drivers.T .- model.parameters.earth_param_set.T_freeze
+    sol_rad = abs.(p.drivers.SW_d) #need to change downward direction to scalar
+    rel_hum =
+        Thermodynamics.relative_humidity.(
+            Ref(model.boundary_conditions.atmos.thermo_params),
+            p.drivers.thermal_state,
+        )
+    wind_speed = p.drivers.u
+    dt = model.parameters.Δt
+    β = density.α
```
kmdeck, line 228: why the name change?

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+    dY.snow.P_avg = β .* (dprecipdt_snow .- Y.snow.P_avg)
+    dY.snow.T_avg = β .* (air_temp .- Y.snow.T_avg)
+    dY.snow.R_avg = β .* (sol_rad .- Y.snow.R_avg)
+    dY.snow.Qrel_avg = β .* (rel_hum .- Y.snow.Qrel_avg)
+    dY.snow.u_avg = β .* (wind_speed .- Y.snow.u_avg)
+end
+
+end
+
+#do tests need to be made for these functions? since it is an extension?
```
kmdeck, line 240: we have neural snow model extension tests currently:
https://github.com/CliMA/ClimaLand.jl/blob/main/test/standalone/Snow/tool_tests.jl

it would be good to add some tests for these functions

src/standalone/Snow/Snow.jl
a-charbon: You are correct, no outer constructor is needed

src/standalone/Snow/snow_parameterizations.jl
a-charbon: I thought it felt more consistent with the rest of the functions but I agree the change isn't really "changing" anything

src/standalone/Snow/snow_parameterizations.jl
a-charbon: changed to "compact" instead of "compat"

experiments/standalone/Snow/snowmip_simulation.jl
a-charbon: Note to myself to check a model simulation without this line, I think I ran into problems without it

src/standalone/Snow/snow_parameterizations.jl
a-charbon: Note to self to check with you about some of the examples of these

ext/neural_snow/NeuralSnow.jl
a-charbon: Let's chat about this one

ext/neural_snow/NeuralSnow.jl
a-charbon: visual aesthetic, no actual reason

ext/neural_snow/NeuralSnow.jl
```diff
@@ -0,0 +1,240 @@
...
+"""
+function update_density_prog!(
+    density::NeuralDepthModel,
+    model::SnowModel,
+    dY,
+    Y,
+    p,
+)
+    #Get Inputs:
+    #is this too many allocations?
```
kmdeck, line 214: yes, anything with a dot (broadcast) over a field needs to be in an expression that involves x .= ...

so that there are no allocations.

You can do this like:
```
    @. dY.snow.P_avg = β * (abs(p.drivers.P_snow) .- Y.snow.P_avg)
    @. dY.snow.T_avg = β * (p.drivers.T - model.parameters.earth_param_set.T_freeze - Y.snow.T_avg)
    @. dY.snow.R_avg = β * (p.drivers.SW_d - Y.snow.R_avg) # SW_d is positive by definition
    @. dY.snow.Qrel_avg = β * (Thermodynamics.relative_humidity(
            Ref(model.boundary_conditions.atmos.thermo_params),
            p.drivers.thermal_state,
        ) - Y.snow.Qrel_avg)
    @. dY.snow.u_avg = β * (wind_speed - Y.snow.u_avg)
```

the way it is written, it is not possible to do this for dY.snow.Z. But you could add in an aux var
```
 p.snow.predicted_dzdt .=  dzdt(density, Y)
```
which is updated here (if you like) and then compute

```
@. dY.snow.Z = clip_dZdt(swe, z, dswedt, predicted_dzdt, dt)
```

PR #908, opened 2024-11-02
run snowy land for 20 years

juliasloan25:
## Purpose 
run snow land longrun for 20 years

### DIFF
--- a/experiments/long_runs/snowy_land.jl
+++ b/experiments/long_runs/snowy_land.jl
@@ -384,7 +384,7 @@ end
 function setup_and_solve_problem(; greet = false)
 
     t0 = 0.0
-    tf = 60 * 60.0 * 24 * 365 * 20
+    tf = 60 * 60.0 * 24 * 365 * 4
     Δt = 450.0
     nelements = (101, 15)
     if greet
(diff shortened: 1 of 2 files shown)


kmdeck:
I think we need to increase the slurm time in the buildkite pipeline

experiments/long_runs/snowy_land.jl
```diff
@@ -384,7 +384,7 @@ end
 function setup_and_solve_problem(; greet = false)
 
     t0 = 0.0
-    tf = 60 * 60.0 * 24 * 365
+    tf = 60 * 60.0 * 24 * 365 * 4
```
kmdeck, line 387: I realized that we need to change the plotting! Right now it is assuming that we have one year, and are plotting indices [1,4,7,10] (indices refer to months). But we probably want to plot these in the last year.

We can remake the plots without rerunning the sim by looking here:
Using cached git repo /scratch/clima/slurm-buildkite/climaland-long-runs/2288/climaland-long-runs

juliasloan25:
> I realized that we need to change the plotting! Right now it is assuming that we have one year, and are plotting indices [1,4,7,10] (indices refer to months). But we probably want to plot these in the last year.
> 
> We can remake the plots without rerunning the sim by looking here:
> Using cached git repo /scratch/clima/slurm-buildkite/climaland-long-runs/2288/climaland-long-runs

I re-ran it - here are the plots during the fourth year:
bio
water
other

kmdeck:
@juliasloan25 close in lieu of #1039 ?

PR #929, opened 2024-11-22
use ClimaArtifacts Lehmann 2008 evaporation artifact

juliasloan25:
## Purpose 
Use Lehmann et al 2008 evaporation dataset, added to ClimaArtifacts in https://github.com/CliMA/ClimaArtifacts/pull/82

## Content

----

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -179,9 +179,9 @@ function topmodel_data_path(; context = nothing)
 end
 
 """
-    lehmann_assouline_or2008_evaporation_data(; context=nothing)
+    lehmann2008_evaporation_data(; context=nothing)
 
-Local path to file containing measured evaporation rate as a function of time
+Returns the path to file containing measured evaporation rate as a function of time
 for bare soil.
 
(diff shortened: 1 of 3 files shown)


src/Artifacts.jl
```diff
@@ -194,9 +194,9 @@ function topmodel_data_path(; context = nothing)
 end
 
 """
-    lehmann_assouline_or2008_evaporation_data(; context=nothing)
+    lehmann2008_evaporation_data(; context=nothing)
 
-Local path to file containing measured evaporation rate as a function of time
+Returns the path to file containing measured evaporation rate as a function of time
 for bare soil.
```
kmdeck, line 200: Can mention here that it refers to Figure 8 of the paper.

kmdeck (approved):
thank you! Looks good

src/Artifacts.jl
juliasloan25: Further down in the docstring it says "Data was originally collected by ... and presented
in Figure 8 of that work."

PR #931, opened 2024-11-25
Update soil water bc and snow conductivity to match paper

kmdeck:
## Purpose 
This PR makes two changes which bring the soil and snow model in line with the equations we have in the paper.
1. Evaporation/condensation is include in the "influx" term from which we then compute runoff and infiltration. Previously, we computed infiltration/runoff from Precip+snowmelt (if applicable), and then set the BC as infiltration+evap. Now we compute infiltration/runoff from Precip+snowmelt+evap, and set the BC = infiltration. This allows us to handle heavy condensation events also, which I think caused NaN before, and so I also stopped setting Evap and subl to 0 if condensation was occuring.
2. Non-dimensionalize the coefficients for thermal conductivity in snow. If the snow density = ice density, we get kappa_snow = kappa_ice exactly

### DIFF
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -481,19 +481,22 @@ function soil_boundary_fluxes!(
 ) where {FT}
     bc = soil.boundary_conditions.top
     p.soil.turbulent_fluxes .= turbulent_fluxes(bc.atmos, soil, Y, p, t)
-    Soil.Runoff.update_runoff!(
-        p,
-        bc.runoff,
-        p.drivers.P_liq .+ p.snow.water_runoff .* p.snow.snow_cover_fraction .+
-        p.excess_water_flux,
-        Y,
-        t,
-        soil,
-    )
-    @. p.soil.top_bc.water =
-        p.soil.infiltration +
+    # influx = maximum possible rate of infiltration given precip, snowmelt, evaporation/condensation
+    # but if this exceeds infiltration capacity of the soil, runoff will
+    # be generated.
+    # Use top_bc.water as temporary storage to avoid allocation
+    influx = p.soil.top_bc.water
+    @. influx =
+        p.drivers.P_liq +
+        p.snow.water_runoff * p.snow.snow_cover_fraction +
+        p.excess_water_flux +
         (1 - p.snow.snow_cover_fraction) *
         p.soil.turbulent_fluxes.vapor_flux_liq
+    # The update_runoff! function computes how much actually infiltrates
+    # given influx and our runoff model bc.runoff, and updates
+    # p.soil.infiltration in place
+    Soil.Runoff.update_runoff!(p, bc.runoff, influx, Y, t, soil)
+    @. p.soil.top_bc.water = p.soil.infiltration
     @. p.soil.top_bc.heat =
         (1 - p.snow.snow_cover_fraction) * (
             p.soil.R_n +
--- a/src/standalone/Snow/snow_parameterizations.jl
+++ b/src/standalone/Snow/snow_parameterizations.jl
@@ -165,15 +165,23 @@ Computes the thermal conductivity, given the density
 of the snow, according to Equation
 5.33 from Bonan's textbook, which in turn is taken from
 Jordan (1991).
+
+We have adjusted the original equation to make the coefficients
+non-dimensional by multiplying by the first by x = ρ_ice/ρ_ice
+and the second by x², with ρ_ice in kg/m³.
+
+When ρ_snow = ρ_ice, we recover κ_snow = κ_ice.
 """
 function snow_thermal_conductivity(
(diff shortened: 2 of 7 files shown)


juliasloan25:
> This allows us to handle heavy condensation events also, which I think caused NaN before, and so I also stopped setting Evap and subl to 0 if condensation was occuring.

Do you think this is related to the set of NaNs we were debugging yesterday, or is it separate?

src/integrated/land.jl
```diff
@@ -481,19 +481,12 @@ function soil_boundary_fluxes!(
...
-        p.excess_water_flux,
-        Y,
-        t,
-        soil,
-    )
-    @. p.soil.top_bc.water =
-        p.soil.infiltration +
-        (1 - p.snow.snow_cover_fraction) *
-        p.soil.turbulent_fluxes.vapor_flux_liq
+    influx = @. p.drivers.P_liq +
```
juliasloan25, line 484: Is influx here different from infiltration? It seems like it is the water available at the surface that may infiltrate into the soil, and it gets separated into runoff and infiltration, but I'm not 100% sure. It would be good to add a comment clarifying

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -170,9 +170,11 @@ function snow_thermal_conductivity(
     parameters::SnowParameters{FT},
 ) where {FT}
     _κ_air = FT(LP.K_therm(parameters.earth_param_set))
+    _ρ_ice = FT(LP.ρ_cloud_ice(parameters.earth_param_set))
     κ_ice = parameters.κ_ice
     return _κ_air +
-           (FT(7.75e-5) * ρ_snow + FT(1.105e-6) * ρ_snow^2) * (κ_ice - _κ_air)
+           (FT(0.07) * (ρ_snow / _ρ_ice) + FT(0.93) * (ρ_snow / _ρ_ice)^2) *
```
juliasloan25, line 176: Are the coefficients here from Bonan's book? And do they change in this PR because the dimensionality of the thermal conductivity changed?

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -170,9 +170,11 @@ function snow_thermal_conductivity(
     parameters::SnowParameters{FT},
 ) where {FT}
     _κ_air = FT(LP.K_therm(parameters.earth_param_set))
+    _ρ_ice = FT(LP.ρ_cloud_ice(parameters.earth_param_set))
```
juliasloan25, line 173: update docstring too

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -987,8 +987,6 @@ function soil_turbulent_fluxes_at_a_point(
         )
         x::FT = 4 * K_sfc * (1 + Ẽ0 / (4 * K_c))
         Ẽ *= x / (Ẽ0 + x)
-    else
-        Ẽ *= 0 # condensation, set to zero
```
juliasloan25, line 991: Is it okay to leave evap/sublimation unchanged if condensation is occurring? Maybe in this case they'll tend towards 0 in the equations solved in `evaporation` anyway?

src/standalone/Snow/snow_parameterizations.jl
kmdeck: The original coefficients were from Bonan's book, yes!  but it's not ideal to have coefficients with units  (the first would be per unit density, the second per unit density squared, but it might not be clear if these are kg/m^3 or something else for density). so I basically multiplied 7.75e-5 rho_snow by rho_ice/rho_ice, or 0.07 = 917*7.75e-5.

Also, it is clear now that if you plug in rho_snow = rho_ice, you get back k_air + (0.07+0.93)*(k_ice - k_air) = k_ice.

src/integrated/land.jl
kmdeck: I added comments in all 4 files with this.

src/standalone/Snow/snow_parameterizations.jl
kmdeck: I updated the doc string

kmdeck:
> > This allows us to handle heavy condensation events also, which I think caused NaN before, and so I also stopped setting Evap and subl to 0 if condensation was occuring.
> 
> Do you think this is related to the set of NaNs we were debugging yesterday, or is it separate?

I think it is seperate, because I turned off condensation in the current version of main (this PR would turn it back on)

src/standalone/Soil/energy_hydrology.jl
kmdeck: let's discuss in person!

juliasloan25 (approved):

PR #938, opened 2024-12-03
use soil grids parameters

kmdeck:
## Purpose 
Use the new soilgrids parameters in place of the constant values we were using previously

## To-do
double check long runs

## Content
Added soilgrids artifact
Use artifact in spatially varying parameters (default)

----

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -60,6 +60,26 @@ function soil_params_artifact_folder_path(; context = nothing)
 end
 
 """
+    soil_grids_params_artifact_path(; lowres = true, context)
+
+Return the path to the file that contains the soil texture parameters
+needed for the Balland and Arp (2005) thermal conductivity model.
+
+Returns a ~1 degree version by default (lowres = true).
+"""
+function soil_grids_params_artifact_path(; context = nothing, lowres = true)
+    if lowres
+        dir = @clima_artifact("soilgrids_lowres", context)
+        file = "soil_solid_vol_fractions_soilgrids_lowres.nc"
+        return joinpath(dir, file)
+    else
+        dir = @clima_artifact("soilgrids", context)
+        file = "soil_solid_vol_fractions_soilgrids.nc"
+        return joinpath(dir, file)
+    end
+end
+
+"""
     experiment_fluxnet_data_path(
         site_ID;
         context = nothing,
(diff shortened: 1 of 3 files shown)


src/Artifacts.jl
```diff
@@ -59,6 +59,25 @@ function soil_params_artifact_folder_path(; context = nothing)
...
+    context = nothing,
+    lowres = true,
+)
+    if lowres
+        return @clima_artifact("soilgrids_lowres", context)
+    else
+        return @clima_artifact("soilgrids", context)
+    end
+end
+
```
imreddyTeja, line 80: If there is only one file in each artifact, we should probably follow the existing pattern and make this 

`soil_grids_params_artifact_data()` or `soil_grids_params_data_path`

and directly return the path to the lowres or full-resolution artifact.  See the `topmodel_data_path` and `lehmann2008_evaporation_data` functions.

src/simulations/spatial_parameters.jl
```diff
@@ -280,39 +281,59 @@ function default_spatially_varying_soil_parameters(
...
+    )
+    # we require that the sum of these be less than 1 and equal to or bigger than zero.
+    # The input should satisfy this almost exactly, but the regridded values may not.
+    function texture_sum_norm(ν_ss_gravel, ν_ss_quartz, ν_ss_om)
+        texture_sum = ν_ss_gravel + ν_ss_quartz + ν_ss_om 
+        texture_norm = texture_sum <= 1 ? 1 : texture_sum
+    end
+    @. ν_ss_gravel = ν_ss_gravel / texture_sum_norm(ν_ss_gravel, ν_ss_quartz, ν_ss_om)
+    @. ν_ss_om = ν_ss_om / texture_sum_norm(ν_ss_gravel, ν_ss_quartz, ν_ss_om)
+    @. ν_ss_quartz = ν_ss_quartz / texture_sum_norm(ν_ss_gravel, ν_ss_quartz, ν_ss_om)
```
imreddyTeja, line 335: If the sum of the three relative volumes is greater than 1, then value returned by `texture_sum_norm(ν_ss_gravel, ν_ss_quartz, ν_ss_om)` will be different for each of the 3 calls, and the rescaling will be less for `ν_ss_quartz` than for `ν_ss_om`, which will be less than `ν_ss_gravel`. The final result will then still sum to greater than 1. Is this intentional?

kmdeck: good catch!! I need to compute the sum and store it, and then use it for all three computations.

src/Artifacts.jl
kmdeck: ok! I fixed this

imreddyTeja (approved):

PR #939, opened 2024-12-04
use ARS222; run longruns for 4 years

kmdeck:
## Purpose 
This PR switches all of our experiments and tutorials to use ARS222. They were previously using ARS111, but this fails to converge in some cases, which we first discovered in multi-year global runs. For example, see the following plot where the solver fails to converge and instead oscillates unphysically before breaking.

Note that two runs (soil_canopy_tutorial.jl and global_soil_canopy.jl) were previously using ARS343 and are unified here with the other runs to use ARS222.

This PR also extends the longruns to run for 4 years, so we can see the behavior in the time range where we've been seeing NaNs.

### DIFF
--- a/.buildkite/longruns_gpu/pipeline.yml
+++ b/.buildkite/longruns_gpu/pipeline.yml
@@ -37,7 +37,7 @@ steps:
           - "snowy_land_longrun_gpu/*pdf"
         agents:
           slurm_gpus: 1
-          slurm_time: 15:00:00
+          slurm_time: 20:00:00
         env:
           CLIMACOMMS_DEVICE: "CUDA"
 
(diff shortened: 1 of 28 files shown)


juliasloan25:
Many SoilCanopy runs in [buildkite](https://buildkite.com/clima/climaland-ci/builds/6217) become unstable using ARS222. Some are run with `dt=450s` and others `dt=900s`. These were all previously passing with ARS111, so this doesn't bode well for ARS222.

kmdeck:
> Many SoilCanopy runs in [buildkite](https://buildkite.com/clima/climaland-ci/builds/6217) become unstable using ARS222. Some are run with `dt=450s` and others `dt=900s`. These were all previously passing with ARS111, so this doesn't bode well for ARS222.

ohhhh no. I wonder why. the canopy should be simpler! might be worth thinking more about?

PR #943, opened 2024-12-05
Add progress log in snowy_land long run

Sbozzolo:
Print out simulation progress every 1000 iterations (for snowy land)

Now printing:
```
┌ Info: Progress
│   simulation_time = "1 day, 1 hour"
│   n_steps_completed = 200
│   wall_time_per_step = "118 milliseconds, 370 microseconds"
│   wall_time_total = "2 hours, 18 minutes"
│   wall_time_remaining = "2 hours, 17 minutes"
│   wall_time_spent = "23 seconds, 674 milliseconds"
│   percent_complete = "0.3%"
│   estimated_sypd = "10.408"
│   date_now = 2024-12-09T15:04:38.224
└   estimated_finish_date = 2024-12-09T17:22:30.224
```

The first two reports report 0 because the code tries to not include compilation time. I think it should be possible to improve this and get all reports with some numbers, but this is left for future work


kmdeck (approved):
looks great! Thanks for adding this.

My only comment is that we might want to make this less frequent. Right now, every 1000 steps is every 5 days, which is somewhat arbitrary. Maybe every week, or every 30 days?

Sbozzolo:
> looks great! Thanks for adding this.
> 
> My only comment is that we might want to make this less frequent. Right now, every 1000 steps is every 5 days, which is somewhat arbitrary. Maybe every week, or every 30 days?

Yes, my plan is to do something like atmos (where it "ramps up" initially, but then becomes every X % of the simulation). But for this, I would like to add a little bit more in ClimaUtilities, which will take a little while

imreddyTeja (commented):
On the long runs build kite, the most recent runs of snowy_land on other branches do not throw a circular dependency warning.  The snow_land runs from this branch throw several of the circular dependency warnings.

Sbozzolo:
> On the long runs build kite, the most recent runs of snowy_land on other branches do not throw a circular dependency warning. The snow_land runs from this branch throw several of the circular dependency warnings.

One possible reason for this is that other branches were run before the recent update to CUDA that leads to the circular depedency warning

imreddyTeja (approved):

PR #947, opened 2024-12-10
Replace Formatting with Format

### LINKED ISSUE #937: replace Formatting.jl with Format.jl
## Describe the bug
Formatting.jl is no longer maintained and should be replaced. See the warning from our doc build below:
```
┌ Formatting
│  ┌ Warning: DEPRECATION NOTICE
│  │ 
│  │ Formatting.jl has been unmaintained for a while, with some serious
│  │ correctness bugs compromising the original purpose of the package. As a result,
│  │ it has been deprecated - consider using an alternative, such as
│  │ `Format.jl` (https://github.com/JuliaString/Format.jl) or the `Printf` stdlib directly.
│  │ 
│  │ If you are not using Formatting.jl as a direct dependency, please consider
│  │ opening an issue on any packages you are using that do use it as a dependency.
│  │ From Julia 1.9 onwards, you can query `]why Formatting` to figure out which
│  │ package originally brings it in as a dependency.
│  └ @ Formatting ~/.julia/packages/Formatting/3VxOt/src/Formatting.jl:12
```

imreddyTeja:
## Purpose 
Formatter.jl is no longer maintained. This PR replaces its usage with Format.jl
Closes #937 

## Content

----


juliasloan25 (commented):
Thanks Teja! I wonder if we even need to keep the Format.jl dependency. It looks like it's only used in one place for [met_drivers_FLUXNET.jl](https://github.com/CliMA/ClimaLand.jl/blob/035b8113741c4772bb943ecfdc6195e7470134f6/experiments/integrated/fluxnet/met_drivers_FLUXNET.jl#L23) and the corresponding lib/ClimaLandSimulations file [make_drivers.jl](https://github.com/CliMA/ClimaLand.jl/blob/035b8113741c4772bb943ecfdc6195e7470134f6/lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/make_drivers.jl#L12).

Dates.jl also has a function `Dates.format` that might do what we need. Or if we're only keeping Format around for one function, we could potentially implement it ourselves. Maybe @AlexisRenchon or @Espeer5 will know if we added Format.jl just for the two calls linked above?

Espeer5:
> Thanks Teja! I wonder if we even need to keep the Format.jl dependency. It looks like it's only used in one place for [met_drivers_FLUXNET.jl](https://github.com/CliMA/ClimaLand.jl/blob/035b8113741c4772bb943ecfdc6195e7470134f6/experiments/integrated/fluxnet/met_drivers_FLUXNET.jl#L23) and the corresponding lib/ClimaLandSimulations file [make_drivers.jl](https://github.com/CliMA/ClimaLand.jl/blob/035b8113741c4772bb943ecfdc6195e7470134f6/lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/make_drivers.jl#L12).
> 
> Dates.jl also has a function `Dates.format` that might do what we need. Or if we're only keeping Format around for one function, we could potentially implement it ourselves. Maybe @AlexisRenchon or @Espeer5 will know if we added Format.jl just for the two calls linked above?

Hmm it's been a while since I wrote this. I think the problem was that the data would automatically be read in as a number and formatting as "##e##" or something like this... I will make a separate issue to work on this as I also want to generalize these scripts to read a specified start date so that the ClimaArtifact for the fluxtower data can be updated to include the full data sets. In the process I will see if I can remove the dependency on Format.

juliasloan25:
> > Thanks Teja! I wonder if we even need to keep the Format.jl dependency. It looks like it's only used in one place for [met_drivers_FLUXNET.jl](https://github.com/CliMA/ClimaLand.jl/blob/035b8113741c4772bb943ecfdc6195e7470134f6/experiments/integrated/fluxnet/met_drivers_FLUXNET.jl#L23) and the corresponding lib/ClimaLandSimulations file [make_drivers.jl](https://github.com/CliMA/ClimaLand.jl/blob/035b8113741c4772bb943ecfdc6195e7470134f6/lib/ClimaLandSimulations/src/Fluxnet/fluxnet_utilities/make_drivers.jl#L12).
> > Dates.jl also has a function `Dates.format` that might do what we need. Or if we're only keeping Format around for one function, we could potentially implement it ourselves. Maybe @AlexisRenchon or @Espeer5 will know if we added Format.jl just for the two calls linked above?
> 
> Hmm it's been a while since I wrote this. I think the problem was that the data would automatically be read in as a number and formatted as "##e##" or something like this... I will make a separate issue to work on this as I also want to generalize these scripts to read a specified start date so that the ClimaArtifact for the fluxtower data can be updated to include the full data sets. In the process I will see if I can remove the dependency on Format.

That sounds like a good plan! In the meantime, we can go ahead with this PR to at least remove the docs warning about Formatting.jl

juliasloan25 (approved):

imreddyTeja:
> > Hmm it's been a while since I wrote this. I think the problem was that the data would automatically be read in as a number and formatted as "##e##" or something like this... I will make a separate issue to work on this as I also want to generalize these scripts to read a specified start date so that the ClimaArtifact for the fluxtower data can be updated to include the full data sets. In the process I will see if I can remove the dependency on Format.

I briefly looked into it when replacing Formatting with Format. It seems like format is used because to parse something in the `yyyymmddHHMM` form, Dates needs a string. The string representation of large floats in is scientific notation. I think we could get around this by converting the dates that are read in to ints before converting them to strings. 

For example:

```
DateTime.(format.(driver_data[2:end, 1]), "yyyymmddHHMM")
```

could be

```
DateTime.(string.(Int.(driver_data[2:end, 1])), "yyyymmddHHMM")
```

This should be fine because the original would still error if a non-whole number is given.

PR #952, opened 2024-12-13
Update jacobian to include energy from water fluxes

juliasloan25:
## Purpose 
To improve the stability of our longruns, we want to include the `dT_rhoe/dvartheta_l` term in our jacobian approximation. We have found that, in the case of high water flux near saturation, the `dpsi/dtheta` term that this depends on becomes very large and unstable to step explicitly. To demonstrate that the water flux energy is the culprit, we tried removing this term from the explicit tendency and saw improved stability.

See #953 for plots justifying the need for this change

## To-do
- [x] update jacobian construction to accommodate off-diagonal terms
- [x] update jacobian calculation to include `dT_rhoe/dvartheta_l` term for `EnergyHydrology` model

### DIFF
--- a/experiments/standalone/Soil/energy_hydrology_saturated.jl
+++ b/experiments/standalone/Soil/energy_hydrology_saturated.jl
@@ -1,0 +1,133 @@
+import SciMLBase
+using Statistics
+import ClimaTimeSteppers as CTS
+using ClimaCore
+import ClimaParams as CP
+using ClimaLand
+using ClimaLand.Domains: Column
+using ClimaLand.Soil
+import ClimaLand
+import ClimaLand.Parameters as LP
+
+FT = Float32;
+earth_param_set = LP.LandParameters(FT)
+
+vg_n = FT(2.9)
+vg_α = FT(6)
+K_sat = FT(4.42 / 3600 / 100) # m/s
+hcm = vanGenuchten{FT}(; α = vg_α, n = vg_n)
+ν = FT(0.43)
+θ_r = FT(0.045)
+S_s = FT(1e-3)
+ν_ss_om = FT(0.0)
+ν_ss_quartz = FT(1.0)
+ν_ss_gravel = FT(0.0)
+
+params = ClimaLand.Soil.EnergyHydrologyParameters(
+    FT;
+    ν,
+    ν_ss_om,
+    ν_ss_quartz,
+    ν_ss_gravel,
+    hydrology_cm = hcm,
+    K_sat,
+    S_s,
+    θ_r,
+    earth_param_set,
+);
+
+surface_water_flux = WaterFluxBC((p, t) -> -K_sat / 10)
+bottom_water_flux = WaterFluxBC((p, t) -> 0.0);
+
+surface_heat_flux = HeatFluxBC((p, t) -> 0.0)
+bottom_heat_flux = HeatFluxBC((p, t) -> 0.0);
+
+boundary_fluxes = (;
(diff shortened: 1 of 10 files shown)


experiments/standalone/Soil/energy_hydrology_saturated.jl
```diff
@@ -0,0 +1,136 @@
...
+set_initial_cache!(p, Y, t0);
+sol_dt_small = SciMLBase.solve(prob, ode_algo; dt = 100.0);
+init_soil!(Y, z, soil.parameters);
+set_initial_cache!(p, Y, t0);
+sol_dt_large = SciMLBase.solve(prob, ode_algo; dt = 900.0);
+@test sum(isnan.(sol_dt_large.u[end])) == 0
+
+norm(x) = sqrt(mean(parent(x .^ 2)))
+rmse(x, y) = sqrt(mean(parent((x .- y) .^ 2)))
+@test rmse(sol_dt_small[end].soil.ϑ_l, sol_dt_large[end].soil.ϑ_l) /
```
kmdeck, line 133: @test vs @assert?

kmdeck (approved):

experiments/standalone/Soil/energy_hydrology_saturated.jl
juliasloan25: maybe it's better to use assert, and then we can remove the Test dependency in this file

src/shared_utilities/implicit_timestepping.jl
```diff
@@ -156,12 +171,22 @@ function ImplicitEquationJacobian(Y::ClimaCore.Fields.FieldVector)
...
+        explicit_blocks...,
+    )
 
-
-    matrix = MatrixFields.FieldMatrix(implicit_blocks..., explicit_blocks...)
-
-    # Set up block diagonal solver for block Jacobian
-    alg = MatrixFields.BlockDiagonalSolve()
+    # Choose algorithm based on whether off-diagonal blocks are present
+    # TODO change this to use dispatch on model type
```
juliasloan25, line 181: Can we delete this TODO for now and just keep track of it in an issue? (I'll open one)

experiments/standalone/Soil/energy_hydrology_saturated.jl
```diff
@@ -0,0 +1,136 @@
...
+)
+Y, p, cds = initialize(soil);
+
+set_initial_cache! = make_set_initial_cache(soil)
+exp_tendency! = make_exp_tendency(soil)
+imp_tendency! = make_imp_tendency(soil);
+jacobian! = ClimaLand.make_jacobian(soil);
+jac_kwargs = (; jac_prototype = ImplicitEquationJacobian(Y), Wfact = jacobian!);
+
+
```
juliasloan25, line 73:
```suggestion
```

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -373,6 +374,28 @@ function ClimaLand.make_compute_jacobian(model::EnergyHydrology{FT}) where {FT}
...
+                        hydrology_cm,
+                        Y.soil.ϑ_l,
+                        ν - Y.soil.θ_i, #ν_eff
+                        θ_r,
+                        S_s,
+                    ),
+                )
+            ) - (I,)
+
+
```
juliasloan25, line 397:
```suggestion
```

src/shared_utilities/implicit_timestepping.jl
kmdeck: thank you!

PR #957, opened 2024-12-16
Remove Plots from buildkite

kmdeck:
## Purpose 
- Switches to CairoMakie for the rest of our experiments, and removes Plots from Project toml and manifest
- alters some parameters of the US-MOz site 
- removes an allocation from a previous PR in soil_canopy_root_interactions
- removes bug in previous fluxnet plotting which would make column as absent if < 0.5% was missing. Now we increase the threshold to 10%.

----


Espeer5 (approved):
Looks good to me!

PR #958, opened 2024-12-16
Misc fixes to canopy model, diagnostics, and simulations

kmdeck:
## Purpose 
Improve the canopy model, implement a diagnostic, and fix a bug in the timestepper of a few canopy alone simulations

## Content 
1. Change to canopy model: use the harmonic mean of K and areas at the interface/face between two centers (how we interpolate center to face). This ensures that if either layer has zero K or zero area, that the flux between them is zero. Previously, we were using the harmonic mean of K and the area of the top layer instead of a mean. Using the arithmetic mean does not give the correct behavior if K or area is zero. [this is in line with the paper, now]
2. Use the implicit tendency in canopy alone runs (that this was not happening is a bug)
3. Output the leaf water potential as a diagnostic.

----

### DIFF
--- a/src/diagnostics/default_diagnostics.jl
+++ b/src/diagnostics/default_diagnostics.jl
@@ -117,7 +117,7 @@ function default_diagnostics(
             "trans",
             "clhf",
             "cshf",
-            # "lwp", # last(p.canopy.hydraulics.ψ) errors
+            "lwp",
             # "fa", # return a Tuple
             "far",
             "lai",
@@ -264,7 +264,7 @@ function default_diagnostics(
             "trans",
             "clhf",
             "cshf",
-            # "lwp", # last(p.canopy.hydraulics.ψ) errors
+            "lwp",
             # "fa", # return a Tuple
             "far",
             "lai",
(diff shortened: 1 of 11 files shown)


src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -595,22 +604,15 @@ function make_compute_exp_tendency(
...
-            # "AI/max(AI x dz, eps(FT))"
-            AIdz =
-                max.(
-                    getproperty(area_index, model.compartment_labels[i]) .* (
-                        model.compartment_surfaces[ip1] -
-                        model.compartment_surfaces[i]
-                    ),
-                    eps(FT),
-                )
+            # To prevent dividing by zero, change AI/(AI x dz)" to AI/max(AI x dz, eps(FT))"
```
kmdeck, line 607: These changes were for readability

kmdeck: this area index should be the "center" area index, not the face, so no interp c 2 f needed

src/diagnostics/default_diagnostics.jl
```diff
@@ -264,7 +264,7 @@ function default_diagnostics(
             "trans",
             "clhf",
             "cshf",
-            # "lwp", # last(p.canopy.hydraulics.ψ) errors
+            "lwp", # last(p.canopy.hydraulics.ψ) errors
```
juliasloan25, line 267: do we still need the comment here?

juliasloan25 (approved):

src/diagnostics/default_diagnostics.jl
kmdeck: I fixed this!

PR #961, opened 2024-12-19
use dt=600s, tf=4years in longruns

juliasloan25:
## Purpose 
After the most recent jacobian update, the longruns are stable after 4 years using dt=400s, but unstable using dt=900s. This PR tests out an intermediate dt value and also extends the length of longruns to use 4 years.


kmdeck:
@juliasloan25 could we close this? I think we can try 600 s again, but on current main.

juliasloan25:
> @juliasloan25 could we close this? I think we can try 600 s again, but on current main.

Yeah, let's close it!

PR #965, opened 2024-12-20
Kd/rain on snow

kmdeck:
## Purpose 
Snow is made of liquid water and ice. Previously, we tracked the total water content in the snow via snow water equivalent (symbol S), which is the volume of liquid water you would get if the snow melted, per ground area. The liquid water mass fraction (symbol q_l) was obtained diagnostically and could only be nonzero if the snow was at the freezing temperature. 

This approach handled melting of snow indirectly (conversion of ice to liquid water occured when we computed q_l diagnostically, based only on the total energy per mass of the snow), and then the loss of liquid water from the snowpack ("runoff" ) was included via a term in the ODE for S which corresponded to damping q_l to zero.

This worked fine, but it made it hard to address how rain falling on snow affects the liquid water content of the snow. We ignored this, and all liquid precip went straight to the soil (equivalent to saying it percolated through the snow instantaneously).

The equation for S was:
`dS/dt = P_snow + snow_cover_fraction(Vapor Flux - Runoff)`

This PR adds a prognostic variable for snow liquid water (symbol S_l = q_l S). This allows us to better account for rain falling on snow. The equations we have are now:
```
dS/dt = P_snow + snow_cover_fraction(P_liq + Vapor Flux - Runoff)
dS_l/dt = (P_liq + Vapor Flux * q_l - Runoff + Snow Melt) * snow cover fraction
```
Where we assume that q_l * Vapor flux = evaporation of liquid water and the snow melt term has to be parameterized.

Note that now q_l can be nonzero even if T < T_freeze, but that it will decrease in time due to refreezing.

I ran the long runs - 88-> 92 NaNs in SWE @juliasloan25 

----

### DIFF
--- a/experiments/standalone/Snow/snowmip_simulation.jl
+++ b/experiments/standalone/Snow/snowmip_simulation.jl
@@ -42,7 +42,7 @@ ClimaComms.init(context)
 include(
     joinpath(climaland_dir, "experiments/standalone/Snow/process_snowmip.jl"),
 )
-savedir = generate_output_path("experiments/standalone/Snow/")
+savedir = generate_output_path("experiments/standalone/Snow/$(SITE_NAME)")
 t0 = FT(0.0)
 tf = FT(seconds[end])
 ndays = (tf - t0) / 3600 / 24
@@ -51,8 +51,7 @@ ndays = (tf - t0) / 3600 / 24
 domain = ClimaLand.Domains.Point(; z_sfc = FT(0))
 
 #density_model = NeuralSnow.NeuralDepthModel(FT)
-density_model = Snow.ConstantDensityModel(ρ)
-depths = z[snow_data_avail]
+density_model = Snow.MinimumDensityModel(ρ)
 
 parameters = SnowParameters{FT}(
     Δt;
@@ -69,7 +68,8 @@ Y, p, coords = ClimaLand.initialize(model)
 
 # Set initial conditions
 Y.snow.S .= FT(SWE[1]) # first data point
-#Y.snow.Z .= FT(depths[1]) #uncomment if using NeuralDepthModel instead of ConstantDensityModel
+Y.snow.S_l .= 0 # this is a guess
+#Y.snow.Z .= FT(depths[1]) #uncomment if using NeuralDepthModel instead of MinimumDensityModel
 Y.snow.U .=
     ClimaLand.Snow.energy_from_q_l_and_swe(FT(SWE[1]), FT(0), parameters) # with q_l = 0
 
@@ -85,7 +85,6 @@ ode_algo = CTS.IMEXAlgorithm(
         update_j = CTS.UpdateEvery(CTS.NewTimeStep),
     ),
 );
-
 
 prob = SciMLBase.ODEProblem(
     CTS.ClimaODEFunction(
@@ -128,12 +127,16 @@ evaporation = [
 R_n = [parent(sv.saveval[k].snow.R_n)[1] for k in 1:length(sol.t)];
 water_runoff =
     [parent(sv.saveval[k].snow.water_runoff)[1] for k in 1:length(sol.t)];
+phase_change_flux =
+    [parent(sv.saveval[k].snow.phase_change_flux)[1] for k in 1:length(sol.t)];
 rain = [parent(sv.saveval[k].drivers.P_liq)[1] for k in 1:length(sv.t)];
 snow = [parent(sv.saveval[k].drivers.P_snow)[1] for k in 1:length(sv.t)];
 scf =
     [parent(sv.saveval[k].snow.snow_cover_fraction)[1] for k in 1:length(sol.t)];
-
+ρ = [parent(sv.saveval[k].snow.ρ_snow)[1] for k in 1:length(sol.t)];
+z = [parent(sv.saveval[k].snow.z_snow)[1] for k in 1:length(sol.t)];
 S = [parent(sol.u[k].snow.S)[1] for k in 1:length(sol.t)];
+S_l = [parent(sol.u[k].snow.S_l)[1] for k in 1:length(sol.t)];
 U = [parent(sol.u[k].snow.U)[1] for k in 1:length(sol.t)];
 t = sol.t;
 
@@ -142,10 +145,10 @@ days = start_day .+ floor.(t ./ 3600 ./ 24)
 doys = days .% 365 # doesn't account for leap year
 
 obs_swes = Vector{Union{Float64, Missing}}(missing, length(doys))
-obs_swes[snow_data_avail] .= mass[snow_data_avail] ./ 1000
+obs_swes[mass_data_avail] .= mass[mass_data_avail] ./ 1000
 
 obs_tsnows = Vector{Union{Float64, Missing}}(missing, length(doys))
-obs_tsnows[snow_data_avail] = T_snow .+ 273.15
+obs_tsnows[mass_data_avail] = T_snow .+ 273.15
 
 obs_df = DataFrame(
     doy = doys,
@@ -171,10 +174,12 @@ ax1 = CairoMakie.Axis(fig[1, 1], ylabel = "Water Vol/Area ground")
 
 xlims!(ax1, 0, ndays)
 CairoMakie.hidexdecorations!(ax1, ticks = false)
-CairoMakie.lines!(ax1, daily, S, label = "Model")
+CairoMakie.lines!(ax1, daily, S, label = "Model SWE")
+CairoMakie.lines!(ax1, daily, S_l, label = "Model SWE_l")
+
 CairoMakie.scatter!(
     ax1,
-    seconds[snow_data_avail] ./ 24 ./ 3600,
+    seconds[mass_data_avail] ./ 24 ./ 3600,
     SWE,
     label = "Data",
     color = :red,
@@ -215,15 +220,15 @@ CairoMakie.hidexdecorations!(ax1, ticks = false)
 CairoMakie.lines!(ax1, daily, T, label = "Model")
 CairoMakie.scatter!(
     ax1,
-    seconds[snow_data_avail] ./ 24 ./ 3600,
+    seconds[mass_data_avail] ./ 24 ./ 3600,
     T_snow .+ 273.15,
     label = "Snow T",
     color = :red,
 )
 CairoMakie.scatter!(
     ax1,
-    seconds[snow_data_avail] ./ 24 ./ 3600,
-    Tair[snow_data_avail],
+    seconds[mass_data_avail] ./ 24 ./ 3600,
+    Tair[mass_data_avail],
     label = "Atmosphere T",
     color = :orange,
 )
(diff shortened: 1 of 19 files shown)


experiments/standalone/Snow/snowmip_simulation.jl
```diff
@@ -52,10 +52,10 @@ domain = ClimaLand.Domains.Point(; z_sfc = FT(0))
 
 #density_model = NeuralSnow.NeuralDepthModel(FT)
 density_model = Snow.ConstantDensityModel(ρ)
-depths = z[snow_data_avail]
```
kmdeck, line 55: this was already computed in process_snowmip.jl

src/standalone/Snow/Snow.jl
```diff
@@ -313,13 +322,12 @@ function ClimaLand.make_update_aux(model::SnowModel{FT}) where {FT}
     function update_aux!(p, Y, t)
         parameters = model.parameters
 
+        @. p.snow.q_l = min(max(Y.snow.Sl / max(Y.snow.S, eps(FT)), 0), 1)
+
         update_density!(parameters.density, parameters, Y, p)
```
kmdeck, line 327: density depends on q_l now, but we can re think this

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -208,9 +207,9 @@ function snow_bulk_temperature(
     _T_ref = FT(LP.T_0(parameters.earth_param_set))
     _LH_f0 = FT(LP.LH_f0(parameters.earth_param_set))
     cp_s = specific_heat_capacity(q_l, parameters)
-    _ρcD_g = parameters.ρcD_g
+    _ΔS = parameters.ΔS
     return _T_ref +
-           (U + (1 - q_l) * _LH_f0 * _ρ_l * SWE) / (_ρ_l * SWE * cp_s + _ρcD_g)
+           (U + (1 - q_l) * _LH_f0 * _ρ_l * SWE) / (_ρ_l * cp_s * (SWE + _ΔS))
```
kmdeck, line 212: The only place we need dS, to prevent dividing by a small number.

src/standalone/Snow/Snow.jl
```diff
@@ -346,42 +353,56 @@ function ClimaLand.make_update_boundary_fluxes(model::SnowModel{FT}) where {FT}
...
+        @. p.snow.phase_change_flux = phase_change_flux(
+            Y.snow.U,
+            Y.snow.S,
+            Y.snow.S_l,
+            p.snow.applied_energy_flux,
+            model.parameters,
+        )
+        @. p.snow.liquid_water_flux +=
+            p.snow.phase_change_flux * p.snow.snow_cover_fraction
+        @. p.snow.liquid_water_flux = clip_water_flux(
```
kmdeck, line 380: I didnt store this an an "applied liquid water flux" because we dont need it anywhere else. but, the difference between applied and total energy/water fluxes are required to conserved energy and mass.

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -296,27 +262,24 @@ function runoff_timescale(z::FT, Ksat::FT, Δt::FT) where {FT}
...
 """
-function volumetric_internal_energy_liq(FT, parameters)
+function volumetric_internal_energy_liq(T::FT, parameters) where {FT}
     _ρ_l = FT(LP.ρ_cloud_liq(parameters.earth_param_set))
     _T_freeze = FT(LP.T_freeze(parameters.earth_param_set))
     _cp_l = FT(LP.cp_l(parameters.earth_param_set))
     _T_ref = FT(LP.T_0(parameters.earth_param_set))
 
-    I_liq = _ρ_l * _cp_l * (_T_freeze .- _T_ref)
+    I_liq = _ρ_l * _cp_l * (T .- _T_ref)
```
kmdeck, line 276: we are no longer guaranteed to only have liquid water at the freezing temperature

a-charbon (commented):
Left some comments!

src/standalone/Snow/Snow.jl
kmdeck: The desired behavior: if there is enough heat to melt all remaining snow, S_l -> S, and then the "runoff" term damps S_l to zero.

That said, I have not clipped S_l directly (it's hard to guarantee because we can only clip tendencies of prognostic variables, not the prognostic variables themselves) but we do clip q_l to be <=1, we clipe S_l <=S in any diagnostic function using it.

kmdeck: I changed it to this: 
```
function clip_liquid_water_flux(S_l::FT, S::FT, liquid_water_flux::FT, applied_water_flux::FT, Δt::FT) where {FT}
    predicted_S = S - applied_water_flux * Δt
    predicted_S_l = S_l - liquid_water_flux * Δt
    if predicted_S_l < 0
        return S_l / Δt
    elseif predicted_S_l > predicted_S
        return (S_l - predicted_S)/Δt
    else
        return liquid_water_flux
    end
end
```

src/standalone/Snow/boundary_fluxes.jl
```diff
@@ -80,7 +80,7 @@ end
 """
     snow_boundary_fluxes!(
         bc::AtmosDrivenSnowBC,
-        prognostic_land_components::Val{(:snow,)},
+v        prognostic_land_components::Val{(:snow,)},
```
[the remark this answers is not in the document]
kmdeck: removed

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -406,12 +404,13 @@ end
...
-    density::ConstantDensityModel,
+    density::ConstantDensityModel{FT},
     params::SnowParameters,
     Y,
     p,
-)
-    p.snow.ρ_snow .= density.ρ_snow
+) where {FT}
+    _ρ_l = FT(LP.ρ_cloud_liq(params.earth_param_set))
+    @. p.snow.ρ_snow = density.ρ_snow * (1 - p.snow.q_l) + _ρ_l * p.snow.q_l
```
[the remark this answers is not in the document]
kmdeck: I renamed this to ConstantDryDensityModel

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -325,27 +288,65 @@ Runoff occurs as the snow melts and exceeds the water holding capacity.
...
+    _LH_f0 = FT(LP.LH_f0(parameters.earth_param_set))
+    _ρ_liq = FT(LP.ρ_cloud_liq(parameters.earth_param_set))
+    _cp_i = FT(LP.cp_i(parameters.earth_param_set))
+    _cp_l = FT(LP.cp_l(parameters.earth_param_set))
+    _T_ref = FT(LP.T_0(parameters.earth_param_set))
+    _T_freeze = FT(LP.T_freeze(parameters.earth_param_set))
+    if energy_excess > 0
+        return -energy_excess / parameters.Δt / _ρ_liq /
+               ((_cp_l - _cp_i) * (_T_freeze - _T_ref) + _LH_f0)
+    elseif energy_excess < 0 && S_l > eps(FT) # predicted energy is below energy at freezing point, but liquid water remains, freeze some
```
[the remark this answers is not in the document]
kmdeck: I combined them.

For the second question, we would alter the temperature indirectly. The total energy flux is applied to the energy, and then we compute the change in S_l, and then we update both variables. the newly computed T would be below freezing.

src/standalone/Snow/Snow.jl
kmdeck: I believe this is addressed with the density passed begin the dry snow density.

src/standalone/Snow/Snow.jl
```diff
@@ -114,7 +114,7 @@ end
                       z_0b = FT(0.00024),
                       α_snow = FT(0.8),
                       ϵ_snow = FT(0.99),
-                      θ_r = FT(0.08),
+                      θ_r = FT(0.0),
```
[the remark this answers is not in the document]
kmdeck: " do we really want our default method to have 0.00 and all water instantaneously becomes runoff?)" This is not the case. We estimate a timescale (a guess at the percolation time through the snowpack, it increases with depth) and the liquid water damps to zero on that timescale. 

We can damp to a nonzero number if T=Tfreeze and damp to zero if T > Tfreeze. If T< Tfreeze, the water should refreeze.

I like the idea of SWE_ice and SWE_liq but Im not sure Im up for implementing it now.

src/standalone/Snow/snow_parameterizations.jl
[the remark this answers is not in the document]
kmdeck: You wrote ": if you picked θ_r = 95% and picked ice_matrix_constant_density = 100 kg/m^3, and reach SWE = 1 cm and SWE_liq = 0.95cm, even though SWE_liq is an allowable amount w.r.t θ_r, you have 0.95cm of water "inside" an ice matrix that is only 0.5cm tall, which is unphysical."

But I dont think this is right...If SWe_ice - 0.5cm, it doenst mean the ice matrix is 0.5cm tall. it means that if you melted the ice matrix it would create 0.5cm of water. it can still be tall enough to accomodate the liquid water if the porosity is very large.

src/standalone/Snow/snow_parameterizations.jl
kmdeck: Im not sure I understand...we can discuss more offline.

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -374,49 +383,40 @@ function energy_from_T_and_swe(S::FT, T::FT, parameters) where {FT}
...
+
 end
 
 """
-    update_density_prog!(density::AbstractDensityModel{FT}, model::SnowModel{FT}, Y, p) where {FT}
+    update_density_prog!(density::MinimumDensityModel, model::SnowModel, Y, p)
 
 Updates all prognostic variables associated with density/depth given the current model state.
-This is the default method for the constant density model,
+This is the default method for the constant dry density model,
```
a-charbon, line 415: needs to be updated name for the minimum density model

src/standalone/Snow/snow_parameterizations.jl
a-charbon: Handled in offline discussion.

a-charbon (approved):
see message in offline chat regarding the one open comment, plus one cosmetic fix marked. PR looks ready otherwise.

src/standalone/Snow/snow_parameterizations.jl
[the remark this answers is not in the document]
a-charbon: handled in offline discussion - considerations resulting from this question remains for another day

src/standalone/Snow/snow_parameterizations.jl
kmdeck: good catch!

src/standalone/Snow/Snow.jl
kmdeck: Edit - we have changed the interp of this density parameter

PR #967, opened 2024-12-24
Reduce allocations

Sbozzolo:
This commit reduces allocations by changing turbulent_fluxes and net_radiation to be in place. In addition to this, the workaround introduced to enable GPU-compatibility in energy_hydrology resulted in lots of additional allocations, so I modified that too. 

These allocations are due to the fact that there are broadcasted expression hidden in simple function calls, so that Julia cannot properly do operations in place.

There are still allocations. One of the easy ones to tackle is in surface_specificy_humidity for snow, which accounts for a large fraction of the total allocations for the snowy_land benchmark.

Closes #966

### DIFF
--- a/docs/tutorials/standalone/Bucket/coupled_bucket.jl
+++ b/docs/tutorials/standalone/Bucket/coupled_bucket.jl
@@ -61,7 +61,7 @@
 
 # These are stored in the [BucketModel](https://clima.github.io/ClimaLand.jl/dev/APIs/Bucket/#ClimaLand.Bucket.BucketModel) object,
 # along with [BucketParameters](https://clima.github.io/ClimaLand.jl/dev/APIs/Bucket/#ClimaLand.Bucket.BucketParameters).
-# In order to compute turbulent surface fluxes, we call [turbulent_fluxes](https://clima.github.io/ClimaLand.jl/dev/APIs/shared_utilities/#ClimaLand.turbulent_fluxes),
+# In order to compute turbulent surface fluxes, we call [turbulent_fluxes!](https://clima.github.io/ClimaLand.jl/dev/APIs/shared_utilities/#ClimaLand.turbulent_fluxes!),
 # with arguments including `prescribed_atmos`. Since this argument is of the type `PrescribedAtmosphere`, the method of `turbulent_fluxes` which is executed is one which computes the turbulent surface fluxes
 # using MOST. We have a similar function for [net_radiation](https://clima.github.io/ClimaLand.jl/dev/APIs/shared_utilities/#ClimaLand.net_radiation) and which computes the net radiation based on the prescribed downwelling radiative fluxes, stored in an argument
 # `prescribed_radiation`, which is of type `PrescribedRadiation`.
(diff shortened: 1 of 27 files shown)


src/shared_utilities/drivers.jl
```diff
@@ -264,24 +266,27 @@ function turbulent_fluxes(
...
-        d_sfc,
-        p.drivers.thermal_state,
-        u_air,
-        h_air,
-        atmos.gustiness,
-        model.parameters.z_0m,
-        model.parameters.z_0b,
-        Ref(model.parameters.earth_param_set),
-    )
+    dest .=
```
kmdeck, line 269: We may not need the Ref(), because of [this](https://github.com/CliMA/ClimaLand.jl/blob/cd785c8f3c5cf984d2663bf1101db791c6e2971f/src/shared_utilities/Parameters.jl#L31)  in which case we could do @. dest = turb(). That might be a very marginal improvement?

docs/tutorials/standalone/Bucket/coupled_bucket.jl
```diff
@@ -74,31 +74,33 @@
 # struct CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT} end
 # ```
 
-# Then, we have defined a new method for `turbulent_fluxes` and `net_radiation` which dispatch for these types,
+# Then, we have defined a new method for `turbulent_fluxes!` and `net_radiation!` which dispatch for these types,
 # and simply return the fluxes that the coupler has updated `p.bucket.turbulent_fluxes` and `p.bucket.R_n` with.
```
kmdeck, line 78: "and simply update `dest` in place with the fluxes that the coupler has updated `p.bucket.turbulent_fluxes` and `p.bucket.R_n` with."

Reading this, I actually think this function can do nothing in this case. Because we are setting p.bucket.Rn .= p.bucket.Rn, basically!

We can leave alone for now and do quick followup PR or you can change it? up to you

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -645,7 +645,7 @@ function ClimaLand.source!(
     _ρ_i = FT(LP.ρ_cloud_ice(earth_param_set))
     Δz = model.domain.fields.Δz # center face distance
 
-    # Wrap hydrology and earth parameters in one struct to avoid type inference failure
+    # Wrap hydrology and earth parameters in one struct to avoid type inference failure. This is allocating! We should remove it.
```
kmdeck, line 648: This was something @juliasloan25 implemented because otherwise we had the GPU error on central. it's a climacore bug if I understood correctly - we couldnt broadcast over hydrology_cm and earth_param_set separately.

but maybe we can refactor the function - hydrology_cm and earth_param_set are packaged inside of model.parameters already

kmdeck (commented):
I think this looks good - to be thoroughness, maybe we can compare the buildkite (regular CI) plots for e.g. Ozark, snow col de porte, and snow soil to the current main?

kmdeck:
also, if you dont mind, you can assign me an issue re dealing with the snow specific humidity allocation and the allocation in HydrologyEarthParameters

src/standalone/Soil/energy_hydrology.jl
Sbozzolo: Yes, I didn't tackle it because it leads to inference issues.

kmdeck (approved):

PR #968, opened 2025-01-02
rename LandHydrologyModel as SoilSnowModel

### LINKED ISSUE #959: Naming of `LandHydrology` and `LandHydrologyModel` is confusing
We have two model types with very similar names:
- `LandHydrology`, which consists of a soil model and a surface water model
- `LandHydrologyModel`, which consists of a soil model (EnergyHydrology) and a snow model (SnowModel)

We should rename these to avoid confusion.

juliasloan25:
## Purpose 
Previously, we had two different model types with very similar names: `LandHydrology` and `LandHydrologyModel`. This PR renames the latter as `SoilSnowModel` to avoid confusion.

closes #959


kmdeck (approved):
good catch!

PR #970, opened 2025-01-02
add state NaN checker

### LINKED ISSUE #969: Add a state NaN checker
Recently, we've been working to improve stability of global longruns. To check if a simulation is unstable, we visually inspect output plots from throughout the simulation and try to tell if there are any NaNs. This is prone to errors in coastal areas where it's hard to distinguish NaNs coming from the simulation vs interpolation in plotting, and is only a qualitative estimate of correctness. We should also have a function that quantitatively checks how many NaNs are present for each state variable at the end of a simulation, and displays this information.

A next step could be outputting the spatial location of the NaNs, but we can add that as needed.

juliasloan25:
## Purpose 
closes #969

Adds a function that quantitatively checks how many NaNs are present in the state, and displays this information to the user. This can be used to inspect `sol.u[end]` after a simulation has run, to see if any NaNs were produced at the end of the simulation.

If no NaNs are found, this information is logged in an `info` statement. If NaNs are found, this information is logged in a `warn` statement.

## To-do
- [x] add function to recursively check NaNs in state
- [x] add tests
- [x] add optional mask argument
- [x] add monthly callback to longruns


src/shared_utilities/utils.jl
```diff
@@ -492,3 +492,34 @@ function isdivisible(
...
+function count_nans_state(state::ClimaCore.Fields.FieldVector)
+    for pn in propertynames(state)
+        state_new = getproperty(state, pn)
+        @info "Checking NaNs in $pn"
+        count_nans_state(state_new)
+    end
+    return nothing
+end
+
+function count_nans_state(state::ClimaCore.Fields.Field)
```
kmdeck, line 517: would we like to make this also take the land mask as a second argument?

kmdeck (commented):
do you want to also add this to our global runs in this PR? soil, soil canopy, soil canopy snow? perhaps we can add it as a callback to check every month e.g.

src/shared_utilities/utils.jl
```diff
@@ -492,3 +492,40 @@ function isdivisible(
...
+in the state variables where the mask is 1. This is intended to be used with
+the land/sea mask, to avoid counting NaNs over the ocean. Note this assumes
+the mask is 1 over land and 0 over ocean.
+"""
+function count_nans_state(state::ClimaCore.Fields.FieldVector, mask = nothing)
+    for pn in propertynames(state)
+        state_new = getproperty(state, pn)
+        @info "Checking NaNs in $pn"
+        count_nans_state(state_new, mask)
+    end
```
Sbozzolo, line 517: Maybe
```julia
any(isnan, state)
```
this works

juliasloan25: That code runs, but we really want to be able to count the number of NaNs so we can have a better sense of which are over ocean vs land. 

```
count(isnan, Y)
count(isnan, parent(Y))
```
fail with scalar indexing, and checking a specific field like this
```
count(isnan, Y.soilco2.C)
```
fails when trying to iterate:
```
ERROR: MethodError: no method matching iterate(::ClimaCore.Fields.Field{ClimaCore.DataLayouts.VIJFH{Float64, 15, 2, CUDA.CuArray{…}}, ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace{ClimaCore.Grids.ExtrudedFiniteDifferenceGrid{…}, ClimaCore.Grids.CellCenter}})
...
```

Unfortunately it seems we need to stick with `count(isnan, parent(Y.soilco2.C))`

juliasloan25:
Looking at the snowy land longrun [here](https://buildkite.com/clima/climaland-long-runs/builds/2898#0194529e-5782-444a-8b4c-47e15aff4a3b), we see that from months 1-5, only `Y.soilco2.C` contains NaNs, the number of which increases from 2516460 to 2800335 during this period. At month 6, NaNs also appear in the soil, canopy, and snow sections of the state, but much fewer than in the soilco2 model (60, 4, and 4 NaNs in each of the state variables of these models, respectively).

I also interactively checked the state immediately after initialization, and there are no NaNs in any of the state, including the soilco2 model.

I'm not sure what to make of this but I'll continue looking into it. Investigating these NaNs is separate from the purpose of this PR, which is to provide the infrastructure to be able to see the NaNs, so this doesn't need to block merging this PR.

src/shared_utilities/utils.jl
Sbozzolo: `sum(isnan.(Y))` might work because falses are 0s and trues are 1s

juliasloan25: Good point! Thank you

juliasloan25: Something isn't right here. With an array, `sum(isnan...)`  works:
```julia
julia> arr = Float32.([0, 1, 2, 3, 4])
5-element Vector{Float32}:
 0.0
 1.0
 2.0
 3.0
 4.0

julia> sum(isnan, arr)
0

julia> arr[1] = NaN
NaN

julia> sum(isnan, arr)
1
```

With a Field, `sum(isnan...)` doesn't work:
```julia
    domain = ClimaLand.Domains.SphericalShell(;
        radius = FT(2),
        depth = FT(1.0),
        nelements = (10, 5),
        npolynomial = 3,
    )

    # Construct some fields
    space = domain.space.subsurface
    var1 = Fields.zeros(space)

julia> sum(isnan, var1)
0.0f0

julia> sum(isnan, parent(var1))
0

    # Add some NaNs to the fields without scalar indexing
    ArrayType = ClimaComms.array_type(ClimaComms.device())
    var1_copy = Array(parent(var1))
    var1_copy[1] = NaN
    parent(var1) .= ArrayType(var1_copy)

julia> sum(isnan, var1)
9.488383f-5

julia> sum(isnan, parent(var1))
1
```

I thought this might be coming from the incorrect max/min of fields we saw from ClimaCore, but this is using ClimaCore v0.14.22 so it should include the fix in https://github.com/CliMA/ClimaCore.jl/pull/2106. I'll look into it more tomorrow

Sbozzolo: There might be a bug with `sum(f, field)` syntax. I don't think it is well tested and widely used.

Maybe as a first step, you can try seeing what happens if you run `sum(f.(field))`

juliasloan25: > Also note that
> 
> ```
>  # Add some NaNs to the fields without scalar indexing
>     ArrayType = ClimaComms.array_type(ClimaComms.device())
>     var1_copy = Array(parent(var1))
>     var1_copy[1] = NaN
>     parent(var1) .= ArrayType(var1_copy)
> ```
> 
> This is pretty much the definition of scalar indexing, you just implementing it yourself instead of letting CUDA do the work.

I did it like this because CUDA isn't a dependency in the Test environment and I didn't want to bring it in

Sbozzolo: You can use `ClimaComms.allowscalar` for that

kmdeck: hey - not sure if this is the issue at hand, but sum(field) is actually the integral of the field over the domain. while sum(array) is the thing you would expect. or at least, this was historically true with ClimaCore

Sbozzolo: Indeed. I didn't know that this was the behavior!

src/shared_utilities/utils.jl
```diff
@@ -492,3 +492,109 @@ function isdivisible(
...
+    if num_nans > 0
+        @warn "$num_nans NaNs found"
+    else
+        verbose && @info "No NaNs found"
+    end
+    return nothing
+end
+
+"""
+    NaNCheckCallback(nancheck_frequency::Union{AbstractFloat, Dates.Period},
```
kmdeck, line 545: we might rename nancheck_frequency to nancheck_interval, because frequency I think implies a rate, not an interval. But only if you agree! it's very minor

src/shared_utilities/utils.jl
```diff
@@ -492,3 +492,109 @@ function isdivisible(
...
+    dt,
+)
+    # TODO: Move to a more general callback system. For the time being, we use
+    # the ClimaDiagnostics one because it is flexible and it supports calendar
+    # dates.
+
+    if nancheck_frequency isa AbstractFloat
+        # Assume it is in seconds, but go through Millisecond to support
+        # fractional seconds
+        nancheck_frequency_period = Dates.Millisecond(1000nancheck_frequency)
```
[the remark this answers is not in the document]
kmdeck: but they seem like the same thing

kmdeck:
> Looking at the snowy land longrun [here](https://buildkite.com/clima/climaland-long-runs/builds/2898#0194529e-5782-444a-8b4c-47e15aff4a3b), we see that from months 1-5, only `Y.soilco2.C` contains NaNs, the number of which increases from 2516460 to 2800335 during this period. At month 6, NaNs also appear in the soil, canopy, and snow sections of the state, but much fewer than in the soilco2 model (60, 4, and 4 NaNs in each of the state variables of these models, respectively).
> 
> I also interactively checked the state immediately after initialization, and there are no NaNs in any of the state, including the soilco2 model.
> 
> I'm not sure what to make of this but I'll continue looking into it. Investigating these NaNs is separate from the purpose of this PR, which is to provide the infrastructure to be able to see the NaNs, so this doesn't need to block merging this PR.

soil co2 has problems - we dont check it in any diagnostic or map right now, so we arent catching the NaNs. it's something @AlexisRenchon  and I need to go back to.

could some of the NaNs be over the ocean?

agreed that that is a separate problem!

juliasloan25:
> > Looking at the snowy land longrun [here](https://buildkite.com/clima/climaland-long-runs/builds/2898#0194529e-5782-444a-8b4c-47e15aff4a3b), we see that from months 1-5, only `Y.soilco2.C` contains NaNs, the number of which increases from 2516460 to 2800335 during this period. At month 6, NaNs also appear in the soil, canopy, and snow sections of the state, but much fewer than in the soilco2 model (60, 4, and 4 NaNs in each of the state variables of these models, respectively).
> > I also interactively checked the state immediately after initialization, and there are no NaNs in any of the state, including the soilco2 model.
> > I'm not sure what to make of this but I'll continue looking into it. Investigating these NaNs is separate from the purpose of this PR, which is to provide the infrastructure to be able to see the NaNs, so this doesn't need to block merging this PR.
> 
> soil co2 has problems - we dont check it in any diagnostic or map right now, so we arent catching the NaNs. it's something @AlexisRenchon and I need to go back to.
> 
> could some of the NaNs be over the ocean?
> 
> agreed that that is a separate problem!

It could be over the ocean, but from the info we get from this checker it's hard to tell.

Before these runs, I expected we would see NaNs over the ocean in most (all?) variables after 1 timestep. But I guess even over the ocean things can be stable for some time and only break later on. If this is the case, I'm not sure how helpful the single number of NaNs we provide will be, since it's impossible to distinguish if they're over ocean or not. We could solve this by including the mask in our check though.

src/shared_utilities/utils.jl
```diff
@@ -492,3 +492,109 @@ function isdivisible(
...
+end
+
+function count_nans_state(
+    state::ClimaCore.Fields.Field;
+    mask = nothing,
+    verbose = false,
+)
+    num_nans =
+        isnothing(mask) ? count(==(1), isnan.(parent(state))) :
+        count(==(1), isnan.(parent(state)) .* parent(mask))
```
Sbozzolo, line 535: Let's at least mark with some comments that this is not the best pattern and shouldn't be replicated.

Sbozzolo (approved):

src/shared_utilities/utils.jl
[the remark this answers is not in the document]
juliasloan25: I renamed all of these to `nancheck_interval`

PR #971, opened 2025-01-03
unset memory pool for gpu

kmdeck:
## Purpose 
From Gabriele: I flipped a switch and unset JULIA_CUDA_MEMORY_POOL=none. This is setting we have to run with MPI, but this is a setting we would want to not use because it disables a useful feature in CUDA to reuse allocated memory. Once I disabled that, the smaller jobs got more than twice as fast, because I removed one of the sources of inefficiency.

This PR unsets JULIA_CUDA_MEMORY_POOL.

## To-do
test

## Content
Updates our three pipelines to unset JULIA_CUDA_MEMORY_POOL. If we add tests with MPI in the future, we would need to unset this for specific experiments only. Right now this PR unsets for all experiments in each buildkite run.

----

### DIFF
--- a/.buildkite/longruns_gpu/pipeline.yml
+++ b/.buildkite/longruns_gpu/pipeline.yml
@@ -23,6 +23,7 @@ steps:
     env:
       JULIA_NUM_PRECOMPILE_TASKS: 8
       JULIA_MAX_NUM_PRECOMPILE_FILES: 50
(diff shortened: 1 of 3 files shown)


kmdeck:
possibly superceded by https://github.com/CliMA/ClimaLand.jl/pull/966, will discuss next week with @Sbozzolo.

PR #973, opened 2025-01-06
Add MODIS LAI driver & use in experiments

Espeer5:
## Purpose 

This PR adds driver utilities to run simulations using MODIS LAI from 2008 as opposed to era5 LAI. Corresponding PR adding MODIS LAI data to ClimaArtifacts is [here](https://github.com/CliMA/ClimaArtifacts/pull/94).

## To-do

 - [x] Verify MODIS LAI works correctly where used in snowy land benchmark
 - [x] Clarify which experiments/benchmarks should use MODIS LAI by default

## Content

 - MODIS LAI artifact added to Artifacts.toml
 - Utility to get the MODIS LAI artifact path added to src/Artifacts.jl
 - Utility to construct the LAI function from MODIS LAI data added to drivers.jl

----

### DIFF
--- a/experiments/benchmarks/snowy_land.jl
+++ b/experiments/benchmarks/snowy_land.jl
@@ -231,12 +231,12 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
     photosynthesis_args =
         (; parameters = Canopy.FarquharParameters(FT, is_c3; Vcmax25 = Vcmax25))
     # Set up plant hydraulics
-    era5_lai_artifact_path =
-        ClimaLand.Artifacts.era5_lai_forcing_data2008_folder_path(; context)
-    era5_lai_ncdata_path =
-        joinpath(era5_lai_artifact_path, "era5_2008_1.0x1.0_lai.nc")
-    LAIfunction = ClimaLand.prescribed_lai_era5(
-        era5_lai_ncdata_path,
+    modis_lai_artifact_path =
+        ClimaLand.Artifacts.modis_lai_forcing_data2008_path(; context)
+    modis_lai_ncdata_path =
+        joinpath(modis_lai_artifact_path, "Yuan_et_al_2011_1x1.nc")
+    LAIfunction = ClimaLand.prescribed_lai_modis(
+        modis_lai_ncdata_path,
         surface_space,
         start_date;
         time_interpolation_method = time_interpolation_method,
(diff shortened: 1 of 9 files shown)


experiments/benchmarks/snowy_land.jl
```diff
@@ -225,12 +225,22 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
...
-        era5_lai_ncdata_path,
+    # era5_lai_artifact_path =
+        # ClimaLand.Artifacts.era5_lai_forcing_data2008_folder_path(; context)
+    modis_lai_artifact_path = 
+        ClimaLand.Artifacts.modis_lai_forcing_data2008_path(; context)
+    # era5_lai_ncdata_path =
+        # joinpath(era5_lai_artifact_path, "era5_2008_1.0x1.0_lai.nc")
+    modis_lai_ncdata_path = 
+        joinpath(modis_lai_artifact_path, "Yuan_et_al_2011_1x1.nc")
+    # LAIfunction = ClimaLand.prescribed_lai_era5(
```
Espeer5, line 236: For now I opted to leave these changes very clearly temporary - I'm not sure whether we want to change certain runs to use MODIS, or all of them, or none of them. The main point for now was just to test that the format of the data is all correct and it will run the experiment correctly on the buildkite runs.

kmdeck:
A note - we dont run the benchmarks by default (or the long runs). You need to add the label to the PR, and then make a push to the branch to trigger them. it's a little annoying but better than running them every time

Espeer5:
I'm really not sure what's happening with the buildkite here - no changes to the manifest or anything so I don't know why there are so many failures in the packages/includes/imports?

kmdeck:
> I'm really not sure what's happening with the buildkite here - no changes to the manifest or anything so I don't know why there are so many failures in the packages/includes/imports?

I had trouble today to, and cleared the depot. I tried rebuilding here - let's see if it runs now

kmdeck (commented):
I think this looks good from the code point of view! I would also think we can use modis wherever we use ERA5 now.  I just want to check that the number of NaNs in the simulation did not increase (would check when you rebase it and restart the long runs) and that the SYPD did not change much

@juliasloan25 a note for the OKR to run climaland for 20 years, we only have one year of MODIS data. But we can start with ERA5 data to assess stability as we are doing.

kmdeck (approved):

Espeer5:
@kmdeck any ideas what's going on with the buildkite here? Seems uinrelated to these changes.

kmdeck:
@Espeer5 I am clearing the depot. That usually fixes this type of bug.

Also - sent some messages on slack. There are a lot of NaNs in the sim now, but I think that is from separate changes in main, not this PR. Im verifying that today so I think we can still go ahead and merge this.

PR #977, opened 2025-01-11
Clear `ALL_DIAGNOSTICS` in `define_diagnostics`

imreddyTeja:
This commit makes the `define_diagnostics` function clear `ALL_DIAGNOSTICS`. The dictionary is only cleared when it has data, and a warning is given. This prevents each `add_diagnostic_variable` call in `define_diagnostics` from throwing overwrite warnings.

This only changes the resulting diagnostics when a user calls `add_diagnostic_variable!` and then calls `define_diagnostics!` or `default_diagnostics`. This desired result could still be achieved by swapping the orders of the call.

## Content

----


AlexisRenchon (approved):
This is great! Thank you Teja

Sbozzolo:
I am not sure if this is the best solution. Users could add their own diagnostics in `ALL_DIAGNOSTICS`, and this would remove them, while only adding back the default ones.

PR #980, opened 2025-01-13
reduce allocations in snow and neural snow

### LINKED ISSUE #976: Reduce allocations in snowy land
The main big remaining allocations:
- [ ] `snow_depth` in `snow_parameterizations.jl` PR #980
- [X] `surface_specific_humidty` in `snow_parametrizations.jl`PR #978  
- [ ] ground albedo allocations  PR #981

kmdeck:
## Purpose 
Closes #976 

@a-charbon does this close 960 from your point of view? It may not be entirely general, but it works for the use cases we have

## Content

----

### DIFF
--- a/ext/neural_snow/NeuralSnow.jl
+++ b/ext/neural_snow/NeuralSnow.jl
@@ -9,7 +9,7 @@ import ClimaLand.Snow:
     density_prog_names,
     update_density!,
     update_density_prog!,
-    snow_depth
+    snow_depth!
 import ClimaLand.Parameters as LP
 using Thermodynamics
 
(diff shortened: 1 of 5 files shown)


ext/neural_snow/NeuralSnow.jl
```diff
@@ -114,14 +114,33 @@ ClimaLand.Snow.density_prog_names(m::NeuralDepthModel) =
...
-This is sufficient to enable dynamics of the auxillary variable `ρ_snow` without extension of update_density!, and avoids
-redundant computations in the computation of runoff.
+
+This function clips the snow depth to be between 0 and SWE.
 """
-ClimaLand.Snow.snow_depth(m::NeuralDepthModel, Y, p, params) = Y.snow.Z
+function snow_depth!(z_snow, m::NeuralDepthModel{FT}, Y, p, params) where {FT}
+    z_snow .= min(Y.snow.Z, Y.snow.S) # z cannot be larger than SWE
+    z_snow .= max(z_snow, eps(FT)) # z must be positive
+end
```
Sbozzolo, line 127:
```suggestion
function snow_depth!(z_snow, m::NeuralDepthModel{FT}, Y, p, params) where {FT}
    z_snow .= min(Y.snow.Z, Y.snow.S) # z cannot be larger than SWE
    z_snow .= max(z_snow, eps(FT)) # z must be positive
    return nothing
end
```

I think it's a good idea to return `nothing` in this type of functions. This ensures that the compiler can easily infer the return type (it's always `nothing`), and prevents the function from being accidentally used in a non-in-place way (e.g., `snowdep = snow_depth!(z)`

kmdeck: ok! I changed this.

a-charbon (approved):
I don't think it closes the issue, we should leave it open as the redundancy remains (just in another way/place). However, this handles the allocation concerns and should be merged, and we can return to that issue another time.

PR #981, opened 2025-01-14
remove allocations in ground albedo of snowy land

kmdeck:
## Purpose 
Closes part of #976

## Content

----

### DIFF
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -205,6 +205,7 @@ lsm_aux_vars(m::LandModel) = (
     :T_sfc,
     :ϵ_sfc,
     :α_sfc,
+    :α_ground,
 )
 
 """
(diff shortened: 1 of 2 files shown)


src/integrated/land.jl
```diff
@@ -591,8 +592,10 @@ function Canopy.ground_albedo_PAR(
     p,
     t,
 )
-    return @. (1 - p.snow.snow_cover_fraction) * p.soil.PAR_albedo +
-              p.snow.snow_cover_fraction * ground.α_snow
+    @. p.scratch1 =
+        (1 - p.snow.snow_cover_fraction) * p.soil.PAR_albedo +
+        p.snow.snow_cover_fraction * ground.α_snow
+    return p.scratch1
```
Sbozzolo, line 598: I think that this is a little risky. Other lines in the functions that call this function might modify the value of `p.scratch`, which would affect the result of this field.

We should try keeping the scope of scratch variables visible limited and assume that something else will overwrite them at the next line

kmdeck: good call. I fixed this!

Sbozzolo (approved):

PR #982, opened 2025-01-14
remove HydrologyEarthParameters struct

### LINKED ISSUE #975: The workaround for GPU inference failure allocates new fields
https://github.com/CliMA/ClimaLand.jl/blob/cd785c8f3c5cf984d2663bf1101db791c6e2971f/src/standalone/Soil/energy_hydrology.jl#L648-L650

This line is allocating a new field at every step (on GPUs), which should be avoided, if possible.

juliasloan25:
## Purpose 
This PR removes the `HydrologyEarthParameters` struct, which was introduced in https://github.com/CliMA/ClimaLand.jl/pull/739 to fix a GPU inference issue. Here, we instead update the interface of `phase_change_source` to take in the full model parameters as its final argument. This both fixes the inference issue we saw before, and removes the allocating use of `HydrologyEarthParameters`.

closes #975

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@ ClimaLand.jl Release Notes
 main
 -------
 
+- Change `Soil.phase_change_source` interface to take model parameters
+  as the final argument; remove `HydrologyEarthParameters` struct.
+  PR[#982](https://github.com/CliMA/ClimaLand.jl/pull/982)
 
 v0.15.7
 -------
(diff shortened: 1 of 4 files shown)


kmdeck:
What a mysterious bug! I thought it was from broadcasting over too many structs, but here it seems like we passed another as an argument, and it fixes it

juliasloan25:
> What a mysterious bug! I thought it was from broadcasting over too many structs, but here it seems like we passed another as an argument, and it fixes it

Unfortunately I don't think we fixed it, it looks like the most recent [buildkite](https://buildkite.com/clima/climaland-ci/builds/6438#_) has the same error that we were seeing [originally](https://buildkite.com/clima/climaland-ci/builds/5416#019253d6-1b25-4293-a047-5e4eed4af2d1).

kmdeck:
closed by PR #1042

PR #984, opened 2025-01-16
add soilco2.C to :short diagnostics

juliasloan25:
## Purpose 
It would be helpful to see soilco2.C throughout our longrun simulations to seen when/where it's NaN, especially since our other prognostic variables don't depend on it directly. This PR updates the "essential" diagnostics specified by the `output_vars = :short` flag for the SoilCanopyModel and LandModel to include soilco2.C.


AlexisRenchon (approved):

kmdeck (commented):
do we also need to define a compute method?

juliasloan25:
> do we also need to define a compute method?

It's already defined here https://github.com/CliMA/ClimaLand.jl/blob/2d8f19048c37afb501cdc837965807a6fa274009/src/diagnostics/land_compute_methods.jl#L428

PR #985, opened 2025-01-16
run soilcanopy longrun for 1 month

juliasloan25:
## Purpose 
trigger longruns for only 3 months with new soilco2 diagnostic


kmdeck:
@juliasloan25 is there more to do here, or can we close this? (maybe this was to debug NaNs in soil co2?)

PR #988, opened 2025-01-21
Remove rho_c_D_g parameter

kmdeck:
## Purpose 
To prevent division by zero in snow bulk temperature, we had previously followed the UEB model with the parameter "rho *c * D" for the ground, i.e. the amount of mass of ground interacting thermally with the snow (assumed to be at the same temp as the snow).

In the ground heat flux between snow and soil, we used this parameter to solve for the thickness of the layer of soil interacting thermally with the snow.

After reading the draft, Tapio preferred that we change this to a small "delta S" parameter - i.e. to keep the interpretation numerical, not physical, and limited only to the snow model.

This makes that change.

## Content

----

### DIFF
--- a/src/integrated/soil_snow_model.jl
+++ b/src/integrated/soil_snow_model.jl
@@ -247,53 +247,17 @@ function update_soil_snow_ground_heat_flux!(
     soil_domain,
     FT,
 )
-    # Thermal conductivities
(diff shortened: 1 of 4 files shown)


a-charbon (approved):
Changes look good.

PR #989, opened 2025-01-21
Integrating NeuralDepthModel for global runs

a-charbon:
## Purpose 
 To set up the code necessary to run the NeuralDepthModel for global scale computation

## To-do
 handle Flux 0.14-0.15 update bug and/or use more efficient NeuralSnowModel, and update ClimaLand Utilities to be able to work with these variables in-bulk

## Content

----
- [ x] I have read and checked the items on the review checklist.


.buildkite/Project.toml
```diff
@@ -46,4 +46,4 @@ cuDNN = "02a925ec-e4fe-4b08-9a7e-0d78e3d38ccd"
 ClimaAnalysis = "0.5.12"
 ClimaTimeSteppers = "0.7"
 Statistics = "1"
-Flux = "0.15"
+Flux = "0.14.25"
```
kmdeck, line 49: I think we can do `Flux = "~0.14"`

kmdeck: (being less restrictive is better!)

experiments/long_runs/snowy_land.jl
```diff
@@ -49,6 +49,9 @@ import NCDatasets
 
 using Poppler_jll: pdfunite
 
+using CSV, DataFrames, HTTP, Flux, StatsBase, cuDNN, BSON
```
kmdeck, line 52: I would suggest reverting this file back (git checkout origin/main experiments/long_runs/snowy_land.jl), because for us to merge this, we need to check if the long runs are stable for multiple years with this addition.

kmdeck (approved):
Looks good pending the changes to the long run file! Maybe we can also make this the default for the snowmip simulation script, as in:
https://github.com/CliMA/ClimaLand.jl/pull/990

PR #992, opened 2025-01-24
debug soilco2

juliasloan25:
## Purpose 
Trigger longrun of SoilCanopyModel for 1 month only, so we can see where NaNs appear and then set up a regional simulation there

## Content

----


juliasloan25:
This was deprioritized so we haven't worked on it. We can close this for now, and reopen a similar PR when we want to return to this

PR #993, opened 2025-01-27
Small parameterization changes + fluxnet site parameters

kmdeck:
## Purpose 
Improve certain parameterizations and parameter values

## To-do
A future branch will contain the scripts for fluxnet paper plots - these are tailored to certain sites based on what data is available. Our current run_fluxnet is more general. We can decide to merge them or not

## Content

----

### DIFF
--- a/experiments/integrated/fluxnet/US-Ha1/US-Ha1_simulation.jl
+++ b/experiments/integrated/fluxnet/US-Ha1/US-Ha1_simulation.jl
@@ -7,7 +7,10 @@ variables for running the simulation."""
 # Column dimensions - separation of layers at the top and bottom of the column:
 dz_bottom = FT(1.5)
 dz_top = FT(0.025)
-
+dz_tuple = (dz_bottom, dz_top)
+nelements = 20
+zmin = FT(-10)
+zmax = FT(0)
 # Stem and leaf compartments and their heights:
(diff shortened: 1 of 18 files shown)


AlexisRenchon (approved):
This PR looks good to me!

AlexisRenchon:
I approved you PR but if you want we can add months name, like this
```
            ax = Axis(
                fig_seasonal_cycle[1, 1],
                xlabel = "Month of the year",
                ylabel = "$short_name [$(ClimaAnalysis.units(var))]",
                xticks = (
                    1:1:12,
                    [
                        "Jan",
                        "Feb",
                        "Mar",
                        "Apr",
                        "May",
                        "Jun",
                        "Jul",
                        "Aug",
                        "Sep",
                        "Oct",
                        "Nov",
                        "Dev",
                    ],
                ),
            )
```
(or similar)

you can also do `CairoMakie.xlims!(ax, 1, 12)` if you don't want blank space left and right (but that is a bit personal preference)

for the first figure (volumetric water and temperature), you masked the x-axis tick label for the top 2 subplots, but not the 3rd
could be "day of the year" instead of "days"
alternatively, you could use xticks as above to put months name at `1:30:365`, and call it month of the year (you can change ticks to actual month length instead of steps 30)

PR #998, opened 2025-01-30
Higher resolution output

kmdeck:
## Purpose 
We currently store 180x90 points in lat/lon, but this causes poor quality figures along coastlines in particular, and doesnt make use of the higher resolution of our simulation (~0.2 degrees). This increases the resolution of the stored output. For monthly output the extra size is OK.

## Content
We have ~101x101x4x4 points in our simulation. The output now has 570x285 points. This is a ratio of ~1 output point to simulation point

----

### DIFF
--- a/experiments/long_runs/land_region.jl
+++ b/experiments/long_runs/land_region.jl
@@ -376,6 +376,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (10, 10, 15))
         subsurface_space,
         outdir;
         start_date,
+        num_points = n_elements,
     )
 
     diags = ClimaLand.default_diagnostics(
--- a/experiments/long_runs/bucket.jl
+++ b/experiments/long_runs/bucket.jl
@@ -124,11 +124,14 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 7))
     updatefunc = ClimaLand.make_update_drivers(drivers)
 
     # ClimaDiagnostics
-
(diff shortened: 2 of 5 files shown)


experiments/long_runs/land_region.jl
```diff
@@ -374,6 +374,7 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (10, 10, 15))
         subsurface_space,
         outdir;
         start_date,
+        num_points = n_elements,
```
juliasloan25, line 377: why does this one use `n_elements`?

kmdeck: this one is regional! so I thought it might make the most sense to use n_elements (nx,ny,nz) directly. For global runs, we specify n_elements = (n_horizontal, nz). But, we can also make the global runs use something based on n_elements (and n_polynomial)

juliasloan25: right, good point!

juliasloan25: I'm okay with hardcoding the points for the global runs since we rarely change the resolution, but it would be good to add in-line comments explaining that with these values we get 1:1 output to simulation points

juliasloan25 (approved):
Looks good, my only suggestion would be to add the in-line comments explaining why we use these numbers

kmdeck:
> Looks good, my only suggestion would be to add the in-line comments explaining why we use these numbers

thanks for the quick review! I added comments in

PR #1011, opened 2025-02-05
Add boundary fluxes terms to the soil jacobian when running with prescribed atmos

kmdeck:
## Purpose 
Adds infrastructure for adding d(boundary flux)/dY to the Jacobian for the soil model when Atmos-Driven fluxes are used.

Currently sets the contributions to zero, no change in simulations are expected. I turned on the "run benchmarks" to see if there is any impact.

## Content
Adds space to the cache for dLHFdT, dSHFdT, dGHFdT. Currently only the last one is updated/nonzero. The other two are set to zero

These terms will then be used in dFlux/dT dT/d rho e. We allocated space for those in the cache, but they are currently set to zero.

These terms are then added into the jacobian for the soil model.

----

### DIFF
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -198,10 +198,9 @@ lsm_aux_vars(m::LandModel) = (
     :atmos_energy_flux,
     :atmos_water_flux,
     :ground_heat_flux,
-    :effective_soil_sfc_T,
+    :dghfdT_soil,
     :sfc_scratch,
     :subsfc_scratch,
-    :effective_soil_sfc_depth,
     :T_sfc,
     :ϵ_sfc,
     :α_sfc,
@@ -215,7 +214,6 @@ The types of the additional auxiliary variables that are
 included in the land model.
 """
 lsm_aux_types(m::LandModel{FT}) where {FT} = (
-    FT,
     FT,
     FT,
     FT,
@@ -259,7 +257,6 @@ lsm_aux_domain_names(m::LandModel) = (
     :surface,
     :surface,
     :subsurface,
-    :surface,
     :surface,
     :surface,
     :surface,
@@ -526,6 +523,15 @@ function soil_boundary_fluxes!(
         ) +
         p.excess_heat_flux +
         p.snow.snow_cover_fraction * p.ground_heat_flux
+
+    # ρc_sfc is stored in scratch! after we use it below, it may be overwritten
+    ρc_sfc = ClimaLand.Soil.get_ρc_sfc(Y, p, soil.parameters)
+    # Get the local geometry of the face space, then extract the top level
+    local_geometry_faceN =
+        ClimaLand.get_local_geometry_faceN(soil.domain.space.subsurface)
+    @. p.soil.dfluxBCdY.heat =
+        covariant3_unit_vector(local_geometry_faceN) * 0 / ρc_sfc # ∂F∂T ∂T∂ρe
+    @. p.soil.dfluxBCdY.water = covariant3_unit_vector(local_geometry_faceN) * 0
     return nothing
 end
 
--- a/src/integrated/soil_canopy_model.jl
+++ b/src/integrated/soil_canopy_model.jl
@@ -363,11 +363,11 @@ the presence of the canopy modifies the soil BC.
 function soil_boundary_fluxes!(
     bc::AtmosDrivenFluxBC{<:PrescribedAtmosphere, <:PrescribedRadiativeFluxes},
     prognostic_land_components::Val{(:canopy, :soil, :soilco2)},
-    soil::EnergyHydrology{FT},
+    soil::EnergyHydrology,
     Y,
     p,
     t,
-) where {FT}
+)
     bc = soil.boundary_conditions.top
     turbulent_fluxes!(p.soil.turbulent_fluxes, bc.atmos, soil, Y, p, t)
     # influx = maximum possible rate of infiltration given precip, snowmelt, evaporation/condensation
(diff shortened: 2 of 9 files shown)


src/standalone/Soil/energy_hydrology.jl
```diff
@@ -332,42 +332,35 @@ function ClimaLand.make_compute_jacobian(model::EnergyHydrology{FT}) where {FT}
...
+            topBC_op_heat = Operators.SetBoundaryOperator(
+                top = Operators.SetValue(zero(FT)),
+                bottom = Operators.SetValue(
+                    Geometry.Covariant3Vector(zero(FT)),
+                ),
+            )
+            topBC_op_water = topBC_op_heat
+        end
+        # Add term from top boundary condition before applying divergence
+        # Note: need to pass 3D field on faces to `topBC_op`. Interpolating `K` to faces
```
kmdeck, line 359: When you wrote "this is inefficient", why is that?

src/integrated/land.jl
```diff
@@ -198,10 +198,9 @@ lsm_aux_vars(m::LandModel) = (
     :atmos_energy_flux,
     :atmos_water_flux,
     :ground_heat_flux,
-    :effective_soil_sfc_T,
```
kmdeck, line 201: These should have been removed in a previous PR - not used

src/integrated/land.jl
```diff
@@ -526,6 +523,36 @@ function soil_boundary_fluxes!(
...
         p.snow.snow_cover_fraction * p.ground_heat_flux
+
+    (; ρc_ds, earth_param_set) = soil.parameters
+    _σ = LP.Stefan(earth_param_set)
+
+    T_sfc = ClimaLand.Domains.top_center_to_surface(p.soil.T)
+    ρc = p.soil.sub_sfc_scratch
+    @. ρc =
+        volumetric_heat_capacity(p.soil.θ_l, Y.soil.θ_i, ρc_ds, earth_param_set)
+    ρc_sfc = ClimaLand.Domains.top_center_to_surface(ρc)
```
juliasloan25, line 534: I think this is allocating - is there another scratch space we could use?

src/integrated/soil_canopy_model.jl
```diff
@@ -383,6 +383,35 @@ function soil_boundary_fluxes!(
...
+
+    # Compute terms needed for derivatives
+    (; ρc_ds, earth_param_set) = soil.parameters
+    _σ = LP.Stefan(earth_param_set)
+
+    T_sfc = ClimaLand.Domains.top_center_to_surface(p.soil.T)
+    ρc = p.soil.sub_sfc_scratch
+    @. ρc =
+        volumetric_heat_capacity(p.soil.θ_l, Y.soil.θ_i, ρc_ds, earth_param_set)
+    ρc_sfc = ClimaLand.Domains.top_center_to_surface(ρc)
```
juliasloan25, line 395: same comment as above w.r.t. allocations

src/integrated/soil_snow_model.jl
```diff
@@ -379,6 +379,36 @@ function soil_boundary_fluxes!(
         ) +
         p.excess_heat_flux +
         p.snow.snow_cover_fraction * p.ground_heat_flux
+
```
juliasloan25, line 382: Since this section of code is the same for LandModel, SoilCanopyModel, and SoilSnowModel, it would be good to pull it out into a shared location to remove duplicated code. Maybe it could be two functions like `get_rho_sfc` and `get_local_geometry_faceN` since accessing those two is identical between the 3 models. The `dfluxBCdY` expressions vary a little bit (with vs without snow), so we can leave those in these model-specific functions. What do you think?

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -388,12 +388,15 @@ function ClimaLand.set_dfluxBCdY!(
...
     # Get the local geometry of the face space, then extract the top level
-    levels = ClimaCore.Spaces.nlevels(Domains.obtain_face_space(axes(p.soil.K)))
+    levels = ClimaCore.Spaces.nlevels(
+        ClimaLand.Domains.obtain_face_space(axes(p.soil.K)),
+    )
     local_geometry_faceN = ClimaCore.Fields.level(
-        Fields.local_geometry_field(Domains.obtain_face_space(axes(p.soil.K))),
+        ClimaCore.Fields.local_geometry_field(
+            ClimaLand.Domains.obtain_face_space(axes(p.soil.K)),
+        ),
```
juliasloan25, line 397: It should be fine to continue using just `Domains` here rather than `ClimaLand.Domains` - it's used in lines 382-387 above (doesn't really matter though)

src/integrated/soil_snow_model.jl
juliasloan25: (just saw that it's also used for EnergyHydrologyModel)

src/integrated/land.jl
kmdeck: oh! I think top_center_to_surface is OK - because it is just pulling out a view of the field.

src/integrated/soil_snow_model.jl
kmdeck: yes! it's a good idea. I'll do that

src/standalone/Soil/energy_hydrology.jl
kmdeck: @juliasloan25 do you recall what this was about?

src/integrated/land.jl
juliasloan25: hmm I see

src/standalone/Soil/energy_hydrology.jl
juliasloan25: I think it's because we call `interpc2f_op(p.soil.K)` and immediately zero out the values, so we could use any field defined on faces. If we have a field that's already on faces we can avoid doing the extra interpolation.

src/integrated/soil_snow_model.jl
```diff
@@ -379,6 +379,16 @@ function soil_boundary_fluxes!(
...
         p.excess_heat_flux +
         p.snow.snow_cover_fraction * p.ground_heat_flux
+
+    # Compute terms needed for derivatives
+    # ρc_sfc is stored in scratch! after we use it below, it may be overwritten
+    ρc_sfc = ClimaLand.Soil.get_ρc_sfc(Y, p, soil.parameters)
+    # Get the local geometry of the face space, then extract the top level
+    local_geometry_faceN =
+        ClimaLand.get_local_geometry_faceN(soil.domain.space.subsurface)
+    @. p.soil.dfluxBCdY.heat =
```
juliasloan25, line 389: are these multiplied by 0 just to exclude the contributions for now?

juliasloan25 (approved):

src/integrated/soil_snow_model.jl
kmdeck: yes! I wasnt sure if they would be initialized to zero, maybe so

kmdeck: I guess also it was sort of a stand in for dF/dT

PR #1013, opened 2025-02-05
Replace canopy temperature with surface temperature in short diagnostics

### LINKED ISSUE #1003: Default diagnostic: T_sfc instead of canopy temperature in default diagnostics
T_sfc is defined everywhere whereas t_canopy is set to NaN if there is no canopy

AlexisRenchon:
closes #1003 

T_sfc is defined everywhere whereas t_canopy is set to NaN if there is no canopy


kmdeck (approved):
im sorry I missed this! This is a good change!

Issue #876, opened 2024-10-22
O2.6.5 Add Hyperspectral Radiative Transfer to Canopy Model


Espeer5:
## Purpose

Our current radiative transfer model divides incoming radiation into 2 bands: the photosynthetically active band, and the near-infrared. band. This is a practice typical of older canopy models, but newer models implement much more granular radiative transfer models, banded into small wavelength ranges. Newer models also incorporate leaf optical properties to achieve more sophisticated modeling. The science for these models is included in the earlier [CliMA Land Model](https://github.com/CliMA/Land).

This is part of ongoing work on the Big Leaf canopy model.

## Cost/benefits/risks
The main cost of this effort is developer hours. Implementing this model will be nontrivial and will require time. 

Another cost could be efficiency. Moving from 2 bands to many wavelength bands will increase the complexity of the radiative transfer model. This could, especially if poorly designed, cause the overall canopy model speed to decrease.

These costs are clearly outweighed by the increased scientific value of the more modern radiative transfer model.

## Producers
@Espeer5 @braghiere 

## Components
The main deliverable is a new radiative transfer model option which can be plugged into the Canopy model, similar to the Beer model or the TwoStream model.  This will need to interact with the photosynthesis and other Canopy components as TwoStream/Beer RT models do.

## Inputs
The model will take as input maps of leaf optical properties and will need to take in the soil albedo. The model will ingest the downwelling radiation across the wavelength spectrum over the course of the simulation.

## Results and deliverables
All science from the radiative transfer model of the earlier [CliMA Land Model](https://github.com/CliMA/Land) is implemented and available in ClimaLand.jl. This is used with all appropriate global input data in the Land model global runs, with realistic (though uncalibrated) results and no NANs over the globe.

## Task breakdown
TODO

## Reviewers
@braghiere @kmdeck @AlexisRenchon

PR #1014, opened 2025-02-05
Allow n-band radiative transfer

Espeer5:
## Purpose 

This PR extends our radiative transfer functionality from 2 bands (PAR and NIR) to an n-band scheme. This will allow pairing with ClimaAtmos which provides 16 bands. Represents step one towards #876.

This PR must be a large number of lines due to updating of all of the experiments and tests.

## To-do

 - [x] Update fundamental radiation structs
 - [x] Update radiative transfer computations in models
 - [ ] Repair all experiments and tests to work with new scheme
 - [ ] Add a test for using up to 16 bands of radiation
 - [ ] Check benchmarks and long runs to ensure no significant regressions
 - [ ] Verify GPU compatibility is not broken

## Content

 - Adds a tuple-based type for the discretization of the λ spectrum
 - All leaf & soil optics types are converted to tuple-valued fields
 - Uses solar irradiance curve to allocate incoming shortwave to radiation bands
   - Also to compute PAR quantities from spectral absorption for photosynthesis
 - Updates all tests and experiments to work with new structs / types

----


Espeer5:
Can anyone help me out with the GPU compatibility issue here?

It's a bit odd to me because the same broadcast structure is used over `ntuple` in `canopy_sw_rt_beer_lambert` as in the `canopy_sw_rt_two_stream`, but only one of them seems to cause the issue. I'm a little stumped by it.

Sbozzolo:
I can help with the GPU compat, but I have a couple of other things that I need to finish first.

juliasloan25:
Is it possible to apply `canopy_sw_rt_two_stream` for each element of `\alpha_soil`, `\rho_leaf`, `\tau_leaf` (just broadcasting normally) instead of wrapping them in tuples to get around the broadcasting?

Espeer5:
> Is it possible to apply `canopy_sw_rt_two_stream` for each element of `\alpha_soil`, `\rho_leaf`, `\tau_leaf` (just broadcasting normally) instead of wrapping them in tuples to get around the broadcasting?

Possibly. I was originally doing this but had some issues handling it correctly when α/ρ/τ were tuple-valued-fields vs. plain tuples. But you're right this must be the issue. Thanks

PR #1017, opened 2025-02-06
Make syntax changes for new CTS version

a-charbon:
## Purpose 

image
The above image details the conversation surrounding the issue, as well as a recommended fix. It was determined that using "collect()" or "Array()" around t0:dt:tf is also good practice instead of just passing t0:dt:tf. Changes have been made within simulation files to make sure no errors appear when  using the new version of ClimaTimeSteppers.

## Content

----
- [ x] I have read and checked the items on the review checklist.

### DIFF
--- a/experiments/standalone/Soil/richards_runoff.jl
+++ b/experiments/standalone/Soil/richards_runoff.jl
@@ -157,7 +157,7 @@ prob = SciMLBase.ODEProblem(
     p,
 )
 save_every = 100
-saveat = Array(t0:(save_every * dt):tf)
+saveat = [t0, tf]
 sv = (;
     t = Array{Float64}(undef, length(saveat)),
     saveval = Array{NamedTuple}(undef, length(saveat)),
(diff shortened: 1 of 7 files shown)


kmdeck (approved):
Looks good! Ill check the buildkite plots once they finish to make sure it looks the same as it did and then we can merge

experiments/standalone/Soil/richards_runoff.jl
```diff
@@ -168,7 +168,13 @@ drivers = ClimaLand.get_drivers(model)
 updatefunc = ClimaLand.make_update_drivers(drivers)
 driver_cb = ClimaLand.DriverUpdateCallback(updateat, updatefunc)
 cb = SciMLBase.CallbackSet(driver_cb, saving_cb)
-sol = @time SciMLBase.solve(prob, ode_algo; dt = dt, saveat = dt, callback = cb)
+sol = @time SciMLBase.solve(
+    prob,
+    ode_algo;
+    dt = dt,
+    saveat = saveat,
```
kmdeck, line 175: I think this is failing because the saveat is at every 100 dt, but we take < 100 steps. We can make saveat =[t0,tf] here.

PR #1020, opened 2025-02-10
Trigger long run from previous point

kmdeck:
starting from this branch https://github.com/CliMA/ClimaLand.jl/pull/998


kmdeck:
We established that the NaNs were not present at this state.

PR #1022, opened 2025-02-10
add to snowy land diags shortlist

juliasloan25:
## Purpose 
Adds LWD, SHF, LHF to snowy land default diagnostics shortlist. We need output of these from the global snowy land longrun to make plots for the paper. Alternatively, we could run the longruns with the option `output_vars = :long` and use the outputs from that run, but I'd rather have these vars output for each longrun, at least while we're working on the paper plots.


kmdeck:
merged in PR #1021

PR #1024, opened 2025-02-11
run with more diagnostics

kmdeck:
## Purpose 
Diagnose ET changes

## Content

----


src/diagnostics/default_diagnostics.jl
```diff
@@ -329,20 +329,23 @@ function default_diagnostics(
             "gpp",
             "ct",
             "lai",
-            "sco2",
```
kmdeck, line 332: this is mostly NaN right now anyways... we can add it back later when we have time to debug it

src/diagnostics/land_compute_methods.jl
```diff
@@ -78,7 +78,20 @@ end
...
+
+function compute_canopy_transpiration!(
+    out,
+    Y,
+    p,
+    t,
+    land_model::Union{SoilCanopyModel{FT}, LandModel{FT}},
+) where {FT}
+    if isnothing(out)
+        return p.canopy.energy.turbulent_fluxes.transpiration .* 1000
```
juliasloan25, line 90: could you add a comment explaining why `.* 1000`?

experiments/long_runs/snowy_land.jl
```diff
@@ -492,7 +492,7 @@ for (group_id, group) in
     CairoMakie.save(joinpath(root_path, "$(group_name).png"), fig)
 end
 
-short_names = ["gpp", "swc", "et", "ct", "swe", "si"]
+short_names = ["gpp", "swc", "et", "ct", "swe", "si", "trans", "msf"]
```
juliasloan25, line 495: we could also add these to `short_names_bio`/`short_names_other` above too if we want the heatmaps to be in the bio/other png outputs

juliasloan25 (approved):
thank you!

src/diagnostics/land_compute_methods.jl
kmdeck: added!

experiments/long_runs/snowy_land.jl
kmdeck: right now i never look at those plots...but all of these are plotted in "figures.pdf"

PR #1026, opened 2025-02-12
change domain depth to 10m to reduce spinup

kmdeck:
## Purpose 
Change domain depth from 50m to 10m to reduce spinup time.

Tapio said that the seasonal cycle penetrates to about 7m in soil. so 10 m is sufficiently deep (and we may need longer than 2 years to spin up).

## To do:
- run for several years and update our spun up initial conditions

## Changes in long run
After:
image
image
image

Before:
image
image
image

----

### DIFF
--- a/experiments/benchmarks/land.jl
+++ b/experiments/benchmarks/land.jl
@@ -12,8 +12,8 @@
 # saved at the bottom of this file
 
 # Simulation Setup
-# Number of spatial elements: 101 in horizontal, 15 in vertical
-# Soil depth: 50 m
+# Number of spatial elements: 101 in horizontal, 10 in vertical
+# Soil depth: 10 m
 # Simulation duration: 6 hours
 # Timestep: 450 s
 # Timestepper: ARS111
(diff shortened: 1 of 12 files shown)


kmdeck:
After talking with @juliasloan25, it seems like the change in ET and GPP from year 1 to year 2 are the same here as when using a 50m domain depth. Therefore, we will close this for now and return to later as needed.

juliasloan25 (approved):
Looks good! Do we want to change `nelements` too?

kmdeck:
> Looks good! Do we want to change `nelements` too?

good call! Ill fix that too

kmdeck:
> After talking with @juliasloan25, it seems like the change in ET and GPP from year 1 to year 2 are the same here as when using a 50m domain depth. Therefore, we will close this for now and return to later as needed.

We have since realized that the deeper soil layers and leaf water potential spin up slowly, leading to the slow spinup in GPP, especially. so, there is an advantage to having a smaller soil domain.

PR #1027, opened 2025-02-13
initial full land calibration

juliasloan25:
## Purpose 
Initial calibration of full land model radiation

## To-do
- [ ] log form for priors?
- [ ] get ICs from saved values at t=1 year (add to ClimaArtifacts)
  - compare sim output, observation data to make sure time is aligned
- [ ] can we run calibration and throw out first few months for spinup?
  - use ClimaAnalysis.window to select dates after spinup, just make sure dates are still aligned

## Content
- [x] update priors
- [x] update loss function

----


kmdeck:
Closing since Alexis incorporated these changes to his branch @juliasloan25

PR #1029, opened 2025-02-14
Spun up IC

kmdeck:
## Purpose 
Add initial conditions for soil and snow. 

These are read from the netcdf file added to clima artifacts: https://github.com/CliMA/ClimaArtifacts/pull/107 and https://github.com/CliMA/ClimaArtifacts/pull/112

While the spin-up in ET seems to be reduced, the GPP spin-up is still there. it is hard to diagnose what state variable are not spun up yet to cause this. This also adds in diagnostics for lwp and total water content, to test what is leading to the slow spinup of GPP.

I also removed an allocation in the LAI diagnostic, and removed some diagnostics that we never use to test if have fewer makes it run faster

## Content

----

### DIFF
--- a/experiments/long_runs/snowy_land.jl
+++ b/experiments/long_runs/snowy_land.jl
@@ -1,7 +1,7 @@
 # # Global run of land model
 
 # The code sets up and runs ClimaLand v1, which
-# includes soil, canopy, and snow, for 730 days on a spherical domain,
+# includes soil, canopy, and snow, on a spherical domain,
 # using ERA5 data as forcing. In this simulation, we have
 # turned lateral flow off because horizontal boundary conditions and the
 # land/sea mask are not yet supported by ClimaCore.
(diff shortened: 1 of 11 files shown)


experiments/long_runs/snowy_land.jl
```diff
@@ -1,7 +1,7 @@
 # # Global run of land model
 
 # The code sets up and runs ClimaLand v1, which
-# includes soil, canopy, and snow, for 730 days on a spherical domain,
+# includes soil, canopy, and snow,  on a spherical domain,
```
ph-kev, line 4: Can we remove the extra space here?

ph-kev (approved):

src/diagnostics/default_diagnostics.jl
```diff
@@ -328,21 +328,16 @@ function default_diagnostics(
         snowyland_diagnostics = [
             "gpp",
             "ct",
-            "lai",
```
kmdeck, line 331: I removed ones we never look at - since Gabriele said that adding diagnostics can slow things down

AlexisRenchon (approved):
This is really great to have! and nice to see more content in src/simulations. Thank you Kat!

PR #1030, opened 2025-02-14
Add ITime

ph-kev:
This PR adds ITime to the ClimaLand simulations.


src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -343,7 +344,8 @@ function ClimaLand.Canopy.set_canopy_prescribed_field!(
     t,
 ) where {FT}
     (; LAIfunction, SAI, RAI) = component.parameters.ai_parameterization
-    evaluate!(p.canopy.hydraulics.area_index.leaf, LAIfunction, floor(t))
+    t = t isa ITime ? t : floor(t)
```
Sbozzolo, line 347: I think you can remove the `floor` here. It's probably there for historical reasons

Sbozzolo (approved):
Amazing, thank you!

PR #1037, opened 2025-02-20
Single turbulent fluxes call for soil

kmdeck:
## Purpose 
Simplify the turbulent flux computation from soil.

Previously, we computed surface fluxes 2x, once for soil water and once for soil ice. These have the same temperature, but different specific humidities. What is easier and probably makes more sense it compute a single specific humidity and compute surface fluxes once, since that is more physically realistic. This PR makes it so we use the saturated humidity over ice or liquid water depending on the temperature of the soil.

If the temperature is above freezing, liquid water is evaporating, and we apply an extra factor to account for dry soil resistance to evaporation. If the temperature is below freezing, frozen water is sublimating.

## Changes to long runs
The long run from this PR: https://buildkite.com/clima/climaland-long-runs/builds/3533
- does not affect or only improves stability (8 NaNs at the end vs 92)
Figures look indistinguishable from current main

### DIFF
--- a/docs/tutorials/standalone/Soil/evaporation.jl
+++ b/docs/tutorials/standalone/Soil/evaporation.jl
@@ -62,7 +62,7 @@ e = rh * esat
 q = FT(0.622 * e / (101325 - 0.378 * e))
 precip = (t) -> 0.0
 T_atmos = (t) -> T_air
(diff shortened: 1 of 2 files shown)


szy21 (approved):
The code change looks good to me.

PR #1040, opened 2025-02-20
add mask to NaN checker

kmdeck:
## Purpose 
Adds land/sea mask that will be used Clima wide
uses this in the nanchecker
fixes a bug with the masked nanchecker where a 2d mask couldnt be used with a 3d field

----

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -374,6 +374,33 @@ function ilamb_dataset_path(filename; context = nothing)
 end
 
 """
+    landseamask_file_path(; resolution = "60arcs", context=nothing)
+
+Construct the file path for the landsea mask data NetCDF file,
+using 60arcseconds by default.
+
+The other options available include "30arcs" (30arcseconds) or "1deg" (1degree).
+"""
+function landseamask_file_path(; resolution = "60arcs", context = nothing)
+    @assert resolution in ["60arcs", "30arcs", "1deg"]
+    filename = "landsea_mask.nc"
+    if resolution == "60arcs"
+        return joinpath(
+            @clima_artifact("landsea_mask_60arcseconds", context),
+            filename,
+        )
+    elseif resolution == "30arcs"
+        return joinpath(
+            @clima_artifact("landsea_mask_30arcseconds", context),
+            filename,
+        )
+    else
+        return joinpath(@clima_artifact("landsea_mask_1deg", context), filename)
+    end
+
+end
+
+"""
     earth_orography_file_path(; context=nothing)
 
 Construct the file path for the 60arcsecond orography data NetCDF file.
--- a/src/shared_utilities/utils.jl
+++ b/src/shared_utilities/utils.jl
@@ -3,7 +3,13 @@ import SciMLBase
 import ClimaDiagnostics.Schedules: EveryCalendarDtSchedule
 import Dates
 
-export FTfromY, count_nans_state
+using ClimaComms
+using ClimaUtilities.ClimaArtifacts
+import Interpolations
+import ClimaUtilities.SpaceVaryingInputs: SpaceVaryingInput
+import ClimaUtilities.Regridders: InterpolationsRegridder
+
+export FTfromY, call_count_nans_state
 
 """
      heaviside(x::FT)::FT where {FT}
@@ -494,25 +500,27 @@ function isdivisible(
 end
 
 """
-    count_nans_state(state, mask::ClimaCore.Fields.Field = nothing, verbose = false)
+    call_count_nans_state(state, mask::ClimaCore.Fields.Field = nothing, verbose = false)
 
-Count the number of NaNs in the state, e.g. the FieldVector given by
+This calls the function which counts the number of NaNs in the state, e.g. the FieldVector given by
 `sol.u[end]` after calling `solve`. This function is useful for
-debugging simulations to determine quantitatively if a simulation is stable.
+debugging simulations to determine quantitatively if a simulation is stable; this is only
+for use in non-column simulations.
 
 If this function is called on a FieldVector, it will recursively call itself
 on each Field in the FieldVector. If it is called on a Field, it will count
-the number of NaNs in the Field and produce a warning if any are found.
+the number of NaNs in the Field and produce a warning if any are found; this behavior
+is different for 2d and 3d fields, which is why we dispatch on the `space`.
 
 If a ClimaCore Field is provided as `mask`, the function will only count NaNs
 in the state variables where the mask is 1. This is intended to be used with
 the land/sea mask, to avoid counting NaNs over the ocean. Note this assumes
-the mask is 1 over land and 0 over ocean.
+the mask is 1 over land and 0 over ocean, and that the mask is a 2D field.
 
 The `verbose` argument toggles whether the function produces output when no
 NaNs are found.
 """
-function count_nans_state(
+function call_count_nans_state(
     state::ClimaCore.Fields.FieldVector;
     mask = nothing,
     verbose = false,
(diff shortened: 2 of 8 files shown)


src/shared_utilities/utils.jl
```diff
@@ -529,6 +529,15 @@ function count_nans_state(
     state::ClimaCore.Fields.Field;
     mask = nothing,
     verbose = false,
+)
+    return count_nans_state(state, axes(state); mask = mask, verbose = verbose)
+end
+
+function count_nans_state(
+    state::ClimaCore.Fields.Field,
+    space::ClimaCore.Spaces.AbstractSpectralElementSpace;
```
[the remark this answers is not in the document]
kmdeck: space is just used for dispatching between 3d and 2d fields

src/Artifacts.jl
```diff
@@ -373,6 +373,20 @@ function ilamb_dataset_path(filename; context = nothing)
...
+"""
+    landseamask_file_path(; resolution = 30, context=nothing)
+
+Construct the file path for the 30arcsecond landsea mask data NetCDF file.
+
+"""
+function landseamask_file_path(; context = nothing)
+    filename = "landsea_mask.nc"
+    return joinpath(
+        @clima_artifact("landsea_mask_30arcseconds", context),
```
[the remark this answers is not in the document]
kmdeck: ok! Ill switch back to that. that failed on the central cluster for some reason

src/shared_utilities/utils.jl
```diff
@@ -542,6 +564,36 @@ function count_nans_state(
...
+    state::ClimaCore.Fields.Field,
+    space::ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace;
+    mask = nothing,
+    verbose = false,
+)
+    # Note: this code uses `parent`; this pattern should not be replicated
+    surface_state = ClimaLand.Domains.top_center_to_surface(state)
+    num_nans =
+        isnothing(mask) ? round(sum(isnan, parent(surface_state))) :
+        round(sum(isnan, parent(surface_state)[Bool.(parent(mask))]; init = 0))
```
kmdeck, line 588: @juliasloan25 does this need to be a regular array to support the indexing?  not sure why that would affect the gpu long runs, though, and not the GPU unit tests.

src/shared_utilities/utils.jl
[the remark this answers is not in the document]
juliasloan25: I see. A lot of code is duplicated between the two methods, with the exception of calling vs not calling `top_center_to_surface`. Could you change the interface of the top-level `count_nans_state` (line 530) to take in the space and dispatch off of it there instead? In the 2D case it can call `count_nans_state(state; mask, verbose)`, and in the 3D case it can call `count_nans_state(top_center_to_surface(state); mask, verbose)`, where this function contains the code that's used in the 2D case currently

src/simulations/spatial_parameters.jl
```diff
@@ -8,6 +8,60 @@ function mean(x::AbstractArray{T}) where {T}
...
+    regridder_type = :InterpolationsRegridder,
+    extrapolation_bc = (
+        Interpolations.Periodic(),
+        Interpolations.Flat(),
+        Interpolations.Flat(),
+    ),
+)
+    @assert (resolution == "60arcs") ||
+            (resolution == "30arcs") ||
+            (resolution == "1deg")
```
juliasloan25, line 47: we can leave this out here since it's already checked in `landseamask_file_path`

src/simulations/spatial_parameters.jl
```diff
@@ -8,6 +8,60 @@ function mean(x::AbstractArray{T}) where {T}
...
+    )
+
+Reads in the default Clima 60arcsecond land/sea mask, regrids to the
+`surface_space`, and treats any point with a land fraction < threshold
+as ocean.
+
+A 1degree (resolution = "1deg") and 30arcsecond (resolution = "30arcs") mask
+are also available.
+"""
+function landsea_mask(
```
juliasloan25, line 34: I wonder if this should be in utils.jl instead, since it could apply to other fields in addition to spatial parameters

juliasloan25 (commented):
Thanks Kat, this looks great! I left some suggestions about restructuring the code, let me know if anything is unclear

src/shared_utilities/utils.jl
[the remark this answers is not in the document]
kmdeck: good idea! Ill make this change

src/Artifacts.jl
```diff
@@ -373,6 +373,35 @@ function ilamb_dataset_path(filename; context = nothing)
...
+
+Construct the file path for the landsea mask data NetCDF file,
+using 60arcseconds by default.
+
+The other options available include "30arcs" (30arcseconds) or "1deg" (1degree).
+"""
+function landseamask_file_path(; resolution = "60arcs", context = nothing)
+    @assert (resolution == "60arcs") ||
+            (resolution == "30arcs") ||
+            (resolution == "1deg")
```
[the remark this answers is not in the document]
kmdeck: much nicer, thanks!

src/simulations/spatial_parameters.jl
kmdeck: ok! Ill move it there

src/shared_utilities/utils.jl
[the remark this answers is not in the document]
kmdeck: ah, I realized that we cant do this, because we are not guaranteed that the property of the FieldVector will be a Field. It could be another FieldVector, in which case, we could not call axes()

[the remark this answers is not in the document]
kmdeck: I used an if statement instead of dispatch, but it reduces the duplicated code, and throws a more meaningful error if the mask space is different from the 2d space of the state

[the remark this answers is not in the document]
kmdeck: Ok, I got around this by defining the first function as "call_count_nans_state" vs the "count_nans_state" function itself. the first handles the field/field vector dispatching, and does the surface state vs state choice based on an if statement.

let me know what you think!

src/shared_utilities/utils.jl
```diff
@@ -494,46 +500,84 @@ function isdivisible(
...
+    verbose = false,
+)
+    if typeof(axes(state)) <: ClimaCore.Spaces.AbstractSpectralElementSpace
+        return count_nans_state(state; mask = mask, verbose = verbose)
+    elseif typeof(axes(state)) <: ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace
+        return count_nans_state(
+            ClimaLand.top_center_to_surface(state);
+            mask = mask,
+            verbose = verbose,
+        )
```
juliasloan25, line 548: Instead of the if/else here, we could pass `axes(state)` to `count_nans_state`, since now we know `state` is a Field. Then we can dispatch on the space type and call `top_center_to_surface(state)` in the 3D case

src/shared_utilities/utils.jl
```diff
@@ -494,46 +500,84 @@ function isdivisible(
...
+    elseif mask isa Nothing
+        count_nans_state(state; verbose = verbose)
```
juliasloan25, line 550: I also think we can omit this check - if the mask is nothing we can just pass it through to the function normally and it should be ignored

src/shared_utilities/utils.jl
```diff
@@ -494,46 +500,84 @@ function isdivisible(
...
+    else
+        @error(
+            "Cannot use a mask in NaN counting unless your field is 2d or 3d."
+        )
```
juliasloan25, line 554: With the comments above we can remove this case too. Trying to do this would raise a method error that nothing is defined for this space type

src/shared_utilities/utils.jl
kmdeck: ok! could you clarify how this is different from what I had before?

kmdeck: If we call count_nans_state with the space, we end up having to duplicate the counting code. But we can call call_count_nans_state with the space, and then call count_nans_state. I guess this is a combo of the previous + current. let me try it!

juliasloan25 (approved):
Thank you!!

PR #1042, opened 2025-02-21
Remove `minimum` and close Issue #975

kmdeck:
## Purpose 
Removes allocation of parameter combo struct Issue #975 
Closes PR #982 
Removes "minimum" computation by computing this once and storing it

## To-do
update benchmark times

## Content

----

### DIFF
--- a/src/shared_utilities/Domains.jl
+++ b/src/shared_utilities/Domains.jl
@@ -111,7 +111,7 @@ struct Column{FT} <: AbstractDomain{FT}
     boundary_names::Tuple{Symbol, Symbol}
     "A NamedTuple of associated ClimaCore spaces: in this case, the surface space and subsurface center space"
     space::NamedTuple
-    "Fields associated with the coordinates of the domain that are useful to store"
+    "Fields and field data associated with the coordinates of the domain that are useful to store"
     fields::NamedTuple
 end
 
@@ -171,7 +171,7 @@ function Column(;
     end
     surface_space = obtain_surface_space(subsurface_space)
     space = (; surface = surface_space, subsurface = subsurface_space)
-    fields = get_additional_domain_fields(subsurface_space)
+    fields = get_additional_coordinate_field_data(subsurface_space)
     return Column{FT}(
         zlim,
         (nelements,),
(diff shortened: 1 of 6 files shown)


src/standalone/Soil/soil_hydrology_parameterizations.jl
```diff
@@ -85,10 +85,9 @@ function update_albedo!(bc::AtmosDrivenFluxBC, p, soil_domain, model_parameters)
         )
         @. S_sfc = ∫H_S_e_dz / ∫H_dz
     else
+        @. p.soil.sub_sfc_scratch = effective_saturation(ν, p.soil.θ_l, θ_r)
```
kmdeck, line 88: this removes an allocation too, but I dont think this branch is being used with our current config

src/shared_utilities/Domains.jl
```diff
@@ -926,13 +926,15 @@ function get_additional_domain_fields(subsurface_space)
...
     d = depth(subsurface_space)
+    Δz_min = minimum(Δz_top)
     fields = (;
         z = z,
         Δz_top = Δz_top,
         Δz_bottom = Δz_bottom,
         z_sfc = z_sfc,
         depth = d,
         Δz = Δz,
+        Δz_min = Δz_min,
```
Sbozzolo, line 937: `Δz_min` here is in `fields`. Is it a Field?

Sbozzolo (approved):

src/shared_utilities/Domains.jl
kmdeck: no :( but it was a handy place to stick it. maybe I can rename "fields" to "field_data"?

Sbozzolo: To me it looks like it should be related to space/domain

kmdeck: Yes, all of these are related to the domain/space. We have domain.space.surface, domain.space.subsurface, domain.fields, etc. The fields are all related to the coordinate field.  So it's just a question of where to stick this additional precomputed quantitity. since it is related to coordinates, I think sticking it in fields is OK, but the name is misleading.

kmdeck: I renamed the function to "get_additional_coordinate_field_data" and clarify that `fields` holds both fields and field data.

Sbozzolo: Sounds good!

Sbozzolo (approved):

PR #1043, opened 2025-02-21
fix bug with iwc name

kmdeck:
## Purpose 
Our diagnostics name "iwc_1m" fails in plotting because it is thought to be "iwc" at 1minute. This changes the name by removing "_1m".

## Content

----


Sbozzolo (approved):
You can also call it `iwc1m` is you want to preserve the information

PR #1044, opened 2025-02-21
Correct Vcmax and Rd for C4 photosynthesis

kmdeck:
## Purpose 
Correct Vcmax, light limited rate, and Rd for C4 photosynthesis

## Content
c3/c4 specific Vcmax and Rd functions

----


AlexisRenchon (approved):
This looks good to me! Thank you

ext/CreateParametersExt.jl
```diff
@@ -164,6 +164,18 @@ function FarquharParameters(
     )
     parameters = CP.get_parameter_values(toml_dict, name_map, "Land")
     FT = CP.float_type(toml_dict)
+    C4_parameters = (
+        :fC4 => FT(0.025),
```
braghiere, line 168: Good catch!

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -557,29 +557,34 @@ end
...
-the rate of electron transport (`J`).
+in units of moles CO2/m^2/s.
+
+For C4 plants, this is a function of APAR and a efficiency parameter E, see Equation 11.70 of
+ G. Bonan's textbook, Climate Change and Terrestrial Ecosystem Modeling (2019).
 """
-function c4_light_assimilation(J::FT, _...) where {FT}
-    Aj = J
+function c4_light_assimilation(::FT, ::FT, ::FT, APAR::FT, E::FT) where {FT}
+    Aj = APAR * E
```
braghiere, line 587: I see you changed it to what is in Bonan's book, where E = 0.05 mol.mol-1, which is what is done in CLM5. 
Screenshot 2025-02-27 at 10 38 21 AM

But in JULES, 
Screenshot 2025-02-27 at 10 39 08 AM

I think we took Aj = J from my implementation in CARDAMOM, but since we're not doing exactly like in Bonan, e.g., our co-limitation is simply the min between Aj and Ac, instead of co-limitation function given by 11.33 in Bonan, I think we should follow the previous one to be consistent.

braghiere (commented):
Everything looks good to me, not sure about the Aj = J vs. Aj = APAR *E. I left a comment there.

src/standalone/Vegetation/canopy_parameterizations.jl
kmdeck: Hmm let's discuss a bit more.

Current Clima: J * factor; C4: J
JULES: efficiency *APAR * factor, C4: efficency * APAR
CLM: J*factor, C4: efficency * APAR

It seems like the current approach is inconsistent with CLM and JULES, because J is not efficiency *APAR (it is the root of a quadratic equation, involving those things, but not quite the same). I'm not sure how adding another limiting rate changes how we compute this limiting rate, so that we differ from CLM in that seems unrelated.

braghiere: I guess it’s ok for now. It’s a Frankenstein model anyway :) hopefully quantum yield is another parameter we could calibrate. BTW, we just have to make sure quantum yield is consistently used across the model, e.g., in SIF.

Issue #1038, opened 2025-02-20
Changes needed for full land model coupling with ClimaAtmos


juliasloan25:
We need the following changes to ClimaLand source code so we can couple the full land model with the ClimaAtmos model: 
- move `diffuse_fraction` update out of `update_aux!` and into a function that dispatches on radiation driver type
  - in the `CoupledAtmosphere` case, this will be updated by the coupler so we don't need to do anything in ClimaLand
- functions to compute `q`, `\rho` from land, atmos inputs, and to return land thermo state
  - function signatures like: `get_\rho(T_land, T_atmos, \rho_atmos)`, `get_q(T_land, \rho_land)`
  - call these functions inside land turbulent flux calculation function? already has access to atmos thermo state
- use cos(zenith angle) instead of zenith angle (`\theta_s`) so we don't have to convert atmos value using arccos
  - this isn't strictly required for coupling, but will make things cleaner :)

kmdeck:
@juliasloan25 would we like this"function signatures like: get_\rho(T_land, T_atmos, \rho_atmos), get_q(T_land, \rho_land)" to return a tuples (one element of the tuple for each land component), and for the bucket to return a tuple of 1, or for bucket to return a scalar?

PR #1046, opened 2025-02-25
Update the diffuse fraction in drivers; use cosine of zenith angle; unneeded h access

kmdeck:
## Purpose 
Partly address Issue #1038 
- Use cosine (zenith angle) instead of zenith angle so streamline with atmos
- update the diffuse fraction in `update_drivers`, store in p.drivers,  rather than storing in canopy aux and updating in canopy functions.
- remove access of atmos.h in Canopy update aux


src/shared_utilities/drivers.jl
```diff
@@ -995,12 +1006,34 @@ function make_update_drivers(r::PrescribedRadiativeFluxes{FT}) where {FT}
...
+                p.drivers.P,
+                p.drivers.q,
+                p.drivers.SW_d,
+                p.drivers.cosθs,
+                r.thermo_params,
+            )
         else
-            p.drivers.θs .= FT(0)
+            p.drivers.cosθs .= FT(0)
+            p.drivers.frac_diff .= FT(0)
```
imreddyTeja, line 1033: Is this behavior different than before? It seems like to replicate the old behavior, this should set
`p.drivers.cos\thetas .= FT(1)`  and still update `p.drivers.frac_diff` with a call to `diffuse_fraction`

src/diagnostics/define_diagnostics.jl
```diff
@@ -560,17 +560,6 @@ function define_diagnostics!(land_model)
             compute_snowfall!(out, Y, p, t, land_model),
     )
 
-    # Solar zenith angle
-    add_diagnostic_variable!(
```
imreddyTeja, line 564: If we still might use this diagnostic, we could keep this and define `compute_solar_zenith_angle!` to take the inverse cos of `p.drivers.\thetas`

src/shared_utilities/drivers.jl
```diff
@@ -995,12 +1006,34 @@ function make_update_drivers(r::PrescribedRadiativeFluxes{FT}) where {FT}
...
+            p.drivers.cosθs .= FT.(cos.(r.θs(t, r.start_date)))
+            @. p.drivers.frac_diff = diffuse_fraction(
+                t,
+                r.start_date,
+                p.drivers.T,
+                p.drivers.P,
+                p.drivers.q,
+                p.drivers.SW_d,
+                p.drivers.cosθs,
+                r.thermo_params,
```
imreddyTeja, line 1029: We either need to check that `!isnothing(r.thermo_params)` here, or make it a required argument with no default in the `PrescribedRadiativeFluxes` constructor (similar to what `PrescribedAtmosphere` does)

src/shared_utilities/drivers.jl
```diff
@@ -450,8 +451,17 @@ struct PrescribedRadiativeFluxes{
...
-        args = (SW_d, LW_d, start_date, θs)
+    "Thermodynamic parameters"
+    thermo_params::TP
+    function PrescribedRadiativeFluxes(
+        FT,
+        SW_d,
+        LW_d,
+        start_date;
+        θs = nothing,
+        thermo_params = nothing,
```
imreddyTeja, line 462: The `PrescribedAtmosphere` constructor takes in `earth_param_set` and then extracts `thermo_params` from it. Both of these are fine, but I was wondering if the `PrescribedAtmosphere` constructor was designed that way for a reason.

imreddyTeja (commented):
Looks good. I just had a couple of comments and questions. Thanks!

src/shared_utilities/drivers.jl
kmdeck: hmmm the only reason I can think of is that in PA it then kind of hides the thermo_param extraction from the user (every simulation needs the earth param set, so they have that handy in their script already, but not all need the thermo_params). It's minor but I can make the change so that the two are the same.

src/shared_utilities/drivers.jl
kmdeck: The reason why I dont want to make it required is because the zenith angle and diffuse fraction are only required by the canopy model right now (and integrated runs). But e.g. soil standalone doesnt need these. And we cannot compute diffuse fraction without zenith angle. Ill adopt the check you suggested.

src/shared_utilities/drivers.jl
kmdeck: hmmm youre right, but we shouldnt be calling the diffuse fraction with a constant zenith angle. The two are used together or not used. Maybe NaN is better indicate that they are not being used..?

src/diagnostics/define_diagnostics.jl
kmdeck: We are not using it now, and I dont really forsee that we need it, so I removed it. It is simple to add if we need to.

src/shared_utilities/drivers.jl
imreddyTeja: I think it's fine to leave it as is.

src/shared_utilities/drivers.jl
imreddyTeja: This is fine then. Another option is just not updating them at all.

I thought the zenith angle is used for calculations other than the diffuse angle. If so, then does changing the cosine zenith angle from 1 to zero matter?

kmdeck: hmmm this is a little confusing and not great.

the zenith angle is used for things other than the diffuse fraction in the canopy model. However, if the zenith angle is needed at all, it should not be `nothing`. so we shouldnt ever use the value of p.drivers.costhetas in a simulation if zenith angle is nothing. For example, the soil model only needs SW_d and LW_d, and never accesses p.drivers.costhetas. The canopy model needs zenith angle, so it cant be nothing, because it will access p.drivers.costhetas. Thinking about this, now, this is kind of dangerous, because someone could accidentally set up the canopy model without a zenith angle and get a zenith angle that does not change in time. 

because of that, I think I will update them to NaN, so it more clearly fails if someone accesses zenith angle but did not provide a way to update it...

PR #1047, opened 2025-02-25
Make fields of struct concrete instead of abstract

ph-kev:
This PR makes fields of structs concrete instead of abstract type.


kmdeck:
@ph-kev  even if this didnt make a huge difference in performance, should we merge it?

ph-kev:
> @ph-kev even if this didnt make a huge difference in performance, should we merge it?

I think this is a breaking change to ClimaCoupler, so maybe merge it in with the other breaking changes?

Sbozzolo:
It would be good to merge this

Sbozzolo:
Could you open the corresponding PR in ClimaCoupler and fix the compatibility there?

Sbozzolo (approved):

PR #1049, opened 2025-02-26
calibration refactor

AlexisRenchon:
will write something very soon


Espeer5:
What is the issue mentioned here in item one of the todo list with the code for the monthly variances?

AlexisRenchon:
> What is the issue mentioned here in item one of the todo list with the code for the monthly variances?

there was an Artifact problem, I had to comment out https://github.com/CliMA/ClimaLand.jl/blob/ar/ct/bucket_calibration/src/Artifacts.jl#L60 or ClimaLand wouldn't compile. I haven't looked at this for a few days

AlexisRenchon:
@Sbozzolo @kmdeck @nefrathenrici @odunbar @juliasloan25 

I know this is a lot of reviewers, I wanted you all to have an opportunity to look and comment this important PR. 
I suggest we try to merge this, but we can fix some things if needed, and make a good to do list for what's next (I started it in the comment above).

Espeer5:
> > What is the issue mentioned here in item one of the todo list with the code for the monthly variances?
> 
> there was an Artifact problem, I had to comment out https://github.com/CliMA/ClimaLand.jl/blob/ar/ct/bucket_calibration/src/Artifacts.jl#L60 or ClimaLand wouldn't compile. I haven't looked at this for a few days

On which system? I've tested this on the HPC, maybe it's because the artifact isn't available on Derecho? If you can show me an error message I can check it out.

experiments/calibration/shared/make_training_locations.jl
```diff
@@ -0,0 +1,19 @@
+# Define the locations longitudes and latitudes
+# Needs to be defined once, both for g(θ) and ERA5 target
+t0 = 0.0
+tf = 60 * 60.0 * 24 * 366
+Δt = 900.0
+nelements = (50, 5)
```
odunbar, line 6: I'm concerned about the redefinitions of these objects. Especially given that they are supposed to be in a "Shared" file, they are redefined within the experiments.

experiments/calibration/global_bucket/forward_model.jl
```diff
@@ -0,0 +1,176 @@
...
+    params_path = parameter_path(caldir, iteration, member)
+    params = toml.parsefile(params_path) # should load a Dict, that needs to be converted to namedtuple
+
+    diagnostics_dir = joinpath(ensemble_member_path, "global_diagnostics")
+    diagdir =
+        ClimaUtilities.OutputPathGenerator.generate_output_path(diagnostics_dir)
+    t0 = 0.0
+    tf = 60 * 60.0 * 24 * 366
+    Δt = 900.0
+    nelements = (50, 5)
```
odunbar, line 167: I'm concerned about the redefinitions of these objects. Especially given that they are also defined in a "Shared" file, they are redefined within the experiments.

experiments/calibration/global_bucket/observation_map.jl
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    # Combine all observations into a single observation
+    full_obs = EKP.combine_observations(obs_list)
+    obs = EKP.get_obs(full_obs)
+    return obs
+end
+
+function CAL.observation_map(iteration)
+    single_member_dims = (2400,) # 2 var * 12 months * 100 locations
```
odunbar, line 44: Is this not inferrable?

experiments/calibration/global_land/observation_map.jl
```diff
@@ -0,0 +1,42 @@
...
+                ),
+            ).data
+        push!(vars_global_average, var_global_average)
+    end
+
+    return vcat(vars_global_average...)
+end
+
+function CAL.observation_map(iteration)
+    single_member_dims = (4 * 12,)
```
odunbar, line 26: Is this not inferrable

experiments/calibration/shared/observation_data_global.jl
```diff
@@ -0,0 +1,26 @@
...
+    var_global_average =
+        ClimaAnalysis.average_lon(
+            ClimaAnalysis.weighted_average_lat(
+                ClimaAnalysis.apply_oceanmask(var),
+            ),
+        ).data
+    push!(observations_vecs, var_global_average)
+end
+observations = vcat(observations_vecs...)
+observations[1:24] .*= -1 # different sign convention for lhf and shf
```
odunbar, line 26: `1:24` this is a very likely source of human error in future, if any resolutions change etc.

experiments/calibration/shared/observation_data_locations.jl
```diff
@@ -0,0 +1,41 @@
...
+
+# Make the ERA5 target
+ERA5_target = []
+close = (x, y) -> abs(x - y) < 5e-1
+for (lon, lat) in training_locations
+    # Fetch slices of lhf and shf era5 data from the era5 dataset
+    lat_ind, lon_ind = findall((x) -> close(x, lat), era5_ds["latitude"][:])[1],
+    findall((x) -> close(x, lon + 180), era5_ds["longitude"][:])[1]
+    lhf_loc = vec(era5_ds["mslhf"][lon_ind, lat_ind, :][:, end, :])
+    shf_loc = vec(era5_ds["msshf"][lon_ind, lat_ind, :][:, end, :])
```
odunbar, line 16: Just a check on the relative grid resolutions of the data and the forward model. Do we have both ERA5 and CL interpolated (via local averaging for example) to the same lat-lon grid here?

otherwise if the CL grid is coarser than the ERA5 grid, the value from direct slicing here at a specfic lat-lon in ERA5  will represent a different quantity than CL run at that lat-lon

odunbar (commented):
Looks like a great start!  I have not run it yet - I will try this later! 

I left a couple of comments, 
- Primarily i am concerned that there are a lot of floating/locally defined values that i think may well become sources of human error down the line. 
- I also wanted to check that the interpolations between the data and CL observation maps are consistent. (e.g. do we account for the CL column at a lat-lon site and the ERA5 column at a lat-lon site having different area)

experiments/calibration/global_bucket/calibrate_bucket.jl
```diff
@@ -0,0 +1,72 @@
...
+using ClimaLand
+dir = pkgdir(ClimaLand)
+import ClimaCalibrate: forward_model, parameter_path, path_to_ensemble_member
+import ClimaCalibrate as CAL
+using Distributed
+addprocs(
+    CAL.PBSManager(2),
+    q = "main", # not prio
+    A = "UCIT0011",
+    l_select = "ngpus=4",
```
nefrathenrici, line 12:
```suggestion
    l_select = "ngpus=1:ncpus=4",
```
I think you only need one GPU per task, but I may be mistaken.

experiments/calibration/global_bucket/forward_model.jl
```diff
@@ -0,0 +1,176 @@
...
+        nelements = nelements,
+        npolynomial = 1,
+        dz_tuple = FT.((1.0, 0.05)),
+    )
+    surface_space = domain.space.surface
+    subsurface_space = domain.space.subsurface
+
+    start_date = DateTime(2008)
+    # Forcing data
+    # Forcing data
```
nefrathenrici, line 65:
```suggestion
```
This comment is duplicated

experiments/calibration/global_bucket/forward_model.jl
```diff
@@ -0,0 +1,176 @@
...
+    diag_cb = ClimaDiagnostics.DiagnosticsCallback(diagnostic_handler)
+
+    driver_cb = ClimaLand.DriverUpdateCallback(updateat, updatefunc)
+    return prob, SciMLBase.CallbackSet(driver_cb, diag_cb), nc_writer
+end
+
+# Explain things here - tutorial style
+"""
+
+"""
```
nefrathenrici, line 155:
```suggestion
```
I assume you did this in the documentation PR, this could be removed

experiments/calibration/global_land/calibrate_land.jl
```diff
@@ -0,0 +1,63 @@
...
+using ClimaLand
+dir = pkgdir(ClimaLand)
+import ClimaCalibrate: forward_model, parameter_path, path_to_ensemble_member
+import ClimaCalibrate as CAL
+using Distributed
+addprocs(
+    CAL.PBSManager(2),
+    q = "main", # not prio
+    A = "UCIT0011",
+    l_select = "ngpus=4",
```
nefrathenrici, line 12: Same comment as above.

src/Artifacts.jl
```diff
@@ -57,6 +57,16 @@ function era5_lai_covers_data_folder_path(; context = nothing)
...
+#"""
+#    era5_surface_data_1979_2024_path(; context)
+#
+#Return the path to the folder that contains the ERA5 monthly surface data
+#from 1979 to 2024.
+#"""
+#function era5_surface_data_1979_2024_path(; context = nothing)
+#    return @clima_artifact("era5_monthly_averages_surface_single_level_1979_2024", context)
+#end
+
```
nefrathenrici, line 69: This snippet could be removed.

nefrathenrici (commented):
This PR looks good to me, I just left a few comments. I want to echo @odunbar 's concerns about centralizing the experiment configuration, and more systematic observation processing for future experiments.

experiments/calibration/README.md
```diff
@@ -0,0 +1,120 @@
...
+│   ├── calibrate_bucket.pbs
+│   └── calibrate_land.pbs
+└── shared
+    ├── make_training_locations.jl
+    ├── observation_data_global.jl
+    ├── observation_data_locations.jl
+    └── rand_locations.jl
+```
+
+## calibrate\_bucket.jl and calibrate\_land.jl
```
Sbozzolo, line 27:
```suggestion
## `calibrate_bucket.jl` and `calibrate_land.jl`
```

experiments/calibration/README.md
```diff
@@ -0,0 +1,120 @@
...
+    ensemble_size,
+    n_iterations,
+    observations,
+    noise,
+    prior,
+    caldir,
+)
+```
+
+### forward\_model
```
Sbozzolo, line 44:
```suggestion
### `forward_model`
```

experiments/calibration/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+to work, a `forward_model(iteration, member)` function needs to be defined (in forward\_model.jl), this will run
+the model (bucket or land) and save the diagnostics.
```
Sbozzolo, line 47:
```suggestion
`forward_model.jl` defines `forward_model(iteration, member)`, the main function that determines the model being run (e.g., bucket or land) and saves the output diagnostics.
```

experiments/calibration/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+### observation\_map
```
Sbozzolo, line 49:
```suggestion
### `observation_map.jl`
```

experiments/calibration/global_bucket/calibrate_bucket.jl
```diff
@@ -0,0 +1,72 @@
+ENV["JULIA_WORKER_TIMEOUT"] = "300.0"
```
Sbozzolo, line 1: Please add a comment to explain this

experiments/calibration/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+Furthermore, an `observation_map(iteration)` function needs to be defined (in observation\_map.jl), this will
+generate your loss from the model output, in the same format as your observations from data, contained in shared/observation\_data.
+For each member (parameters sample), a vector will be created from the function `process_member_data(root_path)` located in the script
+observation\_map.jl.
```
Sbozzolo, line 54:
```suggestion
`observation_map.jl` defines `observation_map(iteration)`, a function that generate the loss from the model output. The loss function has to have the same format as your observations from data, contained in `shared/observation_data.jl`.
For each member (parameters sample), a vector will be created from the function `process_member_data(root_path)` located in the script
`observation_map.jl`.
```

experiments/calibration/README.md
```diff
@@ -0,0 +1,120 @@
...
+│   │   │   │   ├── lhf_1M_average.nc
+│   │   │   │   ├── lwu_1M_average.nc
+│   │   │   │   ├── shf_1M_average.nc
+│   │   │   │   └── swu_1M_average.nc
+│   │   │   └── output_active -> output_0000
+│   │   └── parameters.toml
+```
+
+each iteration contains folders for each member, inside which you can find the parameters value inside parameters.toml, and forward\_model outputs
+inside global\_diagnostics.
```
Sbozzolo, line 108:
```suggestion
each iteration contains folders for each member, inside which you can find the parameters value inside `parameters.`toml, and `forward_model` outputs
inside `global_diagnostics.jl`.
```

experiments/calibration/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Running jobs on Derecho
+
+To run a calibration on Derecho, run the .pbs job scripts contained in experiments/calibration/job\_scripts.
```
Sbozzolo, line 112:
```suggestion
To run a calibration on Derecho, submit the .pbs job scripts contained in `experiments/calibration/job_scripts`.
```

experiments/calibration/global_bucket/forward_model.jl
```diff
@@ -0,0 +1,176 @@
+import TOML as toml
```
Sbozzolo, line 1: Why did you need to do this?

experiments/calibration/global_bucket/observation_map.jl
```diff
@@ -0,0 +1,59 @@
...
+    lhf = get(simdir; short_name = "lhf")
+    shf = get(simdir; short_name = "shf")
+
+    # Initialize an empty list to store observations
+    obs_list = []
+    # Loop over each location
+    for (lon, lat) in training_locations
+        # Slice lhf and shf at the given longitude and latitude
+        lhf_loc = ClimaAnalysis.slice(lhf, lon = lon, lat = lat)
+        shf_loc = ClimaAnalysis.slice(shf, lon = lon, lat = lat)
```
Sbozzolo, line 14:
```suggestion
        lhf_loc = ClimaAnalysis.slice(lhf; lon, lat)
        shf_loc = ClimaAnalysis.slice(shf; lon, lat)
```

experiments/calibration/shared/observation_data_global.jl
```diff
@@ -0,0 +1,26 @@
+import ClimaAnalysis
```
Sbozzolo, line 1: One thing you need to watch out for is where the data is defined in time.

For example, 2024-02-01 could mean average for the month of february or for the month of january (as I think it's our case right now).

What we do in the leaderboard code is moving everything to the first of the month to line things up
(https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/ClimaEarth/leaderboard/data_sources.jl)

However, this depends on your data sources, so make sure you understand them well.

experiments/calibration/shared/observation_data_locations.jl
```diff
@@ -0,0 +1,41 @@
+era5_ds = Dataset(
+    joinpath(
+        ClimaLand.Artifacts.era5_surface_data2008_path(),
+        "era5_monthly_surface_fluxes_200801-200812.nc",
+    ),
+)
+
+# Make the ERA5 target
+ERA5_target = []
```
Sbozzolo, line 9: You could use ClimaAnalysis to increase the level of abstraction for these operations (instead of dealing with the NCDataset directly)

experiments/calibration/shared/rand_locations.jl
```diff
@@ -0,0 +1,39 @@
...
+    landsea_mask = SpaceVaryingInput(
+        datapath,
+        "landsea_mask",
+        surface_space;
+        regridder_type,
+    )
+
+    # Get the corresponding latitude and longitude values
+    lat = ClimaCore.Fields.coordinate_field(surface_space).lat
+    lon = ClimaCore.Fields.coordinate_field(surface_space).long
```
Sbozzolo, line 22: You don't need to use ClimaCore Fields for this: you can directly look at the NetCDF file, which is more straightforward

experiments/calibration/shared/rand_locations.jl
```diff
@@ -0,0 +1,39 @@
...
+using StaticArrays
+
+# Returns a 2 SVectors of (lat, lon) tuples for n random locations on the land
+# surface. The two sets of locations are designed to be used as a training and
+# validation set.
+function rand_locations(surface_space, regridder_type, n = 100)
+    # Load the landsea mask data
+    datapath = ClimaLand.Artifacts.topmodel_data_path()
+    landsea_mask = SpaceVaryingInput(
+        datapath,
```
Sbozzolo, line 14: Is `topmodel_data_path` the land sea mask?

experiments/calibration/shared/rand_locations.jl
```diff
@@ -0,0 +1,39 @@
...
+        surface_space;
+        regridder_type,
+    )
+
+    # Get the corresponding latitude and longitude values
+    lat = ClimaCore.Fields.coordinate_field(surface_space).lat
+    lon = ClimaCore.Fields.coordinate_field(surface_space).long
+
+    # Find the coordinates of 2n random land locations
+    land_inds = rand(findall(x -> x == 1.0, Array(parent(landsea_mask))), 2 * n)
```
Sbozzolo, line 25: If you are using `rand`, make sure to set the random seed.

experiments/calibration/README.md
tapios: Is `observation_map` the observation operator `H` (mapping model output `x` to observables `y`) or generating the loss function (i.e., `norm(y - H(x))` where `norm` is a suitable weighted norm)? Please be clear here.

experiments/calibration/global_bucket/forward_model.jl
```diff
@@ -0,0 +1,176 @@
...
+function CAL.forward_model(iteration, member)
+    ensemble_member_path = path_to_ensemble_member(caldir, iteration, member)
+    params_path = parameter_path(caldir, iteration, member)
+    params = toml.parsefile(params_path) # should load a Dict, that needs to be converted to namedtuple
+
+    diagnostics_dir = joinpath(ensemble_member_path, "global_diagnostics")
+    diagdir =
+        ClimaUtilities.OutputPathGenerator.generate_output_path(diagnostics_dir)
+    t0 = 0.0
+    tf = 60 * 60.0 * 24 * 366
```
tapios, line 165: Length of day (86400 s) should be a parameter somewhere (e.g., ClimaParameters), to it's clear this is an integer number of days.

experiments/calibration/global_land/forward_model.jl
```diff
@@ -0,0 +1,442 @@
+import TOML as toml
```
tapios, line 1: Seems odd to set all of this up here. It seems to me it should be a standard model configuration that is being called here, and that should be set up wherever standard model configs live.

experiments/calibration/shared/observation_data_global.jl
```diff
@@ -0,0 +1,26 @@
...
+    "era5_monthly_surface_fluxes_200801-200812.nc",
+)
+
+era5_lhf = ClimaAnalysis.OutputVar(filename, "mslhf")
+era5_shf = ClimaAnalysis.OutputVar(filename, "msshf")
+era5_lwu = ClimaAnalysis.OutputVar(filename, "msuwlwrf")
+era5_swu = ClimaAnalysis.OutputVar(filename, "msuwswrf")
+
+observations_vecs = []
+for var in [era5_lhf, era5_shf, era5_lwu, era5_swu]
```
tapios, line 16: If the following implies the loss function is based on a global average, e.g., terms like gavg(lhf_obs) - gavg(lhf_sim), it's not right. The averages appearing in the loss functions should be spatially resolved time averages, e.g., seasonal_average(lhf_obs) - seasonal_average(lhf_sim). The global mean does not contain sufficient information for meaningful model calibration.

tapios (commented):
A couple things I didn't see (they may there and I missed them):
1. There should be a sufficient spin up period (several years) for equilibration of the land model with the prior set of parameters, or it should be initialized with an equilibrated state.
2. After each parameter update by EKP, we need an additional burn-in period (at least 3 months) that then is discarded, before data are being accumulated for the loss function.
3. The error covariance should have reasonable (location dependent) variances. An easy way to obtain it is from the initial spinup period, by accumulating the relevant averages (e.g., seasonal) and assembling them into a covariance matrix (can be diagonal at first). This is essential to properly weight, e.g., LHF in dry vs moist regions.

@odunbar Can you please look over this and make sure the above gets done properly? We'd like to get these calibrations done as soon as we can.

experiments/calibration/shared/observation_data_global.jl
kmdeck: Are we sure the sign convention is different? positive should be up in CL, and that is standard for LHF and SHF

experiments/calibration/shared/observation_data_locations.jl
```diff
@@ -0,0 +1,41 @@
...
+close = (x, y) -> abs(x - y) < 5e-1
```
kmdeck, line 10: This threshold should depend on the grid spacing. Maybe it should be find closest? What happens if many points are close enough by this metric?

experiments/calibration/shared/rand_locations.jl
kmdeck: Right now CL doesnt really use a land sea mask except to set soil parameters over the ocean. We use a mask from a soil param dataset for that. In the future we could use the actual Clima land sea mask, but in the future we shouldnt have to do this. 

Otherwise,  in plotting, in computing bias and gloal rmse through ClimaAnalysis (mask defined in CA).

That said, here, we should probably be using the Clima land sea mask. Once Charlie's PR is merged and we stop doing computations over the ocean and need a mask to make the land domain, we can change everything to be etopo2022.

experiments/calibration/shared/rand_locations.jl
kmdeck: I agree that using the netcdf file of diagnostic output + mask, both from netcdf, would be simpler, and map directly to output data. 

As is, I dont understand how indices taken from the parent array of the coordinate space will map onto the diagnostics matrix, since those are not at exactly the same resolution. I dont think we will be getting the same locations in both cases. And likewise for ERA5 data, that will also need other indices.

experiments/calibration/global_land/observation_map.jl
```diff
@@ -0,0 +1,42 @@
...
+function process_member_data(simdir)
+
+    lhf = get(simdir; short_name = "lhf")
+    shf = get(simdir; short_name = "shf")
+    lwu = get(simdir; short_name = "lwu")
+    swu = get(simdir; short_name = "swu")
+
+    vars_global_average = []
+    for var in [lhf, shf, lwu, swu]
+        var_global_average =
```
kmdeck, line 13: How can we be sure that the ERA5 and CL times will align?

kmdeck:
Thank you, Alexis! I also think this is a good start. Like Ollie and Gabriele, I think we need to ensure that the times align and the locations align between Clima and ERA5, and removing some of the duplicate code. That feels like a should-do before merging this. 

I added a couple of other "to -dos" for the future too.

We are working on abstracting the model setup (we have forcing setup, domain setup, initial condition setup, and some parameter setup abstracted, but not the entire thing yet). I can make a PR in climaland doing that so that the forward model code gets simpler, at least for full land.

AlexisRenchon:
> > > What is the issue mentioned here in item one of the todo list with the code for the monthly variances?
> > 
> > 
> > there was an Artifact problem, I had to comment out https://github.com/CliMA/ClimaLand.jl/blob/ar/ct/bucket_calibration/src/Artifacts.jl#L60 or ClimaLand wouldn't compile. I haven't looked at this for a few days
> 
> On which system? I've tested this on the HPC, maybe it's because the artifact isn't available on Derecho? If you can show me an error message I can check it out.

@Espeer5 On Derecho, yes, I believe the problem is that there is no hash in Artifact.toml

```
ERROR: LoadError: Cannot locate artifact 'era5_monthly_averages_surface_single_level_1979_2024' for x86_64-linux-gnu-libgfortran5-cxx11-libstdcxx30-julia_version+1.11.3 in '/glade/u/home/arenchon/GitHub/ClimaLand.jl/Artifacts.toml'
```

PR #1056, opened 2025-03-03
relax SoilDrivers atmos type annotation; compute SnowModel humidity in model-specific method

juliasloan25:
## Purpose 
To couple the full LandModel with ClimaAtmos, there are a handful of miscellaneous changes to the ClimaLand code that need to be made. This PR makes a couple of them (see Content below for details).

These changes shouldn't be behavior changing.

Part of #1038 

## Content
- [x] relax the type restriction for SoilDrivers to allow a `CoupledAtmosphere` in addition to `PrescribedAtmosphere` ([here](https://github.com/CliMA/ClimaLand.jl/blob/b8c7255f446d0fe766d0196ca581deb74cd29ec4/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl#L288))
  - without this, we can't set up a SoilCO2Model with a CoupledAtmosphere
- [x] move `snow_surface_specific_humidity` call out of `update_aux!` and into `surface_specific_humidity`, which  
  - can't access `p.drivers.thermal_state` in `CoupledAtmosphere` case, so the call to `snow_surface_specific_humidity` needs to be within a function where we've dispatched on the atmos type. `surface_specific_humidity` is only called in the `turbulent_fluxes!` method for `PrescribedAtmosphere`, so moving it there fixes this

----

### DIFF
--- a/src/standalone/Snow/Snow.jl
+++ b/src/standalone/Snow/Snow.jl
@@ -50,7 +50,7 @@ for use within an `AbstractSnowModel` type. Currently we support a
 `MinimumDensityModel` model and the `NeuralDepthModel`.
 
 Since depth and bulk density are related via SWE, and SWE is a prognostic variable
-of the snow model, only depth or bulk density can be independently modeled at one time. 
+of the snow model, only depth or bulk density can be independently modeled at one time.
 This is why they are treated as a single "density model" even if the parameterization is actually
 a model for snow depth.
 The snow depth/density model can be diagnostic or introduce additional prognostic variables.
@@ -203,7 +203,7 @@ prognostic_vars(m::SnowModel) =
 
 """
     density_prog_vars(::AbstractDensityModel)
-    
+
 A default method for adding prognostic variables to the snow model as required
 by the density model choice.
 """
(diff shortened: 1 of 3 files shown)


src/standalone/Snow/Snow.jl
```diff
@@ -500,6 +498,71 @@ function ClimaLand.get_drivers(model::SnowModel)
...
+    @. p.snow.q_sfc = snow_surface_specific_humidity(
+        p.snow.T_sfc,
+        p.snow.q_l,
+        p.drivers.thermal_state,
+        model.parameters,
+    )
+
+    T_sfc = surface_temperature(model, Y, p, t)
+    ρ_sfc = surface_air_density(atmos, model, Y, p, t, T_sfc)
+    q_sfc = surface_specific_humidity(model, Y, p, T_sfc, ρ_sfc)
```
kmdeck, line 538: Could we maybe move this:
``` 
@. p.snow.q_sfc = snow_surface_specific_humidity(
        p.snow.T_sfc,
        p.snow.q_l,
        p.drivers.thermal_state,
        model.parameters,
    )
```

into `surface_specific_humidity` for snow? We might not need an additional method, then.

juliasloan25: Ahh yes! Good point

kmdeck (approved):
looks good! thanks!

PR #1060, opened 2025-03-05
Add very long run of snowy land

ph-kev:
This PR adds the option to run a very long run of snowy land (10 years in length). This is accomplished by using the environment variable in both Buildkite and Julia.

A long run of 10 years is detected by the environment variable `LONG_RUN` in the Buildkite steps when uploading the pipeline. By default, `LONG_RUN="false"`. This environment variable can be replaced when making a new build or scheduled build in the "Environmental Variables" option. In `pipeline.yml`, we use a conditional based on the value of `LONG_RUN` to determine what group step to run. For the very long run, we also pass `LONG_RUN` as an environment variable which Julia can use to figure out which code to run for `snowy_land.jl`.

Todo
- [ ] Make a schedule for running long run of 10 years on Buildkite.


.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -84,3 +85,19 @@ steps:
...
+          - julia --color=yes --project=.buildkite experiments/long_runs/snowy_land.jl
+        artifact_paths:
+          - "snowy_land_longrun_gpu/*png"
+          - "snowy_land_longrun_gpu/*pdf"
+        agents:
+          slurm_gpus: 1
+          slurm_time: 24:00:00
+        env:
+          CLIMACOMMS_DEVICE: "CUDA"
+          LONG_RUN: "true"
```
ph-kev, line 103: I wasn't sure if this is a good name to use for the environment variable.

experiments/long_runs/snowy_land.jl
```diff
@@ -74,9 +81,19 @@ function setup_prob(
     subsurface_space = domain.space.subsurface
 
     # Forcing data
-    era5_artifact_path =
-        ClimaLand.Artifacts.era5_land_forcing_data2008_folder_path(; context)
-    era5_ncdata_path = joinpath(era5_artifact_path, "era5_2008_1.0x1.0.nc")
+    if LONG_RUN
```
ph-kev, line 84: This keeps the behavior that the long run in the CI repeats the same forcing data for 2008. We wouldn't need this if and else statement if we are okay with not using the same year for forcing data.

experiments/long_runs/snowy_land.jl
```diff
@@ -74,9 +81,19 @@ function setup_prob(
...
+            date(tf);
+            context,
+        )
+    else
+        era5_artifact_path =
+            ClimaLand.Artifacts.era5_land_forcing_data2008_folder_path(;
+                context,
+            )
+        era5_ncdata_path = joinpath(era5_artifact_path, "era5_2008_1.0x1.0.nc")
+    end
```
ph-kev, line 96: This keeps the behavior that the long run in the CI repeats the same forcing data from 2008. We wouldn't need this if and else statement if we are okay with not repeating the era5 forcing data from 2008.

ph-kev, line 96: This keeps the behavior that the long run in the CI repeats the same forcing data from 2008. We wouldn't need this if and else statement if we are okay with not repeating the era5 forcing data from 2008.

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -84,3 +85,19 @@ steps:
           slurm_time: 00:30:00
         env:
           CLIMACOMMS_DEVICE: "CUDA"
+
+  - group: "Very long run of snowy land"
+    if: build.env("LONG_RUN") == "true"
```
Sbozzolo, line 90: Would it be possible to simply check if `LONG_RUN` is in the env?

experiments/long_runs/snowy_land.jl
```diff
@@ -50,7 +50,14 @@ import NCDatasets
 using Poppler_jll: pdfunite
 
 const FT = Float64;
-time_interpolation_method = LinearInterpolation(PeriodicCalendar())
+# If you want to do a very long run locally, you can do `export LONG_RUN="true"`
+# in the terminal and run this script. If you want to do a very long run on
+# Buildkite manually, then make a new build and pass `LONG_RUN="true"` as an
+# environment variable.
+const LONG_RUN =
```
Sbozzolo, line 57: Would it be possible to simply check if `LONG_RUN` is in the env?

src/Artifacts.jl
```diff
@@ -67,6 +99,29 @@ function modis_lai_forcing_data2008_path(; context = nothing)
...
+
+Find the appropriate files of MODIS LAI data to run a simuation starting on
+`start_date` and ending on `final_date`.
+"""
+function find_modis_year_paths(start_date, final_date; context = nothing)
+    # Get the year of the start and final dates
+    year0 = Dates.year(start_date)
+    yearf = Dates.year(final_date)
+    modis_lai_forcing_data_path =
+        modis_lai_forcing_data2008_path(context = context)
```
Sbozzolo, line 113: Why does `modis_lai_forcing_data2008_path` contain multiple years?

Sbozzolo (commented):
Can you also document this somewhere?

.buildkite/longruns_gpu/pipeline.yml
ph-kev: I updated it to check if the environment variable exists or not. You still need to assign a value to it though, so when initializing a new build, it looks like `LONGER_RUN=`.

src/Artifacts.jl
ph-kev: The artifact got updated, but the function name wasn't updated too. I added a commit to update the name of this function.

Sbozzolo (approved):

PR #1062, opened 2025-03-06
Add functions which compute the land surface fluxes

kmdeck:
## Purpose 
Partially Address Issue #1038 

## To-do
- [x] test from coupler
  - momentum and moisture fluxes are zero, but energy flux is non-zero

## Content
- create "coupler compute turbulent fluxes" for each component which compute the fluxes and update the land cache in place with those fluxes.
- allocate space for `ρτxz` and `ρτyz` in soil, snow, and canopy model caches when running with a coupled atmosphere by extending `boundary_vars`, `boundary_var_types`, and `boundary_var_domain_names`
  - this change also includes moving the location of canopy turbulent fluxes from `p.energy.turbulent_fluxes` to `p.turbulent_fluxes` (affects many files)
  - right now we don't add any resistance terms to the momentum fluxes returned by SF.jl, but we'll improve this later https://github.com/CliMA/ClimaLand.jl/issues/1070
- add a flag `compute_momentum_fluxes` to `turbulent_fluxes_at_a_point` so we can return momentum fluxes in the coupled case but not in the uncoupled case
- add `atmos` as an argument to `surface_specific_humidity` so we can access `thermal_state` correctly based on the atmosphere type for snow

## Flux improvements
In this PR, we found a couple of ways we can improve our flux exchanges and calculations in both coupled and standalone runs - see these issues for more details #1084 #1085

----

### DIFF
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -194,9 +194,37 @@ struct CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT} end
 """
     CoupledAtmosphere{FT} <: AbstractAtmosphericDrivers{FT}
 
-To be used when coupling to an atmosphere model.
+To be used when coupling to an atmosphere model. Contains fields that
+are used to compute surface fluxes in the coupled setup.
+
+When constructed without a space, the struct doesn't contain anything,
+but it still acts as a flag that fluxes have been updated by the coupler
+and don't need to be recomputed.
+When constructed with a space, the struct contains the fields needed to compute
+surface fluxes in the coupled setup, which are accessed by ClimaCoupler.
 """
-struct CoupledAtmosphere{FT} <: AbstractAtmosphericDrivers{FT} end
+struct CoupledAtmosphere{FT} <: AbstractAtmosphericDrivers{FT}
+    "Atmospheric horizontal wind velocity vector at the reference height (m/s)"
+    u::Union{Nothing, Fields.Field}
+    "Atmospheric reference height (m), relative to surface elevation"
+    h::Union{Nothing, Fields.Field}
+    "Minimum wind speed (gustiness; m/s), which is always a spatial constant"
+    gustiness::Union{Nothing, FT}
+    "Atmospheric thermodynamic state at the reference height"
+    thermal_state::Union{Nothing, Fields.Field}
+    "Create a `CoupledAtmosphere` with default values"
+    function CoupledAtmosphere{FT}() where {FT}
+        return new{FT}(nothing, nothing, nothing, nothing)
+    end
+    function CoupledAtmosphere{FT}(space) where {FT}
+        return new{FT}(
+            Fields.zeros(SVector{2, FT}, space),
+            Fields.zeros(space),
+            FT(1), # gustiness is always a spatial constant, for now
+            Fields.zeros(Thermodynamics.PhaseEquil{FT}, space),
+        )
+    end
+end
 
 """
     compute_ρ_sfc(thermo_params, ts_in, T_sfc)
@@ -260,7 +288,7 @@ function turbulent_fluxes!(
 )
     T_sfc = surface_temperature(model, Y, p, t)
     ρ_sfc = surface_air_density(atmos, model, Y, p, t, T_sfc)
-    q_sfc = surface_specific_humidity(model, Y, p, T_sfc, ρ_sfc)
+    q_sfc = surface_specific_humidity(atmos, model, Y, p, T_sfc, ρ_sfc)
     β_sfc = surface_evaporative_scaling(model, Y, p)
     h_sfc = surface_height(model, Y, p)
     r_sfc = surface_resistance(model, Y, p, t)
@@ -288,6 +316,54 @@ function turbulent_fluxes!(
     return nothing
 end
 
+"""
+    coupler_compute_turbulent_fluxes!(dest, atmos::CoupledAtmosphere, model::AbstractModel, Y::ClimaCore.Fields.FieldVector, p::NamedTuple, t)
+
+This function computes the turbulent surface fluxes for a coupled simulation.
+This function is very similar to the default method of `turbulent_fluxes!`,
+but it is used with a `CoupledAtmosphere` which contains all the necessary
+atmosphere fields to compute the surface fluxes, rather than some being stored in `p`.
+
+This function is intended to be called by ClimaCoupler.jl when computing
+fluxes for a coupled simulation with the integrated land model.
+"""
+function coupler_compute_turbulent_fluxes!(
+    dest,
+    atmos::CoupledAtmosphere,
+    model::AbstractModel,
+    Y::ClimaCore.Fields.FieldVector,
+    p::NamedTuple,
+    t,
+)
+    T_sfc = surface_temperature(model, Y, p, t)
+    ρ_sfc = surface_air_density(atmos, model, Y, p, t, T_sfc)
+    q_sfc = surface_specific_humidity(atmos, model, Y, p, T_sfc, ρ_sfc)
+    β_sfc = surface_evaporative_scaling(model, Y, p)
+    h_sfc = surface_height(model, Y, p)
+    r_sfc = surface_resistance(model, Y, p, t)
+    d_sfc = displacement_height(model, Y, p)
+
+    dest .=
+        turbulent_fluxes_at_a_point.(
+            T_sfc,
+            q_sfc,
+            ρ_sfc,
+            β_sfc,
+            h_sfc,
+            r_sfc,
+            d_sfc,
+            atmos.thermal_state,
+            atmos.u,
+            atmos.h,
+            atmos.gustiness,
+            model.parameters.z_0m,
+            model.parameters.z_0b,
+            model.parameters.earth_param_set,
+            return_momentum_fluxes = true,
+        )
+    return nothing
+end
+
 
 """
     turbulent_fluxes_at_a_point(T_sfc::FT,
@@ -303,7 +379,8 @@ end
                                 gustiness::FT,
                                 z_0m::FT,
                                 z_0b::FT,
-                                earth_param_set::EP,
+                                earth_param_set::EP;
+                                return_momentum_fluxes = false,
                                ) where {FT <: AbstractFloat, P}
 
 Computes turbulent surface fluxes at a point on a surface given
@@ -337,12 +414,13 @@ function turbulent_fluxes_at_a_point(
     r_sfc::FT,
     d_sfc::FT,
     ts_in,
-    u::FT,
+    u::Union{FT, SVector{2, FT}},
     h::FT,
     gustiness::FT,
     z_0m::FT,
     z_0b::FT,
-    earth_param_set::EP,
+    earth_param_set::EP;
+    return_momentum_fluxes = false,
 ) where {FT <: AbstractFloat, EP}
     thermo_params = LP.thermodynamic_parameters(earth_param_set)
     ts_sfc = Thermodynamics.PhaseEquil_ρTq(thermo_params, ρ_sfc, T_sfc, q_sfc)
@@ -357,11 +435,11 @@ function turbulent_fluxes_at_a_point(
     # In this we have neglected z_0m and z_0b (i.e. assumed they are small
     # compared to Δh).
     state_sfc = SurfaceFluxes.StateValues(FT(0), SVector{2, FT}(0, 0), ts_sfc)
-    state_in = SurfaceFluxes.StateValues(
-        h - d_sfc - h_sfc,
-        SVector{2, FT}(u, 0),
-        ts_in,
-    )
+    # u is already a vector when we get it from a coupled atmosphere, otherwise we need to make it one
+    if u isa FT
+        u = SVector{2, FT}(u, 0)
+    end
+    state_in = SurfaceFluxes.StateValues(h - d_sfc - h_sfc, u, ts_in)
     # The following line wont work on GPU
     #    h - d_sfc - h_sfc < 0 &&
     #        @error("Surface height is larger than atmos height in surface fluxes")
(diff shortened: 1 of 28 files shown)


src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -289,6 +289,40 @@ function ClimaLand.turbulent_fluxes!(
     return nothing
 end
 
+
+function ClimaLand.coupler_compute_turbulent_fluxes!(
```
kmdeck, line 293: lots of duplication, but I felt it was confusing to have a method for `turbulent_fluxes!` which is not called by land. but, we can disuss

src/shared_utilities/drivers.jl
```diff
@@ -289,6 +289,43 @@ function turbulent_fluxes!(
 end
 
 
+function coupler_compute_turbulent_fluxes!(
```
kmdeck, line 292: This is basically a copy of `turbulent_fluxes!` with `atmos::PrescribedAtmos`. We also have `turbulent_fluxes!` with `atmos::CoupledAtmos` which does nothing (as the value has been precomputed by coupler. These functions are called in the `update_boundary_fluxes` function of the land model.

I liked giving this another name, though, because it is called in a different part of the code and used in a different way, sort of.

src/shared_utilities/drivers.jl
```diff
@@ -289,6 +289,43 @@ function turbulent_fluxes!(
...
+    dest,
+    atmos::NamedTuple,
```
juliasloan25, line 294: Will `atmos` be the atmos model state for this function? I assume so because where this is called we pass in the atmos state, but then in this function we access `atmos.thermal_state`, `atmos.u`, `atmos.h`, etc. which I wouldn't be present in the atmos state.

In the coupler, before we compute fluxes we retrieve from the atmosphere `u`, surface height `z_sfc`, first atmos center height `z_int`, and atmos thermo state `thermo_state_int`, then pass these into the functions that compute fluxes. Maybe here, `atmos` could instead be a NamedTuple `atmos_properties` or something like that, which contains these 4 properties.

juliasloan25: With the changes in https://github.com/CliMA/ClimaCoupler.jl/pull/1220, this function will work as-is

src/shared_utilities/drivers.jl
```diff
@@ -625,7 +678,7 @@ This requires the field `ρ_sfc` to be present in the cache `p` under the name
 of the model.
 """
 function surface_air_density(
-    atmos::CoupledAtmosphere,
+    atmos::Union{CoupledAtmosphere, NamedTuple},
```
juliasloan25, line 681: I changed this because when we compute turbulent fluxes from the coupler, we pass a NamedTuple as "atmos" here

src/standalone/Bucket/Bucket.jl
```diff
@@ -553,6 +553,31 @@ function ClimaLand.get_drivers(model::BucketModel)
...
+This requires the field `ρ_sfc` to be present in the cache `p` under the name
+of the model.
+"""
+function surface_air_density(
+    atmos::CoupledAtmosphere,
+    model::BucketModel,
+    Y,
+    p,
+    _...,
+)
```
juliasloan25, line 575: This was in `src/shared_utilities/drivers.jl` before, but for the full land model we dispatch on `NamedTuple` instead of `CoupledAtmosphere`. This means that this function only applies to the bucket, so I moved it here

src/standalone/Bucket/bucket_parameterizations.jl
```diff
@@ -29,12 +29,18 @@ function ClimaLand.surface_temperature(model::BucketModel, Y, p, t)
...
 
 """
-    ClimaLand.surface_specific_humidity(model::BucketModel, Y, p)
+    ClimaLand.surface_specific_humidity(atmos, model::BucketModel, Y, p)
 
 a helper function which returns the surface specific humidity for the bucket
 model, which is stored in the aux state.
 """
-function ClimaLand.surface_specific_humidity(model::BucketModel, Y, p, _...)
+function ClimaLand.surface_specific_humidity(
```
kmdeck, line 37: Did this function need to change?

src/standalone/Bucket/Bucket.jl
```diff
@@ -553,6 +553,31 @@ function ClimaLand.get_drivers(model::BucketModel)
     return (model.atmos, model.radiation)
 end
 
+"""
+    ClimaLand.surface_air_density(
+                    atmos::CoupledAtmosphere,
+                    model::AbstractModel,
```
kmdeck, line 559: can update this

src/standalone/Bucket/Bucket.jl
```diff
@@ -553,6 +553,31 @@ function ClimaLand.get_drivers(model::BucketModel)
...
+function surface_air_density(
+    atmos::CoupledAtmosphere,
+    model::BucketModel,
+    Y,
+    p,
+    _...,
+)
+    model_name = ClimaLand.name(model)
+    model_cache = getproperty(p, model_name)
+    return model_cache.ρ_sfc
```
kmdeck, line 578: We can simplify this since it is not bucket specific - `return p.bucket.rho_sfc`?

src/standalone/Snow/boundary_fluxes.jl
```diff
@@ -140,3 +140,46 @@ function snow_boundary_fluxes!(
...
+        ::AtmosDrivenSnowBC{<:CoupledAtmosphere, <:CoupledRadiativeFluxes},
+        ::ClimaLand.TopBoundary,
+    ) where {FT}
+
+An extension of the `boundary_var_types` method for AtmosDrivenSnowBC. This
+specifies the type of the additional variables.
+
+This method includes the additional momentum fluxes needed by the atmosphere.
+These are updated in place when the coupler computes turbulent fluxes,
+rather than in `soil_boundary_fluxes!`.
```
kmdeck, line 169: `snow_boundary_fluxes!`

src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -408,13 +443,75 @@ function canopy_turbulent_fluxes_at_a_point(
...
+        ::AtmosDrivenCanopyBC{<:CoupledAtmosphere, <:CoupledRadiativeFluxes},
+        ::ClimaLand.TopBoundary,
+    ) where {FT}
+
+An extension of the `boundary_var_types` method for AtmosDrivenCanopyBC. This
+specifies the type of the additional variables.
+
+This method includes the additional momentum fluxes needed by the atmosphere.
+These are updated in place when the coupler computes turbulent fluxes,
+rather than in `soil_boundary_fluxes!`.
```
kmdeck, line 501: `canopy_boundary_fluxes!`

src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -303,7 +336,8 @@ end
         gustiness::FT,
         z_0m::FT,
         z_0b::FT,
-        earth_param_set::EP,
+        earth_param_set::EP;
+        compute_momentum_fluxes = false,
```
kmdeck, line 340: we might use a different name here - momentum_fluxes are always computed, but in some cases they are not returned

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -97,6 +105,33 @@ function ClimaLand.surface_specific_humidity(model::SnowModel, Y, p, _...)
...
+
+Returns the precomputed specific humidity over snow as a weighted
+fraction of the saturated specific humidity over liquid and frozen
+water.
+
+In the case of a CoupledAtmoshere, the atmosphere thermal state is accessed
+from the provided atmosphere NamedTuple.
+"""
+function ClimaLand.surface_specific_humidity(
+    atmos::NamedTuple,
```
kmdeck, line 119: I'm a little confused about the changes to this function, maybe we can discuss?

src/standalone/Bucket/bucket_parameterizations.jl
juliasloan25: I had to change the `surface_specific_humidity` interface to take in `atmos` so we can dispatch on it for the snow case. Since the default `turbulent_fluxes!` in drivers.jl calls this function, I had to change it for the bucket too

src/standalone/Vegetation/Canopy.jl
```diff
@@ -354,11 +354,42 @@ function initialize_auxiliary(model::CanopyModel{FT}, coords) where {FT}
...
+        )
+    )
     p = ClimaLand.add_dss_buffer_to_aux(p, model.domain)
     return p
 end
 
+"""
+    initialize_boundary_vars(model::CanopyModel{FT}, coords)
+
+Add boundary condition-related variables to the cache.
```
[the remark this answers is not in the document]
juliasloan25: Yeah, we should unify it for all fluxes. Would the others just be `p.canopy.radiative_transfer.LW_n`, `p.canopy.radiative_transfer.SW_n`?

I see these aux vars for radiative transfer:
```
ClimaLand.auxiliary_vars(model::Union{BeerLambertModel, TwoStreamModel}) =
    (:nir_d, :par_d, :nir, :par, :LW_n, :SW_n, :ϵ, :G, :K)
```

and these for energy:
```
ClimaLand.auxiliary_vars(model::AbstractCanopyEnergyModel) =
    (:fa_energy_roots, :∂LW_n∂Tc, :∂qc∂Tc)
```

src/standalone/Vegetation/canopy_boundary_fluxes.jl
juliasloan25: Good point! I'll change it to `return_momentum_fluxes`

src/shared_utilities/drivers.jl
```diff
@@ -194,9 +194,32 @@ struct CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT} end
...
+    "Atmospheric reference height (m), relative to surface elevation"
+    h::Union{Nothing, Fields.Field}
+    "Minimum wind speed (gustiness; m/s), which is always a spatial constant"
+    gustiness::Union{Nothing, FT}
+    "Atmospheric thermodynamic state at the reference height"
+    thermal_state::Union{Nothing, Fields.Field}
+    "Create a `CoupledAtmosphere` with default values"
+    function CoupledAtmosphere{FT}() where {FT}
+        return new{FT}(nothing, nothing, nothing, nothing)
+    end
```
juliasloan25, line 213: I included this constructor so we still support the previous behavior of constructing via `CoupledAtmosphere()`

src/shared_utilities/drivers.jl
```diff
@@ -194,9 +194,32 @@ struct CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT} end
 """
     CoupledAtmosphere{FT} <: AbstractAtmosphericDrivers{FT}
 
-To be used when coupling to an atmosphere model.
-"""
-struct CoupledAtmosphere{FT} <: AbstractAtmosphericDrivers{FT} end
+To be used when coupling to an atmosphere model. Contains fields that
+are used to compute surface fluxes in the coupled setup, since
+these are computed within ClimaLand.
```
kmdeck, line 199: Maybe just "To be used when coupling to an atmosphere model. Contains fields that
are used to compute surface fluxes in the coupled setup." - the last clause might make people thing CL computes the fluxes (but really the coupler does, using CL functions...?)

kmdeck: perhaps describe the behavior when all are Nothing (how this is used)?

src/shared_utilities/drivers.jl
```diff
@@ -194,9 +194,32 @@ struct CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT} end
...
+    gustiness::Union{Nothing, FT}
+    "Atmospheric thermodynamic state at the reference height"
+    thermal_state::Union{Nothing, Fields.Field}
+    "Create a `CoupledAtmosphere` with default values"
+    function CoupledAtmosphere{FT}() where {FT}
+        return new{FT}(nothing, nothing, nothing, nothing)
+    end
+    function CoupledAtmosphere{FT}(space) where {FT}
+        return new{FT}(
+            Fields.similar(Fields.zeros(space), SVector{2, FT}),
```
kmdeck, line 216: I think you can also call
`Fields.zeros(SVector{2, FT}, space)`

src/standalone/Bucket/Bucket.jl
kmdeck: Ooo we might need to be careful here - and remove this method? I am looking here: https://github.com/CliMA/ClimaLand.jl/blob/1176a957dfd72ed69e809b8e8501dbabd52c9a1b/src/standalone/Bucket/Bucket.jl#L409C9-L409C12

and wondering if this now becomes circular (doing nothing)? Maybe we want the default method. But maybe this is OK because the coupler has already updated p.bucket.rho_sfc, which is why this does nothing.

kmdeck (commented):
I think this looks good! Just the one question about the bucket method. Before you merge, could you run the long runs just to be sure nothing changes? 

Thanks!

src/standalone/Bucket/Bucket.jl
juliasloan25: The last sentence is right - the coupler will update air density [here](https://github.com/CliMA/ClimaCoupler.jl/blob/eb8d2ddd589bea906484683e843a8c8e34ea1af5/src/FieldExchanger.jl#L171), so we don't need to compute it at all for a coupled bucket.

Actually now that I look at it again, the air density of the integrated land model will also get updated in the coupler at each timestep, so maybe this method does actually work for the coupled full model as well.

src/shared_utilities/drivers.jl
juliasloan25: good point!

src/standalone/Bucket/Bucket.jl
juliasloan25: Right now, the coupler computes \rho and updates it for the bucket, but not for the integrated land model. We can unify these two by adding the atmosphere thermal state to `CoupledAtmosphere` in the bucket case, and computing `\rho` within ClimaLand for all cases. Kat and I will work on this in a separate PR.

juliasloan25 (approved):

PR #1063, opened 2025-03-07
Use initial conditions for soil and snow from 10 year run

kmdeck:
## Purpose 
This PR changes the artifact used for the soil and snow initial conditions to be the IC at the end of a 10 year run.

## Content
Change artifact toml

----


experiments/long_runs/snowy_land.jl
```diff
@@ -465,60 +465,26 @@ setup_and_solve_problem(; greet = true);
...
+    "trans",
+    "msf",
+]
 
 include(
     joinpath(
         pkgdir(ClimaLand),
         "experiments",
-        "LONGER_RUNs",
+        "long_runs",
```
kmdeck, line 487: @ph-kev I believe this is a bug, can you confirm?

AlexisRenchon (approved):
Looks great! Thank you!

Issue #795, opened 2024-09-25
Refactor (from scratch) ClimaLandSimulations


AlexisRenchon:
ClimaLandSimulations is a high level API to run ClimaLand simulations. 
For example, running a a one column (fluxnet site) simulation via ClimaLandSimulations is as simple as
```
run_fluxnet("US-MOz")
```
using default for model config etc, which can be changed via option kwarg

To discuss/address
- Too many dependencies
- lib or ext?
- (improve speed! 1 column is slow... implicit solver will improve it?)
- Add PFTs params (that can be adjusted)
- Use gap-filled data (our current met data gapfilling is bad, slow and annoying)
- Add global run in ClimaLandSimulations

PR #1065, opened 2025-03-10
Model function

kmdeck:
## Purpose 
- Creates a helper function for setting up the forward model
- adds the mask to artifacts as prep for using it in the domain
- changed path returned by era5 artifact to include filename, since that is different between lowres and high res. 

## Content

----

### DIFF
--- a/experiments/benchmarks/land.jl
+++ b/experiments/benchmarks/land.jl
@@ -65,9 +65,8 @@ function setup_prob(t0, tf, Δt; nelements = (101, 15))
     start_date = DateTime(2008)
 
     # Forcing data
-    era5_artifact_path =
+    era5_ncdata_path =
         ClimaLand.Artifacts.era5_land_forcing_data2008_folder_path(; context)
-    era5_ncdata_path = joinpath(era5_artifact_path, "era5_2008_1.0x1.0.nc")
     atmos, radiation = ClimaLand.prescribed_forcing_era5(
         era5_ncdata_path,
         surface_space,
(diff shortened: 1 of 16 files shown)


AlexisRenchon:
I think this is really useful and an important PR, thanks Kat. It seems to address this issue: https://github.com/CliMA/ClimaLand.jl/issues/795

kmdeck:
> I think this is really useful and an important PR, thanks Kat. It seems to address this issue: #795

thank you, Alexis! I think this helps to address it for global simulations with spatially varying parameters read from a file, but with scalar parameters (global constants) passed in with some flexibility. Ultimately many of those can get moved to ClimaParameters  and it will get even simpler. But, the setup for fluxtower sites will be different since we no longer read the spatially varying parameters from a file. (or maybe we could set it up somehow similar?)

I realized this function is also helpful for unit tests, which is mostly why I made it :)

AlexisRenchon:
> > I think this is really useful and an important PR, thanks Kat. It seems to address this issue: #795
> 
> thank you, Alexis! I think this helps to address it for global simulations with spatially varying parameters read from a file, but with scalar parameters (global constants) passed in with some flexibility. Ultimately many of those can get moved to ClimaParameters and it will get even simpler. But, the setup for fluxtower sites will be different since we no longer read the spatially varying parameters from a file. (or maybe we could set it up somehow similar?)
> 
> I realized this function is also helpful for unit tests, which is mostly why I made it :)

I agree!
I think we can make this PR about global runs, we will think about single site in another one.

AlexisRenchon (approved):
lgtm! Maybe let's make an issue for next steps

PR #1069, opened 2025-03-14
try with better IC

kmdeck:
## Purpose 
Investigate NaNs that appear rapidly (in first month)

We like the flexibility of reading initial conditions from a net cdf file and interpolation to the model grid, because it lets us use any model grid.

However, this can lead to unphysical initial states, and it requires a somewhat complex set of steps (now in the initial condition function).

**_In the future, we should save both the interpolated state diagnostic as well as the field vector itself_**
## Content
We now do the following to make sure the IC are "stable/physical" for soil.
- theta_l > theta_r
- theta_l < \nu
- theta_i < nu- \theta_l
- temperature < max(T_atmos)
- temperature > min(T_atmos)

----

### DIFF
--- a/experiments/long_runs/snowy_land.jl
+++ b/experiments/long_runs/snowy_land.jl
@@ -341,7 +341,6 @@ function setup_prob(
     Y, p, cds = initialize(land)
 
     ic_path = ClimaLand.Artifacts.soil_ic_2008_50m_path(; context = context)
-    ClimaLand.set_soil_initial_conditions!(Y, ν, θ_r, subsurface_space, ic_path)
     evaluate!(p.snow.T, atmos.T, t0)
     ClimaLand.set_snow_initial_conditions!(
         Y,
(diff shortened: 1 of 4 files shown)


experiments/long_runs/snowy_land.jl
```diff
@@ -438,7 +447,7 @@ function setup_and_solve_problem(; greet = false)
     days = 24hours # days in seconds
     years = 366days # years in seconds - 366 to make sure we capture at least full years
     # 10 years in seconds for very long run and 2 years in seconds otherwise
-    tf = LONGER_RUN ? 10years : 2years
+    tf = LONGER_RUN ? 10years : 1years
```
kmdeck, line 450: Ill change this back before mergin

src/simulations/initial_conditions.jl
```diff
@@ -42,45 +44,93 @@ function set_soil_initial_conditions!(
...
+    Y.soil.ρe_int .= ClimaLand.Soil.volumetric_internal_energy.(
+        Y.soil.θ_i,
+        ρc_s,
+        T,
+        soil.parameters.earth_param_set,
+    )
     return nothing
 end
 
+"""
```
kmdeck, line 82: Will add doc string

kmdeck: (I just dont want to push because I dont want to trigger a long run, but want to keep the latest build with the long run that ran!)

juliasloan25 (approved):

PR #1071, opened 2025-03-14
Adds functions for models to compute total energy and water volume(mass) per area

kmdeck:
## Purpose 
Add helper functions which compute the total water volume and energy per unit area on land.
partially closes Issue #1038 

## To-do
- Follow up PR - add conservation test
- Follow up PR: fix https://github.com/CliMA/ClimaLand.jl/issues/1078

## Content
Creates the `total_energy_per_area!` and `total_liq_water_vol_per_area!` functions for AbstractModels
Creates the specific method of these for: Richards, EnergyHydrology, Snow, Canopy
Tests these methods for standalone models
Creates a method for LandModels
Tests this method for the snowy land model

----

### DIFF
--- a/src/standalone/Soil/energy_hydrology.jl
+++ b/src/standalone/Soil/energy_hydrology.jl
@@ -1098,3 +1098,58 @@ function soil_turbulent_fluxes_at_a_point(
         vapor_flux_ice = Ẽ_i,
     )
 end
+
+"""
+    ClimaLand.total_liq_water_vol_per_area!(
+        surface_field,
+        model::EnergyHydrology,
+        Y,
+        p,
+        t,
+)
+
+A function which updates `surface_field` in place with the value for
+the total liquid water volume per unit ground area for the `EnergyHydrology`.
+
+The water in all phases is accounted for by converting ice volume to liquid water
+volume using the ratio of the density of ice to the density of water.
+"""
+function ClimaLand.total_liq_water_vol_per_area!(
+    surface_field,
+    model::EnergyHydrology,
+    Y,
+    p,
+    t,
+)
+    earth_param_set = model.parameters.earth_param_set
+    _ρ_liq = LP.ρ_cloud_liq(earth_param_set)
+    _ρ_ice = LP.ρ_cloud_ice(earth_param_set)
+    ClimaCore.Operators.column_integral_definite!(
+        surface_field,
+        Y.soil.ϑ_l .+ Y.soil.θ_i .* _ρ_ice ./ _ρ_liq, # this line allocates
+    )
+    return nothing
+end
+
+"""
+    ClimaLand.total_energy_per_area!(
+        surface_field,
+        model::EnergyHydrology,
+        Y,
+        p,
+        t,
+)
+
+A function which updates `surface_field` in place with the value for
+the total energy per unit ground area for the `EnergyHydrology`.
+"""
+function ClimaLand.total_energy_per_area!(
+    surface_field,
+    model::EnergyHydrology,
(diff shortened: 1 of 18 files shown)


src/standalone/Soil/energy_hydrology.jl
```diff
@@ -1123,3 +1123,55 @@ function soil_turbulent_fluxes_at_a_point(
...
+    Y,
+    p,
+    t,
+)
+    earth_param_set = model.parameters.earth_param_set
+    _ρ_liq = LP.ρ_cloud_liq(earth_param_set)
+    _ρ_ice = LP.ρ_cloud_ice(earth_param_set)
+    ClimaCore.Operators.column_integral_definite!(
+        surface_field,
+        Y.soil.ϑ_l .+ Y.soil.θ_i .* _ρ_ice ./ _ρ_liq, # this line allocates
```
juliasloan25, line 1151: I think we have subsurface scratch space - we could pass that into this function and use it to store this quantity. But that would likely require changing the function interface, and these functions will only be called to check conservation so they aren't super performance critical

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -765,4 +766,45 @@ function transpiration_per_ground_area(
...
+    surface_field .= 0
+    n = length(labels)
+    surface_field .+=
+        dz[1] .* getproperty(area_index, labels[1]) .*
+        Y.canopy.hydraulics.ϑ_l.:1
+    for i in 2:n
+        surface_field .+=
+            dz[i] .* getproperty(area_index, labels[i]) .*
+            Y.canopy.hydraulics.ϑ_l.:($i)
+    end
```
juliasloan25, line 806:
```suggestion
    for i in 1:n
        surface_field .+=
            dz[i] .* getproperty(area_index, labels[i]) .*
            Y.canopy.hydraulics.ϑ_l.:($i)
    end
```

test/runtests.jl
```diff
@@ -49,6 +49,9 @@ end
 @safetestset "Neural Snow model tools tests" begin
     include("standalone/Snow/tool_tests.jl")
 end
+@safetestset "Snow integrated content" begin
```
juliasloan25, line 52: I think this is a little more clear:
```suggestion
@safetestset "Snow integrated water, energy content" begin
```

src/standalone/Soil/energy_hydrology.jl
kmdeck: We do have access - stored in `p`, but...it's not a default aux variable (only added for some boundary conditions). I could make it a default, and not add it again for other BC

juliasloan25: Hmm... It would be nice to have it available, independent of the BCs. But I'm happy to keep this PR as-is and address that separately

src/standalone/Vegetation/PlantHydraulics.jl
kmdeck: good call 😅

kmdeck:
> This looks great, thank you Kat! It would be helpful to clarify that, although `total_liq_water_vol_per_area!` returns the volume of liquid water, it includes contributions from all phases.
> 
> Could you also update the NEWS please?

yes! will do for both

src/standalone/Soil/energy_hydrology.jl
kmdeck: Since this is a relatively large PR already, in terms of number of files changed, I opened an issue and linked in the PR description

PR #1073, opened 2025-03-18
Update leaderboard for longer run of snowy land

ph-kev:
This PR updates the leaderboard for ClimaLand and uses the last 12 months of either the observational and simulation data for computing RMSEs.

For example, if the observational data is only defined up to November 2013 and the simulation ended at the end of 2018, then the plot would look like this.
et_bias_plot

Similarly, the plot of the RMSE and bias would look like this.
global_rmse_and_bias_graphs

## TODO

- [x] Make a new release of ClimaAnalysis
- [x] Add check for whether all short names are valid in data sources. This is because there might be missing variables depending on whether we are using ILAMB or ERA5 data.

## Summary by CodeRabbit

- **New Features**
  - Updated artifact configuration to include a new dataset for ERA5 monthly averages (1979–2024) with direct download access.
  - Enhanced data handling by adding extra simulation variables and multi-source observational support.
  - Introduced monthly and seasonal leaderboard views with improved time windowing and visualization.
  - Adjusted simulation start dates based on run duration.
  - Provided a helper for easy retrieval of the ERA5 monthly averages dataset.
  - Updated project dependencies to the latest version of ClimaAnalysis (0.5.16).

### DIFF
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.11.4"
 manifest_format = "2.0"
-project_hash = "4fe4935d5d45f04880ac419a664997320feb94e4"
+project_hash = "556cc366c0974a7c1af7cf9d7cb51d21d3254f47"
 
 [[deps.ADTypes]]
 git-tree-sha1 = "e2478490447631aedba0823d4d7a80b2cc8cdb32"
@@ -365,9 +365,9 @@ weakdeps = ["SparseArrays"]
 
 [[deps.ClimaAnalysis]]
 deps = ["Artifacts", "Dates", "Interpolations", "NCDatasets", "NaNStatistics", "OrderedCollections", "Reexport", "Statistics", "Unitful"]
-git-tree-sha1 = "1cae1cd626bd636bba0e1a4a681e0296372848bc"
+git-tree-sha1 = "0a9f4a6de147d0a9dc09f2e8f702f76dd6777e07"
 uuid = "29b5916a-a76c-4e73-9657-3c8fd22e65e6"
-version = "0.5.15"
+version = "0.5.16"
 weakdeps = ["GeoMakie", "Makie"]
 
     [deps.ClimaAnalysis.extensions]
--- a/.buildkite/Project.toml
+++ b/.buildkite/Project.toml
@@ -41,7 +41,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Thermodynamics = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 
 [compat]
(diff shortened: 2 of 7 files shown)


ph-kev:
These are all the new plots generated by the leaderboard.
ERA5_global_rmse_and_bias_graphs
ERA5_seasonal_avg_over_all_years
ERA5_seasonal_global_rmse_and_bias_graphs

Sbozzolo:
@kmdeck @AlexisRenchon @juliasloan25 

How do these plots look like?

AlexisRenchon:
These plots look really good to me. 
I just have suggestions but they are not essential at all, so please feel free to ignore.
- some redundant information could free space, for example the colorbars are the same horizontally for the second plot. (however, I understand it can be helpful to have them next to each plot, for this may be personal preference). likewise for the x-axis on the 3rd plot.
- labels could use latex or unicode instead of ^-2
- legend for 1st and 3rd figures could have individual years in blue (maybe saying year x to year y), why make final year black? (and why not average of all year black instead?)

again, these are just thoughts and the figures are great as is already.

Sbozzolo:
> legend for 1st and 3rd figures could have individual years in blue (maybe saying year x to year y), why make final year black? (and why not average of all year black instead?)

We can easily add the average in black and keep all the years in blue

ph-kev:
@coderabbitai review

ph-kev:
@coderabbitai full review

kmdeck:
thanks! I think these plots look good. A couple questions:
- do we automatically get both seasonal and monthly cycles?
- is ERA5 the default?

kmdeck:
The final year looks different from the others in a statistical way (it's not near the middle of the other years pretty consistently over all fluxes). Does this mean the fluxes are shifting over time (spin up not complete when we start)?

ph-kev:
> thanks! I think these plots look good. A couple questions:
> 
>     * do we automatically get both seasonal and monthly cycles?
> 
>     * is ERA5 the default?

Both seasonal and monthly cycles should be produced. Both ERA5 and ILAMB plots are produced, but I can change it to just produce ERA5 plots if that is more useful.

kmdeck:
> > thanks! I think these plots look good. A couple questions:
> > ```
> > * do we automatically get both seasonal and monthly cycles?
> > 
> > * is ERA5 the default?
> > ```
> 
> Both seasonal and monthly cycles should be produced. Both ERA5 and ILAMB plots are produced, but I can change it to just produce ERA5 plots if that is more useful.

All good is as. Thank you!

ph-kev:
> The final year looks different from the others in a statistical way (it's not near the middle of the other years pretty consistently over all fluxes). Does this mean the fluxes are shifting over time (spin up not complete when we start)?

I believe that this interpretation is correct. However, this is using old data.

Sbozzolo (approved):

PR #1018, opened 2025-02-07
global plots for paper [WIP]

juliasloan25:
## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #1076, opened 2025-03-18
add global heatmaps, bias plots, seasonal cycle plots to longruns

juliasloan25:
## Purpose 
https://github.com/CliMA/ClimaLand.jl/pull/1018/ added some useful plotting functions to make plots for the ClimaLand paper, for variables LE, H, LW_net, SW_net, LW_u, SW_u. This PR cleans those up and reduces some dependencies on ClimaAnalysis that required changes in that package.

Note that downward longwave and shortwave radiative fluxes come from a different artifact than the other quantities. This artifact contains hourly data so it's much larger than the artifact containing monthly data we use for the other variables. The hourly data gets averaged to monthly intervals to match the other data, but this takes some time to run. Because of this, by default upward LW and SW are plotted; net LW and SW can be selected by setting `plot_net_radiation = true`, but this takes longer to produce the plots. In the future this could be fixed by adding downward LW and SW to the monthly artifact and re-generating it.
 
## Content
- [x] add functions to get simulation data and observation data from ERA5
- [x] add function to generate global heatmaps and bias plots vs ERA5
- [x] add function to generate seasonal cycle plot
- [x] make these plots by default for all longruns

### DIFF
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.11.3"
 manifest_format = "2.0"
-project_hash = "3f4ceb30ba99ebd32a38b3cd7ce9703605f14393"
+project_hash = "ff55f579bdc785e1d8b4b219e86fee530635079c"
 
 [[deps.ADTypes]]
 git-tree-sha1 = "e2478490447631aedba0823d4d7a80b2cc8cdb32"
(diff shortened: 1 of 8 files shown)


juliasloan25:
Running locally with the data from longrun [#3785](https://buildkite.com/clima/climaland-long-runs/builds/3785), the scripts produce the following plots:
global_plots_bias
seasonal_cycle

AlexisRenchon:
I have more of a general comment, these plots are very useful, and the functions makes them generalizeable already, so I think they should be in a src folder somewhere (ClimaLand? ClimaAnalysis?), so that we can call them (they should be extensions I guess so that Makie is not a direct dependency) - for example, it is nice to call these plots for output of calibration files...

PR #1088, opened 2025-03-26
tag v0.15.12

juliasloan25:
## Purpose 
Tag a new release so ClimaCoupler can use the flux calculation functions that were recently merged in. Also updates dependencies:
```
    Updating `~/clima/ClimaLand.jl/.buildkite/Project.toml`
  [052768ef] ↑ CUDA v5.6.1 ⇒ v5.7.1
  [29b5916a] ↑ ClimaAnalysis v0.5.13 ⇒ v0.5.15
  [d414da3d] ↑ ClimaCore v0.14.26 ⇒ v0.14.28
  [5c42b081] ↑ ClimaParams v0.10.22 ⇒ v0.10.24
  [595c0a79] ↑ ClimaTimeSteppers v0.8.2 ⇒ v0.8.3
  [b3f4f4ca] ↑ ClimaUtilities v0.1.23 ⇒ v0.1.24
  [85f8d34a] ↑ NCDatasets v0.14.6 ⇒ v0.14.7
  [0bca4576] ↑ SciMLBase v2.75.1 ⇒ v2.79.0
  [b60c26fb] ↑ Thermodynamics v0.12.9 ⇒ v0.12.10
  [02a925ec] ↑ cuDNN v1.4.1 ⇒ v1.4.2
```

PR #1089, opened 2025-03-26
fix full land coupled boundary_var_types

juliasloan25:
## Purpose 
In the PR adding flux calculation functions for the coupled full land model #1062, I incorrectly extended `boundary_var_types` for EnergyHydrology with a CoupledAtmosphere. I'm not sure where the complex types there came from, but they're fixed now (tested from ClimaCoupler and pass). The only difference between the coupled and prescribed atmosphere case is the presence of momentum fluxes in the cache, which is what we want. This change is in `src/standalone/Soil/boundary_conditions.jl`

I'm also tagging a patch release in this PR instead of rebasing #1088 after the `boundary_var_types` fix merges.

Note: The downstream ClimaCoupler test is failing because the interface for `surface_specific_humidity` was changed to take in the atmosphere as the first argument. This will be fixed on the coupler side by changing the call [here](https://github.com/CliMA/ClimaCoupler.jl/blob/1a61e29fa479f8113300b56740a5f43a78c78505/experiments/ClimaEarth/components/land/climaland_bucket.jl#L256).

### DIFF
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -90,9 +90,9 @@ uuid = "4c88cf16-eb10-579e-8560-4a9242c79595"
 version = "0.8.11"
 
 [[deps.ArgCheck]]
-git-tree-sha1 = "680b3b8759bd4c54052ada14e52355ab69e07876"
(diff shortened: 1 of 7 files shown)


kmdeck (approved):
This makes sense! Thanks!

PR #1092, opened 2025-03-28
Remove ArtifactWrappers

Sbozzolo:
https://github.com/CliMA/ClimaArtifacts/pull/116 adds the last two missing artifacts, so now ArtifactWrappers can be removed from ClimaLand


imreddyTeja (approved):
Looks good to me

PR #1093, opened 2025-03-29
dispatch on return_momentum_fluxes type in flux calculations

juliasloan25:
## Purpose 
Our current `turbulent_fluxes_at_a_point` functions are not `isbits` (not GPU compatible) because the return type (4-Tuple vs 6-Tuple) depends on the value of an input (`return_momentum_fluxes`). This PR fixes this by changing `return_momentum_fluxes` from a `Boolean` to a `Val` that we can dispatch on, and it also pulls out a lot of the code in `turbulent_fluxes_at_a_point` into a shared function `turbulent_fluxes_helper` for the two cases.

## To-do
- [ ] test from ClimaCoupler (on GPU 😅)

## Content
- [x] change `return_momentum_fluxes` to a `Val`, dispatch on this in `turbulent_fluxes_at_a_point`
- [x] add `turbulent_fluxes_helper` function to do the shared computation

### DIFF
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -298,6 +298,7 @@ function turbulent_fluxes!(
 
     dest .=
         turbulent_fluxes_at_a_point.(
+            Val(false), # return_momentum_fluxes
             T_sfc,
             q_sfc,
             ρ_sfc,
@@ -345,6 +346,7 @@ function coupler_compute_turbulent_fluxes!(
 
     dest .=
         turbulent_fluxes_at_a_point.(
+            Val(true), # return_momentum_fluxes
             T_sfc,
             q_sfc,
             ρ_sfc,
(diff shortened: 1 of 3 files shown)


.buildkite/Project.toml
```diff
@@ -40,10 +39,9 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
...
 Thermodynamics = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
-cuDNN = "02a925ec-e4fe-4b08-9a7e-0d78e3d38ccd"
 
 [compat]
-ClimaAnalysis = "0.5.12"
+ClimaAnalysis = "0.5.15"
 ClimaTimeSteppers = "0.7, 0.8"
 Statistics = "1"
-Flux = "~0.14"
+Flux = "0.15"
```
kmdeck, line 47: Are these changes needed?

NEWS.md
```diff
@@ -3,6 +3,16 @@ ClimaLand.jl Release Notes
 
 main
 -------
+
+v0.5.13
+-------
+- fix for GPU compatibility of integrated land model coupled
```
kmdeck, line 9: is this related to this PR? Also all the snow changes

kmdeck: maybe a rebase is needed?

src/shared_utilities/drivers.jl
```diff
@@ -405,7 +431,7 @@ from unit conversion of evaporation from a mass to a liquid water volume
 flux. When r_sfc is nonzero, an additional resistance is applied in series
 to the vapor flux (and hence also the latent heat flux).
 """
-function turbulent_fluxes_at_a_point(
+function turbulent_fluxes_helper(
```
kmdeck, line 434: Do you think it would be clearer to call this "compute_turbulent_fluxes_at_a_point"?

It's a long name, but we have a number of "turbulent fluxes" functions now, just wondering if we could find a more descriptive name than "helper"

kmdeck: or maybe "full_turbulent_fluxes_at_a_point" (because it returns momentum fluxes too? not sure)

kmdeck: or maybe - the outer function should be the helper? and the inner one "turbulent_fluxes_at_a_point"? not sure!

juliasloan25: Yes good point! I wanted to keep the outer one as `turbulent_fluxes_at_a_point` to minimize the changes we have to make. I'll rename the helper to `compute_turbulent_fluxes_at_a_point`, if that's okay with you

kmdeck (approved):
looks good! Thank you!

PR #1096, opened 2025-04-04
Create nsight profiles for benchmarks

charleskawczynski:
This PR adds the use of nsight-systems to generate an nsight report for the gpu benchmarks.


kmdeck (approved):
I think this looks fine to merge? Both performance profilers have utility so keeping them both seems good to me.

experiments/benchmarks/snowy_land.jl
```diff
@@ -361,11 +382,11 @@ function setup_prob(t0, tf, Δt; outdir = outdir, nelements = (101, 15))
     return prob, cb
 end
 
-function setup_simulation(; greet = false)
+function setup_simulation(; greet = false, nelements = (101, 15))
     t0 = 0.0
     tf = 60 * 60.0 * 6
     Δt = 450.0
-    nelements = (101, 15)
+    nelements = nelements
```
charleskawczynski, line 389:
```suggestion
```

charleskawczynski (commented):
We may want to skip or warn when running on gpu with the flame graph, which isn't a representative measure for performance because it does not measure compute on the device.

Aside from that, this looks good to me.

PR #1101, opened 2025-04-09
add compat for formatter

kmdeck:
## Purpose 
JuliaFormatter recently release v2. Our current files in main were all formatted with v1. This fixes our compat to v1 so that the formatting check in CI continues to pass. The alternative would be to reformat all files with v2, but this is something we would need to do whenever JuliaFormatter has a major release. Restricting to v1 is simpler in that sense.

## Content

----

## Summary by CodeRabbit

- **Chores**
	- Updated project configuration by adding a compatibility specification for the code formatting dependency to ensure smoother dependency resolution.

### DIFF
--- a/.dev/Project.toml
+++ b/.dev/Project.toml
@@ -1,2 +1,5 @@
 [deps]
 JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
+
+[compat]
+JuliaFormatter = "1"


.dev/Project.toml
```diff
@@ -1,2 +1,5 @@
 [deps]
-JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
\ No newline at end of file
+JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
+
+[compat]
+JuliaFormatter = "=1"
```
Sbozzolo, line 5:
```suggestion
JuliaFormatter = "1"
```

This includes any version in the 1.X series, but not in the 2.X (or 3.X)

Sbozzolo (approved):

PR #1103, opened 2025-04-11
Add default_diagnostic_num_points

Sbozzolo:
When using GL quadrature, it makes little sense to have more than one point per element. Interpolating onto multiple points per element leads to duplicated values.

This doesn't change the long runs.

### DIFF
--- a/src/diagnostics/diagnostic.jl
+++ b/src/diagnostics/diagnostic.jl
@@ -156,3 +156,31 @@ include("define_diagnostics.jl")
 
 # Default diagnostics and higher level interfaces
 include("default_diagnostics.jl")
+
+if pkgversion(ClimaDiagnostics) < v"0.2.13"
+    # Default diagnostic resolution given a Space (approximately one point per
+    # element)
+    function default_diagnostic_num_points(domain::SphericalShell)
+        num_horizontal_elements, num_vertical_elements = domain.nelements
+
+        # 4 panels cover the cubed sphere from -180 to 180 in long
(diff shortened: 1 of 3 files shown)


src/diagnostics/diagnostic.jl
```diff
@@ -156,3 +156,29 @@ include("define_diagnostics.jl")
 
 # Default diagnostics and higher level interfaces
 include("default_diagnostics.jl")
+
+if pkgversion(ClimaDiagnostics) < v"0.2.13"
+    # Default diagnostic resolution given a Space (approximately one point per
+    # element)
+    function default_diagnostic_num_points(domain::SphericalShell)
```
kmdeck, line 163: What do you think about just including this function in ClimaLand, and not using the one in ClimaDiagnostics?

In a way, it is ClimaLand's choice.

Sbozzolo: My goal would be to remove both the functions completely eventually. This setting is now default in the NetCDFWriter, so we could remove both functions once we switch to use `0.2.13` as minimum compat.

kmdeck: ok! sounds good. Is it worth making an issue to remind us to do so?

kmdeck (approved):

src/diagnostics/diagnostic.jl
Sbozzolo: Opened a more general one for ClimaCore as well.

PR #1104, opened 2025-04-11
Add e_per_area to bucket cache

### LINKED ISSUE #1091: Add `energy_per_area` to Bucket cache
In ClimaCoupler, when we want to track conservation for Bucket runs (i.e. for Slabplanet, not for AMIP) we manually construct a new cache for the bucket that contains a field for energy per surface area. Since this is only one additional surface field, it would be nice to add it to the bucket by default, so that we can remove the cache-modifying infrastructure from ClimaCoupler.

imreddyTeja:
## Purpose 
closes #1091 

## Content

----

### DIFF
--- a/src/standalone/Bucket/Bucket.jl
+++ b/src/standalone/Bucket/Bucket.jl
@@ -321,6 +321,7 @@ auxiliary_types(::BucketModel{FT}) where {FT} = (
     FT,
     FT,
     ClimaCore.Geometry.WVector{FT},
+    FT,
 )
 auxiliary_vars(::BucketModel) = (
     :q_sfc,
@@ -336,8 +337,10 @@ auxiliary_vars(::BucketModel) = (
     :snow_melt,
     :infiltration,
     :top_bc_wvec,
+    :e_per_area,
 )
 auxiliary_domain_names(::BucketModel) = (
+    :surface,
     :surface,
     :surface,
     :surface,


kmdeck:
@imreddyTeja we have some functionality for this recently added: total_energy_per_area! and total_liq_water_vol_per_area!

juliasloan25:
> @imreddyTeja we have some functionality for this recently added: total_energy_per_area! and total_liq_water_vol_per_area!

I don't think those functions are extended for the bucket though. It seems like there are 2 options for how we can move this forward to remove `get_new_cache` in the coupler:
- add `e_per_area` to the bucket cache, like Teja has in this PR
- extend `total_energy_per_area!` for the bucket, so the coupler can call that

I slightly prefer allocating space for it in the bucket cache, so we can access that field directly instead of allocating or using temp fields in the coupler

kmdeck:
@juliasloan25 that works! We can add the variables to the bucket cache and update them in ClimaLand using the extended methods of total_energy_per_area! (for the bucket)

juliasloan25:
@imreddyTeja do you want to continue working on this, or do you want me to take it from here?

imreddyTeja:
> @imreddyTeja do you want to continue working on this, or do you want me to take it from here?

I'm fine with either. If I do it, I might need a bit of guidance on synchronizing the changes between ClimaLand and the coupler.

Also a very minor thing: right now the field is called `e_per_area`. Should it be changed to `energy_per_area` per #1091 ?

src/standalone/Bucket/Bucket.jl
```diff
@@ -576,6 +581,38 @@ function surface_air_density(
...
+    p,
+    t,
+)
+    surface_field .= 0
+    ClimaCore.Operators.column_integral_definite!(
+        surface_field,
+        model.parameters.ρc_soil .* Y.bucket.T,
+    )
+
+    surface_field .+=
```
juliasloan25, line 610: This line adds latent heat due to snow melt (I'll add a comment for this)

src/standalone/Bucket/Bucket.jl
```diff
@@ -336,6 +337,7 @@ auxiliary_vars(::BucketModel) = (
     :snow_melt,
     :infiltration,
     :top_bc_wvec,
+    :e_per_area,
```
kmdeck, line 340: Do we also need liquid water volume per area?

In the soil model, we call these "total_energy" and "total_water". Could we use those names here, or else also change the soil aux names to be "e_per_area" and "liq_water_vol_per_area"?

src/standalone/Bucket/Bucket.jl
kmdeck: I believe this is adding the latent energy of snow. I think we could add a comment along the lines of
"All components of the bucket model are at the same temperature, with volumetric specific heat approximated with the parameter rho c_soil. Additional energy is present in the land system due to the latent heat of fusion of frozen water, in the form of snow. We also add in this energy (below)"

kmdeck: and the first term is the energy in soil/ground and the second is in snow

src/standalone/Bucket/Bucket.jl
juliasloan25: The ClimaCoupler method to compute total water doesn't require allocating a new field in the bucket cache, so we don't have the same need to add it here ([this](https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/ClimaEarth/components/land/climaland_bucket.jl#L280-L284) is where it's computed in the coupler). But it could be nice to do the same thing for energy and water, and it would make the access from the coupler simpler. I'm okay with either way - do you have a preference?

I'll rename this to `total_energy` to match the soil!

kmdeck: Thanks!

I'm in favor of adding the water term here too since it makes the coupler work a bit simpler

src/standalone/Bucket/Bucket.jl
juliasloan25: This is helpful, thank you!

src/standalone/Bucket/Bucket.jl
```diff
@@ -576,6 +591,76 @@ function surface_air_density(
...
+"""
+function ClimaLand.total_liq_water_vol_per_area!(
+    surface_field,
+    model::BucketModel,
+    Y,
+    p,
+    t,
+)
+    ρ_cloud_liq = LP.ρ_cloud_liq(model.parameters.earth_param_set)
+    @. surface_field = (Y.bucket.σS + Y.bucket.W + Y.bucket.Ws) * ρ_cloud_liq
```
kmdeck, line 660: This is converting from a liquid volume per area to a mass per area. I think given the function name this should be handled in the coupler...

kmdeck (approved):

PR #1108, opened 2025-04-15
Restore support for ClimaCore < 0.14.30

Sbozzolo:
ClimaCore `0.14.30` introduced support for masks and ClimaLand now uses it by default. This commit restores compatibility with previous versions of ClimaCore restoring the behavior before we had a mask: if users run with previous versions of ClimaCore, they will not get the mask.

This compatibility allows us to replicate previous results.


kmdeck:
I dont exactly see why we need to adjust the compat. With CC 0.14.30, you can still create the land domains without a mask (NoMask())

Sbozzolo:
> I dont exactly see why we need to adjust the compat. With CC 0.14.30, you can still create the land domains without a mask (NoMask()), so you can still run all the previous benchmarks without a mask

Keeping a lower compat allows us to work with a wider range of pacakges, which helps bisecting and isolating problems. If you bump the compat for ClimaCore, you make updating/downgrading individual packages impossible. For example, I could no longer re-run the paper plots, or test if the regression was due ClimaCore or ClimaLand.

This is particularly relevant for ClimaCoupler. Restrictive compats entangle various package versions, which make debugging regressions harder. 

When possible, I think we should try to keep bounds low. In this case, it is just matter of adding two compat functions and adding two checks in the tests, which I argue is a very small price to pay for a much larger compatibility bounds. The functions are also isolated in a `compat.jl` file which can be removed in the future.

kmdeck (approved):
Looks fine to me! Once we are comfortable with the mask (working as it should) and using CC versions well above 0.14.30, we can remove this (i.e. at some point 0.14.30+ will become the new standard, and we wont need to test older versions)

PR #1109, opened 2025-04-15
Automatically pick mask resolution

Sbozzolo:
All the information needed to select the resolution for the land-sea mask is in the space. This commit automatically select the correct file to use when working with land-sea masks.

### DIFF
--- a/src/shared_utilities/Domains.jl
+++ b/src/shared_utilities/Domains.jl
@@ -971,6 +971,77 @@ depth(space::ClimaCore.Spaces.CenterFiniteDifferenceSpace) =
         space.grid.topology.mesh.domain.coord_min
     ).z
 
+"""
+    horizontal_resolution_degrees(domain::AbstractDomain)
+
+Return a tuple with the approximate resolution on the domain in degrees along
+the two directions.
+
+For boxes and planes, the order is `(latitude, longitude)`.
+
+Examples
+=======
+
+```jldoctest
+julia> using ClimaLand
+
+julia> domain = ClimaLand.Domains.SphericalShell(;
+        radius = 6300e3,
+        depth = 15.,
+        nelements = (10, 3),
+        dz_tuple = ((1.0, 0.05)),
+    );
+
+julia> ClimaLand.Domains.average_horizontal_resolution_degrees(domain)
+(9.0, 9.0)
+
+julia> domain = ClimaLand.Domains.Plane(;
+        xlim = (50000.0, 80000.),
+        ylim = (30000.0, 40000.),
+        longlat = (-118.14452, 34.14778),
+        nelements = (20, 3),
+    );
+
+julia> ClimaLand.Domains.average_horizontal_resolution_degrees(domain)
+(0.05839157, 0.20961125)
+```
+"""
+function average_horizontal_resolution_degrees(
+    domain::Union{SphericalShell, SphericalSurface},
+)
+    num_elements_lat = num_elements_lon = first(domain.nelements)
+    quad = ClimaCore.Spaces.quadrature_style(domain.space.surface)
+    num_points_per_element =
+        ClimaCore.Quadratures.unique_degrees_of_freedom(quad)
+
+    # There are 2 full cubed-sphere panels along latitudes, and 4 along
+    # longitudes
+    return (
+        180 / (2 * num_points_per_element * num_elements_lat),
+        360 / (4 * num_points_per_element * num_elements_lon),
+    )
+end
+
+function average_horizontal_resolution_degrees(domain::Union{Plane, HybridBox})
+    isnothing(domain.longlat) &&
+        error("Can only compute resolution with domains in latlong")
+
+    # x is Lat, y is Lon because ClimaCore defines LatLongPoints
+    num_elements_lat, num_elements_lon = domain.nelements[1:2]
+    delta_lat = domain.xlim[2] - domain.xlim[1]
+    delta_lon = domain.ylim[2] - domain.ylim[1]
+
+    quad = ClimaCore.Spaces.quadrature_style(domain.space.surface)
+    num_points_per_element =
+        ClimaCore.Quadratures.unique_degrees_of_freedom(quad)
+
+    return (
+        delta_lat / (num_points_per_element * num_elements_lat),
+        delta_lon / (num_points_per_element * num_elements_lon),
+    )
+end
+
+
 export AbstractDomain
 export Column, Plane, HybridBox, Point, SphericalShell, SphericalSurface
 export coordinates,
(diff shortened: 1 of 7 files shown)


src/shared_utilities/utils.jl
```diff
@@ -686,6 +680,19 @@ function landsea_mask(
         Interpolations.Flat(),
     ),
 )
+    # Assuming 1 degree is 100 km = 1e5 m, we can estimate the resolution in
```
kmdeck, line 683: I am actually adding a helper function for this in a separate PR:
```
"""
    effective_horizontal_resolution(nelements, npolynomial)
Helper function which returns the effective horizontal resolution of 
a global simulation in degrees
"""
function effective_horizontal_resolution(nelements, npolynomial)
    if npolynomial == 0
        return sqrt(360 * 101 / (nelements)^2 / 6)
    else
        return sqrt(360 * 101 / (nelements * npolynomial)^2 / 6)
    end
end
```

Do you want to add this to ClimaLand.Domains.jl instead in this PR?

src/shared_utilities/utils.jl
```diff
@@ -686,6 +680,19 @@ function landsea_mask(
...
+    nodal_distance_in_meters =
+        ClimaCore.Spaces.node_horizontal_length_scale(surface_space)
+    one_degree_in_meters = 100_000
+
+    nodal_distance_in_degrees = nodal_distance_in_meters / one_degree_in_meters
+    nodal_distance_in_arcsec = nodal_distance_in_degrees * 3600
+
+    # Pick the landsea mask at 60 arcseconds if the nodal distance is more than
+    # 120", otherwise pick the higher resolution
+    resultion = nodal_distance_in_arcsec > 120 ? "60arcs" : "30arcs"
```
kmdeck, line 694: resolution

src/shared_utilities/utils.jl
Sbozzolo: I think this would work only for a global simulation, but we also have regional simulations, so I would recommend using the functions I used in this PR

kmdeck: Could you add it to src/shared_utilities/Domains.jl?

Sbozzolo: Yes

charleskawczynski: Shouldn't such a function live in ClimaCore?

src/shared_utilities/utils.jl
```diff
@@ -686,6 +680,19 @@ function landsea_mask(
...
+    nodal_distance_in_meters =
+        ClimaCore.Spaces.node_horizontal_length_scale(surface_space)
+    one_degree_in_meters = 100_000
+
+    nodal_distance_in_degrees = nodal_distance_in_meters / one_degree_in_meters
+    nodal_distance_in_arcsec = nodal_distance_in_degrees * 3600
+
+    # Pick the landsea mask at 60 arcseconds if the nodal distance is more than
+    # 120", otherwise pick the higher resolution
+    resolution = nodal_distance_in_arcsec > 120 ? "60arcs" : "30arcs"
```
charleskawczynski, line 694: I'm not really a fan of the hard-coded things here, but if they really will never change, that's fine with me.

src/shared_utilities/utils.jl
```diff
@@ -686,6 +680,19 @@ function landsea_mask(
         Interpolations.Flat(),
     ),
 )
+    # Assuming 1 degree is 100 km = 1e5 m, we can estimate the resolution in
+    # degrees by taking the nodal_distance_meters and diving by 1e5
+    nodal_distance_in_meters =
+        ClimaCore.Spaces.node_horizontal_length_scale(surface_space)
+    one_degree_in_meters = 100_000
```
charleskawczynski, line 687: Can you explain what this is?

charleskawczynski (commented):
Overall I'm wondering if we can write more generic code that won't need changed

src/shared_utilities/utils.jl
Sbozzolo: What specifically?

src/shared_utilities/utils.jl
Sbozzolo: > Shouldn't such a function live in ClimaCore?

I am not sure how to do this in full generality in ClimaCore. 

The function makes use of an assumption that is true in ClimaLand but not in ClimaCore: that we are on planet Earth (so that 90 degrees are 10e6 meters by definition).
This assumption is also used in our regional simulations, which are boxes with LatLong coordinates. The associated spaces do not contain the radius.

One way we could do this is by taking the difference between adjacent points of the `coordinate_field`, but I don't know how to do that.

Sbozzolo: I realized there's a simple way I can do this

Sbozzolo: I am not sure. There seems to be no easy way to determine how many elements we have along each direction of the horizontal coordinates.

kmdeck:
I think if we make this function take the (ClimaLand) domain, it's straightforward:
`get_horizontal_resolution_degrees(sphere_domain) = sqrt(360*180/sphere_domain.nelements^2/6)`
`get_horizontal_resolution_degrees(hybrid_box) = extracts dx,dy, and center lat from the domain, and computes the resolution in degrees` This can error if center lat/lon not provided.
`get_horizontal_resolution_meters(hybrid_box) = extracts |dx|, and center lat from the domain, and computes the resolution in meters` or just returns |dx| if not center_lat/lon not provided

Sbozzolo:
I changed it to use `Domain`s and to return a tuple with the resolution on the two horizontal directions

src/shared_utilities/Domains.jl
```diff
@@ -982,6 +982,74 @@ depth(space::ClimaCore.Spaces.CenterFiniteDifferenceSpace) =
...
+"""
+function average_horizontal_resolution_degrees(
+    domain::Union{SphericalShell, SphericalSurface},
+)
+    num_elements_lat = num_elements_lon = first(domain.nelements)
+    quad = ClimaCore.Spaces.quadrature_style(domain.space.surface)
+    num_points_per_element =
+        ClimaCore.Quadratures.unique_degrees_of_freedom(quad)
+    return (
+        180 / (2 * num_points_per_element * num_elements_lat),
```
kmdeck, line 1028: Could you add a comment about the 2 vs 4 prefactor for future users?

kmdeck (approved):
thanks!

src/shared_utilities/utils.jl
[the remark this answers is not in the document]
Sbozzolo: The previous behavior was to pick 60 arcs, but I changed to pick 1 deg

src/shared_utilities/Domains.jl
Sbozzolo: Added a comment

PR #1110, opened 2025-04-15
Remove ClimaLandSimulations

Sbozzolo:
Restarting Gabriele's PR since the calibration example was refactored


docs/src/standalone/apps.jl
AlexisRenchon, line 1: This one should not be removed - it is unrelated to ClimaLandSimulations (it is making ParamViz apps)

Sbozzolo: It is using `ClimaLandSimulations` for `ClimaLandSimulations.Dashboards.fluxnet_app();`

I guess we have to move that stuff out

AlexisRenchon: yes - we can remove that line

AlexisRenchon (approved):

PR #1111, opened 2025-04-16
Make npolynomial = 0 by default

Sbozzolo:
We still had some experiments and tests with non-zero `npolynomial`. Diagnostics with the land-sea mask are correct only when `npolynomial` is zero, so this commit ensures that all our tests and experiments are using `npolynomial` of 0. This is also how we want to run `ClimaLand` in general, so there is little point in keeping `npolynomial != 0`.

I left one case with `npolynomial = 3` to check that it is still working.


kmdeck (approved):
looks good, thanks! Maybe you could add a comment to the test where npolynomial > 0 and note that that is the only test testing that (so we dont accidentally remove it later)?

Issue #1106, opened 2025-04-14
Clean up boundary flux and interaction updates


juliasloan25:
### The problem
Right now, we call `update_aux` and `update_boundary_fluxes` in both the explicit tendency and the implicit tendency at each timestep. This leads to a few problems:
- We have unnecessary duplicated function calls (e.g. cache variables that are only used in evaluating the implicit tendency are computed in both the explicit and implicit tendencies)
- In integrated models, interaction/boundary flux terms being computed in both the explicit and implicit tendencies means that component models may use different values in their tendencies, leading to conservation issues.
  - For example, ground heat flux is used when computing the snow explicit tendency and the soil implicit tendency. Since it's updated in `update_boundary_fluxes` in _both_ explicit and implicit tendencies (i.e. twice per timestep), the values used in these two computations will be different.

### The solution
There are 2 components to the solution we've come up with:
### Distinguish between explicit and implicit boundary flux evaluations
Instead of computing all boundary fluxes twice per step, we can compute the explicit boundary fluxes in the explicit tendency, and the implicit ones in the implicit tendency. This will make sure that each boundary flux is only updated once per step, and should resolve any related conservation issues.

Explicit boundary fluxes:
- root extraction
- runoff from snow
- snow tendencies (incl. excess fluxes)
- ground heat flux
- other existing boundary flux updates

Implicit boundary fluxes:
- soil boundary fluxes (LHF, SHF, E_l, E_i)
  - requires moving sublimation source from exp tendency to imp
- canopy boundary fluxes (LHF, SHF, T)
  - requires moving dY.canopy.\vartheta_l update from exp tendency to imp
- radiation
- any cache quantities used only for implicit boundary fluxes or tendency (albedo?)

#### Distinguish between explicit and implicit source terms
Right now, all source terms are computed when we compute the explicit tendency. However, some will depend on quantities computed only during the implicit boundary flux evaluation, so they should be updated during the implicit tendency instead.
 
For example, sublimation is a soil source term that is computed when we compute surface fluxes. As we move surface fluxes to be called in the implicit boundary fluxes method, the source attributed to sublimation must be updated within the implicit tendency, not in the explicit tendency.

We can do this by adding a field to each source struct that labels this source as `explicit` or `implicit`. We can then dispatch on this `Val` to determine whether the source should be evaluated during the explicit or implicit tendency computation.

### Plans for implementation
[...]

PR #1113, opened 2025-04-16
explicit and implicit sources for soil

kmdeck:
## Purpose 
As part of Issue #1106, this labels sources for the soil as explicit or implicit. 

## Content
Adds a boolean tag "explicit" to each concrete soil source. This is set by default and is not meant to be altered by the user (similar how to certain tendency terms are implicitly treated or explicitly treated - it's meant to be hardcoded).
we now compute sources in the implicit tendency as well as the explicit tendency, according to this ^^
I went with a simple `if` statement after talking to Charlie, he felt it might be clearer than adding an additional function to dispatch off of the value of `explicit`, and shouldnt impact performance

----

### DIFF
--- a/docs/tutorials/standalone/Soil/sublimation.jl
+++ b/docs/tutorials/standalone/Soil/sublimation.jl
@@ -158,7 +158,7 @@ init_soil!(Y, z, soil.parameters);
 # Timestepping:
 t0 = Float64(0)
 tf = Float64(24 * 3600 * 4)
-dt = Float64(5)
+dt = Float64(10)
 
 # We also set the initial conditions of the cache here:
 set_initial_cache! = make_set_initial_cache(soil)
@@ -175,7 +175,7 @@ timestepper = CTS.ARS111()
 ode_algo = CTS.IMEXAlgorithm(
     timestepper,
     CTS.NewtonsMethod(
-        max_iters = 1,
+        max_iters = 3,
         update_j = CTS.UpdateEvery(CTS.NewNewtonIteration),
     ),
 )
--- a/src/integrated/soil_snow_model.jl
+++ b/src/integrated/soil_snow_model.jl
@@ -390,9 +390,15 @@ end
     SoilSublimationwithSnow{FT} <: AbstractSoilSource{FT}
 
 Soil Sublimation source type. Used to defined a method
-of `ClimaLand.source!` for soil sublimation with snow present.
+of `ClimaLand.source!` for soil sublimation with snow present;
+treated implicitly
+in ϑ_l, ρe_int but explicitly in θ_i.
+
 """
-struct SoilSublimationwithSnow{FT} <: ClimaLand.Soil.AbstractSoilSource{FT} end
+@kwdef struct SoilSublimationwithSnow{FT} <:
+              ClimaLand.Soil.AbstractSoilSource{FT}
+    explicit::Bool = false
+end
 
 """
      source!(dY::ClimaCore.Fields.FieldVector,
@@ -419,7 +425,10 @@ function ClimaLand.source!(
     @. dY.soil.θ_i +=
         -p.soil.turbulent_fluxes.vapor_flux_ice *
         (1 - p.snow.snow_cover_fraction) *
-        _ρ_l / _ρ_i * heaviside(z + 2 * Δz_top) # only apply to top layer, recall that z is negative
+        _ρ_l / _ρ_i * heaviside(z + 2 * Δz_top) / (2 * Δz_top) # only apply to top layer, recall that z is negative
+    @. dY.soil.∫F_vol_liq_water_dt +=
+        -p.soil.turbulent_fluxes.vapor_flux_ice *
+        (1 - p.snow.snow_cover_fraction) # The integral of the source is designed to be this
     return nothing
 end
 
--- a/src/shared_utilities/Domains.jl
+++ b/src/shared_utilities/Domains.jl
@@ -935,7 +935,7 @@ function get_additional_coordinate_field_data(subsurface_space)
     z_face = ClimaCore.Fields.coordinate_field(face_space).z
     z_sfc = top_face_to_surface(z_face, surface_space)
     d = depth(subsurface_space)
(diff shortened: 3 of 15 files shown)


docs/tutorials/standalone/Soil/sublimation.jl
```diff
@@ -175,7 +175,7 @@ timestepper = CTS.ARS111()
 ode_algo = CTS.IMEXAlgorithm(
     timestepper,
     CTS.NewtonsMethod(
-        max_iters = 1,
+        max_iters = 3,
```
kmdeck, line 178: this is more in line with our "standard"

src/shared_utilities/Domains.jl
```diff
@@ -946,7 +946,7 @@ function get_additional_coordinate_field_data(subsurface_space)
     z_face = ClimaCore.Fields.coordinate_field(face_space).z
     z_sfc = top_face_to_surface(z_face, surface_space)
     d = depth(subsurface_space)
-    Δz_min = minimum(Δz_top)
+    Δz_min = minimum(Δz)
```
kmdeck, line 949: This is a bug

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -1084,7 +1103,7 @@ function soil_compute_turbulent_fluxes_at_a_point(
             q_sat_ice,
         )
         if q_air < q_sat_ice
-            β *= (θ_i_sfc / ν_sfc)^4
+            β *= (θ_i_sfc / ν_sfc)^2
```
kmdeck, line 1106: I'll revert this

src/integrated/soil_snow_model.jl
```diff
@@ -390,9 +390,17 @@ end
     SoilSublimationwithSnow{FT} <: AbstractSoilSource{FT}
 
 Soil Sublimation source type. Used to defined a method
-of `ClimaLand.source!` for soil sublimation with snow present.
+of `ClimaLand.source!` for soil sublimation with snow present;
+treated implicitly
+in ϑ_l but explicitly in θ_i. This is because we compute
```
juliasloan25, line 395: If sublimation is computed in the implicit tendency, why is this source explicit in `theta_i`?

src/standalone/Soil/rre.jl
```diff
@@ -179,6 +179,16 @@ function ClimaLand.make_compute_imp_tendency(model::RichardsModel)
...
+
+        # Source terms, also update the vertical integral of the source: p.soil.∫S_θ_liq_dz
+        p.soil.∫S_θ_liq_dz .= 0
+        # These change ∫S and dY by +=, which is why we zero them out above.
+        for src in model.sources
+            if !src.explicit
+                ClimaLand.source!(dY, src, Y, p, model)
+            end
+        end
+        @. dY.soil.∫F_vol_liq_water_dt += p.soil.∫S_θ_liq_dz # these source terms are stepped implicitly
```
juliasloan25, line 191: Why not update `dY` inside `source!`?

juliasloan25: (same for the explicit sources and for other models)

src/standalone/Soil/soil_heat_parameterizations.jl
```diff
@@ -20,7 +20,7 @@ Returns the thermal timescale for temperature differences across
 a typical thickness Δz to equilibrate.
 """
 function thermal_time(ρc::FT, Δz::FT, κ::FT) where {FT}
-    return ρc * Δz^2 / κ
+    return 3 * ρc * Δz^2 / κ
```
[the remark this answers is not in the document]
juliasloan25: It's okay here if the stability looks good, since we're making other behavioral changes too. But if things look worse or very different maybe we can move it to a separate PR

src/integrated/soil_snow_model.jl
kmdeck: Hmmm so the LHF, SHF, vapor flux, are computed at the same time. so what I intended was for them to be like (e..g. LHF(T_(t+dt), theta_i(t)))

I think this is the case if theta_i does not contribute to the Jacobian. Does that seem right?

juliasloan25: That sounds right! I think the second part of this comment is confusing - "This is because we compute sublimation and evaporation as part of the boundary flux computation for ϑ_l, which occurs in the implicit tendency."

I'm not sure what would make it more clear, maybe just leaving out that sentence and keeping the first part of the comment

src/standalone/Soil/rre.jl
juliasloan25: Okay! I hadn't thought about it before, but it might be nice to have all source contributions to `dY` done inside of `source!` (any maybe for `source!` to not modify `p` if possible but not necessarily :) )

src/integrated/soil_snow_model.jl
kmdeck: okay, I kept the first part only ( and noted we are also implicit in rhoe_int).

src/standalone/Soil/rre.jl
kmdeck: yes! I think this works out. nice! testing now and then will push

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -675,6 +664,7 @@ end
              )
 
 Computes the source terms for phase change.
+explicilty
```
juliasloan25, line 667: typo

juliasloan25 (approved):
This is very clean! My only suggestion is to add a page of documentation listing which sources are explicit/implicit for each model

src/standalone/Soil/energy_hydrology.jl
kmdeck: thanks!

kmdeck:
> This is very clean! My only suggestion is to add a page of documentation listing which sources are explicit/implicit for each model

I added a new docs page - will add more after the next PR (explicit and implicit boundary fluxes)

PR #1120, opened 2025-04-21
add buoyancy flux to cache in coupled case

### LINKED ISSUE #1119: Update boundary flux-related fields for coupler
In coupled runs, the atmosphere requires some additional boundary flux-related fields: `L_MO`, `ustar`, and `buoyancy_flux`. Since fluxes are computed within ClimaLand.jl for the integrated land model, we'll need to extend `coupler_compute_turbulent_fluxes!` within ClimaLand to also return these properties.

These 3 properties are returned by `SurfaceFluxes.surface_conditions`. We'll need to return them from `compute_turbulent_fluxes_at_a_point`, `turbulent_fluxes_at_a_point` (conditionally), and `coupler_compute_turbulent_fluxes`.

Without fixing this, coupled runs using the EDMF atmosphere option will be incorrect

juliasloan25:
## Purpose 
Add buoyancy flux (`buoy_flux`) to the cache of the soil, snow, and canopy models when running with a coupled atmosphere, along with the momentum fluxes which are already added in this case. These quantities are computed and returned in the respective `compute_turbulent_fluxes_at_a_point` methods.

closes #1119

## To-do
- [x] allocate space in caches for the new fields
- [x] return the new fields conditionally from flux calculation functions
- [x] rename `return_momentum_fluxes` option to `return_extra_fluxes`
- [x] test in coupled simulation - tested locally with https://github.com/CliMA/ClimaCoupler.jl/pull/1303

### DIFF
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -298,7 +298,7 @@ function turbulent_fluxes!(
 
     dest .=
         turbulent_fluxes_at_a_point.(
-            Val(false), # return_momentum_fluxes
+            Val(false), # return_extra_fluxes
             T_sfc,
             q_sfc,
             ρ_sfc,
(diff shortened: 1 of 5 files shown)


kmdeck (approved):
looks fine to me!

PR #1121, opened 2025-04-23
Performance improvements

kmdeck:
## Purpose 
This PR computes intermediate quantities in the soil Jacobian for the Richards and EnergyHydrology model. This circumvents performance issues in ClimaCore when many matrix field broadcast operations are fused.

This improves the Richards benchmark time from 1.5s to 0.6s, and the snowy land benchmark time from 3.29s to 0.741s. It changes the long run SYPD as:
snowy land model: 20 SYPD -> ~60 SYPD

## To-do
- understand why more NaNs appear in the full soil long run (this is due to PR #1113; not this current PR)

## Content
- adds bidiagonal matrix row fields to scratch space for the soil model, uses these in the jacobian computation
- does the same with a covariant3vector field
- updates benchmark times

----

### DIFF
--- a/experiments/benchmarks/bucket.jl
+++ b/experiments/benchmarks/bucket.jl
@@ -23,7 +23,8 @@ import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput
 
 import ClimaTimeSteppers as CTS
 import NCDatasets
-using ClimaCore
+import ClimaCore
+@show pkgversion(ClimaCore)
 import ClimaLand
 using ClimaParams
 using ClimaLand.Bucket:
@@ -66,13 +67,10 @@ function setup_prob(t0, tf, Δt; nelements = (200, 7))
     # We set up the problem in a function so that we can make multiple copies (for profiling)
 
     # Set up simulation domain
-    soil_depth = FT(3.5)
-    bucket_domain = ClimaLand.Domains.SphericalShell(;
-        radius = FT(6.3781e6),
-        depth = soil_depth,
-        nelements = nelements,
-        dz_tuple = FT.((1.0, 0.05)),
-    )
+    dz_tuple = FT.((1.0, 0.05))
+    depth = FT(3.5)
+    bucket_domain =
+        ClimaLand.global_domain(FT; nelements, dz_tuple, depth = depth)
     start_date = DateTime(2005)
 
     # Initialize parameters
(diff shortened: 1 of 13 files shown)


experiments/benchmarks/bucket.jl
```diff
@@ -259,7 +260,7 @@ if profiler == "flamegraph"
     end
 
     if get(ENV, "BUILDKITE_PIPELINE_SLUG", nothing) == "climaland-benchmark"
-        PREVIOUS_BEST_TIME = 1.69
+        PREVIOUS_BEST_TIME = 8.26
```
kmdeck, line 263: this is unrelated to this PR

src/standalone/Soil/rre.jl
```diff
@@ -283,6 +283,8 @@ function ClimaLand.auxiliary_vars(soil::RichardsModel)
         :total_water,
         :K,
         :ψ,
+        :bidiag_matrix,
```
kmdeck, line 286: naming suggestions welcome!

experiments/benchmarks/bucket.jl
```diff
@@ -24,6 +24,7 @@ import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput
 import ClimaTimeSteppers as CTS
 import NCDatasets
 using ClimaCore
+ClimaCore.Operators.use_fd_shmem() = false
```
charleskawczynski, line 27:
```suggestion
import ClimaCore
ClimaCore.Operators.use_fd_shmem() = false
```
Ah, this may actually be important. We need to `import` ClimaCore, as `using` gives no indication to julia that we want to override behavior.

kmdeck: Ill fix this!

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -343,6 +343,29 @@ function ClimaLand.make_compute_jacobian(model::EnergyHydrology{FT}) where {FT}
         ∂ϑres∂ϑ = matrix[@name(soil.ϑ_l), @name(soil.ϑ_l)]
         ∂ρeres∂ρe = matrix[@name(soil.ρe_int), @name(soil.ρe_int)]
         ∂ρeres∂ϑ = matrix[@name(soil.ρe_int), @name(soil.ϑ_l)]
+
+        # Derivatives with respect to ϑ:
```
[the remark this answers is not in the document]
kmdeck: or I could give them more informative names so it is less confusing what they are

charleskawczynski (approved):
Looks fine to me. It's probably better that we remove the `Operators.use_fd_shmem() = false` overloading, especially if these benchmarks are using masks (we currently do not support mask-aware fd stencils with shmem, so ClimaCore disables shmem when the mask is not `NoMask`).

src/standalone/Soil/energy_hydrology.jl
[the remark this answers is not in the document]
charleskawczynski: If `bidiag_matrix_scratch` is being computed and then used only locally here, then yeah, that should be safe. If it's used somewhere else, too, then I'd say it might need to be recomputed-- it depends on where it's computed, though.

juliasloan25:
Just a note: this PR is a breaking change because it introduces a `subsurface_face` space in the domain, which needs to be allocated when the coupler sets up the land domain. So our next release should be minor (0.16.0).

This wouldn't be a problem if ClimaCoupler used the `global_domain` function, which we'll switch to soon :)

PR #1123, opened 2025-04-23
split computation in jacobian

kmdeck:
## Content

----


kmdeck:
merging as part of #1121

PR #1132, opened 2025-04-29
Fix calibration coastline errors

AlexisRenchon:
Previously to this commit, we were using `ClimaAnalysis.apply_oceanmask`
to get locations (coordinates) on land to perform global calibration.
This was causing erroneous values at the coast, because it would
sometimes use ocean locations, as the ClimaAnalysis mask slightly differ
from the ClimaLand mask. Furthermore, we use a high threshold (0.99) for
the land mask to ensure only land pixels are used.

For example, longwave up at the coast had many zeros:
image

And, at the same locations, sensible heat had sometimes very large values 
image

### DIFF
--- a/experiments/calibration/make_training_locations.jl
+++ b/experiments/calibration/make_training_locations.jl
@@ -2,25 +2,8 @@ using ClimaAnalysis
 using ClimaLand
 using ClimaLand.Artifacts
 using ClimaComms
+using ClimaCore
 ClimaComms.@import_required_backends
-
-"""
-    diagnostics_lat_lon(nelements)
-
-Return latitudes and longitude for the diagnostics of the land
-model run at a given resolution (`nelements`).
-"""
-function diagnostics_lat_lon(nelements)
-    radius = 0.1 # These don't matter
-    depth = 0.1 # These don't matter
-
-    domain = ClimaLand.Domains.SphericalShell(; radius, depth, nelements)
-    num_long, num_lat, _ =
-        ClimaLand.Diagnostics.default_diagnostic_num_points(domain)
-    longs = collect(range(-180.0, 180.0, length = num_long))
-    lats = collect(range(-90.0, 90.0, length = num_lat))
-    return lats, longs
-end
 
 """
     make_training_locations(nelements)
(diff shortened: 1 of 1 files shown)


ph-kev:
I'm not sure where the plots come from, but if they are from ClimaAnalysis, something that would be worth exploring is this (issue)[https://github.com/CliMA/ClimaAnalysis.jl/issues/198]. The way is the masking is implemented is not that great and could be improved.

AlexisRenchon:
> I'm not sure where the plots come from, but if they are from ClimaAnalysis, something that would be worth exploring is this [issue](https://github.com/CliMA/ClimaAnalysis.jl/issues/198). The way that masking is implemented is not that great and could be improved.

The plots above are plotting the Observation_map from calibration. 
The mask used is indeed `ClimaAnalysis.apply_oceanmask`. 
The problem, we think, is that there might still be locations that are seen as ocean from the perspective of `ClimaLand.landsea_mask`, therefore the Observation_map still has erroneous values (which are seen in the plot, but also more importantly, lead to error in the calibration). 

My plan currently is to further filter out ocean locations from `ClimaLand.landsea_mask` when I chose the locations (coordinates) used to build the Observation_map. But if you issue address this, that could be helpful!

experiments/calibration/make_training_locations.jl
```diff
@@ -15,11 +16,14 @@ function diagnostics_lat_lon(nelements)
...
 
     domain = ClimaLand.Domains.SphericalShell(; radius, depth, nelements)
+
     num_long, num_lat, _ =
         ClimaLand.Diagnostics.default_diagnostic_num_points(domain)
     longs = collect(range(-180.0, 180.0, length = num_long))
     lats = collect(range(-90.0, 90.0, length = num_lat))
-    return lats, longs
+
+    mask = ClimaLand.landsea_mask(domain; threshold = 0.99)
```
kmdeck, line 25: can update the doc string to note that this also returns the ClimaLand mask as a Field, while lat and lon are arrays.

this function is a little confusing because it does two unrelated things; they just happen to both need the `domain`, so it was convenient to package them in one function.

I think it would be clearer to make the change I suggest below, and get rid of this function.

experiments/calibration/make_training_locations.jl
```diff
@@ -35,20 +39,19 @@ globe.
 - The function applies a land mask to identify suitable training locations
 """
 function make_training_locations(nelements)
-    lats, longs = diagnostics_lat_lon(nelements)
-    var = ClimaAnalysis.OutputVar(
-        Dict("long" => longs, "lat" => lats),
-        zeros(length(lats), length(longs)),
-    )
+    lats, longs, mask = diagnostics_lat_lon(nelements)
```
kmdeck, line 42: suggested change:
```
    domain = ClimaLand.Domains.SphericalShell(; 0.1, 0.1, nelements) # values of radius and depth (set to 0.1) do not matter; we just need information from the domain in the latitude/longitude directions.
# If the default number of diagnostic latitude and longitude points is not used when running the simulations, this will need to be changed. 
    num_long, num_lat, _ =
        ClimaLand.Diagnostics.default_diagnostic_num_points(domain)
    longs = collect(range(-180.0, 180.0, length = num_long))
    lats = collect(range(-90.0, 90.0, length = num_lat))
# We also need the mask in order to determine which points were treated as land by the simulation.
# We set the threshold of fractional area of land to be 0.99 here - cells with > 1% ocean are ignored in calibration.
    mask = ClimaLand.landsea_mask(domain; threshold = 0.99)
```

experiments/calibration/make_training_locations.jl
```diff
@@ -35,20 +39,19 @@ globe.
...
 
-    # Apply_oceanmask applies the mask over the ocean
-    vars_in_land = ClimaAnalysis.apply_oceanmask(var)
+    # Use the mask ClimaLand.landsea_mask
```
kmdeck, line 44: # Interpolate the ClimaLand land sea mask (Field) to the diagnostics grid (matrix)

experiments/calibration/make_training_locations.jl
```diff
@@ -35,20 +39,19 @@ globe.
...
+    target_hcoords = [
+        ClimaCore.Geometry.LatLongPoint(lat, lon) for lat in lats, lon in longs
+    ]
+    interpolated_mask =
+        Array(ClimaCore.Remapping.interpolate(mask; target_hcoords))
```
kmdeck, line 50: `# Select only locations where the interpolated mask is equal to 1`

experiments/calibration/make_training_locations.jl
```diff
@@ -35,20 +39,19 @@ globe.
 - The function applies a land mask to identify suitable training locations
```
kmdeck, line 39: Perhaps we can note here that:
- ClimaLand simulates only locations where its mask == 1 (based on a threshold set in the simulations). Since parameters are not defined over the ocean, this can lead to unphysical parameter combinations right at the coastline.
- The interpolated diagnostics are not mask aware; this again means that grid points treated as land by the simulation are combined with grid points that are not acted on by the simulation (typically set to zero, but possibly undefined). Again the result is that grid points on the coastline are not guaranteed to be physically realistic.
Consequently, here we define the mask used to select calibration points using a more stringent criterion (essentially, that the grid point in the simulation is only comprised of land, and the point in the diagnostic is only comprised of land)

kmdeck (commented):
This looks fine to me, I just added/suggested several comments to make so that we recall why we did this in the future

kmdeck (approved):

PR #1135, opened 2025-05-01
Compute zenith angles for CoupledRadiativeFluxes

imreddyTeja:
Related coupler [PR](https://github.com/CliMA/ClimaCoupler.jl/pull/1330)

`p.drivers.cos\thetas` is currently updated by the coupler
when using `CoupledRadiativeFluxes`.

This commit updates the value in `update_drivers!` instead.
To do this, the commit changes `CoupledRadiativeFluxes`
to contain a function that calculates zenith angle in radians
given time from sim start date and sim start date.

This adds an additional constructor for `CoupledRadiativeFluxes`,
which takes in a start date, latitude, and longitude
(and optionally insolation params), and
returns a struct with a zenith calculation function that uses
Insolation.jl to compute zenith angle for the provided lat and lon.

The default function used to compute the zenith angle
also replaces the duplicate defintions of `zenith_function`
used with `PrescribedRadiativeFluxes` and `prescribed_forcing_era5`.

Backwards compatibilty with ClimaCoupler is preserved.

Additionally, this PR fixes a minor bug in the `DriverUpdateCallback`, which happens when the simulation steps beyond the last time in the `updateat` vector.

## Purpose 
Alongside changes to the coupler, this will close [issue](https://github.com/CliMA/ClimaCoupler.jl/issues/1260)

## Content

----

### DIFF
--- a/src/shared_utilities/drivers.jl
+++ b/src/shared_utilities/drivers.jl
@@ -35,7 +35,8 @@ export AbstractAtmosphericDrivers,
     make_update_drivers,
     prescribed_lai_era5,
     prescribed_forcing_era5,
-    prescribed_analytic_forcing
+    prescribed_analytic_forcing,
+    default_zenith_angle
 
 """
      AbstractClimaLandDrivers{FT <: AbstractFloat}
@@ -185,11 +186,116 @@ PrescribedPrecipitation{FT}(liquid_precip) where {FT} =
     PrescribedPrecipitation{FT, typeof(liquid_precip)}(liquid_precip)
 
 """
-    CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT}
+    CoupledRadiativeFluxes{
+        FT,
+        F <: Union{Function, Nothing},
+        T,
+    } <: AbstractRadiativeDrivers{FT}
 
-To be used when coupling to an atmosphere model.
+To be used when coupling to an atmosphere model. Either both `θs` and `start_date`
+must be `nothing`, or both must not be `nothing``.
+
+During the driver update, cosθs is unchanged if `θs` is `nothing`. This behavior differs from
+the `PrescribedRadiativeFluxes` where `cosθs` set to `NaN` if `θs` is `nothing`.
+Otherwise, `θs` recieves the following arguments:
+(time_from_start, `start_date`), and is expected to return zenith angle at the given time.
+$(DocStringExtensions.FIELDS)
 """
-struct CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT} end
+struct CoupledRadiativeFluxes{FT, F <: Union{Function, Nothing}, T} <:
+       AbstractRadiativeDrivers{FT}
+    """Function that fills a climacore field with the zenith angle given the following arguments:
+    (time_from_start, `start_date`)"""
+    θs::F
+    "Start date - the datetime corresponding to t=0 for the simulation"
+    start_date::T
+    function CoupledRadiativeFluxes{FT, F, T}(
+        θs::F,
+        start_date::T,
+    ) where {FT, F, T}
+        (
+            (isnothing(θs) && isnothing(start_date)) ||
+            ((!isnothing(θs) && !isnothing(start_date)))
+        ) || error(
+            "CoupledRadiativeFluxes: `θs` and start_date` must both be `nothing` or both not `nothing`.",
+        )
+        new{FT, F, T}(θs, start_date)
+    end
+end
+
+# This constructor is here to maintain backwards compatability with ClimaCoupler
+# If this constructor is used, the coupler should calulate and update the zenith angle
+CoupledRadiativeFluxes{FT}() where {FT} =
+    CoupledRadiativeFluxes{FT, Nothing, Nothing}(nothing, nothing)
+
+CoupledRadiativeFluxes(::Type{FT}, args...) where {FT} =
+    CoupledRadiativeFluxes{FT}(args...)
+
+"""
+    CoupledRadiativeFluxes{FT}(
+        start_date::Dates.DateTime;
+        latitude,
+        longitude,
+        insol_params=LP.LandParameters(FT).insol_params,
+    )
+Creates a `CoupledRadiativeFluxes` object with a default zenith angle function that uses Insolation.jl
+to compute the zenith angle at a given time and location.
+"""
+function CoupledRadiativeFluxes{FT}(
+    start_date::DT;
+    latitude::LT,
+    longitude::LT,
+    insol_params::IP = LP.LandParameters(FT).insol_params,
+) where {FT, DT, LT, IP}
+    zenith_angle =
+        (t, s) -> default_zenith_angle(
+            t,
+            s;
+            latitude = latitude,
+            longitude = longitude,
+            insol_params = insol_params,
+        )
+    return CoupledRadiativeFluxes{FT, typeof(zenith_angle), DT}(
+        zenith_angle,
+        start_date,
+    )
+end
+
+"""
+    default_zenith_angle(
+        t::T,
+        start_date::Dates.DateTime;
+        latitude::LT,
+        longitude::LT,
+        insol_params::Insolation.Parameters.InsolationParameters{FT},
+    )
+
+Calculate zenith angle with Insolation for the given start date, insolation parameters, latitude,
+and longitude.
+
+`latitude` and `longitude` can be a collections or a Number.
+"""
+function default_zenith_angle(
+    t::T,
+    start_date::Dates.DateTime;
+    latitude::LT,
+    longitude::LT,
+    insol_params,
+) where {T, LT}
+    FT = eltype(latitude)
+    current_datetime =
+        T <: ITime ? date(t) : start_date + Dates.Second(round(t))
+    d, δ, η_UTC =
+        FT.(
+            Insolation.helper_instantaneous_zenith_angle(
+                current_datetime,
+                start_date,
+                insol_params,
+            )
+        )
+    # Reduces allocations by throwing away unwanted values
+    zenith_only = (args...) -> Insolation.instantaneous_zenith_angle(args...)[1]
+    return zenith_only.(d, δ, η_UTC, longitude, latitude)
+end
 
 """
     CoupledAtmosphere{FT} <: AbstractAtmosphericDrivers{FT}
@@ -1144,6 +1250,27 @@ function make_update_drivers(a::PrescribedPrecipitation{FT}) where {FT}
 end
 
 """
+    make_update_drivers(r::CoupledRadiativeFluxes{FT}) where {FT}
+
+Creates and returns a function which updates the driver variables
+in the case of a CoupledRadiativeFluxes.
+
+When `r.θs` is `nothing`, the cosine zenith angle
+not changed, and should be updated by the coupler. This differs from the behavior of
+`PrescribedRadiativeFluxes`, where the cosine zenith angle is set to `NaN` if `θs` is `nothing`.
+
+Otherwise, the cosine zenith angle is computed using `cos.(r.θs(t, r.start_date))`.
+"""
+make_update_drivers(r::CoupledRadiativeFluxes{FT, Nothing}) where {FT} =
+    (p, t) -> nothing
+function make_update_drivers(r::CoupledRadiativeFluxes{FT}) where {FT}
+    function update_drivers!(p, t)
+        p.drivers.cosθs .= cos.(r.θs(t, r.start_date))
+    end
+    return update_drivers!
+end
+
+"""
     make_update_drivers(r::PrescribedRadiativeFluxes{FT}) where {FT}
 
 Creates and returns a function which updates the driver variables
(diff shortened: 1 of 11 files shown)


experiments/integrated/performance/integrated_timestep_test.jl
```diff
@@ -172,6 +172,7 @@ atmos = ClimaLand.PrescribedAtmosphere(
     atmos_h,
     earth_param_set;
 )
+
```
imreddyTeja, line 175: remove this

src/shared_utilities/drivers.jl
```diff
@@ -185,11 +186,117 @@ PrescribedPrecipitation{FT}(liquid_precip) where {FT} =
     PrescribedPrecipitation{FT, typeof(liquid_precip)}(liquid_precip)
 
 """
-    CoupledRadiativeFluxes{FT} <: AbstractRadiativeDrivers{FT}
+    CoupledRadiativeFluxes{FT, F <: Union{Function, Nothing}} <: AbstractRadiativeDrivers{FT}
 
-To be used when coupling to an atmosphere model.
+To be used when coupling to an atmosphere model. If F is a function, cosθs is updated in
+the driver update.
```
kmdeck, line 192: Maybe we could clarify in which situations thetas isa Nothing?

src/shared_utilities/drivers.jl
```diff
@@ -185,11 +186,117 @@ PrescribedPrecipitation{FT}(liquid_precip) where {FT} =
...
+
+Returns a zenith angle function for the given start date, insolation parameters,
+and optionally location.
+
+If `longitude` and `latitude` are not supplied, the returned function takes a time in seconds,
+and latitude and longitude in `FT`, and returns the zenith angle at that time and location.
+If `longitude` and `latitude` are supplied, the returned function takes a time in seconds
+and returns the zenith angle at that time, using the supplied longitude and latitude.
+"""
+function make_zenith_angle(
```
kmdeck, line 243: Is it worth splitting this into two functions? One which does zenith angle at a point, the other global zenith angle?

Im not sure - they are very similar, so grouping into one function is reasonable, but it's also a little convoluted, and repeats a fair amount of code. 

What about something like:
```
# define make_zenith_angle_function_lat_lon_t(start_date, params), returns the function that takes lat/lon/t
function make_zenith_angle_function_t(start_date, params, lat, lon) 
    fn = make_zenith_angle_function_lat_lon_t(start_date, params)
    return (t) -> fn(t, lat, lon)
end
```

src/shared_utilities/drivers.jl
```diff
@@ -1143,6 +1250,17 @@ function make_update_drivers(a::PrescribedPrecipitation{FT}) where {FT}
     return update_drivers!
 end
 
+function make_update_drivers(r::CoupledRadiativeFluxes{FT}) where {FT}
+    # nothing to update if no zenith angle function is provided
+    isnothing(r.θs) && return (p, t) -> nothing
+    function update_drivers!(p, t)
+        # this assumes that the space has lat/lon coordinates
+        coords = Fields.coordinate_field(p.drivers.cosθs)
```
kmdeck, line 1258: this allocates a field each time it is called - maybe we can do what we have in PrescribedRadiativeFluxes?

https://github.com/CliMA/ClimaLand.jl/blob/a02f385cc4dc087ba0965a19f4c11e38a6aa8457/src/shared_utilities/drivers.jl#L1469

And, if we have the make zenith_angle function, we can use that there

src/shared_utilities/drivers.jl
kmdeck: Do we need both options, or is only the one that takes lat/lon as arguments used?

kmdeck (commented):
I left a couple comments; this is a nice change/improvement!

src/shared_utilities/drivers.jl
```diff
@@ -1156,6 +1274,7 @@ function make_update_drivers(r::PrescribedRadiativeFluxes{FT}) where {FT}
         # Next we update the zenith angle and diffuse fraction of light. These are either
         # both required, or neither required.
         if !isnothing(r.θs)
+            # TODO: Remove r.start_date as an argument
```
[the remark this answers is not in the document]
kmdeck: nevermind, I think this makes sense!

src/shared_utilities/drivers.jl
imreddyTeja: `coordinate_field` does not allocate:

```
julia> function trx(tff)
       t = Fields.coordinate_field(tff).lat
       t .+= 0.001
       return
       end
trx (generic function with 1 method)

julia> @allocated trx(tff)
0

File: src/shared_utilities/drivers.jl (hunk 5 above)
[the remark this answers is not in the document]
Comment (imreddyTeja) [reply src/shared_utilities/drivers.jl:1277]: I'm completely fine keeping `start_date` as a field of the driver if you think that is clearer, and I'll make the `CoupledRadiativeFluxes` driver match as well. 

My reasoning was mostly that if a new zenith function is being defined each time the simulation is made, then there's no need to carry the `start_date` around in the driver, but thinking about it now, my change just moves it down a level into the `zenith_function`. One pro of removing start date is it simplifies checking if the construction of the driver is valid (either has a function or doesn't vs. has function, start_date, and insolation params or has none of them.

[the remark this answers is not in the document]
Comment (imreddyTeja) [reply src/shared_utilities/drivers.jl:1277]: Removing `start_date` also matches the current pattern in `PrescribedRadiativeFluxes` of keeping `lat`, `long`, and `insol_params` in the function, rather than in the driver.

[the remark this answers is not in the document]
Comment (imreddyTeja) [reply src/shared_utilities/drivers.jl:1277]: Maybe it would be better to remove the `make_zenith_angle`, and just define a single `zenith_angle(t::T, start_date, long::FT, lat::FT, insole_params)`, and instead have the driver store `start_date` and `insol_params`,  `long`, and `lat` instead of holding `start_date` and a function that is parameterized with `insol_params`,  `long`, and `lat`

File: src/shared_utilities/drivers.jl (hunk 4 above)
Comment (kmdeck) [reply src/shared_utilities/drivers.jl:1258]: oh ok!

File: src/shared_utilities/drivers.jl (hunk 5 above)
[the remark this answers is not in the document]
Comment (kmdeck) [reply src/shared_utilities/drivers.jl:1277]: so, the two options are:
- all in the driver (pro - very explicit about start date, lat, lon, but this part would be more complex "One pro of removing start date is it simplifies checking if the construction of the driver is valid (either has a function or doesn't vs. has function, start_date, and insolation params or has none of them.")
- all in a closure in the function (basically the opposite pros and cons)
?

And in either case, all of our use cases would be able to use the same interface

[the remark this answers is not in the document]
Comment (imreddyTeja) [reply src/shared_utilities/drivers.jl:1277]: Yes I think that's accurate, but we could also keep the current `PrescribedRadiativeFluxes` style, which is a bit of a mix of the two options: insolation params, lat, and lon are in the closure, and start date and the function inside the driver.  The pro of this is no breaking change.  It seems like the "mixed" style has been working OK, so we could also hold off on implementing one of the two above options, and leave that for refactoring later.

Any thoughts on this, @juliasloan25 ?

Currently, the update function for `PrescribedRadiativeFluxes` only returns cosine zenith angle for a single fixed lat/lon, and all the experiments where it is used are on point or column domains. Is this something that will eventually be used with a domain with multiple horizontal points?

File: src/shared_utilities/drivers.jl [hunk 6]
```diff
@@ -185,11 +186,117 @@ PrescribedPrecipitation{FT}(liquid_precip) where {FT} =
...
+    latitude::L = nothing,
+) where {FT, L <: Union{Nothing, FT}}
+    if L <: Nothing
+        return let start_date = start_date, insol_params = insol_params
+            function zenith_angle(t::T, long::FT, lat::FT) where {T}
+                if T <: ITime
+                    current_datetime = date(t)
+                else
+                    current_datetime = start_date + Dates.Second(round(t))
+                end
```
juliasloan25, line 257: I think this will work
```suggestion
                current_datetime = T isa ITime ? date(t) : start_date + Dates.Second(round(t))
```

src/shared_utilities/drivers.jl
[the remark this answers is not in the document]
juliasloan25: I'm inclined to design `CoupledRadiativeFluxes` as similar to `PrescribedRadiativeFluxes` as we can, but I'm not sure I fully understand the differences between the designs you're suggesting. Maybe we could draw it out on a whiteboard?

> Currently, the update function for PrescribedRadiativeFluxes only returns cosine zenith angle for a single fixed lat/lon, and all the experiments where it is used are on point or column domains. Is this something that will eventually be used with a domain with multiple horizontal points?

It looks like `PrescribedRadiativeFluxes` is used in some 3D (box or sphere) experiments too - e.g. `experiments/Standalone/Bucket/bucket_era5.jl` or `experiments/Standalone/Bucket/global_bucket_temporalmap.jl`

src/shared_utilities/drivers.jl
```diff
@@ -1143,6 +1250,17 @@ function make_update_drivers(a::PrescribedPrecipitation{FT}) where {FT}
     return update_drivers!
 end
 
+function make_update_drivers(r::CoupledRadiativeFluxes{FT}) where {FT}
```
juliasloan25, line 1253: Could you add a docstring please?

src/shared_utilities/drivers.jl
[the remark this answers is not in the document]
kmdeck: We do use PrescribedRadiartiveFluxes in all of our global runs as well.  In that case the closure is over lat = field, lon = field, instead of over a point: https://github.com/CliMA/ClimaLand.jl/blob/38c0d26ca6116e82fab95324ef045b69bbca6989/src/shared_utilities/drivers.jl#L1350-L1355

src/shared_utilities/drivers.jl
```diff
@@ -185,11 +187,170 @@ PrescribedPrecipitation{FT}(liquid_precip) where {FT} =
...
+    insol_params::IP
+    function CoupledRadiativeFluxes{FT, F, T, LT, IP}(
+        θs::F,
+        start_date::T,
+        latitude::LT,
+        longitude::LT,
+        insol_params::IP,
+    ) where {FT, F, T, LT, IP}
+        (
+            all(isnothing, (start_date, latitude, longitude, insol_params)) ||
```
juliasloan25, line 228: Should this be `any` instead of `all`? So that it's true if start date, lat, lon, _or_ insolation params is nothing, rather than _and_

src/shared_utilities/drivers.jl
```diff
@@ -1143,6 +1304,32 @@ function make_update_drivers(a::PrescribedPrecipitation{FT}) where {FT}
...
+
+When `r.θs` is `nothing`, the cosine zenith angle
+is set to NaN. Otherwise, the cosine zenith angle is computed using
+`r.θs(t, r.start_date, r.latitude, r.longitude, r.insol_params)`.
+"""
+make_update_drivers(r::CoupledRadiativeFluxes{FT, Nothing}) where {FT} =
+    (p, t) -> p.drivers.cosθs .= FT(NaN)
+function make_update_drivers(r::CoupledRadiativeFluxes{FT}) where {FT}
+    function update_drivers!(p, t)
+        r.θs(
```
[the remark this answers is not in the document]
imreddyTeja: I did not make this change for the `PresrcibedRadiativeFluxes`, so that still allocates, but I modified the zenith angle function to reduce the amount of allocation.

src/shared_utilities/drivers.jl
```diff
@@ -185,11 +187,170 @@ PrescribedPrecipitation{FT}(liquid_precip) where {FT} =
...
+        θs::F,
+        start_date::T,
+        latitude::LT,
+        longitude::LT,
+        insol_params::IP,
+    ) where {FT, F, T, LT, IP}
+        (
+            all(isnothing, (start_date, latitude, longitude, insol_params)) ||
+            !isnothing(θs)
+        ) || error(
```
juliasloan25, line 230: And I think maybe here the `||` should be `&&`, so that this errors if the previous condition is met. I'm not sure though

src/shared_utilities/drivers.jl
```diff
@@ -185,11 +187,170 @@ PrescribedPrecipitation{FT}(liquid_precip) where {FT} =
...
+            "CoupledRadiativeFluxes: `θs` must be defined if `start_date`, `latitude`, `longitude`, or `insol_params` are defined",
+        )
+        new(θs, start_date, latitude, longitude, insol_params)
+    end
+end
+
+CoupledRadiativeFluxes{FT}() where {FT} =
```
juliasloan25, line 237: Could you comment on when this case is meant to be used/why we have it?

src/shared_utilities/drivers.jl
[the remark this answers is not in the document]
juliasloan25: This should have a bang `!` since it's modifying `p.drivers.cos\thetas` now. Maybe it should also be renamed to `r.cos\thetas` if it's computing the cosine now

juliasloan25 (commented):
This looks good! I left a couple of small comments.

I do think it's a little confusing that now we have default functions to update zenith angle and to update cos zenith angle. It's not completely clear to me where each one is used.

src/shared_utilities/drivers.jl
imreddyTeja: That adds backwards support for the coupler, but it can be removed if that is not needed.

src/shared_utilities/drivers.jl
```diff
@@ -185,11 +186,110 @@ PrescribedPrecipitation{FT}(liquid_precip) where {FT} =
...
+    """Function that fills a climacore field with the zentih angle given the following arguments:
+    (time_from_start, `start_date`)"""
+    θs::F
+    "Start date - the datetime corresponding to t=0 for the simulation"
+    start_date::T
+    function CoupledRadiativeFluxes{FT, F, T}(
+        θs::F,
+        start_date::T,
+    ) where {FT, F, T}
+        (!(isnothing(θs) ⊻ isnothing(start_date))) || error(
```
juliasloan25, line 214: I think this is a little more readable:
```suggestion
        ((isnothing(θs) && isnothing(start_date)) || ((!isnothing(θs) && !isnothing(start_date))) || error(
```

juliasloan25 (approved):

imreddyTeja:
@coderabbitai full review

PR #1137, opened 2025-05-05
Set default snow parameters to calibrated values

AlexisRenchon:
This commit sets alpha_0, Delta_alpha, k, beta,
and x0 to calibrated values in src/standalone/Snow/Snow.jl and experiments/long_runs/snowy_land.jl.

### DIFF
--- a/src/standalone/Snow/Snow.jl
+++ b/src/standalone/Snow/Snow.jl
@@ -83,12 +83,12 @@ end
     AbstractAlbedoModel{FT}
 
 Defines the model type for albedo parameterization
-for use within an `AbstractSnowModel` type. 
+for use within an `AbstractSnowModel` type.
 
-These parameterizations are stored in parameters.α_snow, and 
+These parameterizations are stored in parameters.α_snow, and
 are used to update the value of p.snow.α_snow (the broadband
 albedo of the snow at a point).
-stored 
+stored
 """
 abstract type AbstractAlbedoModel{FT <: AbstractFloat} end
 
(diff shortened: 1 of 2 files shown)


src/standalone/Snow/Snow.jl
```diff
@@ -136,8 +136,8 @@ function ZenithAngleAlbedoModel(
     α_0::FT,
     Δα::FT,
     k::FT;
-    β = FT(0),
-    x0 = FT(0.2),
+    β = FT(0.7875), # Calibrated value
```
kmdeck, line 139: I think we can change the constructor to supply defaults for all parameters now, and then use this in the long run.

src/standalone/Snow/Snow.jl
```diff
@@ -293,7 +293,11 @@ function SnowParameters{FT}(
     density::DM = MinimumDensityModel(FT(200)),
     z_0m = FT(0.0024),
     z_0b = FT(0.00024),
-    α_snow::AM = ConstantAlbedoModel(FT(0.8)),
```
kmdeck, line 296: Some of our unit tests fail probably because they use this default.

Others fail probably because they are testing the default of the ZenithAngleAlbedo model and now that has changed.

can test locally with "julia --project=test test/standalone/Snow/test_file_name.jl"  for each of the files in that directory.

kmdeck:
albedo_parameterization
And here is the dependence on rho_snow with mu = 0.5:
albedo_parameterization_mu_fixed

kmdeck:
Note that when we had the simpler model (only a dependence on zenith angle), we got:
albedo_parameterization_simpler


so it looks like most of variation in albedo is now being covered by the snow density portion of the parameterization, rather than the zenith angle portion.

This is probably fine, but we may want to confirm at some point that the fit with all 5 parameters is improved enough, compared with that with only three parameters, so that the extra two parameters are justified. I mention this also because the zenith angle is completely known, so a parameterization based on that should be pretty stable, while snow density computed by the model could change as other things change in the model. @AlexisRenchon

PR #1138, opened 2025-05-05
Make is easier to run a global calibration

AlexisRenchon:
In this commit, a few lines at the top of calibrate_land.jl code are added and can be modified by users to design and run a calibration job.
Previously to this commit, those lines where buried deeper in the code. This makes it very easy and flexible to setup a calibration, and sets up the direction to add further options (like, for example, duration of calibration, noise configuration...)


experiments/calibration/calibrate_land.jl
```diff
@@ -1,5 +1,19 @@
+using Dates # needs to be called again for workers
+
+model = "bucket" # or "bucket"
+const variable_list = ["lhf", "shf"] # variables you want to capture by adjusting your priors
+const n_iterations = 10 # 1 iterations takes ~ 1.5 hour with current settings (resolution, 2 year simulation)
+const spinup_period = Year(1)
+const nelements = (50, 15) # resolution - (horizontal elements (lon,lat), vertical elements (soil discretization))
+const start_date = DateTime(2008, 12, 01) # this is the start of the forward model spinup
+const dirname = "bucket_lhf_shf" # ideally, describe your calibration in a few words
+const caldir = string("/glade/campaign/univ/ucit0011/alexis/", dirname) # you might want to save somewhere else than login
```
Sbozzolo, line 10:
```suggestion
const caldir = joinpath("/glade/campaign/univ/ucit0011/alexis/", dirname) # you might want to save somewhere else than login
```

experiments/calibration/calibrate_land.jl
```diff
@@ -1,5 +1,19 @@
+using Dates # needs to be called again for workers
+
+model = "bucket" # or "bucket"
+const variable_list = ["lhf", "shf"] # variables you want to capture by adjusting your priors
+const n_iterations = 10 # 1 iterations takes ~ 1.5 hour with current settings (resolution, 2 year simulation)
+const spinup_period = Year(1)
+const nelements = (50, 15) # resolution - (horizontal elements (lon,lat), vertical elements (soil discretization))
```
Sbozzolo, line 7: I thought we moved to 100 elements and that the speed was faster now

experiments/calibration/make_training_locations.jl
```diff
@@ -5,30 +5,44 @@ using ClimaComms
 using ClimaCore
 ClimaComms.@import_required_backends
 
+
 """
-    make_training_locations(nelements)
+    diagnostics_lat_lon(nelements)
```
AlexisRenchon, line 10: Note: I re-added this function because we need it in observationseries_era5.jl 

We could return lats and longs from make_training_locations, but that would be confusing I think, because training_locations contains coordinates after the mask is applied, and lats longs before

experiments/calibration/calibrate_land.jl
```diff
@@ -1,5 +1,19 @@
+using Dates # needs to be called again for workers
+
+model = "bucket" # or "bucket"
+const variable_list = ["lhf", "shf"] # variables you want to capture by adjusting your priors
+const n_iterations = 10 # 1 iterations takes ~ 1.5 hour with current settings (resolution, 2 year simulation)
+const spinup_period = Year(1)
```
kmdeck, line 6: does this mean that we could we make this smaller now, and e.g. run for less time per iteration?

kmdeck: e.g. 1.5 years if const spinup_period = 6 month

experiments/calibration/calibrate_land.jl
```diff
@@ -17,37 +31,46 @@ rng_seed = 2
...
 
-# Calibrate the land model with:
-const variable_list = ["swu"] # variables you want to capture by adjusting your priors
-const n_iterations = 10 # 1 iterations takes ~ 1.5 hour with current settings (resolution, 2 year simulation)
-const spinup_period = Year(1)
-# potentially we could add time_for_calibration (currently 1 year)
-
 # Using the following priors:
-include(joinpath(@__DIR__, "priors.jl"))
+include(joinpath(@__DIR__, string("priors_", model, ".jl")))
```
kmdeck, line 35: should this be moved up?

kmdeck: is this how you control which parameters are being calibrated?

experiments/calibration/calibrate_land.jl
```diff
@@ -1,5 +1,19 @@
+using Dates # needs to be called again for workers
+
+model = "bucket" # or "bucket"
```
kmdeck, line 3: maybe model_type = "bucket"?

experiments/calibration/calibrate_land.jl
```diff
@@ -1,5 +1,19 @@
...
+const variable_list = ["lhf", "shf"] # variables you want to capture by adjusting your priors
+const n_iterations = 10 # 1 iterations takes ~ 1.5 hour with current settings (resolution, 2 year simulation)
+const spinup_period = Year(1)
+const nelements = (101, 15) # resolution - (horizontal elements (lon,lat), vertical elements (soil discretization))
+const start_date = DateTime(2008, 12, 01) # this is the start of the forward model spinup
```
kmdeck, line 8: any reason why Dec and not Jan?

experiments/calibration/calibrate_land.jl
AlexisRenchon: yes!

experiments/calibration/calibrate_land.jl
AlexisRenchon: yes, the priors file (priors_land.jl of priors_bucket.jl) contains the parameters to calibrate and their distribution. Why do you want to move it up? We can discuss

experiments/calibration/calibrate_land.jl
AlexisRenchon: ok!

experiments/calibration/calibrate_land.jl
AlexisRenchon: Because the "seasons" we use (DJF, MAM, JJA, SON) start in December and not January

experiments/calibration/calibrate_land.jl
kmdeck: mainly because this is something a user will change in between runs, and there was a comment in the code about how things the user changes are above this section

kmdeck (approved):
looks good! I would like to understand a bit more how the priors file works - it seems somewhat dependent on the forward model code for now (since not all the params have defaults). It would be great if eventually we could set priors or values for any subset of the parameters in priors.jl, and then forward model can then interpret that and use the default values for the ones not in the priors.jl file

AlexisRenchon:
> looks good! I would like to understand a bit more how the priors file works - it seems somewhat dependent on the forward model code for now (since not all the params have defaults). It would be great if eventually we could set priors or values for any subset of the parameters in priors.jl, and then forward model can then interpret that and use the default values for the ones not in the priors.jl file

I agree!! 
I think this is part of making a higher level API to run the global model?

PR #1140, opened 2025-05-05
Global simulations struct

kmdeck:
## Purpose 
See Issue #

## To-do
ID small changes to other pieces of src to enable this
Make those changes in separate PRs
then rebase and merge this

## Content

----

### DIFF
--- a/src/simulations/Simulations.jl
+++ b/src/simulations/Simulations.jl
@@ -1,0 +1,165 @@
+module Simulations
+using ClimaTimesteppers
+using Dates
+using ClimaLand
+
+struct LandSimulation
+    context
+    params
+    model
+    domain
+    timestepper
+    user_callbacks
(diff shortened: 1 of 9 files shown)


src/simulations/Simulations.jl
```diff
@@ -0,0 +1,183 @@
...
+                         canopy_component_args = canopy_component_args,
+                         canopy_model_args = canopy_model_args,
+                         snow_args = snow_args,
+                         snow_model_type = snow_model.type,
+                         land_args = land_input,
+                         )
+    
+    # set initial conditions
+    Y, p, cds = initialize(land)
+    set_ic! = ClimaLand.set_ic_from_file(ic_file)
```
kmdeck, line 120: could also support `set_ic! = function passed in` if the arguments are fixed

PR #1141, opened 2025-05-06
make bc and sources default for soilco2

kmdeck:
## Purpose 
In our land models (soilco2 + canopy +soil or soilco2 + canopy+soil+snow), we always run with atmospheric forcing, and we set up the boundary conditions/fluxes for the components in the LandModel constructor. This PR makes it so that we do the same for soilco2.

This removes some flexibility, but it does not make sense to change the BC for soil co2 from what we already set it to be in every example. Moving it to the constructor makes sense.

Furthermore, we always specify the same source for soilco2, and it doesnt make sense to run without this source, so this PR also moves that to the constructor.

This is useful for simplifying our LandModel setup. Now every component model takes in as external args the domain and their own parameters.
## Content

----


AlexisRenchon (approved):
Looks good! Good idea. Thank you!

PR #1150, opened 2025-05-08
Simpifies the path for certain artifacts; use lowres data from ERA5 if hires not available

kmdeck:
## Purpose 
- Make it so that by default a user can run simulations for 2008 forced by ERA5 (even if hires is not available)
- simplify the functions we use to return paths to forcing data so that the file path can be obtained with a single function, instead of two functions, one which returns the artifact directory, and the other the path.

## Discussion
If a user requests hires ERA5 data from 2008, and it is not downloaded, we now default to using the lowres data (downloads if necesssary). This was facilitated by merging two functions (one which returns the directory, and one which returns the file name) to get the file path into a single function. Note that the lowres and hires data have different directories and filenames.

so, this PR now allows the user to obtain the 2008 data of ERA5 with a single line of code instead of calling two functions and joining paths.

to be consistent, this PR allows the user to obtain a single year of modis data with one line instead of calling two functions, and joining paths

Things that we may wish to revisit or rethink:
1.  it might be nice to get rid of the "2008" specific function for era5 data, and instead follow what I did for modis data and make a function to return a single year (any year), but we only have lowres data for 2008, and it is important to keep that as an option so that people can run locally. so I left this as is.
2. I want to make use of the "find_path_years" functions that @ph-kev wrote, but this would incur a slowdown (since it returns multiple files even if only one year is simulated). One thought is to always use the periodic BC, so that if tf and t0 are the same year, only one file is returned. 

## Content

----


kmdeck:
> I think the `land.jl` file was deleted accidentally? If this wasn't the case, the buildkite pipelines should be updated.

I believe we removed the test from buildkite (as unnecessary) some time ago, but we did not delete the file.

ph-kev (approved):

PR #1151, opened 2025-05-09
ic function cleanup

kmdeck:
## Purpose 
Precursor to adding the simulations struct, which will take an argument set_ic! - a function which sets the initial conditions with arguments (Y, p, t0, model).

## Content

----

### DIFF
--- a/experiments/long_runs/land_region.jl
+++ b/experiments/long_runs/land_region.jl
@@ -306,38 +306,16 @@ function setup_prob(
 
     Y, p, cds = initialize(land)
 
-    init_soil(ν, θ_r) = θ_r + (ν - θ_r) / 2
-
-    Y.snow.S .= 0.0
-    Y.snow.S_l .= 0.0
-    Y.snow.U .= 0.0
-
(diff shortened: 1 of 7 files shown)


experiments/long_runs/land_region.jl
```diff
@@ -306,39 +306,17 @@ function setup_prob(
...
-    evaluate!(Y.canopy.energy.T, atmos.T, t0)
-
+    ic_path = ClimaLand.Artifacts.soil_ic_2008_50m_path(; context = context)
+    set_initial_state! = make_set_initial_state_from_file(ic_path, land)
     set_initial_cache! = make_set_initial_cache(land)
     exp_tendency! = make_exp_tendency(land)
     imp_tendency! = ClimaLand.make_imp_tendency(land)
     jacobian! = ClimaLand.make_jacobian(land)
+
+    set_initial_state!(Y, p, t0, land)
```
kmdeck, line 316: I belive this should work for a regional simulation (reading IC from a file with lat/lon/z), because reading in the ERA5 data works for a regional simulation

Sbozzolo (approved):

PR #1156, opened 2025-05-14
Correct reflected and transmitted fractions

AlexisRenchon:
This fixes a bug in reflected radiation, which led to instances when
abs+refl > 1 (and hence ground absorbed radiation was negative).

It also avoids a division by 1-alpha_soil, which would lead to 1/0 when alpha_soil is zero, and clips alpha_soil > 0.

It also ensures reflected + absorbed + transmitted radiation is equal to 1.

### DIFF
--- a/src/standalone/Vegetation/canopy_parameterizations.jl
+++ b/src/standalone/Vegetation/canopy_parameterizations.jl
@@ -236,15 +236,17 @@ function canopy_sw_rt_two_stream(
     α_soil::FT,
     frac_diff::FT,
 ) where {FT}
-
+    α_soil = max(eps(FT), α_soil) # this prevents division by zero, below.
+    cosθs = max(eps(FT), cosθs) # The insolations package returns θs > π/2 (nighttime), but this assumes cosθs >0
     # Compute μ̄, the average inverse diffuse optical length per LAI
-    μ̄ = 1 / (2G)
+    μ̄ = 1 / max(2G, eps(FT))
 
-    # Clip this to eps(FT) to prevent dividing by zero
-    ω = max(α_leaf + τ_leaf, eps(FT))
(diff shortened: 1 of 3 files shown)


src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -263,7 +263,7 @@ function canopy_sw_rt_two_stream(
     h = √(b^2 - c^2) / μ̄
     σ = (μ̄ * K)^2 + c^2 - b^2
 
-    u₁ = b - c / α_soil
+    u₁ = b - c / max(α_soil, eps(FT))
```
kmdeck, line 266: oh, we dont need this unless alpha_soil = 0! so it should be OK in this limit.

kmdeck: We can just add at the very top of the function:
\alpha_soil = max(eps(FT), alpha_soil)

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -377,7 +377,7 @@ function canopy_sw_rt_two_stream(
 
     # Convert fractional absorption into absorption and return
     # Ensure floating point precision is correct (it may be different for PAR)
-    F_trans = (1 - F_abs - F_refl) / (1 - α_soil)
+    F_trans = (1 - F_abs - F_refl)
```
kmdeck, line 380: This formula is correct as is when alpha_soil < 1.
When alpha_soil = 1, the formula is not correct, I think we should have F_trans = F_abs + F_refl -1

Let's see if we can rework this so the output is correct in either case

braghiere (approved):

PR #1160, opened 2025-05-20
Fix and test Comms methods

kmdeck:
## Purpose 
Fixes a bug with our methods for ClimaComms.device() and context(); adds some tests for functions that are called by these methods.


kmdeck:
long runs: https://buildkite.com/clima/climaland-long-runs/builds/4008

Sbozzolo (approved):
Thank you!

PR #1161, opened 2025-05-23
Fix bottom boundary condition

kmdeck:
## Purpose 
To debug NaNs in the soil model, I restricted the domain depth to 10 m, since that led to more NaNs earlier. I saw in this long run: https://buildkite.com/clima/climaland-long-runs/builds/4009 that the NaNs were appearing in 1 year over South America and Indonesia. I then debugged the ones over South America. 12 NaNs appeared by the end of the simulation.
image

The NaNs were appearing first in the bottom of the domain. Very weird behavior was showing up: water was leaving the domain (as expected from the BC of free drainage), but the temperature in the bottom was increasing (to unphysical levels). This is tied to inconsistent boundary conditions: if water leaves the domain but energy does not, the temperature has to increase to compensate.

This PR adds that change in. I checked here: https://buildkite.com/clima/climaland-long-runs/builds/4011 that the NaNs are gone.
image

I then reverted the domain depth back to 50m. In the next land 10 year long runs, this hopefully will improve stability, since we also saw NaNs in similar locations. At that point, we can decide to make the domain shallower and continue debugging NaNs as makes sense.

Note that a future PR will modify the top bc to account for this type of error also: precipitation is a flux of water into the domain, and this should change the energy flux also.
https://github.com/CliMA/ClimaLand.jl/issues/1163

### DIFF
--- a/src/standalone/Soil/boundary_conditions.jl
+++ b/src/standalone/Soil/boundary_conditions.jl
@@ -10,6 +10,7 @@ using ClimaCore: Geometry
 export TemperatureStateBC,
     MoistureStateBC,
     FreeDrainage,
+    EnergyWaterFreeDrainage,
     HeatFluxBC,
     WaterFluxBC,
     AtmosDrivenFluxBC,
@@ -48,9 +49,14 @@ end
 
 """
     FreeDrainage <: AbstractWaterBC
+
 A concrete type of soil boundary condition, for use at
 the BottomBoundary only, where the flux is set to be
-`F = -K∇h = -K`.
+`F = -K∇h = -K`. 
+
+This is not tied to any boundary condition for the heat equation. 
+To account for the energy flux resulting from free drainage of liquid
+water, please see `EnergyWaterFreeDrainage`.
 """
 struct FreeDrainage <: AbstractWaterBC end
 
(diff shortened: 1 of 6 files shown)


src/standalone/Soil/boundary_conditions.jl
```diff
@@ -569,6 +575,39 @@ function WaterHeatBC(; water, heat)
...
+
+A concrete type of soil boundary condition, for use at
+the BottomBoundary only, where the fluxes are set to be
+`F_liq = -K∇h = -K`, `F_energy = -K ρe_liq`.
+
+That is, this enforces that the free drainage boundary condition
+for liquid water is paired the the corresponding loss of energy
+that that entails.
+"""
+struct EnergyFreeDrainage <: AbstractEnergyHydrologyBC end
```
juliasloan25, line 589: I wonder if `EnergyWaterFreeDrainage` might be a more clear name, to emphasize that this is the BC for both energy and water (not just energy). We could also consider renaming `FreeDrainage` to `WaterFreeDrainage` or having `FreeDrainageWater` and `FreeDrainageWaterEnergy`. I'm not sure what will be best

test/standalone/Soil/soil_bc.jl
```diff
@@ -192,6 +192,83 @@ for FT in (Float32, Float64)
 
         @test abs(flux_expected - flux_int) < eps(FT)
     end
+
+    @testset "Test energy free drainage BC, FT = $FT" begin
```
juliasloan25, line 196:
```suggestion
    @testset "Test water + energy free drainage BC, FT = $FT" begin
```

juliasloan25 (approved):
Amazing! Thank you for debugging this. Just a couple comments about naming, but it's also fine to merge as-is

src/standalone/Soil/boundary_conditions.jl
kmdeck: Im also going to open an issue to unify some naming - we have use "WaterHeat" elsewhere, and I think we should change heat to energy. I also wonder about EnergyHydrology as a named and changing that (you've mentioned that before too - it has no "Model" in the title)

PR #1164, opened 2025-05-27
Add energy from precipitation

kmdeck:
## Purpose 
This adds an energy flux at the top of the soil domain associated with precipitation and snowmelt. 

One important change to highlight is that prior, we computed the runoff and infiltration using `influx = precip + evap`, and the boundary condition was `bc = infiltration`. However, the energy of the vapor flux is already accounted for in the LHF. This complicates finding the energy of the infiltration, since it could in principle also be due to condensation and therefore already accounted for in the LHF. so, this PR considers a liquid influx only when computing infiltration and runoff, so that the `bc = infiltration + vapor flux`, and the energy bc includes `energy of infiltration + LHF`.

For context, we have four different models that this change is affecting:
soil alone (`src/standalone/Soil/boundary_conditions.jl`)
soil + snow (`src/integrated/soil_snow_model.jl`)
soil+ canopy (`src/integrated/soil_canopy_model.jl`)
soil+canopy+snow. (`src/integrated/land.jl`)
## To-do
[X] Rebase after #1176  is merged

A future PR will
- update the liquid infiltration computation to use @lazy and remove corresponding variables from the cache

We can also consider removing duplicate code in `soil_boundary_fluxes!`, since the computations with snow (snow+soil, or snow_soil+canopy) are identical, and since the computations without snow (soil+canopy, soil alone) are nearly identical. On the other hand, these may differ in the future if the presence of a canopy changes the boundary flux computation (eg. interception of rain or snow in the canopy)

## Content

----

### DIFF
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -480,27 +480,45 @@ the presence of the canopy modifies the soil BC.
 function soil_boundary_fluxes!(
     bc::AtmosDrivenFluxBC,
     prognostic_land_components::Val{(:canopy, :snow, :soil, :soilco2)},
-    soil::EnergyHydrology{FT},
+    soil::EnergyHydrology,
     Y,
     p,
     t,
-) where {FT}
+)
     turbulent_fluxes!(p.soil.turbulent_fluxes, bc.atmos, soil, Y, p, t)
-    # influx = maximum possible rate of infiltration given precip, snowmelt, evaporation/condensation
-    # but if this exceeds infiltration capacity of the soil, runoff will
-    # be generated.
-    # Use top_bc.water as temporary storage to avoid allocation
-    influx = p.soil.top_bc.water
-    @. influx =
-        p.snow.water_runoff * p.snow.snow_cover_fraction +
+    # Liquid influx is a combination of precipitation and snowmelt in general
+    liquid_influx =
+        Soil.compute_liquid_influx(p, soil, prognostic_land_components)
+    # This partitions the influx into runoff and infiltration
+    Soil.update_infiltration_water_flux!(
+        p,
+        bc.runoff,
+        liquid_influx,
+        Y,
+        t,
+        soil,
+    )
+    # This computes the energy of the infiltrating water
+    infiltration_energy_flux = Soil.compute_infiltration_energy_flux(
+        p,
+        bc.runoff,
+        bc.atmos,
+        prognostic_land_components,
+        liquid_influx,
+        soil,
+        Y,
+        t,
+    )
+    @. p.soil.top_bc.water =
+        p.soil.infiltration +
         p.excess_water_flux +
         (1 - p.snow.snow_cover_fraction) *
-        (p.drivers.P_liq + p.soil.turbulent_fluxes.vapor_flux_liq)
-    # The update_runoff! function computes how much actually infiltrates
-    # given influx and our runoff model bc.runoff, and updates
-    # p.soil.infiltration in place
-    Soil.Runoff.update_runoff!(p, bc.runoff, influx, Y, t, soil)
-    @. p.soil.top_bc.water = p.soil.infiltration
+        p.soil.turbulent_fluxes.vapor_flux_liq
+    # The actual boundary condition is a mix of liquid water infiltration and
+    # evaporation. The infiltration already has accounted for snow cover fraction,
+    # because the influx it is computed from has accounted for that.
+    # The last term, `excess water flux`, arises when snow melts in a timestep but
+    # has a nonzero sublimation which was applied for the entire step.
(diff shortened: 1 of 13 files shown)


src/integrated/land.jl
```diff
@@ -514,38 +514,132 @@ the presence of the canopy modifies the soil BC.
...
+"""
+   compute_liquid_influx(p,
+                         model,
+                         prognostic_land_components::Val{(:canopy, :snow, :soil, :soilco2)},
+    ) 
+
+Returns the liquid water volume flux at the surface of the soil; this
+will then be partitioned into surface runoff and infiltration:
+influx = infiltration - runoff (runoff > 0; infiltration and influx < 0,
+and infiltration <= influx).
```
tapios, line 578:
```suggestion
and abs(infiltration) <= abs(influx)).
```
(I this what you mean?)

src/integrated/land.jl
```diff
@@ -514,38 +514,132 @@ the presence of the canopy modifies the soil BC.
...
+If the source of the infiltration is purely liquid precipitation,
+we approximate the volumetric internal energy with the volumetric
+internal energy of liquid water at the air temperature at the surface.
+
+Snowmelt is accompanied by an energy loss for the snow, and we use that
+energy flux here (`energy_runoff`). Internally to snow, that is computed using the
+volumetric internal energy of liquid water at the snow temperature.
+
+For a mix of both, we compute the fraction due to each like:
+Infiltration * (P_liq*(1-σ)/Influx * ρe(T_air)+snowmelt_water_flux *σ/Influx * ρe(T_snow))
```
tapios, line 616: Could you please define `σ` here?

src/integrated/soil_canopy_model.jl
```diff
@@ -368,29 +368,110 @@ the presence of the canopy modifies the soil BC.
...
+"""
+   compute_liquid_influx(p,
+                         model,
+                         prognostic_land_components::Val{(:canopy, :soil, :soilco2)},
+    ) 
+
+Returns the liquid water volume flux at the surface of the soil; this
+will then be partitioned into surface runoff and infiltration:
+influx = infiltration - runoff (runoff > 0; infiltration and influx < 0,
+and infiltration <= influx).
```
tapios, line 462: as above?
```suggestion
abs(infiltration) <= abs(influx))```

File: src/integrated/soil_snow_model.jl [hunk 4]
```diff
@@ -346,41 +346,132 @@ conditions, taking into account the presence of snow on the surface.
...
+
+Returns the liquid water volume flux at the surface of the soil; this
+will then be partitioned into surface runoff and infiltration:
+influx = infiltration - runoff (runoff > 0; infiltration and influx < 0,
+and infiltration <= influx).
+
+In a model with snow as a prognostic variable, the influx is a combination
+of the liquid precipitation as a volume flux and the snowmelt leaving the 
+snowpack.
+"""
```
tapios, line 416: Here and elsewhere: I think it would be better not to duplicate the docstrings. A structure that collects the different methods and then uses one docstring for all of them would be cleaner, I think.

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -909,23 +912,99 @@ function soil_boundary_fluxes!(
...
+If the source of the infiltration is purely liquid precipitation,
+we approximate the volumetric internal energy with the volumetric
+internal energy of liquid water at the air temperature at the surface.
+
+Future runoff parameterizations may require allowing for exfiltration of 
+liquid water from the soil, which would have a different energy. To accomodate
+this, the runoff is passed as an argument and this can be used for dispatch in the 
+future.
+
+Furthermore, in couple simulations, the energy flux may be provided directly,
```
tapios, line 988:
```suggestion
Furthermore, in coupled simulations, the energy flux may be provided directly,
```

src/standalone/Soil/boundary_conditions.jl
```diff
@@ -909,23 +912,99 @@ function soil_boundary_fluxes!(
...
+    runoff,
+    atmos::PrescribedAtmosphere,
+    prognostic_land_components::Val{(:soil,)},
+    liquid_influx,
+    model::EnergyHydrology,
+    Y,
+    t,
+)
+    earth_param_set = model.parameters.earth_param_set
+    return @lazy @. p.soil.infiltration * Soil.volumetric_internal_energy_liq(
```
tapios, line 1003: I'm not sure about the `@lazy @.` syntax. Maybe check with @dennisYatunin. Wouldn't `@.lazy()` do? And maybe it's better to leave the `lazy` on the outermost layer (i.e., not inside the function) in any case.

PR #1165, opened 2025-05-27
Fixes bug with spatial parameters, adds soil "longer" run

kmdeck:
## Purpose 
1. Instead of setting nu = 1 in regions where we dont have data, replace with more physical value near mean of porosity, same with Ksat.
2. Fix bug with texture norm :/
3. add soil longer run

## Content

----


.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -94,7 +94,18 @@ steps:
           - "snowy_land_longrun_gpu/*pdf"
         agents:
           slurm_gpus: 1
-          slurm_time: 24:00:00
+          slurm_time: 15:00:00
         env:
           CLIMACOMMS_DEVICE: "CUDA"
           LONGER_RUN: ""
+
+      - label: "Soil"
```
juliasloan25, line 102:
```suggestion
      - label: "Very long run of soil"
```
To distinguish from the other soil longrun

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -94,7 +94,18 @@ steps:
...
+        command:
```
juliasloan25, line 103: maybe we also need this line, which we have for the other "longer run"
```
if: build.env("LONGER_RUN") != null
```

src/simulations/spatial_parameters.jl
```diff
@@ -326,10 +326,10 @@ function default_spatially_varying_soil_parameters(
         regridder_kwargs = (; extrapolation_bc,),
     )
 
-    μ = FT(-6)
+    μ = FT(-5.11)
     K_sat .= masked_to_value.(K_sat, soil_params_mask, 10.0^μ)
 
-    ν .= masked_to_value.(ν, soil_params_mask, 1)
+    ν .= masked_to_value.(ν, soil_params_mask, 0.45)
```
juliasloan25, line 332: Can you add some short comments explaining where these numbers come from? (i.e. mean over the data)

juliasloan25 (approved):
Looks good! Just some small comments

.buildkite/longruns_gpu/pipeline.yml
kmdeck: I renamed the group and added a label so that this mirrors the structure of our "long" runs

PR #1176, opened 2025-06-12
add energy from rain to snow

kmdeck:
## Purpose 
This adds the energy of falling rain to the snow energy fluxes.

It also establishes a function which returns this energy flux for rain and snow. This function can be extended for coupled simulations to pass the energy flux as computed by the atmosphere. However, since we do not have that method yet, I did not specialize to the `atmos::PrescribedAtmosphere` type. 

## Content

----


src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -477,6 +479,42 @@ function energy_from_T_and_swe(S::FT, T::FT, parameters) where {FT}
 
 end
 
+"""
+    volumetric_energy_flux_falling_snow(atmos, p, parameters)
+
+Returns the volumetric energy fluxof falling snow for a PrescribedAtmosphere,
```
tapios, line 485:
```suggestion
Returns the volumetric energy flux of falling snow for a PrescribedAtmosphere,
```

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -477,6 +479,42 @@ function energy_from_T_and_swe(S::FT, T::FT, parameters) where {FT}
...
+approximated as -LH_f0 * _ρ_liq *P_snow. This is per unit volume of liquid water,
```
tapios, line 486:
```suggestion
approximated as -LH_f0 * _ρ_liq * P_snow. This is per unit volume of liquid water, 
```

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -477,6 +479,42 @@ function energy_from_T_and_swe(S::FT, T::FT, parameters) where {FT}
...
+We assume that the sensible heat portion of snow is negligible.
+
+This method can be extended to coupled simulations, where atmos is of type
+CoupledAtmosphere, and the energy flux of the falling snow is passed in the
+cache `p`. In that case, this should specify `atmos::PrescribedAtmosphere`.
+"""
+function volumetric_energy_flux_falling_snow(atmos, p, parameters)
+    _LH_f0 = LP.LH_f0(parameters.earth_param_set)
+    _ρ_liq = LP.ρ_cloud_liq(parameters.earth_param_set)
+    return @lazy @. -_LH_f0 * _ρ_liq * p.drivers.P_snow # per unit vol of liquid water
```
tapios, line 498: I am (perpetually) confused about sign conventions. I trust this is correct, but some explanation in the docstring may be helpful. If fluxes are always defined as positive upward, then P_sno < 0 and the energy flux would be in the same direction as the precipitation flux. The sign here, I suppose, is negative because the latent heat of fusion adds a negative component to the total internal energy. Is that right?

tapios (approved):
Looks good to me!

src/standalone/Snow/snow_parameterizations.jl
kmdeck: yes, exactly! I can define `\rho e  = -_LH_f0 * _ρ_liq` first, and then use that in the return statement. I'll update the docstring to clarify.

src/standalone/Snow/snow_parameterizations.jl
kmdeck: I fixed this.

src/standalone/Snow/snow_parameterizations.jl
kmdeck: I fixed this.

PR #1177, opened 2025-06-13
remove two cache variables

kmdeck:
## Purpose 
In our simulations, we have the cache variables "atmos_energy_flux" and "atmos_water_flux". While useful for tracking conservation (see the experiment script and plot in buildkite that this PR changes), we dont need to store them in the cache. we can compute this after the fact from our other variables in `p`.

## Content

----

### DIFF
--- a/experiments/integrated/fluxnet/snow_soil/simulation.jl
+++ b/experiments/integrated/fluxnet/snow_soil/simulation.jl
@@ -253,11 +253,41 @@ _ρ_i = FT(LP.ρ_cloud_ice(earth_param_set))
 _ρ_l = FT(LP.ρ_cloud_liq(earth_param_set))
 fig = Figure(size = (1600, 1200), fontsize = 26)
 ax1 = Axis(fig[2, 1], ylabel = "ΔEnergy (J/A)", xlabel = "Days")
+function compute_surface_energy_fluxes(p)
+    ρe_flux_falling_snow =
+        Snow.volumetric_energy_flux_falling_snow(atmos, p, land.snow.parameters)
+    ρe_flux_falling_rain =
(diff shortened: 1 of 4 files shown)


experiments/integrated/fluxnet/snow_soil/simulation.jl
```diff
@@ -253,11 +253,41 @@ _ρ_i = FT(LP.ρ_cloud_ice(earth_param_set))
...
+                  p.soil.turbulent_fluxes.shf +
+                  p.soil.R_n
+              ) +
+              p.snow.snow_cover_fraction * (
+                  p.snow.turbulent_fluxes.lhf +
+                  p.snow.turbulent_fluxes.shf +
+                  p.snow.R_n +
+                  ρe_flux_falling_rain
+              ) +
+              ρe_flux_falling_snow
```
imreddyTeja, line 273: should these be swapped?

kmdeck: Both P_liq and P_snow are fluxes in units per ground area. In a given pixel, part of the area is covered by snow, and we can define the snow cover fraction = scf = area of snow/ area of ground. Both soil and snow prognostic variables are per ground area so their boundary conditions/fluxes should be per ground area.

The liquid water fraction falls fractionally on both snow and soil (weighted by their areas of scf and 1-scf). But, by definition, a flux of snow per ground area increases the amount of snow per ground area (the prognostic var), so we dont need an area fraction here.

The same holds for energy fluxes associated with precip: F = energy of snowfall + scf*energy of rain for snow. Eventually we need to add in (1-scf)*energy of rain for the soil, but we had previously neglected it and are adding it in in a following PR.

imreddyTeja (approved):

PR #1179, opened 2025-06-13
Remove 3 vars from canopy cache

kmdeck:
## Purpose 
Remove medlyn term, beta(moisture stress), and stomatal conductance at the leaf level from the canopy cache.

Use lazy broadcasting to compute in the tendencies, compute on the fly in diagnostics when needed

## Content

----

### DIFF
--- a/src/diagnostics/land_compute_methods.jl
+++ b/src/diagnostics/land_compute_methods.jl
@@ -82,7 +82,50 @@ end
 @diagnostic_compute "autotrophic_respiration" Union{SoilCanopyModel, LandModel} p.canopy.autotrophic_respiration.Ra
 
 # Canopy - Conductance
-@diagnostic_compute "stomatal_conductance" Union{SoilCanopyModel, LandModel} p.canopy.conductance.gs
+function compute_stomatal_conductance!(
+    out,
+    Y,
+    p,
+    t,
+    land_model::Union{SoilCanopyModel, LandModel},
+)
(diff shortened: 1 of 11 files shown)


src/diagnostics/land_compute_methods.jl
```diff
@@ -82,7 +82,50 @@ end
...
+    t,
+    land_model::Union{SoilCanopyModel, LandModel},
+)
+    canopy = land_model.canopy
+    (; g1, g0, Drel) = canopy.conductance.parameters
+    earth_param_set = canopy.parameters.earth_param_set
+    thermo_params = LP.thermodynamic_parameters(earth_param_set)
+
+    if isnothing(out)
+        return medlyn_conductance.(
```
kmdeck, line 98: @ph-kev do you know which branch of these we use? If this one, are we allocating all of our diagnostic fields each step?

ph-kev: I think the [documentation](https://clima.github.io/ClimaDiagnostics.jl/dev/user_guide/) for ClimaDiagnostics does a better job of explaining this than I do.

The short answer is that both branches (`if isnothing(out)` and `else`) are called, but allocation of the field should only happen at the first time the quantity is saved (when `out` is `nothing`). Every subsequent time after that should be reusing `out` to reduce allocations.

ph-kev (approved):

PR #1180, opened 2025-06-16
change start and end dates for "longer runs"

kmdeck:
## Purpose 
Now that 10 year runs are stable, we'd like to extend our longer runs and see how things go. 
@ph-kev any reason not to use the years 2000-2019? (observations or forcing data not available? It looks like MODIS is 2000-2020, but I believe we try and grab an extra year of data so I truncated the sim at 2019 instead. Seems like we should be fine on the ERA5 front)

## Content

----

### DIFF
--- a/experiments/long_runs/snowy_land.jl
+++ b/experiments/long_runs/snowy_land.jl
@@ -335,12 +335,12 @@ end
 
 function setup_simulation(; greet = false)
     # If not LONGER_RUN, run for 2 years; note that the forcing from 2008 is repeated.
-    # If LONGER run, run for 10 years, with the correct forcing each year.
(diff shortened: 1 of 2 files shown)


ph-kev (approved):
I don't see any reason why it wouldn't work with 19 years.

ph-kev:
> > It looks like MODIS is 2000-2020, but I believe we try and grab an extra year of data so I truncated the sim at 2019 instead.
> 
> Why do we try to get an extra year of data? Is it just an off-by-one thing at the end of the last year?

This was the PR that added that change: [https://github.com/CliMA/ClimaLand.jl/pull/1117](https://github.com/CliMA/ClimaLand.jl/pull/1117).

PR #1182, opened 2025-06-17
Creating function to be able to run run_fluxnet for any fluxnet site with generic parameters

ThanhNguyen428:
## Purpose 
Added function to allow run_fluxnet.jl to be extrapolated for a fixed set of params and any fluxnet site (stored in any_run_fluxnet.jl). Was tested with US-Ha1 to confirm that it is working.

## To-do

- [ ] Make sure "iterating over empty list" bug is fixed (possibly fixed once rebased to origin)
- [ ] Rebase onto main

## Content

- Implemented function to get data files using site ID
- Confirmed met_drivers_FLUXNET.jl worked with generic fluxnet sites

----

### DIFF
--- a/experiments/integrated/fluxnet/any_run_fluxnet.jl
+++ b/experiments/integrated/fluxnet/any_run_fluxnet.jl
@@ -9,6 +9,7 @@ using Dates
 using Insolation
 using StatsBase
 
+using DelimitedFiles
 using ClimaLand
 using ClimaLand.Domains: Column
 using ClimaLand.Snow
@@ -28,14 +29,12 @@ climaland_dir = pkgdir(ClimaLand)
 include(joinpath(climaland_dir, "experiments/integrated/fluxnet/data_tools.jl"))
 include(joinpath(climaland_dir, "experiments/integrated/fluxnet/plot_utils.jl"))
 
-# Read in the site to be run from the command line
-#if length(ARGS) < 1
-#    error("Must provide site ID on command line")
-#end
-#
-site_ID = "US-MOz"
+include(joinpath(pkgdir(ClimaLand),"experiments/integrated/fluxnet/path_to_data_file.jl"))
+site_ID = "US-Ha1"
+data_path = get_path(site_ID)
 params = Dict("g1" => 1.0, "Vcmax25" => 1.0)
 
+driver_data = readdlm(data_path, ',')
 
 function zenith_angle(
         t,
@@ -68,19 +67,13 @@ function zenith_angle(
       )
 end
 
-# Currently, site_ID can only be "US-MOz", "US-Var", "US-Ha1", "US-NR1"
-# But we want ALL fluxnet sites
-# to do: change the included files to read meta data, all same config otherwise
-# (same parameters, same domains, same resolution, etc.)
-# data files will be read directly from caltech hpc filepath
-
 function run_single_site(params, site_ID) # e.g., run_single_site(params, "US-MOz")
 
     # Read all site-specific domain parameters from the simulation file for the site
     include(
             joinpath(
                      climaland_dir,
-                     "experiments/integrated/fluxnet/$site_ID/$(site_ID)_simulation.jl",
+                     "experiments/integrated/fluxnet/any_site/any_simulation.jl",
                     ),
            )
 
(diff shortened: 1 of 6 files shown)


.buildkite/Manifest-v1.11.toml
```diff
@@ -3401,4 +3401,4 @@ version = "2021.5.5+0"
 deps = ["Artifacts", "JLLWrappers", "Libdl", "Pkg"]
 git-tree-sha1 = "ee567a171cce03570d77ad3a43e90218e38937a9"
 uuid = "dfaa095f-4041-5dcd-9319-2fabd8486b76"
-version = "3.5.0+0"
+version = "3.5.0+0"
```
juliasloan25, line 3404: It would be best to revert this change since it isn't needed for this PR. You can copy this file from main (overwriting this change) using `git checkout main -- .buildkite/Manifest-v1.11.toml`

bucket_restart/output_active
```diff
@@ -0,0 +1 @@
+output_0002
```
juliasloan25, line 1: This is an output file that we don't track in the ClimaLand repo, so we should remove it from here

experiments/integrated/fluxnet/any_run_fluxnet.jl
```diff
@@ -0,0 +1,317 @@
+import SciMLBase
```
juliasloan25, line 1: Just a reminder to make these changes in run_fluxnet.jl and remove this file to avoid code duplication

experiments/integrated/fluxnet/met_drivers_FLUXNET.jl
```diff
@@ -12,15 +12,26 @@ import ClimaComms
 
 context = ClimaComms.context()
 
+
```
juliasloan25, line 15: remove extra line

experiments/integrated/fluxnet/met_drivers_FLUXNET.jl
```diff
@@ -12,15 +12,26 @@ import ClimaComms
...
 )
 
+# Method for getting data file path on HPC
+include(
+    joinpath(
+        pkgdir(ClimaLand),
+        "experiments/integrated/fluxnet/path_to_data_file.jl",
+    ),
+)
+
```
juliasloan25, line 28: remove, use ClimaArtifacts instead (as is done on main)

experiments/integrated/fluxnet/met_drivers_FLUXNET.jl
```diff
@@ -12,15 +12,26 @@ import ClimaComms
...
 data_path = ClimaLand.Artifacts.experiment_fluxnet_data_path(site_ID)
 driver_data = readdlm(data_path, ',')
 
+
```
juliasloan25, line 32: remove extra line

experiments/integrated/fluxnet/met_drivers_FLUXNET.jl
```diff
@@ -12,15 +12,26 @@ import ClimaComms
...
 LOCAL_DATETIME = DateTime.(format.(driver_data[2:end, 1]), "yyyymmddHHMM")
+time_offset = 5 # temporary testing
```
juliasloan25, line 34: delete if this isn't needed anymore

experiments/integrated/fluxnet/path_to_data_file.jl
```diff
@@ -0,0 +1,7 @@
+function get_path(site_ID)
```
juliasloan25, line 1: We should be able to remove this and use our artifacts directly instead

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -29,580 +29,322 @@ include(joinpath(climaland_dir, "experiments/integrated/fluxnet/data_tools.jl"))
 include(joinpath(climaland_dir, "experiments/integrated/fluxnet/plot_utils.jl"))
 
 # Read in the site to be run from the command line
-if length(ARGS) < 1
-    error("Must provide site ID on command line")
+#if length(ARGS) < 1
+#    error("Must provide site ID on command line")
+#end
+#
```
juliasloan25, line 35: Is this commented out on purpose?

juliasloan25: It would be good to continue supporting the previous behavior (i.e. running from the command line). So it would be good to check that at least 2 ARGS are passed in, since now we need `site_ID` and `params`

juliasloan25 (commented):
Thank you Thanhthanh, this is looking great so far! Let me know after you push your most recent changes and address the comments I've made here, and I'll do another review

juliasloan25:
This is looking good! I will review in a moment, but it would also be good to: 1. rebase on current main, and 2. not modify the Manifest in this PR - to do this you can check out the Manifest from main by running `git checkout origin/main -- .buildkite/Manifest-v1.11.toml`

PR #1183, opened 2025-06-18
update label for snowy land longer run

juliasloan25:
## Purpose 
The snowy land `LONGER RUN` was recently changed to run for 19 years instead of 10 years, but we didn't update the label in the buildkite pipeline. It's updated here.

### DIFF
--- a/.buildkite/longruns_gpu/pipeline.yml
+++ b/.buildkite/longruns_gpu/pipeline.yml
@@ -59,7 +59,7 @@ steps:
         artifact_paths: "california_longrun_gpu/*pdf"
         agents:
           slurm_gpus: 1
-          slurm_time: 00:30:00
+          slurm_time: 01:00:00
         env:
           CLIMACOMMS_DEVICE: "CUDA"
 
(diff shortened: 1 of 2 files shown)


.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -86,7 +86,7 @@ steps:
   - group: "Longer runs of Global Land Models"
     if: build.env("LONGER_RUN") != null
     steps:
-      - label: "Snowy Land, 10 years"
+      - label: "Snowy Land, 19 years"
```
kmdeck, line 89: I believe we also need to update the long soil run label too

kmdeck (approved):

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -108,4 +108,4 @@ steps:
           slurm_time: 15:00:00
         env:
```
kmdeck, line 109: Could we also extend the time for the regional run? It keeps failing due to that - maybe 45 minutes?

PR #1184, opened 2025-06-18
make albedo paramterization modular and make it match CLM

kmdeck:
## Content

----


src/standalone/Soil/soil_albedo.jl
```diff
@@ -0,0 +1,169 @@
...
+    "Thickness of top of soil used in albedo calculations (m)"
+    albedo_calc_top_thickness::FT
+end
+
+function CLMTwoBandSoilAlbedo{FT}(;
+    PAR_albedo_dry,
+    NIR_albedo_dry,
+    PAR_albedo_wet,
+    NIR_albedo_wet,
+    albedo_calc_top_thickness = FT(0.02),
```
kmdeck, line 51: this changes from 0.07 to 0.02, which agrees with CLM

src/standalone/Soil/soil_albedo.jl
```diff
@@ -0,0 +1,200 @@
...
+"""
+     CLMTwoBandSoilAlbedo{
+        FT <: AbstractFloat,
+        SF <: Union{FT, ClimaCore.Fields.Field},
+    } <: AbstractSoilAlbedoParameterization
+
+A parameterization for soil albedo: the soil albedo is 
+defined in two bands (PAR and NIR), can spatially vary or
+be set to scalar, and varies with water content at the surface
+θ_sfc, according to CLM:
```
juliasloan25, line 49: If the soil albedo is set to a scalar, is it still adjusted based on the surface water content, or would it be independent?

src/standalone/Soil/soil_albedo.jl
```diff
@@ -0,0 +1,200 @@
...
+
+CLM reference: Lawrence, P.J., and Chase, T.N. 2007. Representing a MODIS consistent land surface in the Community Land Model
+(CLM 3.0). J. Geophys. Res. 112:G01023. DOI:10.1029/2006JG000168.
+"""
+function albedo_from_moisture(
+    θ_sfc::FT,
+    albedo_dry::FT,
+    albedo_wet::FT,
+) where {FT}
+    Δ = max(FT(0.11) - FT(0.4) * θ_sfc, FT(0))
```
juliasloan25, line 108: It would be good to note where 0.11 and 0.4 come from here (maybe they're just the values used by CLM?)

src/standalone/Soil/soil_albedo.jl
```diff
@@ -0,0 +1,200 @@
...
+        PAR_albedo_wet,
+        NIR_albedo_wet,
+        albedo_calc_top_thickness,
+    ) = albedo
+    FT = eltype(soil_domain.fields.Δz_top)
+    # checks if there is at least 1 layer centered within the top soil depth
+    if soil_domain.fields.Δz_min < albedo_calc_top_thickness
+        # We compute ∫H_θ_dz / N, where N =∫H_dz
+        # is a normalization,
+        # and then computing ∫[(H θ)/N]dz.
```
juliasloan25, line 143:
```suggestion
        # and then compute ∫[(H θ)/N]dz.
```

src/standalone/Soil/soil_albedo.jl
```diff
@@ -0,0 +1,200 @@
...
+            albedo_calc_top_thickness + sqrt(eps(FT)),
+            soil_domain.fields.z_sfc - soil_domain.fields.z,
+        )
+        ClimaCore.Operators.column_integral_definite!(N, p.soil.sub_sfc_scratch)
+        # zeros all effective saturation at levels centered lower than boundary
+        @. p.soil.sub_sfc_scratch =
+            ClimaLand.heaviside(
+                albedo_calc_top_thickness + sqrt(eps(FT)),
+                soil_domain.fields.z_sfc - soil_domain.fields.z,
+            ) * p.soil.θ_l / N
```
juliasloan25, line 157: I see that in main we call `effective_saturation` here. I see that now you're dividing by `N` to normalize, but should this be `effective_saturation(ν, p.soil.θ_l, θ_r) / N`?

src/standalone/Soil/soil_albedo.jl
```diff
@@ -0,0 +1,200 @@
...
+        # Now use scratch space to compute normed integral
+        normed_∫H_θ_dz = p.soil.sfc_scratch
+        ClimaCore.Operators.column_integral_definite!(
+            normed_∫H_θ_dz,
+            p.soil.sub_sfc_scratch,
+        )
+        θ_sfc = normed_∫H_θ_dz
```
juliasloan25, line 164: broadcast?
```suggestion
        @. θ_sfc = normed_∫H_θ_dz
```

src/standalone/Soil/soil_albedo.jl
```diff
@@ -0,0 +1,200 @@
...
+        θ_sfc = ClimaLand.Domains.top_center_to_surface(p.soil.θ_l)
+    end
+    @. p.soil.PAR_albedo =
+        albedo_from_moisture(θ_sfc, PAR_albedo_dry, PAR_albedo_wet)
+    @. p.soil.NIR_albedo =
+        albedo_from_moisture(θ_sfc, NIR_albedo_dry, NIR_albedo_wet)
+end
+
+"""
+update_albedo!(bc::AbstractWaterBC, _, p, soil_domain, model_parameters)
```
juliasloan25, line 176: update docstring

ext/CreateParametersExt.jl
```diff
@@ -335,21 +339,12 @@ function EnergyHydrologyParameters(
...
-    NIR_albedo_dry::SF = nothing,
-    PAR_albedo_wet::SF = nothing,
-    NIR_albedo_wet::SF = nothing,
-    PAR_albedo::SFD = 0.2,
-    NIR_albedo::SFD = 0.4,
-    albedo_calc_top_thickness::TD = 0.07,
+    albedo = Soil.ConstantTwoBandSoilAlbedo(
+        CP.float_type(toml_dict)(0.2),
+        CP.float_type(toml_dict)(0.4),
+    ),
```
juliasloan25, line 345: We could also have 0.2 and 0.4 as default values in the `ConstantTwoBandSoilAlbedo` struct that can be optionally overwritten. This might tie into cleaner design/setup for the user.

docs/tutorials/standalone/Soil/sublimation.jl
```diff
@@ -103,8 +101,6 @@ params = ClimaLand.Soil.EnergyHydrologyParameters(
     K_sat,
     S_s,
     θ_r,
-    PAR_albedo,
-    NIR_albedo,
```
juliasloan25, line 107: should we add a `ClimaLand.Soil.ConstantTwoBandSoilAlbedo` here now?

src/standalone/Soil/soil_albedo.jl
kmdeck: I was wondering if I should make our parameterization exactly match CLM, in which case it would use theta_l, and not effective saturation(theta_l). I think it makes more sense to match CLM completely, because the parameters were calibrated using their scheme. But this is a separate change and I can make it in another PR?

src/standalone/Soil/soil_albedo.jl
kmdeck: I think this is OK, it was a rename, but Ill just use this as the var name in the first place instead of normed_∫H_θ_dz

src/standalone/Soil/soil_albedo.jl
```diff
@@ -0,0 +1,169 @@
...
+UCalculates and updates PAR and NIR albedo using the constant parameters provided in `albedo`.
+"""
+function update_albedo!(
+    bc::AtmosDrivenFluxBC,
+    albedo::TwoBandConstantAlbedo,
+    p,
+    soil_domain,
+    model_parameters,
+)
+    @. p.soil.PAR_albedo = albedo.PAR_albedo
```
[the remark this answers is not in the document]
kmdeck: I think they are only initialized in "set_initial_cache" or when we update aux (using this function). I'll leave a note in the docstring, but we wont ever really use this method in global runs.

ext/CreateParametersExt.jl
kmdeck: ok! I added this.

src/standalone/Soil/soil_albedo.jl
```diff
@@ -0,0 +1,221 @@
...
+
+A parameterization for soil albedo: the soil albedo is 
+defined in two bands (PAR and NIR), and can spatially vary or
+be set to scalar. However, it varies temporally due to a
+dependence on soil water content,
+θ_sfc, according to CLM:
+
+α = min(α_wet + Δ(θ_sfc), α_dry), where
+Δ(θ_sfc) = max(θ_int - dαdθ*θ_sfc,0),
+where θ_int = 0.11 and dαdθ = 0.4.
```
braghiere, line 67: In the paper Braghiere et al. (2023b) we validated this scheme vs. the updated one (weighted averages), and we showed a better agreement of the updated one. I wonder if we shouldn't just use the updated one instead: 

https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2023AV000910

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -87,8 +91,7 @@ soil_ps = Soil.EnergyHydrologyParameters(
     z_0m = z_0m_soil,
     z_0b = z_0b_soil,
     emissivity = soil_ϵ,
-    PAR_albedo = soil_α_PAR,
-    NIR_albedo = soil_α_NIR,
+    albedo = albedo,
```
braghiere, line 94: I think we should make the names a bit clearer in terms of what is soil_albedo vs. surface_albedo (canopy +soil), especially because we call emissivity = soil_e, so albedo = soil_albedo? Or is it surface_albedo?

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -582,17 +571,15 @@ function ClimaLand.make_update_aux(model::EnergyHydrology)
             κ_sat_unfrozen,
             ρc_ds,
             earth_param_set,
-            PAR_albedo_dry,
-            NIR_albedo_dry,
-            PAR_albedo_wet,
-            NIR_albedo_wet,
+            albedo,
```
braghiere, line 574: Same here. This should be soil_albedo.

src/standalone/Soil/soil_albedo.jl
kmdeck: Happy to go with what you think is best. My thought was that the parameters we read in from CLM for dry and wet albedos were calibrated with this combination of them, so if we use the CLM parameters we should use their linear combination also. If we use the different linear combination, do we need to recalibrate the soil wet and dry albedo values by color?

src/standalone/Soil/soil_albedo.jl
juliasloan25: I think it makes sense to make that change in a separate PR. So continue using `effective_saturation` here, and later switch to theta_l

juliasloan25 (commented):
I agree with Renato on the albedo -> soil_albedo renaming.

I'm not sure where the best place would be (maybe we need a new page?), but it would be good to explain this new structure in the docs - e.g. the options for soil albedo parameterizations and the default behavior if none is specified.

experiments/integrated/fluxnet/run_fluxnet.jl
kmdeck: I will clarify this. In this case, it is in the soil parameter struct so it is the soil alone.

src/standalone/Soil/energy_hydrology.jl
kmdeck: Ill clarify this

experiments/integrated/fluxnet/run_fluxnet.jl
kmdeck: I changed this also in ozark_pft

src/standalone/Soil/soil_albedo.jl
braghiere: Thanks, that makes sense. Just to clarify: in our 2023b paper we used theta directly from SMAP because using S would require knowing both theta_r and theta_sat, which introduces uncertainty due to spatial variability in soil properties. One advantage of the theta-based scheme was that it was fully observation-driven. 

But I agree. if we're working in a model that defines theta_r and theta_sat, then using S is more consistent with the physics.

lib/ClimaLandSimulations/Manifest.toml
```diff
@@ -0,0 +1,2658 @@
+# This file is machine-generated - editing it directly is not advised
```
juliasloan25, line 1: did you add this file on purpose?

juliasloan25 (approved):
LGTM! I do still think it would be good to have a docs page explaining the parameterization options. I'm okay with starting it here or in another PR

lib/ClimaLandSimulations/Manifest.toml
kmdeck: let me try removing it, and Ill add something to the docs, and then merge on Monday

braghiere (approved):

PR #1191, opened 2025-06-24
Tr/use land simulation struct

### LINKED ISSUE #1190: use simulations struct in most (all?) examples
TODO:

- [x] Convert all dates/times to Time internally (possibly remove `start_date`)
- [x] FluxNet experiments
- [x] All other experiments
- [x] Docs
- [ ] verify outputs of runs

imreddyTeja:
## Purpose 
Closes #1190 

This PR makes all experiments and docs use the `LandSimulation` struct. This also means that all changed simulations will use ITime internally. 

## Content
Add `LandSimulation` constructors that handle ITime conversion.

Make callbacks work with `ITime`

Add kwargs to `LandSimulation`

Drop `FT` as argument to `LandSimulation`

If an experiment/simulation has a start_date anywhere, use start and end date instead of t0 and tf. This makes it so that each simulation only has one notion of time.

TODO:

Some of the output plots do not match because the callback saving order is modified. Fix this before merging. Examples are ozark_soilsnow and snowmip.

#1309 clean up the `set_ic!` functions

----

### DIFF
--- a/docs/src/tutorials/integrated/soil_canopy_tutorial.jl
+++ b/docs/src/tutorials/integrated/soil_canopy_tutorial.jl
@@ -50,6 +50,7 @@ using ClimaLand.Soil
 using ClimaLand.Soil.Biogeochemistry
 using ClimaLand.Canopy
 using ClimaLand.Canopy.PlantHydraulics
+import ClimaLand.Simulations: LandSimulation, solve!
 import ClimaLand
 import ClimaLand.Parameters as LP
 using DelimitedFiles
@@ -65,7 +66,11 @@ earth_param_set = LP.LandParameters(FT);
 site_ID = "US-MOz"
 # Timezone (offset from UTC in hrs)
 time_offset = 7
-start_date = DateTime(2010) + Hour(time_offset)
+start_date = DateTime(2010) + Hour(time_offset) + Day(150) # start mid year
+# Specify the time range and dt value over which to perform the simulation.
+N_days = 100
+end_date = start_date + Day(N_days)
+dt = Float64(30)
 
 # Site latitude and longitude
 lat = FT(38.7441) # degree
@@ -74,11 +79,6 @@ long = FT(-92.2000) # degree
 # Height of the sensor at the site
 atmos_h = FT(32) # m
 
-# Specify the time range and dt value over which to perform the simulation.
-t0 = Float64(150 * 3600 * 24)# start mid year
-N_days = 100
-tf = t0 + Float64(3600 * 24 * N_days)
-dt = Float64(30)
 
 # - We will be using prescribed atmospheric and radiative drivers from the
 #   US-MOz tower, which we read in here. We are using prescribed
@@ -186,8 +186,8 @@ soilco2 = Soil.Biogeochemistry.SoilCO2Model{FT}(domain, drivers);
 # Read in prescribed LAI at the site from global MODIS data
 surface_space = domain.space.surface;
 modis_lai_ncdata_path = ClimaLand.Artifacts.modis_lai_multiyear_paths(;
-    start_date = start_date + Second(t0),
-    end_date = start_date + Second(t0) + Second(tf),
+    start_date = start_date,
+    end_date = end_date,
 );
 LAI = ClimaLand.prescribed_lai_modis(
     modis_lai_ncdata_path,
@@ -273,50 +273,40 @@ canopy = ClimaLand.Canopy.CanopyModel{FT}(
 # Now we can combine the soil and canopy models into a single combined model.
 land = SoilCanopyModel{FT}(soilco2, soil, canopy);
 
-# Now we can initialize the state vectors and model coordinates, and initialize
-# the explicit/implicit tendencies as usual. The Richard's equation time
-# stepping is done implicitly, while the canopy model may be explicitly stepped,
-# so we use an IMEX (implicit-explicit) scheme for the combined model.
-
-Y, p, coords = initialize(land);
-exp_tendency! = make_exp_tendency(land);
-imp_tendency! = make_imp_tendency(land);
-jacobian! = make_jacobian(land);
-jac_kwargs =
-    (; jac_prototype = ClimaLand.FieldMatrixWithSolver(Y), Wfact = jacobian!);
-
 # We need to provide initial conditions for the model:
-Y.soil.ϑ_l = FT(0.4)
-Y.soil.θ_i = FT(0.0)
-T_0 = FT(288.7)
-ρc_s =
-    volumetric_heat_capacity.(
-        Y.soil.ϑ_l,
-        Y.soil.θ_i,
-        land.soil.parameters.ρc_ds,
-        earth_param_set,
-    )
-Y.soil.ρe_int =
-    volumetric_internal_energy.(Y.soil.θ_i, ρc_s, T_0, earth_param_set)
+function set_ic!(Y, p, t0, model)
+    Y.soil.ϑ_l = FT(0.4)
+    Y.soil.θ_i = FT(0.0)
+    T_0 = FT(288.7)
+    ρc_s =
+        volumetric_heat_capacity.(
+            Y.soil.ϑ_l,
+            Y.soil.θ_i,
+            land.soil.parameters.ρc_ds,
+            earth_param_set,
+        )
+    Y.soil.ρe_int =
+        volumetric_internal_energy.(Y.soil.θ_i, ρc_s, T_0, earth_param_set)
 
-Y.soilco2.C .= FT(0.000412) # set to atmospheric co2, mol co2 per mol air
+    Y.soilco2.C .= FT(0.000412) # set to atmospheric co2, mol co2 per mol air
 
-canopy_retention_model = canopy.hydraulics.parameters.retention_model
-ψ_stem_0 = FT(-1e5 / 9800)
-ψ_leaf_0 = FT(-2e5 / 9800)
+    canopy_retention_model = canopy.hydraulics.parameters.retention_model
+    ψ_stem_0 = FT(-1e5 / 9800)
+    ψ_leaf_0 = FT(-2e5 / 9800)
 
-S_l_ini =
-    inverse_water_retention_curve.(
-        canopy_retention_model,
-        [ψ_stem_0, ψ_leaf_0],
-        plant_ν,
-        plant_S_s,
-    )
-for i in 1:2
-    Y.canopy.hydraulics.ϑ_l.:($i) .=
-        augmented_liquid_fraction.(plant_ν, S_l_ini[i])
+    S_l_ini =
+        inverse_water_retention_curve.(
+            canopy_retention_model,
+            [ψ_stem_0, ψ_leaf_0],
+            plant_ν,
+            plant_S_s,
+        )
+    for i in 1:2
+        Y.canopy.hydraulics.ϑ_l.:($i) .=
+            augmented_liquid_fraction.(plant_ν, S_l_ini[i])
+    end
+    evaluate!(Y.canopy.energy.T, atmos.T, t0)
 end
-evaluate!(Y.canopy.energy.T, atmos.T, t0)
 
 # Choose the timestepper and solver needed for the problem.
 timestepper = CTS.ARS343()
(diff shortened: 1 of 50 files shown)


src/shared_utilities/Domains.jl
```diff
@@ -1209,6 +1209,9 @@ end
 function landsea_mask(domain::Domains.AbstractDomain; kwargs...)
     # average_horizontal_resolution_degrees returns a tuple with the resolution
     # along the two directions, so we take the minimum
+    domain isa Union{SphericalShell, SphericalSurface} ||
+        (hasproperty(domain, :longlat) && !isnothing(domain.longlat)) ||
+        return nothing
```
juliasloan25, line 1214:
```suggestion
    (hasproperty(domain, :longlat) && isnothing(domain.longlat)) && return nothing
```

src/shared_utilities/Domains.jl
```diff
@@ -1209,6 +1209,9 @@ end
 function landsea_mask(domain::Domains.AbstractDomain; kwargs...)
```
juliasloan25, line 1209: this function for SphericalShell, SphericalSurface, HybridBox, Plane

new method for Point, Column that returns nothing

src/shared_utilities/utils.jl
```diff
@@ -418,7 +449,16 @@ initial values, don't pass the `initialize` argument to the `DiscreteCallback`
...
-    (t in cb.affect!.saveat) && cb.affect!(integrator)
+    saveat = cb.affect!.saveat
+    if saveat isa Array{<:Dates.DateTime}
+        # If saveat is a vector of DateTimes, we need to convert t to DateTime
+        return (Dates.DateTime(t) in cb.affect!.saveat) &&
+               cb.affect!(integrator)
+    elseif saveat isa Array{<:Number}
+        return (float(t) in cb.affect!.saveat) && cb.affect!(integrator)
+    else
+        return (t in cb.affect!.saveat) && cb.affect!(integrator)
```
juliasloan25, line 460: convert t to eltype of saveat (dispatch on saveat type):
```
function convert_t(saveat, t)
  return eltype(saveat)(t)
end

File: experiments/integrated/fluxnet/run_fluxnet.jl [hunk 4]
```diff
@@ -254,33 +254,23 @@ diags = ClimaLand.default_diagnostics(
...
-    ),
-    Y,
-    (t0, tf),
-    p,
-);
-
-@time sol = SciMLBase.solve(prob, ode_algo; dt = dt, callback = cb);
+updateat = Array(t0:data_dt:tf)
+simulation = LandSimulation(
+    t0,
```
kmdeck, line 263: is it possible to work with dates only here? start_date, stop_date, dt, like we do for global runs?

imreddyTeja: This is definitely possible. Is `start_date` always equivalent to `start_date + Dates.Second(tf)`?

imreddyTeja: If we change to the user passing in an update frequency, then the ability to have non uniform update intervals, or updates that don't span t0 to tf is lost.  I don't think that is ever done. The same goes for the solver saveat.

We could either support passing saveat/updateat as:

Either an Interval or array
Only an interval
Only an array

kmdeck: Specifically for the driver callbacks, I dont see why we would not update for the entire time span.so passing an update frequency would work. For the noninterpsavingcallback, the same might not be true, but in that case, we pass a user_callback and probably cant avoid the t0, tf definitions. since those are rarer use cases (and noninterpsavingcallback isnt really meant to be user facing?) I think this is ok?

For now maybe we can pass the interval as a kwarg, with a default of 3 hours. I'm not 100% sure philosophically where this interval should belong. in the forcing struct itself? or does it make sense to have the simulation? But we can think about that more in the future.

src/shared_utilities/Domains.jl
```diff
@@ -1206,7 +1206,16 @@ function landsea_mask(
     return binary_mask
 end
 
-function landsea_mask(domain::Domains.AbstractDomain; kwargs...)
+# Points and Columns do not have a horizontal dim, so a horizontal mask cannot be applied
+landsea_mask(domain::Union{Point, Column}; kwargs...) = nothing
```
kmdeck, line 1210: Could you explain why this is needed? I think we only call the landsea_mask inside of `global_domain`? or is this inside the nanchecker? I think there is an option to run that without a mask

ph-kev (approved):
This looks good to me

src/diagnostics/default_diagnostics.jl
```diff
@@ -401,3 +418,11 @@ function default_diagnostics(
...
     end
 end
+
+# fallback method id no default diagnostics are defined
+default_diagnostics(
+    land_model::ClimaLand.AbstractModel,
+    start_date;
+    output_writer,
+    average_period = nothing,
+) = []
```
juliasloan25, line 428: This is redundant with the method defined on line 59

src/shared_utilities/Domains.jl
```diff
@@ -1293,6 +1302,7 @@ function global_domain(
     if apply_mask
         surface_space = domain.space.surface # 2d space
         binary_mask = landsea_mask(domain; threshold = mask_threshold)
+        @show binary_mask
```
juliasloan25, line 1305:
```suggestion
```

src/shared_utilities/drivers.jl
```diff
@@ -1831,7 +1835,10 @@ function empirical_diffuse_fraction(
     thermo_params,
 ) where {FT}
     if t isa ITime
-        DOY = Dates.dayofyear(date(t))
+        DOY = Dates.dayofyear(
+            isnothing(t.epoch) ?
+            start_date + Dates.Second(floor(Int64, float(t))) : date(t),
+        )
```
juliasloan25, line 1841: can you add a comment explaining why we need this check?

src/shared_utilities/utils.jl
```diff
@@ -397,18 +398,22 @@ end
...
+        if curr_t == integrator_t
             affect!.saved_values.t[affect!.saveiter] = curr_t
             affect!.saved_values.saveval[affect!.saveiter] =
                 deepcopy(integrator.p)
         end
     end
 end
 
+convert_t_saveat(saveat, t) = eltype(saveat)(t)
+convert_t_saveat(saveat::Vector{T}, t::T) where {T} = t
```
juliasloan25, line 415: do we also need to convert `t` to `eltype(saveat)` in this case?

src/shared_utilities/utils.jl
```diff
@@ -666,6 +684,27 @@ function NaNCheckCallback(
     SciMLBase.DiscreteCallback(cond, affect!)
 end
 
+# no date version
```
juliasloan25, line 687: Could you explain a little more here? Why don't we use this method in all cases?

experiments/standalone/Vegetation/varying_lai_with_stem.jl
```diff
@@ -39,6 +43,15 @@ start_date = DateTime(2010) + Hour(time_offset)
...
 )
+z0_m = FT(2)
+z0_b = FT(0.2)
+
+shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
+    z0_m,
+    z0_b,
+    earth_param_set,
+);
+
```
juliasloan25, line 54: This was recently removed - probably a rebase error
```suggestion
```

experiments/standalone/Vegetation/varying_lai.jl
```diff
@@ -73,41 +77,26 @@ canopy = ClimaLand.Canopy.CanopyModel{FT}(
...
-N_days = 364
-tf = t0 + 3600 * 24 * N_days
-dt = 225.0;
-evaluate!(Y.canopy.energy.T, atmos.T, t0)
-set_initial_cache! = make_set_initial_cache(canopy)
-set_initial_cache!(p, Y, t0);
+function set_ic!(Y, p, t0, model)
+    Y.canopy.hydraulics.ϑ_l.:1 .= augmented_liquid_fraction.(ν, S_l_ini)
+    evaluate!(Y.canopy.energy.T, atmos.T, t0)
+end
```
juliasloan25, line 88: I think it's cleaner and less error-prone to avoid a closure and move parameter access inside of `set_ic!`. And then you can delete the access of these parameters outside the function (lines 81-83)
```suggestion
function set_ic!(Y, p, t0, model)
    (; retention_model, ν, S_s) = model.hydraulics.parameters
    ψ_leaf_0 = FT(-2e5 / 9800)
    S_l_ini = inverse_water_retention_curve(retention_model, ψ_leaf_0, ν, S_s)
    
    Y.canopy.hydraulics.ϑ_l.:1 .= augmented_liquid_fraction.(ν, S_l_ini)
    evaluate!(Y.canopy.energy.T, atmos.T, t0)
end
```

juliasloan25: Actually this applies to pretty much every experiment, so why don't we leave it as an improvement for a followup PR. Could you open an issue for it please?

experiments/integrated/performance/integrated_timestep_test.jl
```diff
@@ -111,7 +111,7 @@ function set_initial_conditions(land, t0)
 
     Y.canopy.energy.T = FT(297.5)
     set_initial_cache!(p, Y, t0)
```
juliasloan25, line 113:
```suggestion
```
this is done inside the `LandSimulation` constructor

juliasloan25 (approved):
This looks great, thank you! I left some small comments but I think it's good to merge, given that you checked the experiment/doc outputs are unchanged

docs/src/tutorials/integrated/soil_canopy_tutorial.jl
```diff
@@ -273,50 +273,40 @@ canopy = ClimaLand.Canopy.CanopyModel{FT}(
...
-    inverse_water_retention_curve.(
-        canopy_retention_model,
-        [ψ_stem_0, ψ_leaf_0],
-        plant_ν,
-        plant_S_s,
-    )
-for i in 1:2
-    Y.canopy.hydraulics.ϑ_l.:($i) .=
-        augmented_liquid_fraction.(plant_ν, S_l_ini[i])
+function set_ic!(Y, p, t0, model)
```
kmdeck, line 277: (just seconding Julia's comment - here we pass in model, but use land, canopy, earth_param_set, etc, without getting them from `model`)

docs/src/tutorials/standalone/Soil/evaporation.jl
```diff
@@ -191,32 +180,28 @@ ode_algo = CTS.IMEXAlgorithm(
...
-saveat = Array(t0:3600.0:tf)
+saveat = Array(start_date:Second(3600.0):end_date);
 sv_hr = (;
-    t = Array{Float64}(undef, length(saveat)),
+    t = Array{DateTime}(undef, length(saveat)),
     saveval = Array{NamedTuple}(undef, length(saveat)),
 )
 saving_cb = ClimaLand.NonInterpSavingCallback(sv_hr, saveat)
 updateat = deepcopy(saveat)
-model_drivers = ClimaLand.get_drivers(soil)
```
kmdeck, line 212: did you want to pass updateat into the simulation here?

src/simulations/Simulations.jl
```diff
@@ -59,11 +60,57 @@ struct LandSimulation{
...
+time-stepping algorithm. During the construction process, the initial conditions
+are set using `set_ic!`, the initial cache is set using `make_set_initial_cache(model)`,
+and driver and diagnostic callbacks are created. If `updateat` is a vector, the drivers
+will be updated at those times; if it is a single time, the drivers will be updated
+at that interval. The order the callbacks are called while stepping the simulation is:
+1. user_callbacks
+2. driver update callback
+3. diagnostics callback
+"""
 function LandSimulation(
```
kmdeck, line 110: I like that all the different methods call this under the hood!

docs/src/tutorials/standalone/Soil/evaporation.jl
imreddyTeja: Yes. Thanks for catching that

src/shared_utilities/drivers.jl
```diff
@@ -285,8 +285,12 @@ function default_zenith_angle(
     insol_params,
 ) where {T, LT}
     FT = eltype(latitude)
-    current_datetime =
-        T <: ITime ? date(t) : start_date + Dates.Second(round(t))
+    current_datetime = if T <: ITime
+        # use start_date if available. some fluxnet simulations use different driver and sim dates
+        isnothing(start_date) ? date(t) : start_date + t.counter * t.period
```
kmdeck, line 290: Youve thought about this a lot more than me, but should we not rely on the simulation time if it is an ITime? and only use the start_date if there is no t.epoch?, as below, when computing DOY

src/diagnostics/default_diagnostics.jl
```diff
@@ -401,3 +418,11 @@ function default_diagnostics(
         )
     end
 end
+
+# fallback method id no default diagnostics are defined
```
kmdeck, line 422: *if

src/shared_utilities/drivers.jl
imreddyTeja: Yes we should! This is a remnant from before the fluxnet refactoring. Before I rebased, one of the fluent simulations ran at a date that differed from the forcing data date. This is no longer the case, and I will remove this.

kmdeck (approved):
Thanks for taking this on! This is a big effort.

Looking through this, I think there are some areas of improvement for future PRs with how we handle callbacks, but this looks ready to merge aside from a few smaller comments.
- What should the user_callbacks default be? What if we made the NaNChecker dispatch off of the domain and not use a mask in the non-global case? (I still am not a huge fan of having to have a landsea_mask method for some domains...) Or what if this was not in the default user_callback set?
- Unifying our callbacks (which I think are all DiscreteCallbacks) - can we have one callback type, but different affect! functions?
- NonInterpSavingCallback could possibly define and make `sv` instead of having that be something the user needs to know how to define
- if different callbacks have potentially different kwargs, maybe we could pass a list of callback types and callback kwargs? instead of the callbacks themselves

kmdeck:
> For the `landsea_mask` point, @juliasloan25 preferred having the method defined for all possible domains, rather than just a subset of them. I'm OK with either of the ways of dealing with default user_callbacks.

ok! sounds fine as is then. thanks again!

Issue #1197, opened 2025-06-30
P model core optimality computation


ychnli:
This is a subissue under the P-model implementation #1185 . This step should simply replicate the behavior of the original Stocker et al. (2020) implementation [Rpmodel](https://github.com/geco-bern/rpmodel/tree/master/R).

PR #1198, opened 2025-06-30
Yl/p model

ychnli:
## Purpose 

This implements of the core P-model computations in the function `compute_pmodel_outputs` (#1197). 

## To-do

- still need to write `update_photosynthesis!` in `pmodel.jl` with appropriate subdaily timestepping equations 
- still need to make sure that stomatal conductance is appropriately updated, since P-model predicts it jointly from the optimality criteria 

## Content

- `compute_pmodel_outputs`and associated helper functions in `pmodel.jl` and `canopy_parameterizations.jl` 
- `test_pmodel.jl` unit tests and test files on ClimaArtifacts

----


src/standalone/Vegetation/pmodel.jl
```diff
@@ -0,0 +1,314 @@
...
+end
+
+"""
+    PModelDrivers{FT<:AbstractFloat}
+
+The required drivers for P-model (Stocker et al. 2020). Drivers are defined as
+external variables used to compute the optimal photosynthetic capacities. 
+$(DocStringExtensions.FIELDS)
+"""
+Base.@kwdef struct PModelDrivers{
```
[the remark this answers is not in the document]
ychnli: By this do you mean just passing in `Y`, `p`, and `canopy` into `compute_pmodel_outputs`? My reason for explicitly bundling them into drivers is so I can call `outputs = compute_pmodel_outputs(parameters, drivers, constants)` for clarity. If you think that this is actually more cumbersome than needed, I can change it.

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1099,3 +1108,459 @@ end
...
+    ) where {FT}
+
+    Computes the intrinsic quantum yield of photosynthesis ϕ (mol/mol) 
+    as a function of temperature T (K) and a calibratable parameter c (unitless). 
+    The functional form is given in Bernacchi et al (2003) and is used in Stocker 
+    et al. (2020) 
+"""
+function intrinsic_quantum_yield(T::FT, c::FT) where {FT}
+    # convert to C
+    T = T - FT(273.15)
```
kmdeck, line 1127: If there is any chance these other constants might vary or be calibrated later, we could move them out of this and stick in the PModel struct (i.e., not hardcoded). We typically try and do this. What does @juliasloan25 think?

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1099,3 +1108,459 @@ end
...
+
+    Computes the CO2 compensation point (`Γstar`), in units Pa, as a function of temperature T (K)
+    and pressure p (Pa). See Equation B5 of Stocker et al. (2020). 
+"""
+function co2_compensation_p(
+    T::FT,
+    To::FT,
+    p::FT,
+    R::FT, 
+    ΔHΓstar::FT = FT(37830),
```
kmdeck, line 1294: Likewise with these - Let's not supply defaults for these, and make them regular arguments?

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1099,3 +1108,459 @@ end
...
+    Ko25::FT, 
+    ΔHkc::FT, 
+    ΔHko::FT, 
+    To::FT, 
+    R::FT
+) where {FT}
+    Kc = MM_Kc(Kc25, ΔHkc, T, To, R)
+    Ko = MM_Ko(Ko25, ΔHko, T, To, R)
+
+    return Kc * (FT(1.0) + po2(p) / Ko)
```
kmdeck, line 1332: we do this very inconsistently in our code, but here you can do 1+po2(p)/Ko, and 1 will take the right float type (i.e., no need for FT()

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1099,3 +1108,459 @@ end
...
+        χ::FT, 
+        ca::FT,
+        A::FT
+    ) where {FT}
+
+    Computes the stomatal conductance (`gs`), in units of mol/m^2/s via Fick's law. 
+    Parameters are the ratio of intercellular to ambient CO2 concentration (`χ`),
+    the ambient CO2 concentration (`ca`), and the assimilation rate (`A`).
+"""
+function compute_gs(
```
kmdeck, line 1392: maybe `compute_pmodel_gs`?

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1099,3 +1108,459 @@ end
...
+    Computes light use efficiency (LUE) in kg C/mol from intrinsic quantum yield (`ϕ0`),
+    moisture stress factor (`β`), and a Jmax modified capacity (`mprime`); see Eqn 17 and 19
+    in Stocker et al. (2020). 
+"""
+function compute_LUE(
+    ϕ0::FT, 
+    β::FT,
+    mprime::FT
+) where {FT} 
+    Mc = FT(0.0120107) # molar mass of carbon (kg/mol)
```
kmdeck, line 1435: Probably this should also go in the parameter struct

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1099,3 +1108,459 @@ end
...
+    Computes the maximum rate of carboxylation assuming optimality and Aj = Ac using 
+    the intrinsic quantum yield (`ϕ0`), absorbed radiation (`I_abs`), Jmax-adjusted capacity
+    (`mprime`), and a Rubisco-limited capacity (`mc`). See Eqns 16 and 6 in Stocker et al. (2020). 
+"""
+function pmodel_vcmax(
+    ϕ0::FT, 
+    I_abs::FT,
+    mprime::FT,
+    mc::FT,
+    βm::FT = FT(1.0)
```
kmdeck, line 1457: do we need this beta_m?

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1099,3 +1108,459 @@ end
...
+    ) where {FT}
+
+    Computes an empirical soil moisture stress factor (`β`) according to the quadratic
+    functional form of Stocker et al. (2020), Eq 21, using the soil moisture (`θ`),
+    AET/PET (`meanalpha`), and parameters `a_hat`, `b_hat`, `θ0`, and `θ1`. Default values
+    are from the original paper. 
+"""
+function quadratic_soil_moisture_stress(
+    θ::FT,
+    meanalpha::FT = FT(1.0),
```
kmdeck, line 1480: these should also be in a struct

src/standalone/Vegetation/pmodel.jl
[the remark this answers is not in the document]
kmdeck: oh - yes - I think Julia is saying we can obtain these directly from p.drivers.P, p.drivers.c_co2, compute canopy T from Y,p, etc.

that is, I think what you have here would require us to make a field of PModelDrivers which duplicates a lot of stuff we already have access to in p.

src/standalone/Vegetation/pmodel.jl
```diff
@@ -0,0 +1,313 @@
...
+        P_air = P_air,
+        VPD = VPD,
+        βm = FT(1.0) # TODO: add a new file for soil moisture stress parameterizations
+    )
+
+    # Use model's parameters and constants
+    parameters = model.parameters
+    constants = model.constants
+
+    outputs = compute_pmodel_outputs(parameters, drivers, constants)
```
kmdeck, line 302: Does this run right now, if the fields of `drivers` are fields? I think you would need to broadcast all the functions inside of  `compute_pmodel_outputs` to do so.

We should be careful here also not to allocate. If `compute_p_model_outputs` is broadcasted, it allocates `outputs`, which is a field of NamedTuples with many keys. Or,  if all the functions in it are broadcasted, we allocate a ton of fields inside of that in addition to `outputs`.

Best would be if we could write a function that holds at a single point in space ( expects a scalar) and returns a scalar. Then we broadcast that, e.g. `@. p.canopy.photosynthesis.Vcmax25 = compute_pmodel_vcmax25(parameters, T_canopy, P, VPD, ...., constants)` and like so for GPP, Rd, etc.
This avoids any allocations.

That may be impractical because these things all rely on the same intermediate quantities. In that case, we might have compute_pmodel_outputs return a named tuple only of what we need (GPP, Rd, An, VCmax25), and compute
output = @. lazy(compute_pmodel_outputs(parameters, T_canopy, P, VPD, ...., constants))

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1099,3 +1108,459 @@ end
...
+        p::FT
+    ) where {FT}
+
+    Computes the density of water in kg/m^3 given temperature T (K) and pressure p (Pa)
+    according to F.H. Fisher and O.E Dial, Jr. (1975) Equation of state of pure water and 
+    sea water, Tech. Rept., Marine Physical Laboratory, San Diego, CA.
+
+    (can consider removing the higher order terms?) 
+"""
+function density_h2o(T::FT, p::FT) where {FT}
```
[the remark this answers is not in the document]
ychnli: Yes, maybe constant by default is acceptable. At typical atmospheric PT conditions the density varies by only order 1% (See below). Where is the constant density of water defined? 

image

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1099,3 +1108,459 @@ end
...
+
+
+"""
+    po2(P_air::FT) 
+
+    Computes the partial pressure of O2 in the air (Pa) given atmospheric pressure. Assumes a 
+    constant mixing ratio of 21000 / 101325. 
+"""
+function po2(P_air::FT) where {FT} 
+    return FT(0.2095) * P_air
```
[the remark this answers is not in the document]
ychnli: how do we add parameters to ClimaParams?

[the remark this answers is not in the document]
kmdeck: I think for now you can stick it in the PModel parameter struct. In the future we'll have to add it to ClimaParams, but this involves a PR to that repo, a release...and better to do that later on in one go

src/standalone/Vegetation/canopy_parameterizations.jl
ychnli: this should be stomatal conductance of CO2, sorry. This differs with the conductance of H2O by a factor Drel = 1.6. Should we change this to output the conductance w.r.t. water by default?

src/standalone/Vegetation/canopy_parameterizations.jl
ychnli: I will resolve these in a new issue, since even the functional form of soil moisture stress is fairly uncertain

ychnli: so I think we should represent it as a separate parameterization entirely

src/standalone/Vegetation/canopy_parameterizations.jl
kmdeck: Let's copy what we do for the Medlyn model

src/standalone/Vegetation/pmodel.jl
```diff
@@ -0,0 +1,337 @@
...
+    N_a = LP.avogadro_constant(earth_param_set)
+    (; _, λ_γ_PAR, Ω) = canopy.radiative_transfer.parameters
+    energy_per_mole_photon_par = planck_h * lightspeed / λ_γ_PAR * N_a
+
+    T_canopy = canopy_temperature(canopy.energy, canopy, Y, p)
+    f_abs = p.canopy.radiative_transfer.par.abs
+    ca = p.drivers.c_co2
+    P_air = p.drivers.P
+    par_d = p.canopy.radiative_transfer.par_d
+    VPD = ClimaLand.vapor_pressure_deficit(
```
kmdeck, line 305: lazy broadcast?

juliasloan25:
The downstream tests are failing because of an unrelated ClimaCoupler problem, but the formatter test can be fixed by running `julia --project=.dev .dev/climaformat.jl`

PR #1199, opened 2025-06-30
Default global plots using simulation struct

kmdeck:
## Purpose 
Moves leaderboard and figures function code to src, and makes the default make_plots(simulation)
function. Uses this in three of our long runs. 

## To-do
propagate to the bucket long run?

## Content

----


ext/LandSimulationVisualization.jl
```diff
@@ -0,0 +1,239 @@
...
+Uses the diagnostic output of the `sim` to create leaderboard plots, comparing the output of the simulation to the ``observations"
+from ERA5 or ILAMB. In the future, other observations can be included - see the leadboard directory included here for details.
+"""
+function make_leaderboard_plots(
+    sim::ClimaLand.Simulations.LandSimulation;
+    outdir = ".",
+    leaderboard_data_sources = ["ERA5", "ILAMB"],
+)
+    make_leaderboard_plots(
+        model,
```
kmdeck, line 32: sim.model

kmdeck:
check_weakdeps = true in aqua compat

ph-kev:
I am going to take a look at the Makie/GeoMakie bug this week. I looked at it a little last week and I think it is a bug from Makie/GeoMakie, unless the code is using `GridLayout` in a weird way, which I don't think it is.

kmdeck:
> I am going to take a look at the Makie/GeoMakie bug this week. I looked at it a little last week and I think it is a bug from Makie/GeoMakie, unless the code is using `GridLayout` in a weird way, which I don't think it is.

thanks!

imreddyTeja (commented):
This looks good to me. Could you rebase off of main after I merge in the `GeoMakie` rollback, and run the long runs again?

test/aqua.jl
```diff
@@ -26,7 +26,7 @@ end
     Aqua.test_undefined_exports(ClimaLand)
     # Insolation is a direct dependency for the package extension
     Aqua.test_stale_deps(ClimaLand; ignore = [:Insolation])
-    Aqua.test_deps_compat(ClimaLand)
+    Aqua.test_deps_compat(ClimaLand; check_weakdeps = false)
```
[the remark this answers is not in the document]
kmdeck: I'll add them and change this back

experiments/long_runs/bucket.jl
```diff
@@ -32,12 +30,15 @@ import ClimaLand.Parameters as LP
...
 using Dates
 import NCDatasets
 
-using Poppler_jll: pdfunite
+using CairoMakie, GeoMakie, Poppler_jll, ClimaAnalysis
+LandSimVis =
+    Base.get_extension(
+        ClimaLand,
+        :LandSimulationVisualization,
+    ).LandSimulationVisualization;
```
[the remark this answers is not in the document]
kmdeck: That's a good point. But making that change should apply then to all of our tutorials and other examples, so I think if we want to do it, we can do it consistently and in another PR?

Project.toml
```diff
@@ -48,13 +56,16 @@ DataFrames = "1.4"
...
+GeoMakie = "< 0.7.13"
 HTTP = "1.10"
 Insolation = "0.9.2"
 Interpolations = "0.15.1, 0.16"
 LazyArtifacts = "1"
 LazyBroadcast = "1.0.0"
 LinearAlgebra = "1"
 NCDatasets = "0.14"
+Poppler_jll = "24"
+Printf = "1.10"
```
kmdeck, line 68: I tried Printf = "1", but got this error in CI, so I updated it:

```
 ERROR: LoadError: Unsatisfiable requirements detected for package Printf [de0858da]:
 Printf [de0858da] log:
 ├─possible versions are: 1.10.10 or uninstalled (package in sysimage!)
 ├─restricted to versions 1.0 by ClimaLand [08f4d4ce], leaving only versions: uninstalled
 │ └─ClimaLand [08f4d4ce] log:
 │   ├─possible versions are: 0.17.0 or uninstalled
 │   └─ClimaLand [08f4d4ce] is fixed to version 0.17.0
```

kmdeck: I dont really understand though, since Printf = "1" should be [1.0.0 - 2.0.0)

imreddyTeja (approved):

PR #1210, opened 2025-07-07
[WIP] Updated land calibration pipeline

ph-kev:
This PR rewrites the land calibration pipeline to use the latest snowy land model, the new `ObservationRecipe` in `ClimaCalibrate`, and ClimaAnalysis for data preprocessing and transformations. With these additions, the pipeline should be 

- simpler to understand
- offload the covariance matrix computation to ClimaCalibrate
- less error prone when adding priors, modifying or adding new observations, modifying the calibration configuration, etc

However, the updated land calibration pipeline is brittle in multiple aspects and these issues affect the other calibration pipelines. In this issue, the issues specific to the land calibration pipeline is listed.

1. **Overwriting parameters is painful.** There is no easy way of overwriting parameters. 

https://github.com/CliMA/ClimaLand.jl/blob/ccc311536364d802e7c5e0806d5b633d8503a977/experiments/calibration/forward_model_land.jl#L52-L75

2. **Configuring the calibration is not straightforward**. To configure the calibration, all the settings are centralized to a single file, but making the user modify what those functions return does not seem ideal. Furthermore, it is not clear whether using functions is the best way to pass those information to the worker processes.

3. **Adding a variable is difficult and not obvious.** To add a new variable, you need to modify three different files, specify what the simulation and observational data are, how they should be preprocessed (e.g. units conversion, shifting dates), and add additional data transformation (e.g. seasonal averages). This process is error prone.


experiments/calibration/generate_observations.jl
```diff
@@ -0,0 +1,163 @@
...
+        left = 2,
+        right = length(lats) - 1,
+        by = ClimaAnalysis.Index(),
+    )
+
+    var = ClimaCalibrate.ObservationRecipe.change_data_type(var, Float32)
+    return var
+end
+
+if abspath(PROGRAM_FILE) == @__FILE__
```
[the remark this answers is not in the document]
juliasloan25: thanks!

experiments/calibration/observation_map.jl
```diff
@@ -1,77 +1,215 @@
...
+        var = ClimaAnalysis.average_season_across_time(var, ignore_nan = false)
+
+        ocean_mask = make_ocean_mask(nelements)
+        var = ocean_mask(var)
+
+        # Replace all NaNs on land with the mean value
+        # This is needed to stop small NaN values from stopping the calibration
+        # TODO: Replace this with mask aware flatten once it is merged in
+        # ClimaAnalysis
+        nanmean_land_val = 500.0
```
nefrathenrici, line 92: Is this an appropriate fill value? It looks like this will apply the same value to different vars with potentially different scales. Maybe we could have a dict/mapping from short names to default fill values, with one overall default if the mapping is not found.

experiments/calibration/observation_map.jl
```diff
@@ -1,77 +1,215 @@
...
+end
+
+"""
+    plot_constrained_params_and_errors(output_dir, ekp, prior)
 
-    return vcat(obs...)
+Plot the constrained parameters and errors from `ekp` and `prior` and save
+them to `output_dir`.
+"""
+function plot_constrained_params_and_errors(output_dir, ekp, prior)
```
nefrathenrici, line 194: convenience plotting functions like these should probably be moved to ClimaCalibrate/EKP in the future, I think they are fine here for now.

experiments/calibration/PBS_calibration.pbs
```diff
@@ -17,4 +17,4 @@ export CLIMACOMMS_DEVICE="CUDA"
 export CLIMACOMMS_CONTEXT="SINGLETON"
 julia --project=.buildkite -e 'using Pkg; Pkg.instantiate(;verbose=true)'
```
nefrathenrici, line 19: This script currently won't work without changes in the julia script. Could you add a comment, something like this: 

> Note: Using this script requires changes to `addprocs` in the `experiments/calibration/run_calibration.jl`

nefrathenrici (approved):
This PR is a big improvement over the previous pipeline. I just had a few comments, I think someone who is more familiar with the land model should approve as well before merging.

docs/src/calibration.md
```diff
@@ -199,33 +201,160 @@ You would start the job with a command such as `qsub name_of_job_script` for PBS
...
+    spinup = Dates.Month(3),
+    nelements = (101, 15),
+    output_dir = "experiments/calibration/land_model",
+    rng_seed = 42,
+)
+```
+
+With the configuration above, a calibration is being done using the `swu`
+observation. The calibration will run for 10 iterations, and each iteration will
+use a minibatch size of 2. The start and end dates of the simulation is
```
kmdeck, line 230: is minibatch defined somewhere here?

docs/src/calibration.md
```diff
@@ -199,33 +201,160 @@ You would start the job with a command such as `qsub name_of_job_script` for PBS
...
+automatically determined by `sample_date_ranges`, `spinup`, and `extend`. The
+amount to spin up the simulation for is three months and the amount to run the
+simulation for past the dates in `sample_date_ranges` is also three months.
+Since the minibatch size is 2, the first iteration of the calibration will run
```
kmdeck, line 281: could you clarify this a bit more? Minibatch = N is in units of years? dos this relate to the number of elements in sample_date_ranges at all?

kmdeck: oh, i see this below

experiments/calibration/api.jl
```diff
@@ -0,0 +1,159 @@
...
+        output_dir::String
+        rng_seed::Int64
+    end
+
+A configuration struct for keeping track of multiple fields that are of interest
+to a user running calibration, or that are needed in multiple places (e.g., for
+ensemble members and generating observations).
+"""
+struct CalibrateConfig{SPINUP <: Dates.Period, EXTEND <: Dates.Period}
+    "The short names of the observations used for calibration"
```
kmdeck, line 21: these should match the short_names we use in the diagnostics?

experiments/calibration/api.jl
```diff
@@ -0,0 +1,159 @@
...
+        rng_seed::Int64
+        minibatch_size::Int64
+        n_iterations::Int64
+    end
+
+A configuration struct for keeping track of multiple fields that are of interest
+to a user running calibration, or that are needed in multiple places (e.g., for
+ensemble members and generating observations).
+"""
+struct CalibrateConfig{SPINUP <: Dates.Period, EXTEND <: Dates.Period}
```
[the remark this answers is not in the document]
juliasloan25: ok, sounds good

experiments/calibration/api.jl
```diff
@@ -0,0 +1,159 @@
...
+            "The start date ($start_date) should be before the end date ($end_date)",
+        )
+    end
+    issorted(sample_date_ranges) ||
+        error("The samples in $sample_date_ranges should be sorted")
+
+    minibatch_size > 0 ||
+        error("The minibatch size ($minibatch_size) should be positive")
+    n_iterations > 0 ||
+        error("The number of iterations ($n_iterations) should be positive")
```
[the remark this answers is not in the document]
juliasloan25: oops my comment was wrong - what you have is right

juliasloan25 (approved):
LGTM!

imreddyTeja (approved):

PR #1214, opened 2025-07-08
Close output writers at the end of the simulation

### LINKED ISSUE #1213: Close NetCDF writers
When the simulation ends, the NetCDF writers are never closed which leave the NetCDF files open.

ph-kev:
closes #1213 - This PR makes it, so the output writers are closed when the simulation crashes or ends.

Otherwise, there might be errors when accessing NetCDF files.

### DIFF
--- a/src/diagnostics/diagnostic.jl
+++ b/src/diagnostics/diagnostic.jl
@@ -150,6 +150,22 @@ function diagnostic_as_vectors(writer::DictWriter, diagnostic; layer = 1)
     return times, vector_layer_n
 end
 
+"""
+    close_output_writers(diagnostics)
+
+Close the output writers in the `diagnostics`, an iterable of
+`ClimaDiagnostics.ScheduledDiagnostic` or `nothing`.
+
+This function should be called at the end of every simulation.
+"""
+function close_output_writers(diagnostics)
+    isnothing(diagnostics) && return nothing
+    for diagnostic in diagnostics
+        close(diagnostic.output_writer)
+    end
+    return nothing
+end
+
 # Do you want to define more diagnostics? Add them here
 include("land_compute_methods.jl")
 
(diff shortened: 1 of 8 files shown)


src/diagnostics/diagnostic.jl
```diff
@@ -150,6 +150,19 @@ function diagnostic_as_vectors(writer::DictWriter, diagnostic; layer = 1)
     return times, vector_layer_n
 end
 
+"""
+    close_diagnostics(sim)
+
+Close the output writers in the diagnostics.
```
kmdeck, line 156: I wonder if this should take `diagnostics` as the argument. Then it can be used also in cases where we dont set up a land simulation.

and then in the other function, we can call close_output_writers(sim.diagnostics).

src/diagnostics/diagnostic.jl
```diff
@@ -150,6 +150,19 @@ function diagnostic_as_vectors(writer::DictWriter, diagnostic; layer = 1)
     return times, vector_layer_n
 end
 
+"""
+    close_diagnostics(diagnostics)
```
kmdeck, line 154: can update docstring function name

kmdeck (approved):
Thanks for adding this!

juliasloan25 (approved):
This is great, thank you Kevin! I think there are a couple more experiments we need to close diagnostics for:
- experiments/calibration/forward_model_bucket.jl
- experiments/calibration/forward_model_land.jl
- experiments/standalone/Bucket/global_bucket_temporalmap.jl
- experiments/longruns/soil.jl

ph-kev:
> This is great, thank you Kevin! I think there are a couple more experiments we need to close diagnostics for:
> 
>     * experiments/calibration/forward_model_bucket.jl
> 
>     * experiments/calibration/forward_model_land.jl
> 
>     * experiments/standalone/Bucket/global_bucket_temporalmap.jl
> 
>     * experiments/longruns/soil.jl

I don't think it is worth it to update the calibration code, since it is unmaintained right now. Also, for the first three files, all of those close the writers in their own way. The last file already close the writers, because I changed `ClimaLand.Simulations.solve!` to close the writers.

I can update the code to use the new function if you want me to.

PR #1211, opened 2025-07-07
split up parameters and remove ModelSetup

kmdeck:
## Purpose 
- Remove ModelSetup
- move domain related things to Domains.jl, and spatial parameters to individual component folders
- split up spatial parameter functions to return subsets relevant to individual parameterizations

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #1217, opened 2025-07-08
Fix texture norm and reset means for soil parameters

kmdeck:
## Purpose 
The branch used in #1211 was based on a old version of main which did not have an important bug fix in the computation of the soil composition parameters. When I rebased, this was lost(and some NaNs reintroduced!). This PR makes our soil spatial parameters match what we had before.

## To-do
Double check that NaNs are gone again

## Content

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@ ClimaLand.jl Release Notes
 
 main
 -------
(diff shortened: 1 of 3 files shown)


juliasloan25 (approved):
good catch!

PR #1218, opened 2025-07-09
Refactor Benchmarks, add NVTX

imreddyTeja:
## Purpose 
- previously, no cpu benchmarks were run, and the artifacts produced in buildkite were not uploaded by the pipeline
- add NVTX annotations

## To-do
Maybe reduce the number of NVTX annotations

## Content
- NVTX added as a dependency. Annotations added to cache updates, implicit tendencies, explicit tendencies, and boundary flux updates. Uses default naming and domains
- adds CPU runs to benchmark pipeline.
- reduce code duplication in the benchmarks. move shared functions to a benchmarking utils file
- add a soil only benchmark
- use the LandSimulation struct in benchmarks, update them to work with new LandParams
- Save integrated gpu profiling results to csv. Fix issue where profiler would stall and builkdite job would run  until its time limit. 
- Change nsys profiling to not profile compilation/setup. Enables additional NVTX callbacks for allocation, freeing memory, and inference.
- Make buildkite job only use 2 gpus at once
- Add a buildkite pipeline that is conditionally triggered based on changes. It only runs gpu benchmarks, and only for a single model per PR

Screenshot 2026-04-23 at 4 25 09 PM

----


experiments/benchmarks/benchmark_sim.jl
```diff
@@ -0,0 +1,169 @@
+# This script contains functions to assess performance, either via Nsight or by running the
+# model multiple times withj the integrated profiler and collecting statistics
```
kmdeck, line 2: "with"

experiments/benchmarks/benchmark_sim.jl
```diff
@@ -0,0 +1,169 @@
...
+using Dates
+import ClimaLand
+using ClimaComms
+using CSV
+
+function parse_commandline()
+    s = ArgParseSettings()
+    @add_arg_table s begin
+        "--profiler"
+        help = "Profiler option: nsight or flamegraph"
```
kmdeck, line 26: nsight or integrated?

experiments/benchmarks/benchmark_sim.jl
daverumph, line 2: Remove extraneous "j"

experiments/benchmarks/benchmark_sim.jl
```diff
@@ -0,0 +1,169 @@
...
+        @info "Done with timing benchmarks. Starting integrated profiling"
+        run_integrated_profiler(device, setup_simulation, outdir)
+        if get(ENV, "BUILDKITE_PIPELINE_SLUG", nothing) ==
+           "climaland-benchmark" && device isa ClimaComms.CUDADevice
+            if average_timing_s > PREVIOUS_GPU_TIME + std_timing_s
+                @info "Possible performance regression, previous average time was $(PREVIOUS_GPU_TIME)"
+            elseif average_timing_s < PREVIOUS_GPU_TIME - std_timing_s
+                @info "Possible significant performance improvement, please update PREVIOUS_GPU_TIME in $(@__DIR__)"
+            end
+            @testset "Performance" begin
```
kmdeck, line 56: this makes the timing a strict check, right? 

is this now something we want to run in order to merge into main?

experiments/benchmarks/benchmark_sim.jl
```diff
@@ -0,0 +1,169 @@
...
+
+# integrated profiling for CUDA
+function run_integrated_profiler(
+    device::ClimaComms.CUDADevice,
+    setup_simulation,
+    outdir,
+)
+    simulation = setup_simulation()
+    p = CUDA.@profile begin
+        for i in 1:20
```
kmdeck, line 100: we run this profiler for 20 steps, but the other external option for 3. should these be the same?

experiments/benchmarks/bucket.jl
```diff
@@ -1,77 +1,65 @@
...
-# This code also assesses performance, either via Nsight or by running the
-# model multiple times and collecting statistics for execution time and allocations
-# to make flame graphs. You can choose between the two
-# by requestion --profiler nsight or --profiler flamegraph. If not provided,
-# flamegraphs are created.
+# See "benchmark_sim.jl" for details on what is benchmarked
 
-# When run with buildkite on clima, without Nisght, this code also compares with the previous best time
-# saved at the bottom of this file
+# When run with buildkite on clima, without Nisght, this code compares with the previous
```
daverumph, line 10: Fix spelling of Nsight

experiments/benchmarks/snowy_land.jl
```diff
@@ -44,42 +36,38 @@ using ClimaLand.Soil
...
-        "--profiler"
-        help = "Profiler option: nsight or flamegraph"
-        arg_type = String
-        default = "flamegraph"
-    end
-    return parse_args(s)
-end
+######################################################################
+## This result is from a benchmark ran on an A100 on the clima cluster
+const PREVIOUS_GPU_TIME = 2.67
```
kmdeck, line 45: the previous best time was 0.67. do we understand the increase?

kmdeck (commented):
Very nice simplification/clean up of the benchmarking we do. I left pretty minor comments, but am wondering if we should require benchmarking (on GPU?) to merge, and if the increase in the benchmark time is expected

experiments/benchmarks/snowy_land.jl
imreddyTeja: Yes. I changed the simulation run for a day on gpu, rather than for 6 hours. This was to reduce the impact of any first call penalties.

daverumph (changes requested):
I found a couple of typos in comments.  I'm not familiar enough with the current testing and performance infrastructure to comment on the chosen settings.

petebachant (commented):
A few minor formatting comments. I'm not able to access the benchmarking project on Buildkite. Is that expected?

petebachant:
By the way, regarding our discussion on type annotations, someone told me last week they can actually make Julia code slower, but I have no idea if that's true.

.buildkite/target/pipeline.yml
```diff
@@ -20,89 +22,127 @@ steps:
       - "julia --project -e 'using Pkg; Pkg.status()'"
 
       - echo "--- Instantiate experiments"
-      - "julia --project=.buildkite -e 'using Pkg; Pkg.develop(;path=\".\"); Pkg.instantiate(;verbose=true)'"
+      - 'julia --project=.buildkite -e ''using Pkg; Pkg.develop(;path="."); Pkg.instantiate(;verbose=true)'''
```
daverumph: The parser can't possibly know about nested single quotes, so this must be wrong.

experiments/benchmarks/README.md
```diff
@@ -0,0 +1,81 @@
+# ClimaLand Benchmarks
+
+This folder contains scripts that setup and benchmark/profile certain ClimaLand models.
```
daverumph: setup -> set up

experiments/benchmarks/README.md
```diff
@@ -0,0 +1,81 @@
...
+## This result is from a benchmark ran with a single process on the clima cluster
+const PREVIOUS_CPU_TIME_S = 100.0
+######################################################################
+
+include("benchmark_utils.jl")
+
+context = ClimaComms.context()
+ClimaComms.init(context)
+device = ClimaComms.device()
+device_suffix = device isa ClimaComms.CPUSingleThreaded ? "cpu" : "gpu"
```
daverumph: Using CPUSingleThreaded for this test is incomplete, since we also have CPUMultiThreaded.  (Not future-proof.)

experiments/benchmarks/README.md
```diff
@@ -0,0 +1,81 @@
...
+by repeatedly running it. Then three steps of the simulation
+are profiled. The results are printed to stdout and saved as csv files.
+
+If ran on buildkite, the mean simulation
+time is compared to the previous best time for that device. If it is significantly slower,
+the test fails.
+
+### Usage with an external profiler
+
+If an external CUDA profiler is detected, the simulation isprofiled for three steps,
```
daverumph: isprofiled -> is profiled

experiments/benchmarks/bucket.jl
```diff
@@ -1,80 +1,59 @@
...
-        default = "flamegraph"
-    end
-    return parse_args(s)
-end
+using ClimaLand: PrescribedAtmosphere, PrescribedRadiativeFluxes
 
 const FT = Float64;
 
+######################################################################
+## This result is from a benchmark ran on an A100 on the clima cluster
```
daverumph: ran -> run, also below

experiments/benchmarks/snowy_land.jl
```diff
@@ -44,43 +29,35 @@ using ClimaLand.Soil
...
-    @add_arg_table s begin
-        "--profiler"
-        help = "Profiler option: nsight or flamegraph"
-        arg_type = String
-        default = "flamegraph"
-    end
-    return parse_args(s)
-end
+######################################################################
+## This result is from a benchmark ran on an A100 on the clima cluster
```
daverumph: ran -> run, also below

daverumph (changes requested):

.buildkite/short_perf/pipeline.yml
```diff
@@ -0,0 +1,106 @@
+# This pipeline is designed as an alternative to the full benchmarking pipeline.
+# It does not run cpu benchmarks, or generate nsys reports.
+# Benchmark tests are conditionally triggered by changes to specific files/directories
+# in comparison to the branch point from main.
+
+# If a file in src/simulations/ or src/shared_utilities is modified, then
+# both the snowy land an bucket model will be tested for performance regression.
```
daverumph: an -> and

experiments/benchmarks/benchmark_utils.jl
```diff
@@ -0,0 +1,250 @@
...
+"""
+    run_profiler(
+        setup_simulation::Function,
+        device::ClimaComms.CUDADevice,
+        outdir::AbstractString;
+        use_external_profiler = CUDA.Profile.detect_cupti(),
+    )
+
+Profile 3 steps of the simulation returned by `setup_simulation`. If `use_external_profiler`
+is true, the output should be handeled by the external profiler, Otherwise, the following are
```
daverumph: handeled --> handled

daverumph (changes requested):
It looks good overall.  My requested changes are all typos in comments, I think.  If you fix those and repush, I'll re-review quickly.

daverumph (approved):

src/integrated/soil_snow_model.jl
```diff
@@ -351,7 +352,7 @@ integrated land surface models; this computes and returns the net
 energy and water flux at the surface of the soil for use as boundary
 conditions, taking into account the presence of snow on the surface.
 """
-function soil_boundary_fluxes!(
+NVTX.@annotate function soil_boundary_fluxes!(
```
kmdeck, line 355: any reason why these function are annotated but not others? also in land.jl

src/standalone/Soil/rre.jl
```diff
@@ -366,7 +366,7 @@ variables `p.soil.variable` in place.
 This has been written so as to work with Differential Equations.jl.
 """
 function ClimaLand.make_update_aux(model::RichardsModel)
```
kmdeck, line 368: Maybe also: https://github.com/CliMA/ClimaLand.jl/blob/3ea0d7c1caf79ef2268f48d8562f108e7b50489f/src/standalone/Soil/energy_hydrology.jl#L427

src/standalone/Vegetation/Canopy.jl
```diff
@@ -1267,7 +1268,7 @@ function ClimaLand.make_compute_jacobian(canopy::CanopyModel)
         x -> make_compute_jacobian(getproperty(canopy, x), canopy),
         components,
     )
-    function compute_jacobian!(W, Y, p, dtγ, t)
+    NVTX.@annotate function compute_jacobian!(W, Y, p, dtγ, t)
```
kmdeck, line 1271: Maybe also
https://github.com/CliMA/ClimaLand.jl/blob/3ea0d7c1caf79ef2268f48d8562f108e7b50489f/src/standalone/Vegetation/Canopy.jl#L1208
https://github.com/CliMA/ClimaLand.jl/blob/3ea0d7c1caf79ef2268f48d8562f108e7b50489f/src/standalone/Vegetation/Canopy.jl#L1222

experiments/benchmarks/snowy_land.jl
```diff
@@ -104,198 +83,60 @@ function setup_prob(
...
-    ClimaTimeSteppers.step!(integrator)
-    ClimaTimeSteppers.step!(integrator)
-else
-    @error("Profiler choice not supported.")
+    simulation = ClimaLand.Simulations.LandSimulation(
+        start_date,
+        stop_date,
+        Δt,
+        land;
+        set_ic!,
```
kmdeck: I think we could use a default set_ic! function - maybe 
https://github.com/CliMA/ClimaLand.jl/blob/3ea0d7c1caf79ef2268f48d8562f108e7b50489f/src/simulations/initial_conditions.jl#L775

experiments/benchmarks/soil.jl
```diff
@@ -0,0 +1,78 @@
...
+
+function setup_soil()
+    toml_dict = LP.create_toml_dict(FT)
+    start_date = DateTime(2008)
+    duration = Hour(6)
+    stop_date = start_date + duration
+    Δt = 450.0
+    nelements = (101, 101, 15)
+    domain = ClimaLand.Domains.global_box_domain(FT; context, nelements)
+    forcing = ClimaLand.prescribed_forcing_era5(
```
kmdeck, line 55: will these only run on clima? if not, we should set "use_lowres_forcing = true". But it's better if we don't need to do that, and can make this as similar as possible to our models runs

docs/Manifest-v1.12.toml
```diff
@@ -456,7 +456,7 @@ version = "0.1.1"
     CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 
 [[deps.ClimaLand]]
-deps = ["ClimaComms", "ClimaCore", "ClimaDiagnostics", "ClimaParams", "ClimaTimeSteppers", "ClimaUtilities", "Dates", "DocStringExtensions", "Insolation", "Interpolations", "LazyArtifacts", "LazyBroadcast", "LinearAlgebra", "NCDatasets", "RootSolvers", "StaticArrays", "SurfaceFluxes", "Thermodynamics"]
+deps = ["ClimaComms", "ClimaCore", "ClimaDiagnostics", "ClimaParams", "ClimaTimeSteppers", "ClimaUtilities", "Dates", "DocStringExtensions", "Insolation", "Interpolations", "LazyArtifacts", "LazyBroadcast", "LinearAlgebra", "NCDatasets", "NVTX", "RootSolvers", "StaticArrays", "SurfaceFluxes", "Thermodynamics"]
```
kmdeck, line 459: not needed for Julia 1.10 Manifest?

experiments/benchmarks/soil.jl
imreddyTeja: This was intended to be run on clima, but  I can change it to use the lowres if it will be used outside the cluster. Otherwise I will just put a note at the top of the file

docs/Manifest-v1.12.toml
imreddyTeja: done

experiments/benchmarks/soil.jl
kmdeck: It's ok to keep as is and just run on Clima. there is a benefit (I think) in case reading the forcing becomes a dominant source of the sypd in the future for some reason

kmdeck (approved):

PR #1219, opened 2025-07-09
Remove CreateParams extension, ClimaParams in Project.toml, expose roughness and emissivity

kmdeck:
## Purpose 
This PR adds ClimaParams to the ClimaLand environment. 

Because of that, we can remove the extension CreateParamsExtension and move that into src.

To make it easier to access default parameter values (to use as kwargs), I made an interface function `get_default_parameter` which returns the default values (from the toml dict). These are now the default options in the Soil EnergyHydrology constructor.

This lets us use the same underlying EnergyHydrologyParameters constructor within the new model constructor (with many defaults from ClimaParams) while choosing roughness length and emissivity from outside the new model constructor. 

Note to @juliasloan25 - I still opted to keep the earth_param_set a required argument. We use it in a lot of places, and it makes me a bit nervous to have a default. what if someone overrides the default when setting up the forcing, but not in creating the soil model? 
## Content

----


nefrathenrici (approved):
This looks good to me, the parameters are simpler without the extension.
Do you think simple tests for the kwarg override behavior would be useful?

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -75,6 +75,193 @@ end
 
 Base.broadcastable(ps::EnergyHydrologyParameters) = tuple(ps)
 
+
+### Interfacing with ClimaParams
+
+SoilSurfaceRoughness(::Type{FT}) where {FT <: AbstractFloat} =
+    SoilSurfaceRoughness(CP.create_toml_dict(FT))
```
juliasloan25, line 82: Usually camel case naming is used for types/constructors in Julia ([style guide](https://github.com/CliMA/ClimaCoupler.jl/pull/1318)). Since these functions return NamedTuples rather than objects, it would probably be good to rename them - maybe just to `soil_surface_roughness` and `soil_emissivity`

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -75,6 +75,193 @@ end
...
+    name_map = (;
+        :soil_momentum_roughness_length => :z_0m,
+        :soil_scalar_roughness_length => :z_0b,
+    )
+    parameters = CP.get_parameter_values(td, name_map, "Land")
+    FT = CP.float_type(td)
+    return (; parameters...)
+end
+
+SoilEmissivity(::Type{FT}) where {FT <: AbstractFloat} =
```
juliasloan25, line 94: Can you add short docstrings please?

src/standalone/Vegetation/photosynthesis.jl
```diff
@@ -271,3 +271,105 @@ Base.broadcastable(m::FarquharParameters) = tuple(m)
...
+            error(
+                "is_c3 has maximum of $(maximum(is_c3)). is_c3 should be between 0 and 1",
+            )
+        end
+        if minimum(is_c3) < 0
+            error(
+                "is_c3 has minimum of $(minimum(is_c3)). is_c3 should be between 0 and 1",
+            )
+        end
+    =#
```
juliasloan25, line 360: Should we remove this code if it's just commented out?

src/standalone/Snow/Snow.jl
```diff
@@ -332,6 +332,36 @@ function SnowParameters{FT}(
     )
 end
 
+## For interfacing with ClimaParams
+
+SnowParameters(::Type{FT}, Δt; kwargs...) where {FT <: AbstractFloat} =
```
juliasloan25, line 337: Can you add a docstring for this like you have for the canopy param constructors?

juliasloan25 (approved):
This looks good! I think it's fine to keep `earth_param_set` as a required argument - I agree we wouldn't have any way to enforce consistency across models

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -75,6 +75,193 @@ end
...
+    SoilEmissivity(CP.create_toml_dict(FT))
+
+function SoilEmissivity(td::CP.AbstractTOMLDict)
+    name_map = (; :emissivity_bare_soil => :emissivity)
+    parameters = CP.get_parameter_values(td, name_map, "Land")
+    FT = CP.float_type(td)
+    return (; parameters...)
```
juliasloan25, line 101:
```suggestion
function get_land_parameter(FT, param_name)
    td = CP.create_toml_dict(FT)
    return td[param_name]
```

juliasloan25: to use as default values in model constructors

src/standalone/Soil/energy_hydrology.jl
kmdeck: no longer needed

src/standalone/Snow/Snow.jl
```diff
@@ -332,6 +332,66 @@ function SnowParameters{FT}(
...
+# TOML Dictionary:
+import ClimaParams as CP
+toml_dict = CP.create_toml_dict(Float32);
+ClimaLand.Canopy.SnowParameters(toml_dict, Δt; ϵ_snow = Float32(0.99), Ksat = Float32(1e-4))
+```
+"""
+SnowParameters(::Type{FT}, Δt; kwargs...) where {FT <: AbstractFloat} =
+    SnowParameters(CP.create_toml_dict(FT), Δt; kwargs...)
+
+function SnowParameters(toml_dict::CP.AbstractTOMLDict, Δt; kwargs...)
```
juliasloan25, line 366: Could you explain in the docstring what `kwargs` is here? Does it correspond to the values of `name_map` in the function?

src/shared_utilities/Parameters.jl
```diff
@@ -57,4 +62,86 @@ thermodynamic_parameters(ps::ALP) = ps.thermo_params
 surface_fluxes_parameters(ps::ALP) = ps.surf_flux_params
 insolation_parameters(ps::ALP) = ps.insol_params
 
+
+# interfacing with ClimaParams
+LandParameters(::Type{FT}) where {FT <: AbstractFloat} =
+    LandParameters(CP.create_toml_dict(FT))
+
+function LandParameters(toml_dict::CP.AbstractTOMLDict)
```
juliasloan25, line 70: docstring please

juliasloan25 (approved):
This looks good to me! To recap the new structure:
- At the Parameters struct level, we take in some parameters explicitly (e.g. porosity for EnergyHydrology), and also kwargs which corresponds to all other parameters stored in this struct.
- At the Model struct level, we take in defaults for all parameters - either via a function defined for this (e.g. soil_vangenuchten_parameters ) or by retrieving the default parameter from the ClimaParams toml dict. Inside the model constructor, these all get passed into the Parameters struct, where they overwrite all (?) kwargs.
- If a user wants to overwrite a parameter, they can do so by passing a kwarg to the Model constructor using the ClimaLand naming, or to the Parameters constructor using the ClimaParams naming.

The only things that stand out to me as potential points of confusion are:
- What does kwargs entail for each Parameter struct? Is it any parameters contained in the struct that aren't explicitly required by the constructor?
- It may be confusing for users to keep track of both the ClimaLand and the ClimaParams name for each parameter. Hopefully they'll be able to interface at the Model struct level, and the Parameter struct constructors can be mostly internal - we could make a note pointing users towards the Model constructors in the Parameter constructor docstrings. We can also provide a link to the ClimaParams.jl repo for people to see the complete name mapping.

kmdeck:
> This looks good to me! To recap the new structure:
> 
>     * At the Parameters struct level, we take in some parameters explicitly (e.g. porosity for EnergyHydrology), and also kwargs which corresponds to all other parameters stored in this struct.
> 
>     * At the Model struct level, we take in defaults for all parameters - either via a function defined for this (e.g. soil_vangenuchten_parameters ) or by retrieving the default parameter from the ClimaParams toml dict. Inside the model constructor, these all get passed into the Parameters struct, where they overwrite all (?) kwargs.
> 
>     * If a user wants to overwrite a parameter, they can do so by passing a kwarg to the Model constructor using the ClimaLand naming, or to the Parameters constructor using the ClimaParams naming.
> 
> 
> The only things that stand out to me as potential points of confusion are:
> 
>     * What does kwargs entail for each Parameter struct? Is it any parameters contained in the struct that aren't explicitly required by the constructor?
> 
>     * It may be confusing for users to keep track of both the ClimaLand and the ClimaParams name for each parameter. Hopefully they'll be able to interface at the Model struct level, and the Parameter struct constructors can be mostly internal - we could make a note pointing users towards the Model constructors in the Parameter constructor docstrings. We can also provide a link to the ClimaParams.jl repo for people to see the complete name mapping.

Almost! A couple comments:
>     * At the Model struct level, we take in defaults for all parameters - either via a function defined for this (e.g. soil_vangenuchten_parameters ) or by retrieving the default parameter from the ClimaParams toml dict. Inside the model constructor, these all get passed into the Parameters struct, where they overwrite all (?) kwargs.

In this case, at the Model struct level, I have only exposed (so far) parameters that I think someone will want to change. Eventually we could expose all of them, if we think we need to. In principle someone could overwrite them by changing the toml dict (though as we discussed before, this is not clear how to do at all). At the Parameter struct level, you have the option to change any of the values using kwargs.

>     * What does kwargs entail for each Parameter struct? Is it any parameters contained in the struct that aren't explicitly required by the constructor?

Yes, I think so. this was how Nat set it up before. I agree it is challenging because it requires the user to know what options are available. One idea which I kind of like, but is a bigger change, is to change these functions to get the defaults using the new `get_default_parameter` function for each kwarg explicitly (like we started to do in the model struct level), and get rid of the name map etc internally. Then it should be clear what is available.

>     * If a user wants to overwrite a parameter, they can do so by passing a kwarg to the Model constructor using the ClimaLand naming, or to the Parameters constructor using the ClimaParams naming.

In all cases, they use the ClimaLand naming. The Parameters constructor takes in e.g. (; climaland_name = X) and overwrites the value obtained via the name map (climaparams_name => climaland_name). climaparams names are only used internally to look up the value, it is then "renamed" to the climaland name.

>     * It may be confusing for users to keep track of both the ClimaLand and the ClimaParams name for each parameter. 

This should not be necessary

PR #1226, opened 2025-07-13
Rollback GeoMakie to 0.7.12 and add upper compat bound

imreddyTeja:
GeoMakie 0.7.13 causes a strange stackoverflow error because of recursion during the long run plots.

## Content

----

### DIFF
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.5"
+julia_version = "1.11.6"
 manifest_format = "2.0"
-project_hash = "c057205b5dfa45d173ab89f45dff9c1d2ab4a915"
+project_hash = "acaaffaaac229b13d375a67bbaf146c0a0ce520e"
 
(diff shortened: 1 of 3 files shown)


.buildkite/Project.toml
```diff
@@ -51,4 +51,5 @@ ClimaDiagnostics = "0.2.13"
 ClimaTimeSteppers = "0.7, 0.8"
 EnsembleKalmanProcesses = "2.4.1"
 Flux = "0.15"
+GeoMakie = "< 0.7.13"
```
imreddyTeja, line 54: Remember to add a comment explaining why (if possible to leave comments)

.buildkite/Manifest-v1.11.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.5"
+julia_version = "1.11.4"
```
ph-kev, line 3: I'm not sure how important this is, but the `julia_version` changes from "1.11.5" to "1.11.4". I am guessing this is because you were using Julia 1.11.4 instead of 1.11.5 when updating the manifest.

ph-kev (approved):

PR #1233, opened 2025-07-16
Add convenience constructors for canopy model components

juliasloan25:
## Purpose 
Adds constructors at the model level for the following canopy components:
- energy
- hydraulics
- autotrophic respiration
- radiative transfer

The components still remaining after this PR are:
- photosynthesis
- stomatal conductance

## Content

----


src/standalone/Vegetation/Canopy.jl
```diff
@@ -47,6 +48,202 @@ include("./canopy_parameterizations.jl")
...
+    domain;
+    start_date = Dates.DateTime(2008, 1, 1),
+    LAIfunction::TVI = ClimaLand.prescribed_lai_modis(
+        ClimaLand.Artifacts.modis_lai_single_year_path(;
+            year = Dates.year(start_date),
+        ),
+        domain.space.surface,
+        start_date,
+    ),
+    SAI::FT = FT(0),
```
kmdeck, line 139: we should also expose (but set as default)
n_stem = 0, n_leaf = 1

I think we also need the canopy height as a required argument (and then need to compute compartment surface and midpoints in kwargs appropriately for n_stem = 0, n_leaf = 1, h_canopy = X.

src/standalone/Vegetation/Canopy.jl
```diff
@@ -47,6 +48,202 @@ include("./canopy_parameterizations.jl")
...
+    S_s::FT = FT(1e-2 * 0.0098), # m3/m3/MPa to m3/m3/m
+    conductivity_model = Weibull{FT}(
+        K_sat = FT(7e-8),
+        ψ63 = FT(-4 / 0.0098),
+        c = FT(4),
+    ),
+    retention_model = LinearRetentionCurve{FT}(a = FT(0.05 * 0.0098)),
+    rooting_depth = ClimaLand.Canopy.clm_rooting_depth(domain.space.surface),
+) where {FT <: AbstractFloat, TVI <: AbstractTimeVaryingInput}
+    ai_parameterization = PrescribedSiteAreaIndex{FT}(LAIfunction, SAI, RAI)
```
kmdeck, line 151: One other thought is to expose this instead of SAI/RAI/LAI, so do:
```
PlantHydraulicsModel{FT}(domain, forcing, h_canopy; SAI = 0.0, RAI = 1.0, ai_parameterization = PrescribedSiteAreaIndex{FT}(forcing.LAI, SAI, RAI)
```

I think ideally the forcing is passed in, because then it is clear that the model only depends on start date via forcing. This is how we pass in atmos and radiative forcing also (in a named tuple, required arg, `forcing`) for soil and snow.

I think then we can do something like:
```
CanopyModel(forcing, domain; hydraulics = PlantHydraulicsModel(FT, domain, forcing...)
```

oh, and one other thing is to match the argument order here: https://github.com/CliMA/ClimaLand.jl/blob/b006c1dca68131a4bacfaaa3e174011a8f50434c/src/standalone/Soil/Soil.jl#L232

so we should remove {FT} and pass as the first argument also (for all of these)

src/standalone/Vegetation/Canopy.jl
```diff
@@ -47,6 +48,202 @@ include("./canopy_parameterizations.jl")
...
+    return PlantHydraulicsModel{FT, typeof(parameters)}(parameters)
+end
+
+## Radiative transfer models
+"""
+    TwoStreamModel{FT}(
+        domain;
+        radiation_parameters = clm_canopy_radiation_parameters(domain.space.surface),
+        ϵ_canopy::FT = LP.get_default_parameter(FT, :canopy_emissivity),
+        λ_γ_PAR::FT = LP.get_default_parameter(FT, :wavelength_per_PAR_photon),
```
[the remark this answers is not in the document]
kmdeck: (also in Beer Lambert)

src/standalone/Vegetation/Canopy.jl
```diff
@@ -47,6 +48,202 @@ include("./canopy_parameterizations.jl")
...
+- albedo and transmissitivy in PAR and NIR bands
+- leaf angle distribution G function parameter χl
+
+Canopy emissivity and wavelength per PAR photon are currently treated
+as constants; these can be passed in as Floats by kwarg.
+Otherwise the default values from ClimaParams.jl are used.
+
+The number of layers in the canopy is set by `n_layers`, which defaults to 20.
+"""
+function TwoStreamModel{FT}(
```
kmdeck, line 188: likewise for the {FT} -> (FT,)

src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -173,7 +173,7 @@ function PlantHydraulicsParameters(;
     S_s::FT,
     conductivity_model,
     retention_model,
-    rooting_depth::Union{Nothing, FT, ClimaCore.Fields.Field},
+    rooting_depth::Union{FT, ClimaCore.Fields.Field},
```
kmdeck, line 176: good catch, not sure why we allowed Nothing!

src/standalone/Vegetation/Canopy.jl
juliasloan25: It looks like in other places we set `h_stem` and `h_leaf`, and compute `h_canopy=hstem + h_leaf`. Should we do that here?

kmdeck: I dont think we need to do that here, but when we need it, we can access h_stem and h_leaf and compute it.

kmdeck: what you have looks good!

src/standalone/Vegetation/Canopy.jl
juliasloan25: I tried changing this PR to pass FT as a regular argument, but it produces a precompilation failure (see below). I think this is because of a type inference failure with many types in the constructor depending on FT (e.g. `conductivity_model = Weibull{FT}(...)`), but FT not being known at compile time.

It seems like I need to pass FT as a type parameter for PlantHydraulicsModel. I'd prefer to do it for all the canopy components to keep them consistent among themselves, then update Soil and Snow to match them next.

```
without type parameter:
Precompiling ClimaLand...
Info Given ClimaLand was explicitly requested, output will be shown live 
ERROR: LoadError: TypeError: in TypeVar, in upper bound, expected Type, got a value of type typeof(ClimaLand.Canopy.PlantHydraulicsModel)
Stacktrace:
 [1] TypeVar(n::Any, ub::Any)
   @ Core ./boot.jl:297
 [2] top-level scope
   @ ~/clima/ClimaLand.jl/src/standalone/Vegetation/Canopy.jl:611
 [3] include(mod::Module, _path::String)
   @ Base ./Base.jl:562
 [4] include(x::String)
   @ ClimaLand ~/clima/ClimaLand.jl/src/ClimaLand.jl:1
 [5] top-level scope
   @ ~/clima/ClimaLand.jl/src/ClimaLand.jl:402
 [6] include
   @ ./Base.jl:562 [inlined]
 [7] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2881
 [8] top-level scope
   @ stdin:6
in expression starting at /Users/jsloan/clima/ClimaLand.jl/src/standalone/Vegetation/Canopy.jl:1
in expression starting at /Users/jsloan/clima/ClimaLand.jl/src/ClimaLand.jl:1
in expression starting at stdin:6
  ✗ ClimaLand
  0 dependencies successfully precompiled in 6 seconds. 253 already precompiled.

ERROR: The following 1 direct dependency failed to precompile:

ClimaLand 
```

src/standalone/Vegetation/Canopy.jl
```diff
@@ -47,6 +47,278 @@ include("./canopy_parameterizations.jl")
...
+
+Creates a FarquharModel using default parameters of type FT.
+
+The following default parameters are used:
+- Vcmax25 = 5e-5 (mol m^-2 s^-1) - maximum carboxylation rate at 25C
+- sc = 5e-6 (Pa^{-1}) - sensitivity to low water pressure in the moisture stress factor [Tuzet et al. (2003)]
+- pc = -2e6 (Pa) - reference water pressure for the moisture stress factor [Tuzet et al. (2003)]
+"""
+function FarquharModel{FT}(;
+    is_c3::MECH,
```
kmdeck, line 103: I think we would also pass the domain here, and then get Vcmax25 and is_c3 from: https://github.com/CliMA/ClimaLand.jl/blob/04effc88b7cb9e1c3b2262e5be198b6782d03a95/src/standalone/Vegetation/spatially_varying_parameters.jl#L147

kmdeck (approved):
this looks good! I just had two comments.

One other thought - when we run single column runs, and pass the domain (now with lat/lon!), we can still use the spatially varying parameters, and extract the values at that lat/lon (ideally with the same functions, like clm_canopy_radiation_parameters(space)? thought maybe we need another method for the point space, not sure.

The other thought is that if someone wanted to provide scalars for these variables by hand, it's a bit odd to pass the domain and not use it, but it's not that big of a deal, I dont think. It may be that in that case they use the other constructor for the parameters and make the model from that, which I think is fine (we can document this).

src/standalone/Vegetation/Canopy.jl
```diff
@@ -47,6 +47,289 @@ include("./canopy_parameterizations.jl")
...
+
+The following additional default parameters are used:
+- sc = 5e-6 (Pa^{-1}) - sensitivity to low water pressure in the moisture stress factor [Tuzet et al. (2003)]
+- pc = -2e6 (Pa) - reference water pressure for the moisture stress factor [Tuzet et al. (2003)]
+"""
+function FarquharModel{FT}(
+    domain;
+    photosynthesis_parameters = clm_photosynthesis_parameters(
+        domain.space.surface,
+    ),
```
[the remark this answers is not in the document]
kmdeck: good point. we can also split up the `clm_photosynthesis_parameters` into two functions, one for c3, and one forVcmax25?

PR #1238, opened 2025-07-17
Move forcing script from fluxnet experiments to src

kmdeck:
## Purpose 
This moves the content of met_drivers_fluxnet and data_tools into src, with some cleanup and alteration of those functions. 

Some notable changes:
- we use the interpolating ability of TVI to handle gaps in data, rather than filling missing values by hand, except for in the plotting, which we can revisit in another PR.
- remove metdrivers fluxnet script in favor of prescribed_forcing_fluxnet function in a new extension
- set IC is a new function in the extension
- get comparison data is in the new extension, this is used in plotting. 

## To-do
Future PR: Move plotting to LandSimulationVisualization, clean that up

## Content

----


experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -21,19 +21,17 @@ import ClimaLand.Parameters as LP
...
-include(joinpath(climaland_dir, "experiments/integrated/fluxnet/data_tools.jl"))
 include(joinpath(climaland_dir, "experiments/integrated/fluxnet/plot_utils.jl"))
-
-# Read in the site to be run from the command line
-if length(ARGS) < 1
-    error("Must provide site ID on command line")
-end
-
-site_ID = ARGS[1]
+site_ID = "US-MOz"
```
juliasloan25, line 34: We should revert this before merging

Project.toml
```diff
@@ -53,8 +56,10 @@ ClimaTimeSteppers = "0.8"
 ClimaUtilities = "0.1.24"
 DataFrames = "1.4"
 Dates = "1"
+DelimitedFiles = "1"
```
juliasloan25, line 59: I thought these packages would go under `weakdeps` rather than `compat` since they're used by the extension. I don't find very good Julia documentation about how to set up package extensions, but I did find this example repo someone made: https://github.com/pebeto/julia_extensions_example/tree/main

ext/FluxnetSimulations.jl
```diff
@@ -0,0 +1,308 @@
...
+    columns = driver_data[1, :]
+    labels = ("TA_F", "VPD_F", "PA_F", "P_F", "WS_F", "LW_IN_F", "SW_IN_F")
+    indices = [findfirst(columns .== label) for label in labels]
+
+    if any(indices .== nothing)
+        nothing_id = findall(indices .== nothing)
+        @error("$(labels[nothing_id]) is missing in the data, but required.")
+    end
+
+    index_map = Dict(zip(labels, indices))
```
juliasloan25, line 56: Can you add some comments for what's being done here? It's not very clear to me why we need this

ext/FluxnetSimulations.jl
```diff
@@ -0,0 +1,308 @@
...
+    LOCAL_DATETIME = DateTime.(format.(driver_data[2:end, 1]), "yyyymmddHHMM")
+    UTC_DATETIME = LOCAL_DATETIME .+ Dates.Hour(hour_offset_from_UTC)
+    DATA_DT = Second(LOCAL_DATETIME[2] - LOCAL_DATETIME[1]).value # seconds
+    seconds_since_start_date =
+        [Second(UTC - start_date).value for UTC in UTC_DATETIME]
+
+    atmos_T_data = driver_data[2:end, index_map["TA_F"]]
```
juliasloan25, line 63: Can you add a note above this section that we don't care about missing data because it'll be handled by the TVI?

juliasloan25: Actually, can the TVI handle values of `Missing` in the data array, or do we have to filter them out like this?

ext/FluxnetSimulations.jl
```diff
@@ -0,0 +1,308 @@
...
+        seconds_since_start_date[not_missing_mask],
+        Float64.(atmos_T_data[not_missing_mask]) .+ 273.15,
+    )
+
+    atmos_P_data = driver_data[2:end, index_map["PA_F"]]
+    not_missing_mask = .~ismissing.(atmos_P_data)
+    atmos_P = TimeVaryingInput(
+        seconds_since_start_date[not_missing_mask],
+        Float64.(atmos_P_data[not_missing_mask]) .* 1000,
+    )
```
juliasloan25, line 75: This pattern is duplicated a lot. We could move it into a function, e.g.:
```suggestion
function data_to_TVI(driver_data, varname, seconds_since_start_date)
    atmos_data = driver_data[2:end, index_map[varname]]
    not_missing_mask = .~ismissing.(atmos_data)
    return TimeVaryingInput(
        seconds_since_start_date[not_missing_mask],
        Float64.(atmos_data[not_missing_mask]) .* 1000,
    )
    
data_to_TVI(driver_data, "PA_F, seconds_since_start_date)
```
and then we can call that function for all the variables

ext/FluxnetSimulations.jl
```diff
@@ -0,0 +1,308 @@
...
+)
+    fluxnet_csv_path = ClimaLand.Artifacts.experiment_fluxnet_data_path(site_ID)
+    driver_data = readdlm(fluxnet_csv_path, ',')
+    columns = driver_data[1, :]
+    labels = ("SWC_F_MDS_1", "TS_F_MDS_1", "TA_F")
+    indices = [findfirst(columns .== label) for label in labels]
+    index_map = Dict(zip(labels, indices))
+
+    LOCAL_DATETIME = DateTime.(format.(driver_data[2:end, 1]), "yyyymmddHHMM")
+    UTC_DATETIME = LOCAL_DATETIME .+ Dates.Hour(hour_offset_from_UTC)
```
juliasloan25, line 243:
```suggestion
    local_datetime = DateTime.(format.(driver_data[2:end, 1]), "yyyymmddHHMM")
    utc_datetime = local_datetime .+ Dates.Hour(hour_offset_from_UTC)
```
we should use the normal naming convention since these arent `const`s

ext/FluxnetSimulations.jl
```diff
@@ -0,0 +1,308 @@
...
+    if index_map["TS_F_MDS_1"] isa Nothing
+        atmos_T_data = driver_data[2:end, index_map["TA_F"]]
+        not_missing_mask = .~ismissing.(atmos_T_data)
+        idx_start = argmin(abs.(UTC_DATETIME[not_missing_mask] .- start_date))
+        T_0 = atmos_T_data[not_missing_mask][idx_start] + 273.15
+    else
+        soil_T_data = driver_data[2:end, index_map["TS_F_MDS_1"]]
+        not_missing_mask = .~ismissing.(soil_T_data)
+        idx_start = argmin(abs.(UTC_DATETIME[not_missing_mask] .- start_date))
+        T_0 = soil_T_data[not_missing_mask][idx_start] + 273.15
```
juliasloan25, line 265: I think a function will help here too, and we can use it for the else and if blocks above too:
```suggestion
function get_var_data(driver_data, varname, utc_datetime, start_date, preprocess_func)
        data = driver_data[2:end, index_map[varname]]
        not_missing_mask = .~ismissing.(data)
        idx_start = argmin(abs.(utc_datetime[not_missing_mask] .- start_date))
        return preprocess_func.(data[not_missing_mask][idx_start])
    end
    celsius_to_kelvin(x) = x + 273.15
    T_0 = get_var_data(driver_data, "TS_F_MDS_1", utc_datetime, start_date, celsius_to_kelvin)
```

ext/FluxnetSimulations.jl
```diff
@@ -0,0 +1,308 @@
+module FluxnetSimulations
```
juliasloan25, line 1: I think a good structure would be to have a main module file but split up the code into different files, like we do for e.g. the Soil module. So the code currently in here would go into something like `ext/FluxnetSimulations.jl/data_cleaning.jl`, and then we include that from this module.

Then Thanhthanh's `get_parameter` and `get_domain_info` functions can go into another file, etc.

Note that even if we do this we're still moving away from the exp/integrated/fluxnet include structure because everything is packaged into functions and in a module :)

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -201,44 +206,7 @@ land = LandModel{FT}(;
...
-    Y.canopy.hydraulics.ϑ_l.:($i) .=
-        augmented_liquid_fraction.(plant_ν, S_l_ini[i])
-end
-
-Y.canopy.energy.T = drivers.TA.values[1 + Int(round(t0 / DATA_DT))] # Get atmos temperature at t0
-
-Y.snow.S .= 0.0
-Y.snow.S_l .= 0.0
-Y.snow.U .= 0.0
+FluxnetSimulations.set_fluxnet_ic!(Y, site_ID, start_date, time_offset, land)
```
juliasloan25, line 209: much nicer!!

Project.toml
ph-kev: You should be able to add a weak dependency with `Pkg.add(name_of_package, target=:weakdeps)`.

ext/FluxnetSimulations.jl
kmdeck: I can add the comments if we dont switch to DataFrames

ext/FluxnetSimulations.jl
kmdeck: I actually need to double check. I think sometimes it is Missing, and sometimes missing data is marked by -9999 or something

TVI cannot handle missing data marked by Missing

ext/FluxnetSimulations.jl
kmdeck: nice! will add

(Note - TVI for spatial fields does take in a preprocess function, but not for 0D. :( )

Project.toml
kmdeck: ^^ This is what I did (then package name and uuid is listed under weak deps). However the compat is separate from that.

We are not required to list compats for weak dependencies, but one of the Aqua tests checks that (we can disable it if we want). If we do want to list compats, though, there is not weak_compats - all compats are in a single list.

Our extensions have added compat in the past (Neural Snow - Flux, LandSimulationVisualization (very recently) - Poppler, Printf, etc)

ext/FluxnetSimulations.jl
kmdeck: I split the code up but can do so more if we prefer

ext/FluxnetSimulations.jl
kmdeck: I added this

ext/FluxnetSimulations.jl
kmdeck: I added a comment in the corresponding location in the new code

ext/FluxnetSimulations.jl
kmdeck: I added some comments

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -63,13 +61,20 @@ include(
...
 )
-
-include(
-    joinpath(
-        climaland_dir,
-        "experiments/integrated/fluxnet/met_drivers_FLUXNET.jl",
-    ),
-)
+start_date = DateTime(2010) + Hour(time_offset)
+(; atmos, radiation, LAI, maxLAI, DATA_DT) =
```
kmdeck, line 65: I need to get maxLAI and DATA_DT in another way.

Ideally just open file once and get the columns we need (transformed)?

ext/fluxnet_sims/data_processing.jl
```diff
@@ -0,0 +1,87 @@
...
+)
+    first_row = skip_header ? 2 : 1
+    var_ids = [column_name_map[varname] for varname in varnames]
+    var_data = driver_data[first_row:end, var_ids]
+    # The time varying input object interpolates over gaps
+    # as needed, so we remove data that is marked as missing here
+    t, v = mask_data(time_in_seconds, var_data)
+    return TimeVaryingInput(
+        Float64.(t),
+        Float64.(preprocess_func.(eachcol(v)...)),
```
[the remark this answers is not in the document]
kmdeck: Right now the data is read in as Any type. Maybe because the data is a mix of Int (-9999) and float values. 

Ill try without it

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -21,19 +21,17 @@ import ClimaLand.Parameters as LP
 import ClimaUtilities.OutputPathGenerator: generate_output_path
 using ClimaDiagnostics
 using ClimaUtilities
+using DelimitedFiles
+using Format
+FluxnetSimulations =
+    Base.get_extension(ClimaLand, :FluxnetSimulations).FluxnetSimulations;
+
```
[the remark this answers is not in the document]
kmdeck: this works! much nicer

.buildkite/Manifest-v1.11.toml
```diff
@@ -481,9 +481,10 @@ deps = ["ClimaComms", "ClimaCore", "ClimaDiagnostics", "ClimaParams", "ClimaTime
 path = ".."
 uuid = "08f4d4ce-cf43-44bb-ad95-9d2d5f413532"
 version = "0.17.1"
-weakdeps = ["BSON", "CSV", "CairoMakie", "ClimaAnalysis", "DataFrames", "Flux", "GeoMakie", "HTTP", "Poppler_jll", "Printf", "StatsBase"]
+weakdeps = ["BSON", "CSV", "CairoMakie", "ClimaAnalysis", "DataFrames", "DelimitedFiles", "Flux", "Format", "GeoMakie", "HTTP", "Poppler_jll", "Printf", "StatsBase"]
 
     [deps.ClimaLand.extensions]
+    FluxnetSimulations = ["DelimitedFiles", "Format"]
```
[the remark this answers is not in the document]
kmdeck: Ill remove this.

ext/fluxnet_sims/data_processing.jl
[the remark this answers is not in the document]
kmdeck: I removed this!

experiments/integrated/fluxnet/run_fluxnet.jl
```diff
@@ -599,11 +562,11 @@ lines!(
     color = "blue",
 )
 
-if drivers.TS.status != absent
+if comparison_data.TS.absent == false
```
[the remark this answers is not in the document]
kmdeck: fixed this and SWC

src/diagnostics/default_diagnostics.jl
```diff
@@ -107,71 +107,71 @@ function default_diagnostics(
 ) where {FT}
 
     define_diagnostics!(land_model)
-
+    possible_diags = [
```
kmdeck, line 110: I had implemented this change for the LandModel in another PR - it speeds up our fluxnet runs by ~30% to use save the diagnostics we need. In this PR, I made the same change for the soil canopy model in order to run the ozark_pft simulation with the same code as the run_fluxnet simulation.

NEWS.md
```diff
@@ -3,6 +3,7 @@ ClimaLand.jl Release Notes
 
 main
 -------
+- Add data handling tools to FluxnetSimulationsExt and use throughout docs and experiments PR[#1259=1238](https://github.com/CliMA/ClimaLand.jl/pull/1238)
```
ph-kev, line 6:
```suggestion
- Add data handling tools to FluxnetSimulationsExt and use throughout docs and experiments PR[#1259](https://github.com/CliMA/ClimaLand.jl/pull/1238)
```

ext/fluxnet_simulations/forcing.jl
```diff
@@ -0,0 +1,413 @@
...
+
+Note that the TimeVaryingInput objects can be used to interpolate in time,
+which is why we drop missing data.
+
+This assumes that the first row of the CSV file is the list of of column names,
+and that these names are:
+"TA_F" (air temperature in C)
+"VPD_F" (air vapor pressure deficit in hPa)
+"PA_F" (air pressure in ?) TODO: CHECK UNITS
+"P_F" (accumulated precipitation in kg/m^2)
```
ph-kev, line 31: Do we know the units of `PA_F`?

ext/fluxnet_simulations/forcing.jl
```diff
@@ -0,0 +1,413 @@
...
+            column_name_map,
+            seconds_since_start_date;
+            preprocess_func = (x) -> -x / ρ_liq / data_dt,
+        )
+        atmos_P_snow = time_varying_input_from_data(
+            data,
+            "P_F",
+            column_name_map,
+            seconds_since_start_date;
+            preprocess_func = (x) -> 0 * x,
```
ph-kev, line 176:
```suggestion
            preprocess_func = x -> zero(x),
```

ph-kev: or just `zero`

ph-kev (approved):

ext/fluxnet_simulations/forcing.jl
kmdeck: thank you! I looked it up - kPa.

ext/fluxnet_simulations/initial_conditions.jl
```diff
@@ -0,0 +1,205 @@
...
+        model::ClimaLand.AbstractLandModel,
+    )
+
+Sets the initial conditions of `Y` using observations from the site `site_ID`, if available,
+using the observations closest to the start_date (in UTC). Since the data from Fluxnet sites
+is provided in local time, we require the offset from UTC in hours `hour_offset_from_UTC`.
+The `model` indicates which how to update it `Y` from these observations,
+via different methods of `set_fluxnet_ic!`.
+"""
+function set_fluxnet_ic!(
```
imreddyTeja, line 16: It seems to me that the general pattern in climaland for functions that modify the model is that they are created with a `make_something_func` or `make_compute_something`. For example, I often see `set_ic! = make_set_initial_state_from_file(...`).

experiments/integrated/fluxnet/fluxnet_simulation.jl
```diff
@@ -1,7 +1,7 @@
 """This file contains the site-generic time variables for running ClimaLand on
 fluxtower sites. These work in tandem with the site-specific timing parameters
 found in the {site-ID}_simulation.jl files in each site directory."""
-
+data_dt = Float64(FluxnetSimulationsExt.get_data_dt(site_ID))
```
juliasloan25, line 4: can you define this in `run_fluxnet.jl` instead? Thanhthanh's PR is removing this file

ext/fluxnet_simulations/initial_conditions.jl
```diff
@@ -0,0 +1,205 @@
...
+
+"""
+     set_fluxnet_ic!(Y, data, columns, Δ_date, model::ClimaLand.Soil.EnergyHydrology)
+
+Sets the values of Y.soil in place with:
+- observed value of SWC at the surface at the observation date closest to the start date
+- no ice (θ_i = 0)
+- an internal energy computed using the above θ_l, θ_i, and the temperature of the soil
+  in the first layer, at the observation date closest to the start date. If the soil 
+  temperature is not available, the air temperature is used.
```
juliasloan25, line 42: To visualize which variables are set:
```suggestion
- \vartheta_l: observed value of SWC at the surface at the observation date closest to the start date
- θ_i: no ice (θ_i = 0)
- \rho e_int: an internal energy computed using the above θ_l, θ_i, and the temperature of the soil
  in the first layer, at the observation date closest to the start date. If the soil 
  temperature is not available, the air temperature is used.
```

ext/fluxnet_simulations/forcing.jl
```diff
@@ -0,0 +1,399 @@
...
+    RH = e / esat
+    α = -10.04
+    β = 1.41
+    γ = 0.09
+    snow_frac = (1.0 / (1.0 + exp(α + β * T_in_C + γ * RH)))
+    return snow_frac
+end
+
+
+function compute_q(P, VPD, T; thermo_params)
```
juliasloan25, line 295: docstring?

ext/fluxnet_simulations/data_processing.jl
```diff
@@ -0,0 +1,195 @@
...
+The NamedTuple has two keys: absent, and values. If the data column
+is missing completely, the value of absent is true, and no values
+are returned. If the data column is present, absent is set to false,
+and the data is returned with the key values. If the column is present
+but the data is missing in each row, absent is set to true.
+
+In the future, we can explore dropping the missing values to be
+consistent with what we do above, but this is consistent with the
+current Fluxnet runs.
+)
```
juliasloan25, line 171:
```suggestion
```

ext/fluxnet_simulations/data_processing.jl
```diff
@@ -0,0 +1,195 @@
...
+`data` specified by `varnames`. 
+
+For example, if you wish to compute specific humidity from temperature,
+pressure, and vpd, you would do:
+varnames = ["TA_F", "PA_F", "VPD_F"] along with a preprocess function of
+the form
+function preprocess_func(T,P,VPD)
+# carries out unit conversion
+# computes q from T, P, VPD
+#returns q
```
juliasloan25, line 93:
```suggestion
# return q
```

juliasloan25 (approved):
Awesome!!

experiments/integrated/fluxnet/fluxnet_simulation.jl
kmdeck: I will call it from run_fluxnet instead.

ext/fluxnet_simulations/forcing.jl
kmdeck: added!

ext/fluxnet_simulations/initial_conditions.jl
kmdeck: Yeah, good point...our current `make_set_initial_state_from_file(ic_path, model)` function would not be sufficient for the fluxnet sites because it would need to also dispatch on the model domain (or call something that did), since site level data is in CSV format while global data is in netcdf format..

for fluxnet, we also have to pick the time corresponding to the initial conditions, while for the global run that was already done (the artifact is a single time slice).

I think we should revisit this

PR #1242, opened 2025-07-19
clip VPD

kmdeck:
## Purpose 
The medlyn term is = 1+g1/sqrt(VPD). When VPD is zero or positive, this errors. As suggested by @trontrytel, we could compute the inverse of this, and then compute the stomatal resistance, but all of our names everywhere are "conductance" so this would be a somewhat big change to make. For now, I clipped VPD to be > 1e-2 Pa. This is 1000x smaller than g1 which seems reasonable. This leads to a conductance which is very large (i.e., a very small stomatal resistance).

## To-do
upgrade thermodynamics compat once the release happens

## Content

----


tapios (approved):
Looks good to me. See my comment about a preference for explicit safe division.

PR #1253, opened 2025-07-22
restructuring domain + site parameters for Fluxnet simulation into a module

ThanhNguyen428:
## To-do

- [x] remove lingering syntax + style issues
- [x] make a test file and test each site's parameters
- [x] move into /ext

## Content

- Created site-specific ```get_domain_info``` and ```get_parameters``` functions for US-Ha1, US-MOz, US-NR1, and US-Var
- Created generic ```get``` functions for undefined sites (currently using US-MOz info)

----

### DIFF
--- a/ext/fluxnet_simulations/generic_site.jl
+++ b/ext/fluxnet_simulations/generic_site.jl
@@ -2,7 +2,168 @@
 #        MODULE FUNCTIONS         #
 ###################################
 
-# TODO
+"""
+    get_domain_info(FT; dz_bottom = FT(1.5), dz_top = FT(0.1),
+        nelements = 20, zmin = FT(-10),  zmax = FT(0))
+
+Gets and returns primary domain information for a generic Fluxnet site, using
+default values corresponding to a 10m deep soil column with 20 layers, with
+a resolution of 10 cm at the top of the domain and 1.5m at the bottom of the domain.
+"""
+function FluxnetSimulations.get_domain_info(
+    FT;
+    dz_bottom = FT(1.5),
+    dz_top = FT(0.1),
+    nelements = 20,
+    zmin = FT(-10),
+    zmax = FT(0),
+)
+
+    dz_tuple = (dz_bottom, dz_top)
+
+    return (; dz_tuple, nelements, zmin, zmax)
+end
+
+function FluxnetSimulations.get_location(
+    site_ID::String;
+    site_info = FluxnetSimulationsExt.get_site_info(site_ID),
+    lat = site_info[:lat],
+    long = site_info[:long],
+    time_offset = site_info[:time_offset],
+)
+    atmos_h = site_info[:atmospheric_sensor_height][1] # take the first height as default
+    return (; time_offset, lat, long, atmos_h)
+end
+
+"""
+    get_parameters(FT, site_ID, domain, pft::Pft; kwargs...)
+
+Gets parameters for a generic Fluxnet site based on PFT (plant functional type),
+supplemented with additional autofilled values from US-MOz (Missouri Ozark) site.
+
+Data sources:
+
+Hydraulics parameters:
+    - Holtzman, Nataniel, et al. 2023, https://doi.org/10.1029/2023WR035481
+"""
+function FluxnetSimulations.get_parameters(
+    FT,
+    site_ID,
+    domain,
+    pft::Pft = ClimaLand.Canopy.clm_dominant_pft(domain.space.surface),
+    ;
+    soil_params_gupta = ClimaLand.Soil.soil_vangenuchten_parameters(
+        domain.space.subsurface,
+        FT,
+    ),
+    soil_params_grids = ClimaLand.Soil.soil_composition_parameters(
+        domain.space.subsurface,
+        FT,
+    ),
+    soil_ν = soil_params_gupta[:ν],
+    soil_K_sat = soil_params_gupta[:K_sat],
+    soil_S_s = FT(1e-2),
+    soil_hydrology_cm = soil_params_gupta[:hydrology_cm],
+    θ_r = soil_params_gupta[:θ_r],
+    ν_ss_quartz = soil_params_grids[:ν_ss_quartz],
+    ν_ss_om = soil_params_grids[:ν_ss_om],
+    ν_ss_gravel = soil_params_grids[:ν_ss_gravel],
+    z_0m_soil = FT(0.01),
+    z_0b_soil = FT(0.01),
+    soil_ϵ = FT(0.98),
+    soil_albedo = ClimaLand.Soil.CLMTwoBandSoilAlbedo{FT}(;
+        ClimaLand.Soil.clm_soil_albedo_parameters(domain.space.surface)...,
+    ),
+    Ω = pft.parameters.Ω,
+    χl = pft.parameters.χl,
+    G_Function = CLMGFunction(χl),
+    α_PAR_leaf = pft.parameters.α_PAR_leaf,
+    λ_γ_PAR = FT(5e-7),
+    α_NIR_leaf = pft.parameters.α_NIR_leaf,
+    τ_PAR_leaf = pft.parameters.τ_PAR_leaf,
+    τ_NIR_leaf = pft.parameters.τ_NIR_leaf,
+    ϵ_canopy = pft.parameters.ϵ_canopy,
+    ac_canopy = pft.parameters.ac_canopy,
+    g1 = pft.parameters.g1,
+    Drel = FT(1.6),
+    g0 = ClimaCore.Fields.field2array(
+        ClimaLand.Canopy.clm_medlyn_g0(domain.space.surface),
+    )[1],
+    Vcmax25 = pft.parameters.Vcmax25,
+    pc = FT(-2.0e6),
+    sc = FT(5e-6),
+    SAI = FT(0),
+    f_root_to_shoot = pft.parameters.f_root_to_shoot,
+    K_sat_plant = pft.parameters.K_sat_plant,
+    ψ63 = pft.parameters.ψ63,
+    Weibull_param = FT(4),
+    a = FT(0.1 * 0.0098),
+    conductivity_model = PlantHydraulics.Weibull{FT}(
+        K_sat_plant,
+        ψ63,
+        Weibull_param,
+    ),
+    retention_model = PlantHydraulics.LinearRetentionCurve{FT}(a),
+    plant_ν = pft.parameters.plant_ν,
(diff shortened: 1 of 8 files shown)


ext/fluxnet_simulations/US-Var.jl
```diff
@@ -1 +1,157 @@
...
+    ld = FT(0.5),
+    G_Function = ConstantGFunction(ld),
+    α_PAR_leaf = FT(0.11),
+    λ_γ_PAR = FT(5e-7),
+    τ_PAR_leaf = FT(0.05),
+    α_NIR_leaf = FT(0.35),
+    τ_NIR_leaf = FT(0.34),
+    ϵ_canopy = FT(0.97),
+    g1 = FT(166),
+    Drel = FT(1.6),
```
kmdeck, line 76: A note that both Drel is universal and can be moved into ClimaParams at some point

ext/fluxnet_simulations/generic_site.jl
```diff
@@ -1 +1,169 @@
+"""
+    get_domain_info(FT, site_ID::Symbol; dz_bottom = FT(1.5), dz_top = FT(0.1),
+        nelements = 20, zmin = FT(-10),  zmax = FT(0))
 
+Gets and returns primary domain information for a generic Fluxnet site,
+using autofilled values from US-MOz (Missouri Ozark) site.
```
kmdeck, line 6: I think here we can just say "using default values corresponding to a 10m deep soil column with 20 layers, with a resolution of 10 cm at the top of the domain and 1.5m at the bottom of the domain." (not tied to US-MOz at all)

ext/fluxnet_simulations/generic_site.jl
```diff
@@ -1 +1,169 @@
...
+
+Data sources:
+
+Conductance parameters:
+    - Wang et al. 2021 https://doi.org/10.5194/gmd-14-6741-2021
+Hydraulics parameters:
+    - Holtzman, Nataniel, et al. 2023, https://doi.org/10.1029/2023WR035481
+"""
+function get_parameters(
```
kmdeck, line 42: I find it a bit odd to have US-MOz as a default because it is so specific. I'm not sure a default makes sense for many of these parameters.

src/ClimaLand.jl
```diff
@@ -424,4 +424,5 @@ import .Diagnostics: default_diagnostics
 
 # Simulations
 include(joinpath("simulations", "Simulations.jl"))
+include("ext/FluxnetSimulationsExt.jl")
```
kmdeck, line 427: This is not needed, right?

test/integrated/fluxnet_sim.jl
```diff
@@ -0,0 +1,457 @@
+using Dates
+using Test
+import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput
+
+using ClimaLand
+const FT = Float64
+
+@testset "US-Ha1 domain info + parameters" begin
```
kmdeck, line 8: We can test extensions, but without adding them to climaland. (e.g. here: https://github.com/CliMA/ClimaLand.jl/blob/f10d8f417484287ad2a5c425c1b74cbd7aa641ea/test/standalone/Snow/tool_tests.jl#L19)

ext/fluxnet_simulations/generic_site.jl
ThanhNguyen428: Is there a different set of parameters we should use for a "default"? Or should we just stick with using the `get_parameters` function based on user-inputted PFT, since I assume all of the other sites are equally as specific as US-MOz?

test/integrated/fluxnet_sim.jl
ThanhNguyen428: does this mean I should rewrite the test here, or move the test somewhere else and rewrite it with a new format in a different location?

ext/fluxnet_simulations/US-Var.jl
```diff
@@ -1 +1,157 @@
...
+Photosynthesis parameters:
+    - CLM 5.0 Tech Note: https://www2.cesm.ucar.edu/models/cesm2/land/CLM50_Tech_Note.pdf
+    - Slevin et al. 2015, https://doi.org/10.5194/gmd-8-295-2015
+Hydraulics parameters:
+    - Holtzman, Nataniel, et al. 2023, https://doi.org/10.1029/2023WR035481
+    - Xu and Baldocchi 2003, doi:10.1016/j.agrformet.2003.10.004
+"""
+function get_parameters(
+    ::Val{:US_Var};
+    time_offset = 8,
```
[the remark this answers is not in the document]
ThanhNguyen428: should they be split for user-accessibility (etc, a struct for user input), or for readability? Julia suggested that the arguments be passed in with default values and allow the user to edit using keyword arguments, and forgo the `struct` format that I originally proposed (for example, having a HydraulicsParameter struct for all hydraulics-related parameters and passing the struct into the function AS the main parameter itself).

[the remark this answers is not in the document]
kmdeck: I think the kwarg approach is good and & in line how we will handle parameters elsewhere, i.e. not packaging into structs or returning structs. But I wonder about grouping things like
time_offset, lat,lon, atmos_h (which are not really parameters) with things that are.

I was thinking more along the lines of:
get_location(site_ID) = (; lat = , long = ; time_offset = )
get_fluxtower_height(site_ID) = (; atmos_h = )
get_parameters(site_ID) = (; the rest...)

src/ClimaLand.jl
juliasloan25: Correct - we can delete this and it should fix the ClimaLand precompilation error that's in CI right now

test/integrated/fluxnet_sim.jl
juliasloan25: You can keep the test here and add this line:
```
FluxnetSimulationsExt = Base.get_extension(ClimaLand, :FluxnetSimulationsExt)
```
Then instead of accessing functions that come from that extension with `ClimaLand.FluxnetSimulationsExt.`, you can access them with `FluxnetSimulationsExt.`

juliasloan25: You can keep the test here and add this line:
```
FluxnetSimulationsExt = Base.get_extension(ClimaLand, :FluxnetSimulationsExt)
```
Then instead of accessing functions that come from that extension with `ClimaLand.FluxnetSimulationsExt.`, you can access them with `FluxnetSimulationsExt.`

ext/fluxnet_simulations/generic_site.jl
kmdeck: I think having a default for a site which returns the PFT/soil type parameters is great:
get_parameters(site_ID) = (; get_pft_params(get_pft(site_ID)..., get_soil_type_param(get_soil_type(site_ID))

and we also have discussed having the default be using the CLM data we already have and looking up the params closest to the lat/long.

but I dont think we should define a default that returns values for a specific site.

test/integrated/fluxnet_sim.jl
kmdeck: You can keep the test here, but FluxnetSimulationsExt does not need to be included in src/ClimaLand.jl. You can access the extension functions by following what is done for NeuralSnow extensions in the test I linked.

ext/fluxnet_simulations/US-Var.jl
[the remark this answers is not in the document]
juliasloan25: I like this approach! It doesn't make a lot of sense to have e.g. `atmos_h` returned with the parameters, but I wasn't sure where else to put it. I think it makes sense to add the 2 new functions Kat suggests: `get_location` to return spatial information about the site (lat, lon, time_offset) and `get_fluxtower_height` for the height.

This will be nice too because we can provide a defaults for `get_parameters` and `get_fluxtower_height`, but throw an error if a user tries to call `get_location(site_ID)` for a site that isn't implemented yet (since these variables are really site-specific and we can't have good defaults for them)

ext/fluxnet_simulations/generic_site.jl
juliasloan25: Okay! It would be good to implement a struct SoilType analogous to the Pft type we currently have. Since that will take a little bit longer than using the existing Pft types, I think it makes sense to do it in another PR. For now we can just use some default values for the soil parameters (e.g. from ozark), or do you think that still doesn't make sense because they're too specific?

ext/fluxnet_simulations/generic_site.jl
```diff
@@ -1 +1,172 @@
...
+"""
+function get_domain_info(
+    FT,
+    site_ID::Symbol;
+    dz_bottom = FT(1.5),
+    dz_top = FT(0.1),
+    nelements = 20,
+    zmin = FT(-10),
+    zmax = FT(0),
+) where {FT}
```
juliasloan25, line 22: FT is passed in as a regular argument here (as opposed to as a type parameter), so you don't need the `where {FT}`

ext/fluxnet_simulations/US-Ha1.jl
```diff
@@ -1 +1,176 @@
...
+    get_location(::Val{:US_Ha1}; kwargs...)
+
+Returns geographical information for US-Ha1 (Massachusetts Harvard Forest) Fluxnet site.
+"""
+function get_location(
+    ::Val{:US_Ha1};
+    time_offset = 5,
+    lat = FT(42.5378),
+    long = FT(-72.1715),
+)
```
juliasloan25, line 39:
```suggestion
function get_location(
    FT,
    ::Val{:US_Ha1};
    time_offset = 5,
    lat = FT(42.5378),
    long = FT(-72.1715),
)
```

ext/fluxnet_simulations/US-Ha1.jl
```diff
@@ -1 +1,176 @@
...
+    - Wang et al. 2021, https://doi.org/10.5194/gmd-14-6741-2021
+Conductance parameters:
+    - Wang et al. 2021, https://doi.org/10.5194/gmd-14-6741-2021
+Photosynthesis parameters:
+    - Holtzman, Nataniel, et al. 2023, https://doi.org/10.1029/2023WR035481
+Hydraulics parameters:
+    - Holtzman, Nataniel, et al. 2023, https://doi.org/10.1029/2023WR035481
+"""
+function get_parameters(
+    ::Val{:US_Ha1};
```
juliasloan25, line 74:
```suggestion
function get_parameters(
    FT,
    ::Val{:US_Ha1};
```

juliasloan25 (commented):
This is looking good, thank you Thanhthanh!

PR #1256, opened 2025-07-23
use get_default_parameter for canopy parameters

juliasloan25:
## Purpose 
In #1233, we added convenience constructors for canopy component models. These constructors provide defaults for many parameters that can easily be overwritten.

However, we forgot to use `get_default_parameter` as the defaults for some parameters that are available via ClimaParams. Note that some parameters are still hardcoded because they aren't in ClimaParams, but eventually we'd like to add them there as well.

### DIFF
--- a/src/shared_utilities/Parameters.jl
+++ b/src/shared_utilities/Parameters.jl
@@ -121,8 +121,8 @@ end
 """
     get_default_parameter(FT, climaparams_name)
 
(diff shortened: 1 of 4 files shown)


juliasloan25:
> I don't have a preference, but you could include the TODO as comments instead of in the documentation.

Sure! I was copying what Kat did for Soil.jl, but I'll move them all to comments and open an issue for it too

PR #1257, opened 2025-07-23
force type specialization in constructors

juliasloan25:
## Purpose 
We prefer to write a function with a type parameter over passing the type as a regular function argument because the method with a parametric type will always be specialized by the compiler. When the type is passed as an argument, the compiler will specialize if the type is used within the function body, but not if it's only passed to other functions. See the [relevant section of the Julia manual](https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing) for more details.

Because our types are large and complex, and because we often pass FT around without using it within a function (e.g. for constructors), we do want to force specialization and therefore type inference. This PR changes constructors to uniformly take FT as a type parameter.

### Examples
These will specialize:
```julia
# used by model constructors
function f{FT}() where {FT}
end

# used by parameter constructors
function f(::Type{FT}) where {FT}
end
```
but these won't:
```julia
# removed in this PR
function f(FT)
end

function f(::Type)
end
```

## To-do
- [x] PrescribedGroundConditions constructor
- [x] EnergyHydrology constructor
- [x] RichardsModel constructor

### DIFF
--- a/docs/src/tutorials/standalone/Canopy/canopy_tutorial.jl
+++ b/docs/src/tutorials/standalone/Canopy/canopy_tutorial.jl
@@ -141,8 +141,7 @@ shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
 
 ψ_soil = FT(0.0)
 T_soil = FT(298.0)
-soil_driver = PrescribedGroundConditions(
-    FT;
+soil_driver = PrescribedGroundConditions{FT}(;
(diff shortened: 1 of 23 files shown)


ph-kev (approved):

nefrathenrici (approved):
Looks good! Thanks for fixing the incorrect parameter docstrings as well.

PR #1259, opened 2025-07-24
add FluxnetSimulations ext

juliasloan25:
## Purpose 
Add a template for the FluxnetSimulations extension, so that multiple people can start adding to it in parallel.

I chose to organize the extension by having a file for each site we test. That way, we can add anything site-specific in one file and see them in one place (I expect these will only be the `get_parameters` and `get_domain_info` functions). The extension will also have functions in new files for data cleaning, running a simulation, etc - see #1238

## To-do
- [x] add ext/FluxnetSimulations.jl file
- [x] add extension to ClimaLand.jl project
- [x] add template files for each site we test, and a generic site for fallback/default methods


ph-kev (approved):
I think the name of the module could be FluxnetSimulationsExt instead of FluxnetSimulations, since it is a package extension. However, there is also LandSimulationVisualization which also doesn't end with Ext.

ThanhNguyen428 (approved):

PR #1267, opened 2025-07-26
Add simple standalone EnergyHydrology, Canopy tutorials

juliasloan25:
## Purpose 
Adds 2 simple simulations using default model constructors for EnergyHydrology and CanopyModel. These will be used as the basic demo tutorials for the MC3 meeting (in colab notebooks).

The simple soil simulation doesn't make plots so it can be as simple as possible. The canopy simulation does, and it looks like this:
[diurnal_timeseries.pdf](https://github.com/user-attachments/files/21778786/diurnal_timeseries.pdf)

## Content
- [x] add simple soil simulation in Getting Started page
- [x] add simple canopy tutorial
- [x] update README (taken from #1311)
- [x] link to stable docs instead of dev - this is more consistent with using our packages as "libraries", as outside users will want to do

----

### DIFF
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -1,61 +1,119 @@
-# Getting Started
-
 ## Installation of Julia and ClimaLand
 
 ClimaLand is provided as a Julia package, so it requires having Julia installed. Information about Julia packages is available on the [Julia website](https://julialang.org/packages/).
 
 First, download and install Julia by following the instructions at [https://julialang.org/downloads/](https://julialang.org/downloads/).
-Then, you can launch a [Julia REPL](https://docs.julialang.org/en/v1/stdlib/REPL/) and install the
-ClimaLand.jl package by running:
+Then, you can launch a [Julia REPL](https://docs.julialang.org/en/v1/stdlib/REPL/) by running `julia --project` and install the
+ClimaLand.jl package by running the following within the REPL:
 
 ```julia
-julia> using Pkg
-julia> Pkg.add(ClimaLand)
-julia> using ClimaLand
+using Pkg
+Pkg.add(ClimaLand)
 ```
 
-A typical land simulation employs several different parameterizations to model the various land-surface processes. Let's start our journey into ClimaLand by looking at one of those.
+## Running a simple soil model
 
-### Parameterization
+Now that we have Julia installed, let's set up and run a simple ClimaLand simulation!
+You can follow along by copying the following code segments into your REPL.
 
-Let's start with a basic example: compute canopy gross photosynthesis (GPP).
-
-```@repl
+First we import the necessary Julia packages:
+```julia
+import ClimaParams as CP
 using ClimaLand
-@doc ClimaLand.Canopy.compute_GPP
+using ClimaLand.Domains
+using ClimaLand.Soil
+import ClimaLand.Simulations: LandSimulation, solve!
+import ClimaLand.Parameters as LP
+using Dates
 ```
 
-As you can see, our parameterization for GPP is located in the [`Canopy`](@ref Photosynthesis) Module, and requires four arguments.
-For example, with An = 5 µmol m⁻² s⁻¹, K = 0.5, LAI = 3 m² m⁻², Ω = 0.7, you can compute GPP like below:
-
-```@repl
-import ClimaLand.Canopy as canopy
-canopy.compute_GPP(5.0, 0.5, 3.0, 0.7)
+Choose a floating point precision, and get the parameter set, which holds constants used across CliMA models.
+Note: we use SI units unless otherwise specified.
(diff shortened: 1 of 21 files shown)


docs/src/getting_started.md
```diff
@@ -1,61 +1,111 @@
...
+```julia
+FT = Float32
+earth_param_set = LP.LandParameters(FT);
 ```
 
-Et voilà!
-
-Note that our package [ParamViz](https://github.com/CliMA/ParamViz.jl) allows interactive visualisation of
-our parameterizations. See examples in the standalone models pages.
+We will run this simulation on a column domain with 1 meter depth, at a random lat/lon location.
```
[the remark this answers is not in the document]
kmdeck: We might link to the Units page (and say "We use SI units unless noted otherwise.")

docs/src/getting_started.md
```diff
@@ -1,61 +1,111 @@
...
+start_date = DateTime(2008);
+end_date = start_date + Second(60 * 60 * 72);
+dt = 1000.0;
+```
 
-ClimaLand contains multiple modules. They are listed below:
+The soil model takes in 2 forcing objects, atmosphere and radiation,
+which we read in from ERA5 data.
+```julia
+era5_ncdata_path =
```
kmdeck, line 55: Maybe at the bottom we should cite this: Hersbach, Hans, et al. "The ERA5 global reanalysis." Quarterly journal of the royal meteorological society 146.730 (2020): 1999-2049.

docs/src/tutorials/standalone/Canopy/simple_canopy.jl
```diff
@@ -0,0 +1,115 @@
...
+era5_ncdata_path =
+    ClimaLand.Artifacts.era5_land_forcing_data2008_path(; lowres = true);
+atmos, radiation = ClimaLand.prescribed_forcing_era5(
+    era5_ncdata_path,
+    surface_space,
+    start_date,
+    earth_param_set,
+    FT,
+);
+ground = PrescribedGroundConditions{FT}();
```
kmdeck, line 55: Just an idea for a future PR: I think there is some overlap between this and the "canopy_tutorial" but we can turn the canopy_tutorial into something more "advanced" - for example, we could force the canopy with the data from a fluxnet site

For this, we might note that this prescribes emissivity, albedo, temperature, and soil moisture

docs/src/tutorials/standalone/Canopy/simple_canopy.jl
```diff
@@ -0,0 +1,115 @@
...
+LAI = TimeVaryingInput((t) -> FT(1.0));
```
kmdeck, line 56: You could refer to the canopy_tutorial for how to prescribe this using MODIS

docs/src/tutorials/standalone/Canopy/simple_canopy.jl
```diff
@@ -0,0 +1,115 @@
...
+    domain,
+    (; atmos, radiation, ground),
+    LAI,
+    earth_param_set,
+);
+
+# Define a function to set initial conditions for the prognostic variables.
+# Since these are specific to the model physics, the contents here differ from
+# the soil example, but the function structure remains the same.
+function set_ic!(Y, p, t0, model)
```
kmdeck, line 73: We might mention that these variables and their meanings are described in the science tutorials?

docs/src/tutorials/standalone/Soil/richards_equation.jl
```diff
@@ -218,6 +218,7 @@ plot(
     legend = :bottomleft,
     title = "Equilibrium test",
 );
+plot!(1e-3 .+ ϑ_l[1], z, label = "porosity");
```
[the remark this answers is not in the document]
kmdeck: (I see that this was already like that!)

kmdeck (approved):
Thank you! these look great

docs/src/tutorials/standalone/Canopy/simple_canopy.jl
juliasloan25: I agree! It would be nice to make the canopy_tutorial.jl more complex now

docs/src/tutorials/standalone/Canopy/simple_canopy.jl
juliasloan25: Good idea. Unfortunately there's no documentation for canopy energy so no description of T I believe.

Also, I know we just restructured the docs, but I think it would be helpful if the science documentation was accessible at the top-level - right now it seems kind of hidden away to me.

docs/src/tutorials/standalone/Soil/richards_equation.jl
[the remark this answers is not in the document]
juliasloan25: I'd prefer to just leave it as it is for now. I think we'ld have to create a field of \nu so I'm not sure there's a really clean way to do it.

PR #1279, opened 2025-07-29
Enable reading spatial data to a point/column

juliasloan25:
## Purpose 
Finish last small tasks to be able to read in spatial data to a point/column. Adds a small test verifying that we can do this and the values are as expected. Note that the tests don't check interpolation, since we assume that ClimaUtilities has correctly implemented and tested their interpolation.

## Content
- [x] update to ClimaUtilities v0.1.25
- [x] extend `use_lowres_clm` for PointSpace
- [x] add tests for reading in 2D data to a PointSpace, and 3D data to a column

----


test/shared_utilities/domains.jl
```diff
@@ -584,5 +586,88 @@ end
...
+        subsurface_space = column.space.subsurface
+
+        # Get the residual water content at this lat/lon using ClimaUtilities
+        soil_params_mask = SpaceVaryingInput(
+            filepath_mask,
+            "α",
+            subsurface_space;
+            file_reader_kwargs = (; preprocess_func = (data) -> data > 0,),
+        )
+        (; θ_r) =
```
kmdeck, line 667: do we still need to mask and set the masked values to 0.09? I thought that was handled in `ClimaLand.Soil.soil_vangenuchten_parameters(subsurface_space, FT)`.

test/shared_utilities/domains.jl
```diff
@@ -584,5 +586,88 @@ end
...
+        zs = ncd["z"][:]
+        lat_ind = 90
+        long_ind = 100
+        lat = FT(lats[lat_ind])
+        long = FT(longs[long_ind])
+
+        θ_r_nc = ncd["θ_r"][long_ind, lat_ind, :]
+
+        # Construct a Column domain at a longlat that appears in the dataset (not testing interpolation)
+        zmin = FT(0)
```
kmdeck, line 653: not used?

test/shared_utilities/domains.jl
kmdeck: I know you arent testing interpolation here, but doesnt this interpolate in z (and make it harder to compared directly to the raw data in the nc file? It must work since the test passes...but maybe you could leave a comment explaining?

kmdeck (approved):
Only some questions about the test. thanks for adding this!

test/shared_utilities/domains.jl
juliasloan25: Oops, I meant to do this for the data we're reading directly from netcdf in the test. I guess for this point it doesn't matter, so I'll just remove it.

PR #1223, opened 2025-07-10
Implement `compute_full_pmodel_outputs` and unit tests

ychnli:
This is a duplicated PR of #1198 because git history complications made it very cumbersome to rebase. See #1198 for previous discussion. The description is copied here (and updated with latest names) for convenience: 

> This implements of the core P-model computations in the function `compute_full_pmodel_outputs` (#1197). 
> 
> ## To-do
> 
> - still need to write `update_photosynthesis!` in `pmodel.jl` with appropriate subdaily timestepping equations 
> - still need to make sure that stomatal conductance is appropriately updated, since P-model predicts it jointly from the optimality criteria 
> 
> ## Content
> 
> - `compute_full_pmodel_outputs` and associated helper functions in `pmodel.jl` and `canopy_parameterizations.jl` 
> - `test_pmodel.jl` unit tests and test files on ClimaArtifacts 

## Important note
 in this PR, the PModel cannot be used yet in the canopy model because subdaily GPP variability is not explicitly separated from longer timescale acclimatized GPP. Instead `compute_full_pmodel_outputs` should be used as a ClimaLand implementation of the [rpmodel package](https://github.com/geco-bern/rpmodel) that models GPP variability assuming adjustment to optimality on weekly timescales.

PR #1281, opened 2025-07-30
Implement subdaily P-model 

### LINKED ISSUE #1206: Subdaily model for GPP
The P-model assumes optimality to model the weekly timescale acclimatization of Rubisco-limited and light-limited photosynthetic capacities Vcmax and Jmax. Faster (subdaily) variation in these capacities can be modeled by a modified-Arrhenius type function (Kattge & Knorr, 2007) and by adjusting temperature-dependent intrinsic quantum yield $\phi_0(T)$ and soil moisture stress $\beta(\theta)$. 

Mengoli et al. ([2022](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2021MS002767)) implement a subdaily version of the P-model where optimal values for Vcmax25, Jmax25, and $\xi$ are computed at local noon and accumulated via an exponential moving average. For each day, the instantaneous photosynthetic capacity is then modeled using the current noontime-acclimated capacities. We will use this formulation to implement `update_photosynthesis!` and `update_conductance!` for the P-model.

ychnli:
## Purpose 

This followup PR to #1223 implements the "subdaily P-model", which explicitly models both the slow (weekly) timescale optimal acclimatization of photosynthetic capacities (Vcmax25, Jmax25) and sensitivity of stomatal closure to dryness ($\xi$), as well as the fast (instantaneous) timescale adjustment of these capacities to temperature (for Vcmax and Jmax) and VPD (for computing $c_i$ from $\xi$). Closes #1206. 

## To-do

- We need a way of ensuring that the P-model callback is included in the callbacks when we run a simulation using the P-model. The model will not work if the callback is not included, but currently we don't check for this
- Implement soil moisture stress as a modular parameterization (PR forthcoming) 

## Content

We follow the scheme introduced in Mengoli et al. (2022), who propose computing optimal parameters at local noon and modeling weekly acclimatization using an exponential moving average. The implementation details are as follows:
- Local noon is computed via the longitude of the domain and represented as seconds UTC. Longitude can be passed in directly as a float, or automatically inferred from the canopy_domain, if it is provided. 
- We do not account for the [equation of time](https://clima.github.io/Insolation.jl/dev/ZenithAngleEquations/#Equation-of-Time) correction to compute local noon, so it should not be interpreted literally as local noon. This introduces an error of at most ~20 minutes, which is unconsequential for the purpose of computing optimal photosynthetic parameters (see Mengoli et al. 2022). As such, local noon is constant throughout the year and solely dependent on longitude.
- We create a new callback `frequency_based_callback` which checks for local noon every simulation timestep `dt` and updates the optimal parameters if $t \in [t_{noon} - dt/2, t_{noon} + dt/2]$
- Because the exponential moving average requires the last computed value, we add new variables to the cache as a NamedTuple `p.canopy.photosynthesis.OptVars` to store these variables. We use a NamedTuple because of shared computation. 
- Use of past values requires some special logic at initialization. Two possible schemes are 1) computing optimal parameters using initial conditions $(Y_0, p_0)$; 2) use pre-set optimal parameters e.g., from a spun-up long run or a map of parameters. Currently we implement option (1). This means that it takes ~1 month of spinup time for the optimal photosynthetic parameters to acclimate to realistic values. 
- Since the P-model also predicts stomatal conductance $g_s$, a `PModelConductance` model is added to `stomatal_conductance.jl`. When constructing a `canopy`, we check that if we use P-model for photosynthesis, we should also be using P-model conductance for physical consistency. 

### Citation
Mengoli, G., Agustí-Panareda, A., Boussetta, S., Harrison, S. P., Trotta, C., & Prentice, I. C. (2022). Ecosystem photosynthesis in land-surface models: A first-principles approach incorporating acclimation. Journal of Advances in Modeling Earth Systems, 14, e2021MS002767. https://doi.org/10.1029/2021MS002767

----


.buildkite/Project.toml
```diff
@@ -2,6 +2,7 @@
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 ArgParse = "c7e460c6-2fb9-53a9-8c5b-16f535851c63"
 BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
```
ph-kev, line 6: You shouldn't need to include BenchmarkTools.

test/standalone/Vegetation/test_pmodel.jl
```diff
@@ -216,3 +223,251 @@ end
...
+            )
+            @test isapprox(
+                outputs_from_intermediate_vars.ci,
+                outputs_full.ci,
+                rtol = rtol,
+                atol = atol,
+            )
+        end
+    end
+end
```
ph-kev, line 473: Could you add a new line to the end of this file?

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1357,22 +1371,59 @@ function optimal_co2_ratio_c3(
...
     ) where {FT}
 
-    Computes the stomatal conductance of CO2 (`gs`), in units of mol CO2/m^2/s
+    Computes the stomatal conductance of CO2 (`gs_co2`), in units of mol CO2/m^2/s
     via Fick's law. Parameters are the ratio of intercellular to ambient CO2 
-    concentration (`χ`), the ambient CO2 concentration (`ca`), and the 
-    assimilation rate (`A`). This is related to the conductance of H2O by a 
+    concentration (`χ`), the ambient CO2 concentration (`ca`, in mol/mol), and the 
+    assimilation rate (`A`, mol m^-2 s^-1). This is related to the conductance of water by a 
     factor Drel = 1.6.
```
juliasloan25, line 1406:
```suggestion
    factor Drel (with default value 1.6). 
```
since the value of Drel can be overriden

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1357,22 +1371,59 @@ function optimal_co2_ratio_c3(
...
+        χ::FT, 
+        ca::FT,
+        A::FT,
+        Drel::FT
+    ) where {FT}
+
+    Computes the stomatal conductance of H2O (`gs_h2o`), in units of mol H2O/m^2/s
+    via Fick's law. Parameters are the ratio of intercellular to ambient CO2 
+    concentration (`χ`), the ambient CO2 concentration (`ca`, in mol/mol), the 
+    assimilation rate (`A`, mol m^-2 s^-1), and the relative conductivity ratio `Drel` (unitless).
```
juliasloan25, line 1423:
```suggestion
Computes the stomatal conductance of H2O (`gs_h2o`), in units of mol H2O/m^2/s
via Fick's law. Parameters are the ratio of intercellular to ambient CO2 
concentration (`χ`), the ambient CO2 concentration (`ca`, in mol/mol), the 
assimilation rate (`A`, mol m^-2 s^-1), and the relative conductivity ratio `Drel` (unitless).
```

src/standalone/Vegetation/pmodel.jl
```diff
@@ -119,21 +129,20 @@ function PModelConstants{FT}(;
...
 ) where {FT <: AbstractFloat}
-
+    earth_param_set = LP.LandParameters(FT)
     return PModelConstants{FT}(
-        R,
+        LP.gas_constant(earth_param_set),
         Kc25,
         Ko25,
-        To,
+        FT(298.15),
```
juliasloan25, line 142: We should be able to get this from ClimaParams using:
```suggestion
        LP.get_default_parameter(:kelvin_25C)
```

src/standalone/Vegetation/pmodel.jl
```diff
@@ -102,13 +115,10 @@ Creates a `PModelConstants` object with default values for the P-model constants
 See Stocker et al. (2020) Table A2 and references within for more information. 
 """
 function PModelConstants{FT}(;
-    R = LP.gas_constant(LP.LandParameters(FT)),
     Kc25 = FT(39.97), # Pa (see note on line 38 above)
     Ko25 = FT(27480), # Pa (see note on line 38 above)
-    To = FT(298.15),
     ΔHkc = FT(79430),
     ΔHko = FT(36380),
-    Drel = FT(1.6),
```
juliasloan25, line 111: I think it's better to expose Drel and Mc here so they can be overwritten. Why did you remove them from the function arguments?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -173,40 +185,110 @@ struct PModel{FT, OPFT <: PModelParameters{FT}, OPCT <: PModelConstants{FT}} <:
...
         constants,
     )
 end
 
-ClimaLand.auxiliary_vars(model::PModel) = (:An, :GPP, :Rd, :Vcmax25)
-ClimaLand.auxiliary_types(model::PModel{FT}) where {FT} = (FT, FT, FT, FT)
-ClimaLand.auxiliary_domain_names(::PModel) =
-    (:surface, :surface, :surface, :surface)
+"""
+New cache variables:
```
juliasloan25, line 211:
```suggestion
In addition to total net carbon assimilation (`An`), total canopy photosynthesis (`GPP`),
and dark respiration (`Rd`), the P-model requires some more cache variables:
```

src/standalone/Vegetation/pmodel.jl
```diff
@@ -288,7 +373,10 @@ function compute_full_pmodel_outputs(
     )
     Vcmax25 = Vcmax / inst_temp_scaling_vcmax25
 
-    Jmax = FT(4) * ϕ0 * I_abs / sqrt((mj / (βm * mprime))^2 - 1)
+    # check for negative arg before taking sqrt
+    arg = (mj / (βm * mprime))^2 - 1
+    arg = arg > 0 ? arg : FT(0)
```
juliasloan25, line 378: What is `arg` here? It would be good to use a more descriptive name

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +431,568 @@ function compute_full_pmodel_outputs(
...
+            Vcmax25_opt = α * OptVars.Vcmax25_opt + (1 - α) * Vcmax25,
+            Jmax25_opt = α * OptVars.Jmax25_opt + (1 - α) * Jmax25,
+        )
+    else
+        return OptVars
+    end
+end
+
+
+function compute_intermediate_pmodel_vars(
```
juliasloan25, line 601: docstring

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +431,568 @@ function compute_full_pmodel_outputs(
...
+        VPD,
+        p.drivers.c_co2,
+        βm,
+        I_abs,
+        local_noon_mask,
+    )
+end
+
+
+function call_update_optimal_EMA(
```
juliasloan25, line 726: docstring

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +431,568 @@ function compute_full_pmodel_outputs(
...
+    I_abs = @. lazy(
+        compute_I_abs(
+            p.canopy.radiative_transfer.par.abs,
+            p.canopy.radiative_transfer.par_d,
+            canopy.radiative_transfer.parameters.λ_γ_PAR,
+            constants.lightspeed,
+            constants.planck_h,
+            constants.N_a,
+        ),
+    )
```
juliasloan25, line 767: It looks like all of this is copied in `set_historical_cache!` too so it could be pulled out into a function they both call to reduce code duplication. But since it's only in 2 places it's ok like this too

src/shared_utilities/utils.jl
```diff
@@ -666,6 +666,65 @@ function NaNCheckCallback(
...
+- `dt`: The model timestep (used for divisibility warning).
+- `func`: A function that is called every time the schedule triggers.
+- `func_args...`: Additional arguments passed to `func`.
+
+The callback uses `ClimaDiagnostics.EveryCalendarDtSchedule` to determine when to
+call the function based on the `frequency`. 
+"""
+function FrequencyBasedCallback(
+    frequency::Union{AbstractFloat, Dates.Period},
+    start_date,
```
juliasloan25, line 695: It would be good to take in t0 (or t_start), since the simulation starts at `start_date + Seconds(t0)` if t0 != 0, like how we do for CheckpointCallback.

Also, could you open an issue to use this instead of CheckpointCallback? This is more general so we can use it for that case too (but it doesn't have to be done in this PR)

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +431,568 @@ function compute_full_pmodel_outputs(
...
+"""
+function make_PModel_callback(
+    ::Type{FT},
+    start_date, 
+    dt, 
+    canopy, 
+    longitude = nothing
+) where {FT <: AbstractFloat}
+    function seconds_after_midnight(t)
+        return Hour(t).value * 3600 + Minute(t).value * 60
```
juliasloan25, line 813: add any remaining seconds?
```suggestion
        return Hour(t).value * 3600 + Minute(t).value * 60 + Second(t)
```

src/standalone/Vegetation/solar_induced_fluorescence.jl
```diff
@@ -72,23 +72,65 @@ function update_SIF!(p, Y, sif_model::Lee2015SIFModel, canopy)
...
-        T_freeze,
-        ΔHJmax,
-        To,
-        θj,
-        ϕ,
-        sif_parameters,
-    )
+
+    # Check if photosynthesis model is PModel and use cached Jmax and J values
+    if isa(canopy.photosynthesis, PModel)
```
juliasloan25, line 78: We could also instead write a function to compute SIF that dispatches on the photosynthesis model type. That would be more in the style of Julia, but it isn't strictly necessary

src/standalone/Vegetation/solar_induced_fluorescence.jl
```diff
@@ -72,23 +72,65 @@ function update_SIF!(p, Y, sif_model::Lee2015SIFModel, canopy)
...
+            θj,
+            ϕ,
+            sif_parameters,
+        )
+    end
 end
 
 Base.broadcastable(m::SIFParameters) = tuple(m)
+
+function compute_SIF_at_a_point(
```
juliasloan25, line 110: docstring please

src/standalone/Vegetation/stomatalconductance.jl
```diff
@@ -108,3 +109,65 @@ function MedlynConductanceParameters(
     G1 = typeof(g1)
     return MedlynConductanceParameters{FT, G1}(; g1, parameters..., kwargs...)
 end
+
+
+#################### P model conductance ####################
+"""
+    PModelConductanceParameters{FT <: AbstractFloat}
+
+The required parameters for the Medlyn stomatal conductance model.
```
juliasloan25, line 118:
```suggestion
The required parameters for the P-model stomatal conductance model.
```

src/standalone/Vegetation/stomatalconductance.jl
```diff
@@ -108,3 +109,65 @@ function MedlynConductanceParameters(
...
+$(DocStringExtensions.FIELDS)
+"""
+Base.@kwdef struct PModelConductanceParameters{FT <: AbstractFloat}
+    "Relative diffusivity of water vapor (unitless)"
+    Drel::FT
+end
+
+Base.eltype(::PModelConductanceParameters{FT}) where {FT} = FT
+
+struct PModelConductance{FT, PMCP <: PModelConductanceParameters{FT}} <:
```
juliasloan25, line 128: Some documentation about how this model works would be helpful

src/standalone/Vegetation/stomatalconductance.jl
```diff
@@ -108,3 +109,65 @@ function MedlynConductanceParameters(
...
+    χ = @. lazy(ci / (c_co2_air * P_air))       # ratio of intercellular to ambient CO2 concentration, unitless
+    @. p.canopy.conductance.r_stomata_canopy =
+        1 / (upscale_leaf_conductance(
+            pmodel_gs_h2o(χ, c_co2_air, An, Drel), # leaf level conductance in mol H2O Pa^-1
+            LAI,
+            T_air,
+            R,
+            P_air,
+        ) + eps(FT)) # avoids division by zero, since conductance is zero when An is zero 
+
```
juliasloan25, line 172:
```suggestion
return nothing
```

test/standalone/Vegetation/test_pmodel.jl
```diff
@@ -216,3 +223,251 @@ end
...
+    z_0b = FT(0.1) # m, Roughness length for scalars
+    h_int = FT(30.0) # m, "where measurements would be taken at a typical flux tower of a 20m canopy"
+    shared_params = SharedCanopyParameters{FT, typeof(earth_param_set)}(
+        z_0m,
+        z_0b,
+        earth_param_set,
+    )
+    start_date = DateTime(2005)
+
+    # Dummy radiative forcing
```
juliasloan25, line 309: Since this is just testing the callback init, you can use this function to get simple analytic forcing:
```
atmos, radiation = prescribed_analytic_forcing(FT)
```

test/standalone/Vegetation/test_pmodel.jl
```diff
@@ -216,3 +223,251 @@ end
...
+    ar_model = AutotrophicRespirationModel{FT}(ar_params)
+
+    # Radiative transfer 
+    rt_params = BeerLambertParameters(FT)
+    rt_model = BeerLambertModel{FT}(rt_params)
+
+    # Photosynthesis 
+    is_c3 = FT(1) 
+    photosynthesis_params = FarquharParameters(FT, is_c3)
+    photosynthesis_model = FarquharModel{FT}(photosynthesis_params)
```
juliasloan25, line 248: It probably makes more sense to use the PModel for this test :) but it will test what you want either way

test/standalone/Vegetation/test_pmodel.jl
```diff
@@ -216,3 +223,251 @@ end
...
+        TimeVaryingInput(u_atmos),
+        TimeVaryingInput(q_atmos),
+        TimeVaryingInput(P_atmos),
+        start_date,
+        h_atmos,
+        earth_param_set;
+        c_co2 = TimeVaryingInput(c_atmos),
+    )
+
+    canopy = ClimaLand.Canopy.CanopyModel{FT}(;
```
juliasloan25, line 351: You can use the constructor Kat and I recently added to make this test much shorter. E.g. see the model setup in [this test](https://github.com/CliMA/ClimaLand.jl/blob/b3fce30e5e9b4e3024f0bbfd13e2e9654b541296/test/standalone/Vegetation/canopy_model.jl#L1361-L1410)

test/standalone/Vegetation/test_pmodel.jl
```diff
@@ -216,3 +223,251 @@ end
...
+            constants,
+            T_canopy,
+            P_air,
+            VPD,
+            ca,
+            βm,
+            I_abs,
+        )
+
+        @testset "Test update_optimal_EMA optimality computation for $FT" begin
```
juliasloan25, line 408: These don't need to be in a nested testset - they can just be in the outer testset (same for the other testset below)

src/shared_utilities/drivers.jl
```diff
@@ -971,7 +971,7 @@ function vapor_pressure_deficit(T_air, P_air, q_air, thermo_params)
         P_air,
         Thermodynamics.PhasePartition(q_air),
     )
-    return es - ea
+    return max(es - ea, 0)
```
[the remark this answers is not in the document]
ychnli: I changed this when I noticed sometimes we were getting negative VPD in fluxnet simulations if the air is supersaturated. I'm not sure if this is standard, but from what I could find online it seems like negative VPD is usually clipped to zero. I can see where it happens in the simulation

src/standalone/Vegetation/Canopy.jl
```diff
@@ -461,6 +487,10 @@ function CanopyModel{FT}(;
         @assert typeof(boundary_conditions.atmos) <: PrescribedAtmosphere{FT}
     end
 
+    if typeof(photosynthesis) <: PModel{FT}
+        @assert typeof(conductance) <: PModelConductance{FT} "When using PModel for photosynthesis, you must also use PModelConductance for stomatal conductance"
+    end
```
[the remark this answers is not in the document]
ychnli: Pmodelconductance should be used if and only if Pmodel is used for photosynthesis. I'll add the other check

src/standalone/Vegetation/pmodel.jl
ychnli: Mc is a physical constant (molar mass of carbon) so should never be changed. Drel is also a physical constant (maybe with some more uncertainty), which realistically will also never be changed.

src/shared_utilities/utils.jl
```diff
@@ -666,6 +666,69 @@ function NaNCheckCallback(
...
+function FrequencyBasedCallback(
+    frequency::Union{AbstractFloat, Dates.Period},
+    start_date::Dates.DateTime,
+    t0::Union{AbstractFloat, Dates.Period},
+    dt::Union{AbstractFloat, Dates.Period};
+    func,
+    func_args...,
+)
+    # Normalize frequency to a Dates.Period
+    frequency_period =
```
kmdeck, line 703: is it possible to test this function in a unit test?

src/standalone/Vegetation/Canopy.jl
```diff
@@ -30,9 +32,10 @@ import ClimaLand:
     make_compute_jacobian,
     get_drivers,
     total_liq_water_vol_per_area!,
-    total_energy_per_area!
+    total_energy_per_area!,
+    FrequencyBasedCallback
 using ClimaLand: PrescribedGroundConditions, AbstractGroundConditions
-using ClimaLand.Domains: Point, Plane, SphericalSurface
+using ClimaLand.Domains: Point, Plane, SphericalSurface, get_long
```
kmdeck, line 38: slightly in favor of the more verbose name, get_longitude

src/shared_utilities/models.jl
```diff
@@ -19,6 +19,7 @@ export AbstractModel,
     auxiliary_domain_names,
     make_set_initial_cache,
     make_update_cache,
+    set_historical_cache!,
```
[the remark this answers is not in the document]
kmdeck: Im confused why this is defined here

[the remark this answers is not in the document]
kmdeck: (I think we dont need this for general models - it can be very tailored to the pmodel for now - so - not defined here)

src/standalone/Vegetation/solar_induced_fluorescence.jl
```diff
@@ -89,6 +153,65 @@ function update_SIF!(p, Y, sif_model::Lee2015SIFModel, canopy)
...
+        Jmax::FT,
+        J::FT,
+        T_freeze::FT,
+        sif_parameters::SIFParameters{FT},
+    ) where {FT}
+    
+Computes the Solar Induced Fluorescence (SIF) at 755 nm in W/m^2 using the Lee et al. 2015 model.
+This function is used for the PModel photosynthesis model, which computes Jmax and J in the cache.
+"""
+function compute_SIF_at_a_point(
```
kmdeck, line 172: I think it would be clearer to rename this compute_SIF_at_a_point_pmodel. Otherwise we rely on dispatch on the number of arguments to pick the right method.

src/standalone/Vegetation/solar_induced_fluorescence.jl
```diff
@@ -68,14 +68,78 @@ function update_SIF!(p, Y, sif_model::Lee2015SIFModel, canopy)
...
     (; λ_γ_PAR,) = canopy.radiative_transfer.parameters
-    energy_per_mole_photon_par = planck_h * c / λ_γ_PAR * N_a
     T_freeze = LP.T_freeze(earth_param_set)
     R = LP.gas_constant(earth_param_set)
-
-    (; ΔHJmax, To, θj, ϕ) = canopy.photosynthesis.parameters
     sif_parameters = sif_model.parameters
+
+    APAR = @. lazy(compute_APAR(f_abs_par, par_d, λ_γ_PAR, c, planck_h, N_a))
+    _update_SIF_dispatch!(
```
kmdeck, line 76: Im not sure about this _ notation. What does it indicate? If we add it here, do we need to add it to all other update! functions?

src/standalone/Vegetation/stomatalconductance.jl
```diff
@@ -108,3 +109,68 @@ function MedlynConductanceParameters(
...
+    @. p.canopy.conductance.r_stomata_canopy =
+        1 / (
+            upscale_leaf_conductance(
+                pmodel_gs_h2o(χ, c_co2_air, An, Drel), # leaf level conductance in mol H2O Pa^-1
+                LAI,
+                T_air,
+                R,
+                P_air,
+            ) + eps(FT)
+        ) # avoids division by zero, since conductance is zero when An is zero
```
kmdeck, line 175: ah, and we dont need to worry about this in the non-p-model because there is a g0, is that right?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -26,28 +32,28 @@ Base.@kwdef struct PModelParameters{FT <: AbstractFloat}
     ϕa1::FT
     """Second order term in temp-dependent intrinsic quantum yield (K^-2)."""
     ϕa2::FT
+    """Timescale parameter used in EMA for acclimation of optimal photosynthetic capacities (unitless).
```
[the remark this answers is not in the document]
kmdeck: since we update the values at noon, is it implicitly in units of 1/d?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -26,28 +32,28 @@ Base.@kwdef struct PModelParameters{FT <: AbstractFloat}
...
+    pc::FT
 end
 
-
 """
     PModelConstants{FT<:AbstractFloat}
 
 The required constants for P-model (Stocker et al. 2020). These are physical
 or biochemical constants that are not expected to change with time or space.
-
```
kmdeck, line 37: seems helpful! can we keep this comment?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -79,21 +85,28 @@ Base.@kwdef struct PModelConstants{FT}
...
-    """Intercellular O2 mixing ratio (mol mol^-1)"""
+    """Intercellular O2 mixing ratio (unitless)"""
     oi::FT
     """First order coefficient for temp-dependent Rd (K^-1)"""
     aRd::FT
     """Second order coefficient for temp-dependent Rd (K^-2)"""
     bRd::FT
     """Constant factor appearing the dark respiration term for C3 plants (unitless)"""
     fC3::FT
+    """Planck constant (J s)"""
```
kmdeck, line 96: This is minor, but we should have access to the `earth_param_set` wherever these are needed. can we get them from there, instead of duplicating?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +421,604 @@ function compute_full_pmodel_outputs(
...
+            bRd,
+            fC3,
+            planck_h,
+            lightspeed,
+            N_a,
+        ) = constants
+
+        # Compute intermediate values
+        ϕ0 =
+            isnan(ϕ0) ? intrinsic_quantum_yield(T_canopy, ϕc, ϕa0, ϕa1, ϕa2) :
```
kmdeck, line 512: is NaN acceptable for phi0? If not, is there a way to adjust the computation of it so that is it not NaN, and we dont need to do the check here?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +421,604 @@ function compute_full_pmodel_outputs(
...
+    T_canopy = canopy_temperature(canopy.energy, canopy, Y0, p)
+    VPD = @. lazy(
+        ClimaLand.vapor_pressure_deficit(
+            p.drivers.T,
+            p.drivers.P,
+            p.drivers.q,
+            LP.thermodynamic_parameters(canopy.parameters.earth_param_set),
+        ),
+    )
+    I_abs = @. lazy(
```
kmdeck, line 703: how come I_abs = compute_APAR(..)? would it be clearer to call this APAR?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +421,604 @@ function compute_full_pmodel_outputs(
...
+            constants.To,
+            constants.Ha_Jmax,
+            constants.Hd_Jmax,
+            constants.aS_Jmax,
+            constants.bS_Jmax,
+            constants.R,
+        )
+
+    @. J = electron_transport_pmodel(
+        isnan(parameters.ϕ0) ?
```
kmdeck, line 957: same question about the NaN here

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +421,604 @@ function compute_full_pmodel_outputs(
...
+            constants.To,
+            constants.Ha_Vcmax,
+            constants.Hd_Vcmax,
+            constants.aS_Vcmax,
+            constants.bS_Vcmax,
+            constants.R,
+        )
+
+    # rubisco limited assimilation rate
+    @. Ac =
```
kmdeck, line 982: can these be lazy and not stored in the cache, or do you need them later?

src/standalone/Vegetation/stomatalconductance.jl
ychnli: yes, that is correct

src/standalone/Vegetation/pmodel.jl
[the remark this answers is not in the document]
ychnli: sorry, you're right to point out this confusion. $\alpha$ is unitless, but the EMA equation $\bar{x}_{t+1} = \alpha\bar{x}_t + (1- \alpha) x_t$ is just the forward Euler discretization of the ODE: $\tau \dfrac{d \bar{x}}{dt} = \bar{x}(t) - x(t)$ where $\alpha = 1 - dt/\tau$. Here the callback updates the equation every local noon, so dt = 1 day, as you point out. $\tau$ = 15 days corresponds to the default $\alpha = 0.933$. I will include this in the documentation and can include a comment in the code

src/standalone/Vegetation/pmodel.jl
ychnli: NaN is acceptable. If NaN, we calculate phi_0 via the temperature-dependent form; otherwise, if specified to a constant by the user, it is taken to be constant

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +421,604 @@ function compute_full_pmodel_outputs(
...
+                T_canopy,
+                T_canopy,
+                To,
+                Ha_Jmax,
+                Hd_Jmax,
+                aS_Jmax,
+                bS_Jmax,
+                R,
+            )
+        verbose && print(
```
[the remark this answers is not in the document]
ychnli: I was mainly using it for debugging. can get rid of it

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +421,604 @@ function compute_full_pmodel_outputs(
...
+        T_canopy,
+        constants.To,
+        P_air,
+        constants.R,
+        constants.ΔHΓstar,
+        constants.Γstar25,
+    )
+    ci = (ξ_opt * ca_pp + Γstar * sqrt(VPD)) / (ξ_opt + sqrt(VPD))
+
+    return (;
```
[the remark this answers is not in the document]
ychnli: They can be and are not needed, but we decided to do it this way because these get repeatedly used throughout the update_photosynthesis function. Each get used ~5 times in subsequent functions. I think we decided to expand the cache rather than repeat calculation. But the calculation isn't very expensive, so I can also do this approach if you think we should rather keep the cache lean

src/standalone/Vegetation/Canopy.jl
```diff
@@ -461,6 +537,14 @@ function CanopyModel{FT}(;
         @assert typeof(boundary_conditions.atmos) <: PrescribedAtmosphere{FT}
     end
 
+    if typeof(photosynthesis) <: PModel{FT}
+        @assert typeof(conductance) <: PModelConductance{FT} "When using PModel for photosynthesis, you must also use PModelConductance for stomatal conductance"
+    end
+
+    if typeof(conductance) <: PModelConductance{FT}
+        @assert typeof(photosynthesis) <: PModel{FT} "When using PModelConductance for stomatal conductance, you must also use PModel for photosynthesis"
+    end
```
juliasloan25, line 546: A more concise way to write this check:
```suggestion
(photosynthesis isa PModel{FT} && conductance isa PModelConductance{FT}) ||
    @error("The PModel option must be used either for both photosynthesis and conductance, or for neither.")
```

src/standalone/Vegetation/solar_induced_fluorescence.jl
juliasloan25: A leading `_` is usually reserved for internal (not user-facing functions) in Julia. This function is technically internal, but as Kat points out we have many internal `update!` functions that don't use this syntax - I think it's better to be consistent with those

test/standalone/Vegetation/test_pmodel.jl
juliasloan25: Hmm it should be working with lat/lon domains. If you're still running into problems with it after rebasing on current main can you show me the error?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -337,4 +421,604 @@ function compute_full_pmodel_outputs(
...
+        T_canopy,
+        p.drivers.P,
+        VPD,
+        p.drivers.c_co2,
+    )
+
+    # compute instantaneous max photosynthetic rates and assimilation rates 
+    @. Jmax =
+        p.canopy.photosynthesis.OptVars.Jmax25_opt * inst_temp_scaling(
+            T_canopy,
```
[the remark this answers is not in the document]
ychnli: this function takes both the current temperature and the "growth temperature". We take both to be the canopy temperature. See docstring:

```
"""
    inst_temp_scaling(
        T_canopy::FT,
        T_acclim::FT = T_canopy,
        To::FT,
        Ha::FT,
        Hd::FT,
        aS::FT,
        bS::FT,
        R::FT
    ) where {FT}

Given Vcmax or Jmax that have acclimated according to T_acclim, this function computes
the instantaneous temperature scaling factor f ∈ [0, ∞) for these maximum rates at the 
instantaneous current temperature T_canopy. To is a reference temperature for the constants
and should be set to 298.15 K (25 °C). By default we assume that T_acclim = T_canopy. 

The parameters (`Ha`, `Hd`, `aS`, `bS`) come from Kattge & Knorr (2007)
"""
```

src/standalone/Vegetation/Canopy.jl
juliasloan25: oops yeah it should be:
```suggestion
(photosynthesis isa PModel{FT} && conductance isa PModelConductance{FT}) ||
    (!(photosynthesis isa PModel{FT}) && !(conductance isa PModelConductance{FT}))
    @error("The PModel option must be used either for both photosynthesis and conductance, or for neither.")
```

kmdeck (approved):
thank you for adding this! It looks great.

PR #1282, opened 2025-07-30
Use global MODIS LAI for all simulations; remove `modis_lai_fluxnet_sites` artifact

### LINKED ISSUE #1224: change LAI artifact used for fluxnet simulations
Up to now, we've been using a MODIS LAI artifact distributed by ORNL for fluxnet simulations. This artifact (which was recently added to [ClimaArtifacts](https://github.com/CliMA/ClimaArtifacts/tree/main/modis_lai_fluxnet_sites)) only has data for 4 sites: US-Ha1, US-MOz, US-NR1, and US-Var. Since we want to be able to run at any site in the world, it would be better to use a global map of LAI and select the value closest to the lat/lon point we're running at.

Instead of using `modis_lai_fluxnet_sites` (which limits the sites we can run at), we should use the artifact `modis_lai`. To run at specific sites, we'll need a way to select the LAI value at the location closest to the lat/lon point we're running at. We can likely use ClimaUtilities.SpaceVaryingInput to do this.

juliasloan25:
## Purpose 
We've been using 2 different MODIS LAI artifacts: one global map for global runs, and one dataset for 4 fluxnet sites to use for site-level simulations. After #1279, we have the ability to read in spatial maps to Point/Column spaces, so we can use the global map for global and site-level runs.

This PR switches all site-level (fluxnet) runs to use the global map, and removes the site-specific LAI artifact.

Closes #1224

## Content
- [x] rename `find_modis_year_paths` -> `modis_lai_multiyear_paths`
- [x] use `modis_lai_multiyear_paths` instead of `modis_lai_single_year_path` (if we have multiple years of data we may as well use it since this artifact isn't too large). Single year artifact still exists and is tested
- [x] `modis_lai_multiyear_paths` only includes an extra year of data if the end date is in December, instead of always including an extra year of data
- [x] in `modis_lai_multiyear_paths`, remove `PeriodicCalendar` extrapolation from `LinearInterpolation`. Since the data spans multiple years, we get uneven time intervals (e.g. between last date of one year and first date of the next year) so we can't use `PeriodicCalendar` with it
- [x] rename `final_date` -> `end_date` in src/Artifacts.toml for consistency
- [x] always pass `longlat` to domain for fluxnet simulations (this required some restructuring of drivers)

## LAI values at the start of the simulation
| site ID | site-level LAI artifact (main) | global LAI artifact (this PR) |
| :------ | :--------------------------: | --------------------------: |
| US-MOz | 0.320855   | 0.343325  |
| US-Ha1 | 0.594469 | 0.532508  |
| US-Var | .0962558  | 1.0596 |
| US-NR1 | 0.488428 | 0.40729  |

Note that the values from the global LAI artifact are identical whether we have `use_lowres_clm` true or false.

## Changes in results
For most variables output from the 4 fluxnet sites, the R-squared value compared to data remained the same (variables are LHF, SHF, LWU, SWU, GPP, ET). RMSD values mostly decreased (suggesting better agreement with data) but RMSD changes range from 2.3% increase (Ha1 GPP) to 20% decrease (Ozark GPP). Overall it seems like this change slightly improves results, but does not have a strong effect.

----

### DIFF
--- a/docs/src/tutorials/integrated/soil_canopy_tutorial.jl
+++ b/docs/src/tutorials/integrated/soil_canopy_tutorial.jl
@@ -62,28 +62,6 @@ FluxnetSimulationsExt =
 const FT = Float32;
 earth_param_set = LP.LandParameters(FT);
 
-# Setup the domain for the model:
-
-nelements = 10
-zmin = FT(-2)
-zmax = FT(0)
-f_root_to_shoot = FT(3.5)
-plant_ν = FT(2.46e-4)
-n_stem = Int64(1)
-n_leaf = Int64(1)
-h_stem = FT(9)
-h_leaf = FT(9.5)
-compartment_midpoints = [h_stem / 2, h_stem + h_leaf / 2]
-compartment_surfaces = [zmax, h_stem, h_stem + h_leaf]
-land_domain = Column(; zlim = (zmin, zmax), nelements = nelements);
-
-# - We will be using prescribed atmospheric and radiative drivers from the
-#   US-MOz tower, which we read in here. We are using prescribed
-#   atmospheric and radiative flux conditions, but it is also possible to couple
-#   the simulation with atmospheric and radiative flux models. We also
-# read in the observed LAI and let that vary in time in a prescribed manner.
-
-
 # First provide some information about the site
 site_ID = "US-MOz"
 # Timezone (offset from UTC in hrs)
@@ -96,6 +74,33 @@ long = FT(-92.2000) # degree
 
 # Height of the sensor at the site
 atmos_h = FT(32) # m
+
+# Setup the domain for the model:
+nelements = 10
+zmin = FT(-2)
+zmax = FT(0)
+f_root_to_shoot = FT(3.5)
+plant_ν = FT(2.46e-4)
+n_stem = Int64(1)
+n_leaf = Int64(1)
+h_stem = FT(9)
+h_leaf = FT(9.5)
+compartment_midpoints = [h_stem / 2, h_stem + h_leaf / 2]
+compartment_surfaces = [zmax, h_stem, h_stem + h_leaf]
+land_domain =
+    Column(; zlim = (zmin, zmax), nelements = nelements, longlat = (long, lat));
+
+# Specify the time range and dt value over which to perform the simulation.
+t0 = Float64(150 * 3600 * 24)# start mid year
+N_days = 100
+tf = t0 + Float64(3600 * 24 * N_days)
+dt = Float64(30)
+
+# - We will be using prescribed atmospheric and radiative drivers from the
+#   US-MOz tower, which we read in here. We are using prescribed
+#   atmospheric and radiative flux conditions, but it is also possible to couple
+#   the simulation with atmospheric and radiative flux models. We also
+# read in the observed LAI and let that vary in time in a prescribed manner.
 
 # Forcing data
 (; atmos, radiation) = FluxnetSimulationsExt.prescribed_forcing_fluxnet(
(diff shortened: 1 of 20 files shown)


experiments/integrated/fluxnet/fluxnet_domain.jl
```diff
@@ -8,7 +8,3 @@ h_canopy = h_stem + h_leaf
 compartment_midpoints =
     n_stem > 0 ? [h_stem / 2, h_stem + h_leaf / 2] : [h_leaf / 2]
 compartment_surfaces = n_stem > 0 ? [zmax, h_stem, h_canopy] : [zmax, h_leaf]
-
-land_domain =
```
juliasloan25, line 12: I moved this into the driver scripts so that we have access to lat and long when we construct it. This file will be removed soon in a different PR, so it doesn't matter that this is duplicating some code.

docs/src/tutorials/integrated/soil_canopy_tutorial.jl
```diff
@@ -223,12 +230,25 @@ photosynthesis_args =
...
-(; LAI, maxLAI) =
-    FluxnetSimulationsExt.prescribed_LAI_fluxnet(site_ID, start_date)
-maxLAI = FT(maxLAI) # convert to the float type of our simulation
+
+# Read in LAI from MODIS data
+surface_space = land_domain.space.surface
+modis_lai_ncdata_path = ClimaLand.Artifacts.modis_lai_multiyear_paths(;
+    context = ClimaComms.context(surface_space),
+    start_date = start_date + Second(t0),
+    end_date = start_date + Second(t0) + Second(tf),
```
ph-kev, line 239: I find it strange that `start_date` isn't actually the start date.

ph-kev (approved):
I am not sure if I am the best person for reviewing this, but everything looks reasonable to me.

imreddyTeja (commented):
Some of the results are slightly changed, which is probably expected.

Have you checked how much/if the LAI values differ for the fluxnet sites when using modis_lai_fluxnet_sites vs the global modes artifact?

juliasloan25:
> Some of the results are slightly changed, which is probably expected.
> 
> Have you checked how much/if the LAI values differ for the fluxnet sites when using modis_lai_fluxnet_sites vs the global modes artifact?

I added a table to the PR description showing the LAI values at the start of the simulation for each site, using the site-level artifact vs the global artifact. They look pretty similar to me. I didn't go through and check the values over time, but I think this is a good enough check to see that the values are reasonable.

imreddyTeja (approved):

juliasloan25:
I realized that the way I'm computing `maxLAI` doesn't recover the previous behavior. Since we evaluate the TVI at a point at the start date to get `LAI_init`, taking `maximum(LAI_init)` just returns the initial value at that point. I can't find a straightforward way to take a maximum of the TVI values over time, and since we use the global artifact now it also isn't very straightforward to take a maximum of the dataset values only at this point in space. We would have to find the index of the lat/lon in the dataset that are closest to our space, and take a maximum over all values at that point over time. This would defeat the purpose of using ClimaUtilities to handle the interpolation and file reading.

I would prefer to stop using `maxLAI` to set `RAI` and `capacity` - maybe we can just use the initial LAI instead? But I'd like to discuss this with Kat before making the change, so I'll wait to merge until next week

ext/fluxnet_simulations/forcing.jl
```diff
@@ -198,32 +200,33 @@ function prescribed_forcing_fluxnet(
...
-    return (; LAI, maxLAI)
+function get_maxLAI_at_site(ncd_path, lat, long)
+    NCDataset(ncd_path) do ds
+        # Get the spacing of latitude and longitude, assuming each is constant
+        dlat = ds["lat"][2] - ds["lat"][1]
+        dlong = ds["lon"][2] - ds["lon"][1]
+
+        # Find the indices of the closest latitude and longitude in the dataset
+        # to the given lat and long
+        j = findfirst(x -> abs(x - lat) < dlat, ds["lat"][:])
```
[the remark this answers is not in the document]
juliasloan25: yes good idea!

PR #1286, opened 2025-08-04
Add modules in src for LandSimVis and FluxnetSimulations

### LINKED ISSUE #1289: Add modules in src for LandSimVis and FluxnetSimulations
These are easier to do than compared to NeuralSnow.

ph-kev:
closes #1289

This PR adds modules in src for FluxnetSimulations and LandSimVis.

As a result, instead of typing
```julia
LandSimulationVisualizationExt =
    Base.get_extension(
        ClimaLand,
        :LandSimulationVisualizationExt,
    ).LandSimulationVisualizationExt;
```
you can now use
```julia
import ClimaLand.LandSimVis as LandSimVis
```

It is a little difficult to do the same for NeuralSnow, so it will be done in a separate PR.

### DIFF
--- a/docs/src/tutorials/integrated/soil_canopy_tutorial.jl
+++ b/docs/src/tutorials/integrated/soil_canopy_tutorial.jl
@@ -53,8 +53,7 @@ using ClimaLand.Canopy.PlantHydraulics
 import ClimaLand
 import ClimaLand.Parameters as LP
 using DelimitedFiles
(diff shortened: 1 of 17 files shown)


kmdeck (approved):
this is a really nice change! Thanks! now I understand why ClimaUtilities is set up this way

juliasloan25 (approved):

Issue #1212, opened 2025-07-07
ongoing docs improvements


juliasloan25:
Remaining documentation improvements, to be done before August 2025.

## To Do 
- add pages about:
  - running with prescribed atmospheres
  -  timestepping information, or at least pointer to CTS.jl (maybe timestepping.jl tutorial is sufficient, but it should be more visible)
  - types of boundary conditions available
  - ClimaLand domains, how to set up sims on each
- update page: running your first simulation
  - replace canopy GPP example with example setting up and running EnergyHydrology sim using default parameterizations and plotting, after KMD constructor is merged
- update page: model structure - add information about simulations, once implemented/in use
  - LandSimulation docstring has good info to start with, but we should keep docs consistent with what's currently used in tutorials/experiments
- (other) use .buildkite env for docs too?

## Desired table of contents
Right now our docs have so many tabs, it might be hard to know where to start. We could restructure to something like:
- Running your first simulation (installation + simple example)
- ClimaLand Simulations (many of our current tutorials, running on MPI, etc)
- Developing with ClimaLand (many of our current tutorials, reorganized)
- Calibrating ClimaLand Parameterizations
- Contributing
- FAQ (we could have question link to miscellaneous other pages? E.g. how to diagnostics work? What units does ClimaLand use? How do I restart a simulation?)

PR #1288, opened 2025-08-05
Fluxnet and global run documentation

kmdeck:
## Purpose
Improve documentation for fluxnet and global runs; simplify table of contents in docs
preview: https://clima.github.io/ClimaLand.jl/previews/PR1288/

Bucket Plots:
[bucket_figures.pdf](https://github.com/user-attachments/files/21759022/bucket_figures.pdf)
[bucket_annual_timeseries.pdf](https://github.com/user-attachments/files/21759023/bucket_annual_timeseries.pdf)

Snow land fluxnet:
[US_NR1_variable_timeseries.pdf](https://github.com/user-attachments/files/21759185/US_NR1_variable_timeseries.pdf)
[US_NR1_diurnal_timeseries.pdf](https://github.com/user-attachments/files/21759186/US_NR1_diurnal_timeseries.pdf)

Soil canopy
[US_MOz_diurnal_timeseries.pdf](https://github.com/user-attachments/files/21760378/US_MOz_diurnal_timeseries.pdf)

## To-do

Future PRs
**- Change default plotting to png - otherwise we cannot show pdf in doc markdown**
## Content
docs: 
- Rearrange docs a bit: create a fluxnet simulation group, and a global sim group, in addition to standalone model group
- add fluxnet documentation (soil canopy and snowy land runs, visualization, forcing data)
- add global run documentation (bucket + snowy land)
- added stand in pages for things we know we need to add
To review: @juliasloan25  all docs/ changes and src/integrated/soil_canopy_model.jl
@AlexisRenchon all docs/ changes and src/integrated/soil_canopy_model.jl

fluxnet and simulations:
- Add make_set_fluxnet_ic function to hide closure from user and be more in line with how we create the ic function in the global case
- added start_date to Simulation struct (used in plotting fluxnet runs)
- small other changes
- @imreddyTeja could review: - ozark_pft.jl
- ozark_soilsnow.jl
- run_fluxnet.jl
- ozark_conservation.jl
- ext/FluxnetSimulationsExt.jl
- ext/LandSimulationVisualizationExt.jl
- ext/fluxnet_simulations/initial_conditions.jl
 - ext/land_sim_vis/plotting_utils.jl
- src/simulations/Simulations.jl

Moved to PR 1315:
- src/standalone/Vegetation/Canopy.jl
- test/integrated/full_land.jl
-test/integrated/full_land_utils.jl
- test/standalone/Soil/Biogeochemistry/biogeochemistry_module.jl
- 

----


src/simulations/Simulations.jl
```diff
@@ -100,12 +101,19 @@ function LandSimulation(
         ),
     ),
 )
-    if !isnothing(diagnostics) &&
-       !isempty(diagnostics) &&
-       first(diagnostics).output_writer.output_dir != outdir
-        @warn "Note that the kwarg outdir and outdir used in diagnostics are inconsistent; using $(first(diagnostics).output_writer.output_dir)"
+    if !isnothing(diagnostics) && !isempty(diagnostics)
+        try
+            ow = first(diagnostics).output_writer.output_dir
```
kmdeck, line 106: this is required since the NetCDFWriter has an output dir (writes to file), while the DictWriter does not.

src/simulations/Simulations.jl
```diff
@@ -171,6 +179,7 @@ function LandSimulation(
     return LandSimulation(
         model,
         timestepper,
+        start_date,
```
kmdeck, line 182: I added this since it was convenient to access from the simulation for the fluxnet default plotting. It's not stored otherwise, but seems like a useful number to be able to access

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -133,9 +133,9 @@ end
 
 
 """
-SoilCO2Model{FT}(;
+SoilCO2Model{FT}(
         domain::ClimaLand.AbstractDomain,
```
kmdeck, line 137: I changed this since our new constructors for Soil and Canopy take the forcing and the domain as arguments, not kwargs. a TODO is to make the forcing for the soil co2 model take the form of a namedtuple.

juliasloan25:
> We should support periodic BC for fluxnet forcing - can we do that?

I don't completely understand how we read in the fluxnet forcing. We can have periodic BCs if we use the TVI constructor we use for `prescribed_forcing_era5` instead of the one we use for `prescribed_forcing_fluxnet`, but I'm not sure if there's something prohibiting us from using that.

kmdeck:
> > We should support periodic BC for fluxnet forcing - can we do that?
> 
> I don't completely understand how we read in the fluxnet forcing. We can have periodic BCs if we use the TVI constructor we use for `prescribed_forcing_era5` instead of the one we use for `prescribed_forcing_fluxnet`, but I'm not sure if there's something prohibiting us from using that.

I wasnt sure because when I worked on this the `time` for the TimeVaryingInput0D had to be a float, in which case a periodic calendar doesnt make sense...but perhaps we can do this now. Ill try in a later PR

src/simulations/Simulations.jl
```diff
@@ -137,10 +138,12 @@ function LandSimulation(
         ),
         ClimaLand.ReportCallback(1000),
     ),
-    diagnostics = ClimaLand.default_diagnostics(model, t0, outdir),
+    diagnostics = ClimaLand.default_diagnostics(model, t0.epoch, outdir),
```
kmdeck, line 141: @imreddyTeja without this change the diagnostics error, with this:
```
┌ Warning: cannot parse time units `seconds since 0.0 seconds (2008-01-01T00:00:00) [counter = 0, period = 1 second, epoch = 2008-01-01T00:00:00]`: ArgumentError: invalid base 10 digit '.' in "0.0"
└ @ CommonDataModel ~/.julia/packages/CommonDataModel/HZ4nE/src/cfvariable.jl:144
┌ Warning: cannot parse time units `seconds since 0.0 seconds (2008-01-01T00:00:00) [counter = 0, period = 1 second, epoch = 2008-01-01T00:00:00]`: ArgumentError: invalid base 10 digit '.' in "0.0"
└ @ CommonDataModel ~/.julia/packages/CommonDataModel/HZ4nE/src/cfvariable.jl:144
┌ Error: ClimaLand simulation crashed. Stacktrace for failed simulation
│   exception =
│    MethodError: no method matching +(::ClimaUtilities.TimeManager.ITime{Int64, Second, DateTime}, ::Millisecond)
│    The function `+` exists, but no method is defined for this combination of argument types.
│    
│    Closest candidates are:
│      +(::Any, ::Any, ::Any, ::Any...)
│       @ Base operators.jl:596
│      +(::ChainRulesCore.NotImplemented, ::Any)
│       @ ChainRulesCore ~/.julia/packages/ChainRulesCore/Vsbj9/src/tangent_arithmetic.jl:24
│      +(::Any, ::ChainRulesCore.NotImplemented)
│       @ ChainRulesCore ~/.julia/packages/ChainRulesCore/Vsbj9/src/tangent_arithmetic.jl:25
│      ...
│    
│    Stacktrace:
│      [1] write_field!(writer::ClimaDiagnostics.Writers.NetCDFWriter{Tuple{Int64, Int64, Int64}, Array{Float64, 3}, Dict{ClimaDiagnostics.ScheduledDiagnostics.ScheduledDiagnostic, Array}, Nothing, ClimaDiagnostics.Writers.LevelsMethod, ClimaUtilities.TimeManager.ITime{Int64, Second, DateTime}}, field::ClimaCore.Fields.Field{ClimaCore.DataLayouts.IJFH{Float64, 1, Array{Float64, 4}}, ClimaCore.Spaces.SpectralElementSpace2D{ClimaCore.Grids.SpectralElementGrid2D{ClimaCore.Topologies.Topology2D{ClimaComms.SingletonCommsContext{ClimaComms.CPUSingleThreaded}, ClimaCore.Meshes.Equiang...
```

imreddyTeja: Was it just one simulation that caused this, or all of them?

AlexisRenchon (changes requested):
This is a dramatic improvement of ClimaLand documentation, thank you Kat! 
I left a bunch of comments below, I don't think we need to address them all in this PR, I leave it up to you. 
I can go through another review later on today if needed, let me know when you think it is ready to merge!


Note: the getting_started.md page (not modified in this PR) has a typo, we could correct it here. Line 17, "journet" instead of "journey". 

In getting_started.md, there's an error due to "About.jl" not being installed. I know we removed it a while ago because of a bug. Maybe it's been fixed now? (I can do this in another PR)

the getting_started.md page is called "running your first simulation" yet doesn't have code to run a simulation (also for another PR maybe - I thought Julia had a PR for it but I can't find it anymore)

The file structure in the code doesn't match the menu structure of the docs. it makes it a bit hard to navigate. 
docs menu: 
image
docs code: 
image

The page `changing_canopy_parameterizations` is empty
image

Maybe we don't want to show the output (or silence warning) of some box of code, e.g., running fluxnet simulations: 
image

Note this seems quite important - some box of warnings are very long... (if needed I can help with this)

The `driver_tutorial` page has a long return box that should be hidden
image

Can we rename "Leaderboard" to "Benchmark"? Benchmark is a well known modelling exercise, to compare model output to observation. Leaderboard is more an internal CliMA goal of showing we're the best model... 

Note: I see that Kevin added warning boxes in the calibration docs. I think this is pretty neat and we should use it throughout the docs 

image

APIs/shared_utilities, simulations, has a blank name (-) (which should be "LandSimulations" I believe)
image

APIs/Soil_Albedo rendering is bugged. It seems to be full of titles (hashtags) 
image

src/simulations/Simulations.jl
kmdeck: I believe it is any long run or sim using the NetCDFWriter. Particularly here it is the new bucket tutorial

docs/src/tutorials/global/bucket.jl
```diff
@@ -0,0 +1,139 @@
...
+# This should have the argument structure (Y,p,t, model)
+# in order to be used by the `LandSimulation` struct, below:
+function set_ic!(Y, p, t, bucket)
+    temp_anomaly_amip(coord) = 40 * cosd(coord.lat)^4
+    T_sfc_0 = 271.0
+    cds = ClimaCore.Fields.coordinate_field(Y.bucket.T)
+    @. Y.bucket.T = T_sfc_0 + temp_anomaly_amip(cds)
+    Y.bucket.W .= 0.15
+    Y.bucket.Ws .= 0.0
+    Y.bucket.σS .= 0.0
```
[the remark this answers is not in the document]
kmdeck: Our simulations struct requires a function of the form set_ic!(Y,p,t,model), so Im not sure how else to do it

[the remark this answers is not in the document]
kmdeck: ill make the suggested coord change though!

docs/src/tutorials/integrated/fluxnet_data.jl
```diff
@@ -0,0 +1,178 @@
+# # Fluxnet forcing data: LAI, radiation, and atmospheric variables
+
+# To access the forcing data (LAI from MODIS, SW_d, LW_d, T_air, q_air,
```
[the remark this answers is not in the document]
kmdeck: nice! thanks

src/integrated/soil_canopy_model.jl
```diff
@@ -75,6 +75,85 @@ struct SoilCanopyModel{
...
+    ),
+    soilco2 = Soil.Biogeochemistry.SoilCO2Model{FT}(
+        domain,
+        Soil.Biogeochemistry.SoilDrivers(
+            Soil.Biogeochemistry.PrognosticMet(soil.parameters),
+            PrescribedSoilOrganicCarbon{FT}(TimeVaryingInput((t) -> 5)),
+            forcing.atmos,
+        ),
+    ),
+    canopy = Canopy.CanopyModel{FT}(
```
imreddyTeja, line 140: Do these need to be `kwargs`? This follows the pattern of the convenience `LandModel` constructor, but how often do only one or two of these need to be overwritten? It seems like they could also just be constructed in the function body.

imreddyTeja (commented):
The new `make_set_fluxnet_initial_conditions` is much cleaner than defining it is the simulation script! 

I left a few comments, mostly about the assumption that `start_date = t0.epoch` (which is almost always the case).

docs/src/tutorials/standalone/Canopy/changing_canopy_parameterizations.jl
```diff
@@ -0,0 +1 @@
+
```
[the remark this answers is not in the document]
kmdeck: I added this to #1212 and also moved Alexis' comments there (that are not fixed here)

src/integrated/soil_canopy_model.jl
kmdeck: I think we need to keep these as kwargs, otherwise the user could not change parameters, parameterizations, etc, in any of the models.

docs/src/tutorials/global/bucket.jl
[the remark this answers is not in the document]
juliasloan25: I just meant accessing coords from Y instead of taking it from the global scope

src/simulations/Simulations.jl
```diff
@@ -141,6 +142,8 @@ function LandSimulation(
     updateat = [promote(t0:(ITime(3600 * 3)):tf...)...],
     solver_kwargs = (;),
 )
+    start_date = t0.epoch
```
imreddyTeja, line 145:
```
start_date = isnothing(t0.epoch) ? nothing : date(t0)
```

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -322,6 +322,33 @@ function make_ocean_masked_annual_timeseries(
...
+Converts an ITime to a date using the epoch
+of the Itime, the counter, and the period (unit)
+of the counter. 
+
+Although the epoch can be different from the start_date,
+we usually think of the simulation time as relative to the start_date,
+and so we warn here if that is not the case 
+"""
+function time_to_date(t::ITime, start_date)
+    start_date != t.epoch &&
```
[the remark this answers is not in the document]
kmdeck: Ah ok. this makes sense. I think ideally (?) if they only set things up a simulation with times, or only with dates, this should not happen. (because we only construct sim time t from passed dates internally and can be consistent, or we construct the start date internally from t0)

src/simulations/Simulations.jl
kmdeck: I fixed this

juliasloan25:
I had the same HTTP error in the ubuntu tests - I retried and it ran ok

AlexisRenchon (approved):

PR #1290, opened 2025-08-05
use nearest neighbor interpolation

kmdeck:
## Purpose 
Switch to using nearest neighbor interpolation by default:
- for spatially varying parameters of the canopy & soil albedo. NN is better than linear because many spatially varying parameters are defined by PFT or soil class.
- for ERA5 forcing. this should let us run low res sims with low res forcing without the error induced by linearly interpolating to an unphysical atmos state
- for the mask
- for MODIS LAI for consistency.

Note that using NN for the initial conditions and soil parameters greatly changed the initial state in a way that lead to some instability. This is plausible since the spun up IC are state variables that correspond to certain parameters, and the new parameters are different. I'm not 100% sure if that is the whole explanation. This requires investigation starting from a non spun up state & future work.
@juliasloan25 @nefrathenrici for visibility
## Content

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@ ClimaLand.jl Release Notes
 
 main
 -------
-- Use nearest neighbor spatial interpolation PR[#1282](https://github.com/CliMA/ClimaLand.jl/pull/1282)
+- Use nearest neighbor spatial interpolation PR[#1290](https://github.com/CliMA/ClimaLand.jl/pull/1290)
 - Use global MODIS LAI for all simulations; remove `modis_lai_fluxnet_sites` artifact PR[#1282](https://github.com/CliMA/ClimaLand.jl/pull/1282)
 - Use ClimaUtilities v0.1.25 to read in spatial data to a point using lat/lon PR[#1279](https://github.com/CliMA/ClimaLand.jl/pull/1279)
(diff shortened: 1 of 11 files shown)


NEWS.md
```diff
@@ -3,6 +3,7 @@ ClimaLand.jl Release Notes
 
 main
 -------
+- Use nearest neighbor spatial interpolation PR[#1282](https://github.com/CliMA/ClimaLand.jl/pull/1282)
```
kmdeck, line 6: Ill update this to 1290

experiments/long_runs/snowy_land.jl
```diff
@@ -206,8 +206,8 @@ function setup_model(FT, start_date, stop_date, Δt, domain, earth_param_set)
     else
         modis_lai_ncdata_path = ClimaLand.Artifacts.modis_lai_multiyear_paths(;
             context = nothing,
-            start_date = start_date + Second(t0),
-            end_date = start_date + Second(t0) + Second(tf),
+            start_date,
+            end_date = stop_date,
```
kmdeck, line 210: this was a bug introduced in an earlier PR but not caught since we dont run long runs each step

ph-kev (approved):

juliasloan25:
Nice! Do we always use nearest neighbor now, or is there somewhere the default is overwritten?

PR #1300, opened 2025-08-07
Use ClimaParams instead of hardcoding values

### LINKED ISSUE #1298: Use ClimaParams instead of hard coded values
For doing calibration, there are instances where you want to calibrate a parameter, but the parameter is a hard coded value. For example, in the current land calibration pipeline, there are parameters that we need to create a variable for, so that it can be overwritten.

https://github.com/CliMA/ClimaLand.jl/blob/631c69bbd75aa0bb7673e96b2225e4780cec9e9c/experiments/calibration/forward_model_land.jl#L51-L75

This is not feasible in the long term and it would be better to transition to using ClimaParams for storing and loading parameters.

## Proposed solution

Instead of hardcoding values in the code, we can instead hardcode the values in a TOML file. This would look something like this:
```
[beta_snow]
value = 0.7
description = "Snow parameter for ZenithAngleAlbedoModel from some paper" # optional
tag = "snow" # optional
```

Then, you can create a TOML dict from it using `ClimaParams.create_toml_dict` and load the parameters from this instead. See [ClimaParams documentation](https://clima.github.io/ClimaParams.jl/dev/toml/) for more information.

For calibration, we can use `ClimaParams.merge_toml_files` to merge the TOML file from ClimaCalibrate and the TOML for the land parameters. Furthermore, this can significantly reduces the amount of the code used for the model interface or forward model as one of the reason that it can't reuse `snowy_land.jl` is because there is no functionality to override parameters.     

## Alternative solution

One other solution is to put all the parameters in ClimaParams. This might not be a good idea, because if any parameter ever need to be changed, then you need to go to ClimaParams, update it, make a new release, and update ClimaParams in ClimaLand.

ph-kev:
closes #1298 - Instead of using hard code values for the constructor, this allows the user to modify a TOML file instead if they want to change the parameters. Furthermore, this should simplify the code needed for the forward model for calibration as it is much easier to change parameter values.

## Limitations

- Due to how ClimaParams is implemented, all the name of the parameter values must be different.
- If there are constructors that takes in different number of parameters, it might be difficult to work with using ClimaParams.

### DIFF
--- a/.buildkite/longruns_gpu/pipeline.yml
+++ b/.buildkite/longruns_gpu/pipeline.yml
@@ -65,17 +65,6 @@ steps:
         env:
           CLIMACOMMS_DEVICE: "CUDA"
 
-      - label: ":snow_capped_mountain: Low-Res Snowy Land"
-        command:
-          - julia --color=yes --project=.buildkite experiments/long_runs/low_res_snowy_land.jl
-        artifact_paths:
-          - "lowres_snowy_land_longrun_gpu/*png"
-        agents:
-          slurm_gpus: 1
-          slurm_time: 3:00:00
-        env:
-          CLIMACOMMS_DEVICE: "CUDA"
-
       - label: ":sunglasses: California regional simulation"
         command:
           - julia --color=yes --project=.buildkite experiments/long_runs/land_region.jl


ph-kev:
I asked for a review, but this is a prototype of what using ClimaParams would look like. I still need to add it for all the runs that we do. If needed, there might a TOML file for each run if a single TOML file does not work because different parameters are used for different simulations.

experiments/long_runs/bucket.jl
```diff
@@ -106,16 +106,23 @@ domain =
...
 # Parameters
 params = LP.LandParameters(FT)
-σS_c = FT(0.2)
-W_f = FT(0.2)
-z_0m = FT(1e-3)
-z_0b = FT(1e-3)
-κ_soil = FT(1.5)
-ρc_soil = FT(2e6)
+
+name_map = (;
```
kmdeck, line 110: Could this map live in src/toml also? otherwise we might have to include the same map in every bucket run

experiments/long_runs/land_region.jl
```diff
@@ -294,7 +283,14 @@ domain = ClimaLand.Domains.HybridBox(;
...
 )
 params = LP.LandParameters(FT)
-model = setup_model(FT, context, start_date, Δt, domain, params)
+default_param_filepath =
+    joinpath(pkgdir(ClimaLand), "toml", "default_parameters.toml")
+toml_dict = CP.create_toml_dict(
+    Float64,
+    default_file = CP.merge_toml_files([default_param_filepath]),
+)
+model = setup_model(FT, context, start_date, Δt, domain, params, toml_dict)
```
kmdeck, line 292: The `params` are also accessible from the toml_dict, right? ultimately it seems redundant to have the earth_param_set and the toml dict

toml/default_parameters.toml
```diff
@@ -0,0 +1,125 @@
+# How do I add a parameter?
+# At the minimum, a parameter needs a parameter name, value, and type. All
+# parameter names must be unique. The description attribute is optional, but it
+# is recommended to add one. These parameters are loaded in ClimaLand using
+# `ClimaParams.get_parameter_values`.
+
+# TOPMODELRunoff model
+
+[f_over]
```
kmdeck, line 9: @juliasloan25 how do you think this fits in with the default values we have in the parameter structs?

experiments/long_runs/land_region.jl
ph-kev: With how it is currently implemented, `params` is not accessible from `toml_dict`, but everything could come from a single object instead of `params` and `toml_dict`. I am not sure how that would be implemented though given the current `LandParameters` struct.

toml/default_parameters.toml
juliasloan25: I do think it's a little confusing to have default parameters defined in two places. I guess the idea is to access all parameters using `get_parameter_values` (i.e. via ClimaParams + these tomls) instead of specifying some default parameters with hardcoded values in the constructors. I think that avoids defining defaults in mulitple places; they're all defined in ClimaParams/these tomls, and the default constructors just access them

ph-kev: I can add that, but I am wondering why users would want this when you can just change the value in the TOML file instead?

experiments/long_runs/bucket.jl
kmdeck: nice! That looks way cleaner. (I think it's unavoidable that some kind of name map is present, though, since the names in the ClimaParams dont match our names in the structs).

having the constructor work with the toml_dict and work with the name map there sounds like a good idea!

src/shared_utilities/Parameters.jl
```diff
@@ -156,4 +163,26 @@ function get_default_parameter(FT, climaparams_name)
...
+"""
+    create_toml_dict(FT, filepaths...; override = false)
+
+Construct a `ParamDict{FT}` struct from `filepaths`.
+
+If `override = false`, then non-unique TOML entries are not allowed. If
+`override = true`, then parameters from later TOML files in `filepaths` will
+overwrite the parameters from earlier TOML files.
+"""
+function create_toml_dict(FT, filepaths...; override = false)
```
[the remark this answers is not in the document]
ph-kev: For calibration, you would supply multiple TOML files, so it would look like

```julia
create_toml_dict(FT, default_params_filepath, calibrate_params, override = true)
```

src/standalone/Bucket/Bucket.jl
```diff
@@ -290,35 +310,34 @@ end
 function BucketModelParameters(
     toml_dict::CP.AbstractTOMLDict;
     albedo,
-    z_0m,
-    z_0b,
-    τc,
-    kwargs...,
+    W_f = toml_dict["land_bucket_capacity"],
+    f_bucket = toml_dict["bucket_capacity_fraction"],
+    z_0b = toml_dict["z_0b"],
```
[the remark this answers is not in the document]
ph-kev: I change it to this

```julia
function BucketModelParameters(
    toml_dict::CP.AbstractTOMLDict;
    albedo,
    W_f = toml_dict["land_bucket_capacity"],
    f_bucket = toml_dict["bucket_capacity_fraction"],
    z_0b = toml_dict["bucket_z_0b"],
    ρc_soil = toml_dict["bucket_soil_heat_capacity"],
    f_snow = toml_dict["critical_snow_fraction"],
    z_0m = toml_dict["bucket_z_0m"],
    σS_c = toml_dict["critical_snow_water_equivalent"],
    p = toml_dict["bucket_beta_decay_exponent"],
    κ_soil = toml_dict["bucket_soil_conductivity"],
    τc = toml_dict["tau_c"],
)
```

experiments/calibration/model_interface.jl
```diff
@@ -90,226 +83,75 @@ function setup_model(
...
-        soil_model_type = soil_model_type,
-        soil_args = soil_args,
-        canopy_component_types = canopy_component_types,
-        canopy_component_args = canopy_component_args,
-        canopy_model_args = canopy_model_args,
-        snow_args = snow_args,
-        snow_model_type = snow_model_type,
-    )
+    # Construct the land model with all default components except for snow
+    land = LandModel{FT}(forcing, LAI, toml_dict, domain, Δt; snow, canopy)
```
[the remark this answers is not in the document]
ph-kev: > Right now, some of the parameter(izations) of snow/canopy are exposed and created using the toml, and some are not, and soil is not. Could you explain why that is?

I am not too sure why this is. I moved everything to ClimaParams while trying to keep everything in the script the same. I would speculate that some of them needed additional keyword arguments, so it was constructed in the script and passed to the function.

src/standalone/Bucket/Bucket.jl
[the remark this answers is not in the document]
ph-kev: I am not sure about renaming `soil_conductivity` and `soil_heat_capacity` because they are already defined in `ClimaParams`.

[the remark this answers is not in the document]
kmdeck: thank you!

docs/src/APIs/parameters.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Store parameters for running long runs
+default_params_filepath =
+    joinpath(pkgdir(ClimaLand), "toml", "default_parameters.toml")
+toml_dict = LP.create_toml_dict(FT, default_params_filepath)
+
+# Store parameters for running bucket model
+bucket_params_filepath =
+    joinpath(pkgdir(ClimaLand), "toml", "bucket_parameters.toml")
+toml_dict = LP.create_toml_dict(FT, bucket_params_filepath)
```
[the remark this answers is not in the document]
kmdeck: thanks! this sounds good to me

src/shared_utilities/Parameters.jl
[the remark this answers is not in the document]
kmdeck: oh ok! that makes sense.

experiments/calibration/model_interface.jl
```diff
@@ -90,226 +83,75 @@ function setup_model(
...
+    canopy = ClimaLand.Canopy.CanopyModel{FT}(
+        surface_domain,
+        canopy_forcing,
+        LAI,
+        toml_dict;
+        prognostic_land_components = (:canopy, :snow, :soil, :soilco2),
+        energy,
+        hydraulics,
+        z_0m = z0_m,
+        z_0b = z0_b,
```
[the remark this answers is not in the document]
kmdeck: thanks!

toml/default_parameters.toml
```diff
@@ -0,0 +1,152 @@
...
+
+[canopy_momentum_roughness_length]
+value = 0.1
+type = "float"
+description = "Roughness length for momentum (m)"
+
+[canopy_scalar_roughness_length]
+value = 0.2
+type = "float"
+description = "Roughness length for scalars (m)"
```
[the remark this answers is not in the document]
kmdeck: I opened an issue to fix these, so what you have in this PR is fine
https://github.com/CliMA/ClimaLand.jl/issues/1351

experiments/long_runs/bucket.jl
```diff
@@ -106,19 +95,13 @@ domain =
 
 # Parameters
 params = LP.LandParameters(FT)
-σS_c = FT(0.2)
```
kmdeck, line 109: I will check if any of these are behavior changing - we can adjust the toml file if needed then in a later PR if the long runs change much

kmdeck (approved):
thank you Kevin! this is very helpful.

Good to merge after we take a look at the long runs

docs/src/APIs/parameters.md
```diff
@@ -0,0 +1,96 @@
...
+`ClimaParams` provides the function [`ClimaParams.log_parameter_information`](https://clima.github.io/ClimaParams.jl/dev/API/#ClimaParams.log_parameter_information)
+which logs the parameters used in a simulation. Before calling
+`ClimaLand.Simulations.solve!`, log the parameters and inspect the generated
+TOML file to verify that the particular parameter you're interested in is being
+used.
+
+!!! note "Logging parameters with versions above ClimaParams 0.11.1"
+    With how `ClimaParams` is used with `ClimaLand`, The
+    `ClimaParams.log_parameter_information` function requires ClimaParams
+    version above 0.11.1 to behave correctly.
```
juliasloan25, line 89: One of the pieces of feedback I've gotten at the workshop this week is that while it is helpful to have defaults provided by the model, users are very interested in easily being able to check which parameters and parameterizations the model is running with.

This section is great, but could we move it to the top of this docs page and generalize the header to "How to check parameter values used in a simulation"? We can keep this FAQ as well, with the response as a briefer reference to this ClimaParams function, and a reminder that kwargs override TOML values if both are present.

cc @kmdeck for visibility of the feedback item

ph-kev: I realized that if a keyword argument is used instead, then this function won't be able to log the parameters since it is not using the value from the dictionary.

ph-kev:
This is the buildkite: https://buildkite.com/clima/climaland-long-runs/builds/4163#_

Make note of it since I need to make some changes, but they should not affect the long runs.

AlexisRenchon (approved):

PR #1302, opened 2025-08-07
Bump to v0.18.0

imreddyTeja:
## Purpose 
Bump ClimaParams and Thermodynamics compat upper bounds

----


.buildkite/Project.toml
```diff
@@ -46,11 +46,11 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
...
 
 [compat]
-ClimaAnalysis = "0.5.17"
-ClimaCalibrate = "0.1"
+ClimaAnalysis = "0.5.19"
+ClimaCalibrate = "0.1.3"
 ClimaDiagnostics = "0.2.13"
 ClimaTimeSteppers = "0.7, 0.8"
-EnsembleKalmanProcesses = "2.4.1"
+EnsembleKalmanProcesses = "2.5.0"
```
juliasloan25, line 53: These should be able to stay at the previous compats, unless there's a reason we need to use the newer versions (e.g. they introduced a new feature we need to use). If that's not the case, we can just `up` the versions in the manifests but keep the compats as they are

juliasloan25 (approved):
Just a comment about the .buildkite compats, but other than that this looks good. Thank you!

.buildkite/Project.toml
imreddyTeja: @ph-kev

juliasloan25: ok it's fine to keep it here then

juliasloan25 (approved):

PR #1304, opened 2025-08-08
Add support for high/low res vG param maps, fix depth varying bug

### LINKED ISSUE #1303: Add support for 1 km resolution soil hydrology param maps
The Gupta 2020, 2022 soil hydrology parameter maps that we use in global runs is remapped to 1 deg x 1 deg from the native 1 km x 1 km. For fluxnet/site level runs we want to sample parameters from the higher resolution map. 

Additionally, the existing low resolution maps are depth-averaged from the native depth grid (4 depths going to 1 meter), which is a bug. 

Changes needed:
- Add support for both high res and low res maps creation in ClimaArtifacts
- Keep the depth dimension 
- Add support for accessing the high res version (which will be undownloadable and will only work on the cluster) for site-level runs

ychnli:
## Purpose 

closes #1303 

## To-do

- PR in climaArtifacts (https://github.com/CliMA/ClimaArtifacts/pull/130)
- high res data still needs to be moved into the correct directory in `/groups/esm/ClimaArtifacts/artifacts/soil_params_gupta2020_2022/`. Currently this directory still contains the old low res, depth-averaged data. 

## Content

- Add support for low/high res soil hydrology data paths in `Artifacts.jl`. Since previous global use cases all use the low res version, `lowres = true` by default. Typical use case for high res might be fluxnet/site level runs
- added a test to check if new soil maps are indeed depth varying (not depth averaged values)

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@ ClimaLand.jl Release Notes
 
 main
 -------
+- Use correct van Genuchten parameters as a function of depth PR[#1304](https://github.com/CliMA/ClimaLand.jl/pull/1304)
 - Fix divide by zero when VPD is zero PR[#1363](https://github.com/CliMA/ClimaLand.jl/pull/1363)
 - Remove Smith Optimality model; use PModel instead PR[#1361](https://github.com/CliMA/ClimaLand.jl/pull/1361)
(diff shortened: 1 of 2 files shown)


kmdeck:
@ychnli this looks good. I think this is ready to merge aside from the comment about the test.

The Clima utilities PR looks good too, I left one comment there

kmdeck (approved):

PR #1306, opened 2025-08-09
use new LandModel constructors

juliasloan25:
## Purpose 
Use the new convenience constructors we recently added for `LandModel`

## To-do
- [x] add another constructor for `LandModel` analogous to those we've added for other models
- [x] reformat SoilCO2Model constructor args to match other model constructors
- [ ] update experiments to use new `LandModel` constructors

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #1310, opened 2025-08-12
fix three bugs: one in canopy, one in run fluxnet, one in fluxnet data

kmdeck:
## Purpose 
This PR fixes three bugs:
- we now use the CLMGFunction for fluxnet runs (as of #1306), but in the method that computes G with that type, if cos(zenith angle) is negative, G can be negative. This occurs when the sun is below the horizon. This then leads to a value of K which is very large and negative, and this is not correct. K is expected to be positive. Before we used a ConstantGFunction which does not depend on zenith angle so the error did not appear.
- the fluxnet params function returns a constant GFunction and chi_l, and this is confusing. For now, in the fluxnet scripts, I removed G_Function, since we make G_function = CLMGFunction(chi_l) in the script.
- there was a bug in the comparison data which returned all NaNs if the data column was missing, but we want it to return an empty named tuple.

## Content

----

### DIFF
--- a/experiments/integrated/fluxnet/ozark_pft.jl
+++ b/experiments/integrated/fluxnet/ozark_pft.jl
@@ -58,7 +58,6 @@ site_ID_val = FluxnetSimulations.replace_hyphen(site_ID)
     soil_α_NIR,
     Ω,
     χl,
-    G_Function,
     α_PAR_leaf,
     λ_γ_PAR,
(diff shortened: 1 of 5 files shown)


juliasloan25:
cc @ThanhNguyen428 for the `get_parameters` change!

juliasloan25 (approved):
thank you!

PR #1314, opened 2025-08-12
add perturbed RH run

kmdeck:
## Purpose 
Adds perturbed RH forcing

Moves the perturbed runs to a separate buildkite run
https://buildkite.com/clima/climaland-long-runs/builds/4151

## Content

----


src/shared_utilities/drivers.jl
```diff
@@ -1082,6 +1083,46 @@ function perturbed_specific_humidity_from_dewpoint(
...
+
+Relative humidity is computed using the Magnus formula.
+
+For more information on the Magnus formula, see e.g.
+Lawrence, Mark G. (1 February 2005).
+"The Relationship between Relative Humidity and the Dewpoint Temperature in Moist Air:
+A Simple Conversion and Applications".
+Bulletin of the American Meteorological Society. 86 (2): 225–234.
+"""
+function perturbed_specific_humidity_from_dewpoint_and_rh(
```
juliasloan25, line 1101:
```suggestion
function perturbed_specific_humidity(
```
I think this is more concise and still readable, but I'm ok either way!

src/shared_utilities/drivers.jl
```diff
@@ -1082,6 +1083,46 @@ function perturbed_specific_humidity_from_dewpoint(
     return q
 end
 
+"""
+    perturbed_specific_humidity_from_dewpoint_and_rh(dewpoint_temperature, temperature, air_pressure, earth_param_set, Δrh)
```
juliasloan25, line 1087: Maybe we should move the perturbed forcing functions into a different file, e.g. `src/shared_utilities/drivers_perturbed.jl` to avoid crowding this file. Since the perturbed ones are only used for experiments

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -95,6 +83,36 @@ steps:
         env:
           CLIMACOMMS_DEVICE: "CUDA"
 
+  - group: "Perturbed Low-res Global Land Models"
+    if: build.env("PERTURBED_RUN") != null
```
juliasloan25, line 87: how does this var get set?

juliasloan25 (approved):
thank you!

src/shared_utilities/drivers.jl
kmdeck: Ill move them to a new file. If they end up being used only for the Berlin workshop, we can also remove them in the future more easily that way

.buildkite/longruns_gpu/pipeline.yml
kmdeck: you set it in buildkite - click on new build in the top right corner, and then add your branch instead of main (or use main), and then add an environment variable
PERTURBED_RUN=""

(it is also how we trigger longer runs if you want to do a one off run of those. Otherwise we set the environment variable in our scheduled long run here: https://buildkite.com/clima/climaland-long-runs/settings/schedules/c7373c40-27cc-4051-b13a-58769f62f5e7)

src/shared_utilities/drivers.jl
kmdeck: I did:
```
ClimaLand.perturbed_rh_specific_humidity_from_dewpoint
ClimaLand.perturbed_temp_specific_humidity_from_dewpoint
```

PR #1317, opened 2025-08-13
Add global P Model runs

juliasloan25:
## Purpose 
Adds 2 new runs:
- regular CI: global SoilCanopyModel with PModel
- longruns: global LandModel with PModel (analogous to snowy_land.jl)

## Content

----


src/shared_utilities/utils.jl
```diff
@@ -738,7 +737,6 @@ call the function based on the `frequency`.
 function FrequencyBasedCallback(
     frequency::Union{AbstractFloat, Dates.Period},
     start_date::Dates.DateTime,
-    t0::Union{AbstractFloat, Dates.Period},
     dt::Union{AbstractFloat, Dates.Period};
```
kmdeck, line 740: @imreddyTeja do we support dt as a Period or can I remove this?

ychnli (approved):

imreddyTeja (changes requested):
There is a long run failure, and it looks like it is not because of CUDA

kmdeck:
> There is a long run failure, and it looks like it is not because of CUDA

thanks for catching this! this a bug in plotting conservation of the soil model, it must have been introduced in one of my earlier PRs with plotting. Ill fix it today in another PR. We only check conservation in the soil model right now.

imreddyTeja (approved):
I just realized the error is unrelated to this PR

PR #1325, opened 2025-08-15
Update calibration tutorials to use `FluxnetSimulations`

nefrathenrici:
## Content

----


docs/src/tutorials/calibration/minimal_working_example.jl
```diff
@@ -1,123 +0,0 @@
...
-# In this tutorial, we demonstrate that we can recover the parameters of a previous model run via our calibration framework,
-# by calibrating a single site (Ozark) latent heat flux (LHF).
-
-# First, we run a simulation of Ozark with known values of parameters g1 and g0, which control the stomatal behavior of leaves
-# and therefore the canopy transpiration, and retrieve latent heat flux (transpiration expressed in energy).
-
-# Then, we use [EnsembleKalmanProcesses.jl](https://github.com/CliMA/EnsembleKalmanProcesses.jl) to perform a calibration
-# by giving a prior distribution of the parameters g1 and g0, different from
-# the known parameters used in the first "perfect model" run. Our goal is that, after calibraton on the perfect model latent heat flux, the parameters
-# will be close to the original ones, therefore demonstrating that we can recover the parameters of the perfect model.
```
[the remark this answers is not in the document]
kmdeck: ok perfect you say this in the new tutorial!

docs/src/tutorials/calibration/obs_site_level_calibration.jl
```diff
@@ -0,0 +1,366 @@
...
+    start_date,
+);
+
+# ## Model Setup
+# 
+# Create an integrated land model that couples canopy, snow, soil, and soil CO2
+# components. This comprehensive model allows us to simulate the full land
+# surface system and its interactions.
+
+# Set up ground conditions and define which components to simulate
```
kmdeck, line 124: Maybe
# Because we are changing the canopy parameters, we need to create a new instance of the `canopy` component model with the new parameters each time. Below we split this up into pieces that change when the parameters change, and to pieces that are constant.

kmdeck: maybe we can just move them into the modelI) function and remove this "# Set up ground conditions and define which components to simulate"

docs/src/tutorials/calibration/perfect_model_site_level_calibration.jl
```diff
@@ -0,0 +1,361 @@
...
+
+# Set up ground conditions and define which components to simulate
+# prognostically
+ground = ClimaLand.PrognosticGroundConditions{FT}()
+prognostic_land_components = (:canopy, :snow, :soil, :soilco2)
+
+# Prepare canopy domain and forcing
+canopy_domain = ClimaLand.Domains.obtain_surface_domain(domain)
+canopy_forcing = (; forcing.atmos, forcing.radiation, ground);
+
```
kmdeck, line 132: Same comment here

kmdeck (approved):
I think as long as the plots/formatting in the markdown look good, this is ready to go! Thank you!

If by chance the docs take too long to build now, you can also comment out the part where the code is runs? and show the plots by adding the png files to the repo? Not sure if it will be a problem.

PR #1328, opened 2025-08-15
Fix P-model docs

ychnli:
- fix math
- add recipes for optimal params

## Content

----


juliasloan25 (approved):
thank you!

Issue #1185, opened 2025-06-20
Add P-model to canopy photosynthesis models


ychnli:
The P-model ([Stocker 2020](https://gmd.copernicus.org/articles/13/1545/2020/)) is a light use efficiency (LUE) model based on a simple optimality criterion. The implementation should share a lot of the code in [optimality_farquhar.jl](https://github.com/CliMA/ClimaLand.jl/blob/main/src/standalone/Vegetation/optimality_farquhar.jl). I believe the main differences are 

- P-model uses a different way of computing J_max limitation 
- P-model includes a soil moisture stress prefactor \beta in the LUE computation
- P-model includes a temperature-dependent intrinsic quantum yield \phi_0(T) 
- Will need to save new aux variables

We can use output from the pmodel implementation in R (https://github.com/geco-bern/rpmodel/tree/master) for testing.

PR #1330, opened 2025-08-15
Modular soil moisture stress, standalone canopy at fluxnet sites

### LINKED ISSUE #655: Moisture factor
Revisit how we compute the moisture stress factor for photosynthesis. In particular, at the Ozark site, it looks quite bad.

### LINKED ISSUE #1205: Soil moisture stress parameterization
The P-model, which uses an empirical soil moisture stress factor $\beta \in [0, 1]$ to downweight the light use efficiency during drought conditions, takes in $\beta$ as a parameter. Currently, there are two different soil moisture stress parameterizations that can be used to model $\beta$, but they take different parameters and have different functional forms. These should be written as a parameterization `soil_moisture_stress.jl` in Vegetation.

ychnli:
## Purpose 

This PR closes #1205 and #1185 and implements methods for running standalone canopy at Fluxnet sites.

## To-do

- Plant hydraulics was previously used only to compute the leaf water potential. Now, if our moisture stress function is not a function of $\psi_{leaf}$, there is no need to solve plant hydraulics. We should allow for this flexibility when creating the Canopy model. We would need to move the root distribution function from `PlantHydraulics.jl` into soil moisture stress.
- The existing soil moisture stress functions are problematic at several fluxnet sites. We need to think about how we should resolve this for site-level simulations. 
- A global run using the piecewise soil moisture stress and prognostic soil 

### Main changes
- **Adds modular soil moisture stress**: new soil moisture stress parameterizations can be added just like other canopy component models. Instead of computing $\beta_m$ wherever it is used, it is now stored in the cache. The pre-existing soil moisture stress which was based on leaf water potential is represented by the `TuzetMoistureStressModel`.  
- **Adds an empirical soil moisture stress functional form** commonly used in other LSMs. This is computed as:
```math
\beta_m = \beta_0 \int_{-H}^0 dz\,  \left( \frac{\theta(z) - \theta_w(z)}{\theta_c(z) - \theta_w(z)} \right)^c R(z) 
```
where $\theta$, $\theta_w$, and $\theta_c$ are (possibly depth-dependent) soil water content, wilting point, and field capacity. The latter are two thresholds that can either be specified by the user or by the soil hydrology model using standard matric potentials. $c > 0$ is a curvature parameter and $\beta_0 \in [0, 1]$ is an "intrinsic moisture stress" used in Mengoli et al. ([2025](https://onlinelibrary.wiley.com/doi/abs/10.1111/gcb.70098)). $R(z)$ is the root probability distribution and integrates to 1.
- Add an option to construct a soil driver for Fluxnet, so that we can run standalone canopy at Fluxnet sites. This is added as a kwarg for `get_data_dates` and `prescribed_forcing_fluxnet`, false by default.
- Previously, we took $\psi_{root}$ as a driver when constructing `PrescribedGroundConditions`. This was used in our plant hydraulics scheme to compute leaf water potential. Now, we instead take the soil water content $\theta_{root}$, which is a variable (`SWC`) provided by most Fluxnet sites. This is converted to $\psi$ via a retention curve supplied by the user. 

### Smaller changes
- rename `FrequencyBasedCallback` to `PeriodicCallback`, which makes more sense
- When using `LandSimulation`, we call `make_PModel_callback` if we are using the P-model via a new method `required_model_callbacks` which dispatches on the land component models.
- Previously, atmospheric CO2 concentration was hardcoded to be a constant in fluxnet. Now, we read in the measured values at each site.
- In general, fluxnet2015 sites do not have all required forcing variables at the beginning of the record. Typically C_CO2 or SWC will be missing at the beginning. Thus, we now find all times where all variables are available, and take the extrema of this set to get the nominal `start_date` and `stop_date`. 
- Some fluxnet2015 sites have non-integer time offsets (e.g., in Australia), which can now be converted to a `Dates.Period`

----


kmdeck:
The majority of these changes have been implemented in other PRs

PR #1332, opened 2025-08-15
Kd/plots and offset

kmdeck:
## Purpose 
In our fluxnet plotting, we make plots comparing climaland diagnostics (saved hourly, with a timestamp at the end of the averaging hour) with half hourly data (output at the start of the averaging window). this leads to a relative phase shift between the two diurnal cycles. To be directly comparable to fluxnet we need to save diagnostics at 30 minute intervals, and return the comparison data at the end of the averaging window (TIMESTAMP_END).

This PR addresses that and fixes a typo in our conservation plots (this only affects soil global runs right now). 

@nefrathenrici @juliasloan25 

## Content

----

### DIFF
--- a/ext/fluxnet_simulations/forcing.jl
+++ b/ext/fluxnet_simulations/forcing.jl
@@ -23,7 +23,9 @@ precipitation that is in snow (if split_precip ==true), and (5)
 making the TimeVaryingInput objects.
 
 Note that the TimeVaryingInput objects can be used to interpolate in time,
-which is why we drop missing data.
+which is why we drop missing data. For the timestamp of the observation, we use
+the halfway point between the timestamp start and timestamp end:
+see https://fluxnet.org/data/fluxnet2015-dataset/fullset-data-product/ for details.
 
 This assumes that the first row of the CSV file is the list of of column names,
 and that these names are:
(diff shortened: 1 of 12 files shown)


ext/fluxnet_simulations/forcing.jl
```diff
@@ -23,7 +23,9 @@ precipitation that is in snow (if split_precip ==true), and (5)
 making the TimeVaryingInput objects.
 
 Note that the TimeVaryingInput objects can be used to interpolate in time,
-which is why we drop missing data.
+which is why we drop missing data. For the timestamp of the observation, we use
```
kmdeck, line 26: for the forcing, using the midpoint makes the most sense, since the average is over the window

ext/fluxnet_simulations/forcing.jl
```diff
@@ -268,7 +296,33 @@ function FluxnetSimulations.get_data_dates(
 )
     fluxnet_csv_path = ClimaLand.Artifacts.experiment_fluxnet_data_path(site_ID)
     (data, columns) = readdlm(fluxnet_csv_path, ','; header = true)
-    local_datetime = DateTime.(string.(Int.(data[:, 1])), "yyyymmddHHMM")
+    # Determine which column index corresponds to which varname
+    varnames = ("TIMESTAMP_START", "TIMESTAMP_END")
```
kmdeck, line 300: to be consistent, I also did the same here

ext/fluxnet_simulations/forcing.jl
```diff
@@ -345,23 +399,38 @@ end
...
+        timestamp_end = true
 )
 
 Gets and returns the a NamedTuple with the comparison
 data for the Fluxnet `site_ID`, given its hour offset from
-UTC, and the value used to indicate missing data (`val`)
+UTC, and the value used to indicate missing data (`val`), and the
+averaging period.
+
+Fluxnet as has two timestamps associated with it: TIMESTAMP_START and TIMESTAMP_END,
```
kmdeck, line 411: comparison data output at the end of the averaging window to compare with Clima diagnostics

AlexisRenchon (approved):

PR #1354, opened 2025-08-20
HPC conditional statement for addprocs, metadata

AlexisRenchon:
This commit adds a conditional statement to execute the appropriate addprocs depending on which HPC we are running the calibration on. 

@nefrathenrici I am not sure if this is the most elegant solution, or if this should be internal to ClimaCalibrate somehow?

### DIFF
--- a/experiments/calibration/run_calibration.jl
+++ b/experiments/calibration/run_calibration.jl
@@ -22,11 +22,26 @@ const CALIBRATE_CONFIG = CalibrateConfig(;
     rng_seed = 42,
 )
 
+# true solution is at 0.96
+priors =
+    [EKP.constrained_gaussian("emissivity_bare_soil", 0.82, 0.12, 0.0, 2.0)]
+prior = EKP.combine_distributions(priors)
+ekp_process = EKP.Unscented(prior)
+ensemble_size = ekp_process.N_ens
 
 if abspath(PROGRAM_FILE) == @__FILE__
-    # Note: Using this script on Derecho requires changes to addprocs to use
-    # the PBSManager
-    addprocs(ClimaCalibrate.SlurmManager())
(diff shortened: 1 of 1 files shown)


experiments/calibration/run_calibration.jl
```diff
@@ -22,11 +22,26 @@ const CALIBRATE_CONFIG = CalibrateConfig(;
...
+prior = EKP.combine_distributions(priors)
+ekp_process = EKP.Unscented(prior)
+ensemble_size = ekp_process.N_ens
 
 if abspath(PROGRAM_FILE) == @__FILE__
-    # Note: Using this script on Derecho requires changes to addprocs to use
-    # the PBSManager
-    addprocs(ClimaCalibrate.SlurmManager())
+    # addprocs is slightly different on derecho or central. Could add more.
+    if contains(gethostname(), "derecho") # Derecho
```
nefrathenrici, line 34:
```suggestion
    if contains(gethostname(), "derecho") # Derecho
```

I think we need a more elegant solution to this problem in ClimaCalibrate, but this is slightly better for now:
```suggestion
    if get_backend() == ClimaCalibrate.DerechoBackend
```

experiments/calibration/run_calibration.jl
```diff
@@ -22,11 +22,26 @@ const CALIBRATE_CONFIG = CalibrateConfig(;
...
+        addprocs(
+            ClimaCalibrate.PBSManager(ensemble_size),
+            q = "main",
+            A = "UCIT0011",
+            l_select = "1:ngpus=1:ncpus=1",
+            l_walltime = "05:30:00",
+        )
+    elseif contains(gethostname(), "cluster") # Caltech central
```
nefrathenrici, line 42: I think this could just be an `else` or use `get_backend()` as above.

PR #1358, opened 2025-08-21
enable plotting comparison between simulations

juliasloan25:
## Purpose 
Adds a keyword arg `diagnostics2` to enable comparing model output between two simulations.

## Content

----

### DIFF
--- a/ext/land_sim_vis/plotting_utils.jl
+++ b/ext/land_sim_vis/plotting_utils.jl
@@ -351,6 +351,7 @@ function LandSimVis.make_diurnal_timeseries(
     diagnostics,
     start_date;
     plot_stem_name = "diurnal_timeseries",
+    diagnostics2 = nothing,
     comparison_data = nothing,
     spinup_date = start_date,
 )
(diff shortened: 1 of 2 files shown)


limamau (commented):
it doesn't compile

juliasloan25:
> it doesn't compile

Please try now.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -385,6 +386,26 @@ function LandSimVis.make_diurnal_timeseries(
             label = "Model",
             color = "blue",
         )
+        # Plot diagnostics from the second model, if provided
+        if !(isnothing(diagnostics2))
```
kmdeck, line 390: maybe we could also add it to the annual timeseries and just regular timeseries plots too?

For the single column runs, it feels less likely ? that we would run twice in the same script/session to get two diagnostics in memory (which we convert to a vector with "diagnostics as vectors"), but eventually when we write to a file (like we do for global runs) I could see it being useful to read in two diagnostic files and plot them on the same plot

PR #1360, opened 2025-08-21
Bugs in canopy/leaf level quantities

kmdeck:
## Purpose 
- Farquhar works entirely in leaf level fluxes, and therefore (1) GPP needs to be upscaled to canopy level from An, and (2) stomatal conductance needs to be upscaled to canopy level with LAI. The Vcmax, Jmax, Rd, etc are all leaf level. However, we pass in APAR at the canopy level, not leaf level
- Pmodel: works entirely in canopy level: uses APAR_canopy, predicts GPP = An. Therefore, conductance computed from An is already leaf level. Jmax,Vcmac, Rd are all canopy level. We pass in the correct APAR, but upscale GPP and conductance when we dont need to.
- Downstream autotrophic respiration and SIF use quantities computed by the photosynthesis models and we need to be consistent with which values are leaf or canopy level
- Diagnostics also need to be consistent between the two (e.g. cannot return Vcmax25 canopy in one case, and leaf level in the other case).

## Content
Diagnostics: GPP (canopy gross photosynthesis), vcmax25_leaf, rd_leaf now clearly labeled. I removed An since this not used anywhere currently.
Bug fix: GPP is gross, not net: we should not subtract Rd from it. This is consistent with pmodel papers and Bonan's book
Bug fix: Farquhar works entirely with leaf level, medlyn conductance uses An at leaf level
Bug fix: Pmodel works entirely with canopy level: no need to upscale to get GPP from A or when computing conductance of the canopy
Helper functions to get Vcmax25, Jmax, etc of leaves from the pmodel - make it clear what is canopy vs leaf in downstream SIF and autotrophic respiration

----

### DIFF
--- a/src/standalone/Vegetation/canopy_parameterizations.jl
+++ b/src/standalone/Vegetation/canopy_parameterizations.jl
@@ -15,7 +15,6 @@ export canopy_sw_rt_beer_lambert, # Radiative transfer
     net_photosynthesis,
     moisture_stress,
     dark_respiration,
-    compute_GPP,
     MM_Kc,
     MM_Ko,
     compute_Vcmax,
@@ -820,22 +819,6 @@ function c3_dark_respiration(
 end
 
 """
-    compute_GPP(An::FT,
-             K::FT,
-             LAI::FT,
-             Ω::FT) where {FT}
-
-Computes the total canopy photosynthesis (`GPP`) as a function of
-the total net carbon assimilation (`An`), the extinction coefficient (`K`),
-leaf area index (`LAI`) and the clumping index (`Ω`).
-"""
-function compute_GPP(An::FT, K::FT, LAI::FT, Ω::FT) where {FT}
-    GPP = An * (1 - exp(-K * LAI * Ω)) / (K * Ω)
-    return GPP
-end
-
-
-"""
     upscale_leaf_conductance(gs::FT, LAI::FT, T::FT, R::FT, P::FT) where {FT}
 
 This currently takes a leaf conductance (moles per leaf area per second)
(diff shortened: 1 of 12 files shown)


src/standalone/Vegetation/photosynthesis.jl
```diff
@@ -228,71 +237,170 @@ function update_photosynthesis!(p, Y, model::FarquharModel, canopy)
...
+            s3,
+            s4,
+            s5,
+            s6,
+            E,
+        ),
+    )
+    # Compute net from gross
+    # See Table 11.5 of G. Bonan's textbook, Climate Change and Terrestrial Ecosystem Modeling (2019).
+    @. An = max(0, β * A_leaf - Rd)
```
kmdeck, line 287: Farquhar saves this since conductance needs it, but pmodel doesnt need it ever, so it's not part of that cache

src/standalone/Vegetation/photosynthesis.jl
```diff
@@ -228,71 +237,170 @@ function update_photosynthesis!(p, Y, model::FarquharModel, canopy)
...
-    return @. lazy(max_electron_transport(Vcmax25, ΔHJmax, T_canopy, To, R))
+    Jmax_leaf =
+        @. lazy(max_electron_transport(Vcmax25, ΔHJmax, T_canopy, To, R))
+    return Jmax_leaf
 end
 
-function get_electron_transport(Y, p, canopy, m::FarquharModel)
-    T_canopy = canopy_temperature(canopy.energy, canopy, Y, p)
 
+function get_J_over_Jmax(Y, p, canopy, m::FarquharModel)
```
kmdeck, line 314: I added this; it was helpful for SIF

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -34,15 +34,16 @@ export canopy_sw_rt_beer_lambert, # Radiative transfer
     compute_Kmm,
     optimal_co2_ratio_c3,
     intercellular_co2_pmodel,
-    gs_co2_pmodel,
-    gs_h2o_pmodel,
+    gcanopy_co2_pmodel,
```
kmdeck, line 37: not sure what to call this. I am pretty sure this is canopy level, do we still want to call it stomatal conductance?

PR #1368, opened 2025-08-22
Photosynthesis cleanup and bug fixes

### LINKED ISSUE #1376: Photosynthesis scaling: fix and clarify leaf and canopy level
Our photosynthesis models may be from leaf level scaled to canopy (e.g., Farquhar) or directly at canopy level (e.g., Pmodel). As a results, some variables (e.g., APAR, An, Aj) may be expressed per m2 of leaf or m2 or ground. We rename those variables, such as APAR_leaf, for clarity, or if a parameterisation allows both options (e.g., APAR_leaf or APAR_canopy), we leave it to APAR and explain the function may be used at leaf or canopy level. 

We also fix some scaling bugs.

kmdeck:
closes #1376 

If this PR is too hard to review because of the movement of functions from photosynthesis.jl to photosynthesis_farquhar.jl and the movement of functions out of canopy_parameterizations.jl, I can move things back and do that in a separate PR

## Purpose 
- Regular Farquhar works entirely in leaf level fluxes, and therefore (1) `GPP` needs to be upscaled to canopy level from A, and (2) stomatal conductance `gs` needs to be upscaled to canopy level with LAI. The Vcmax, Jmax, Rd, etc are all leaf level. However, the bug is that we pass in APAR at the canopy level, not leaf level. so we were blending leaf+canopy in computing A, and sort of upscaling GPP and gs twice.
- Pmodel: works entirely in canopy level: uses APAR_canopy, predicts GPP = A. Therefore, conductance computed from A is already canopy level. Jmax,Vcmax, Rd are all canopy level. The bug is that we pass in the correct APAR, but upscale GPP and conductance when we dont need to.
- Downstream autotrophic respiration and SIF use quantities computed by the photosynthesis models and we need to be consistent with which values are leaf or canopy level. The bug is that right now they get different quantities from the pmodel and farquhar model.
- Diagnostics also need to be consistent between the two. the bug is that we return Vcmax25 canopy in one case, and leaf level in the other case, same for Rd, etc.
- I moved farquhar specific functions and model/param definitions to photosynthesis_farquhar.jl, and moved pmodel specific functions to pmodel.jl, and likewise for autotrophic_respiration.jl. In a future PR I want to move more out of canopy_parameterizations.jl - that file is too long and hard to read.

## To-do
Revisit autotrophic model and SIF model to make sure we are passing the correct things. This PR should be behavior preserving in the sense of passing e.g. Vcmax25 leaf still, An_leaf, Rd_leaf...but Im not sure if that is what we should be passing

## Content
Diagnostics: GPP (canopy gross photosynthesis), vcmax25_leaf, rd_leaf, An_leaf now consistent between pmodel and farquhar
Bug fix: GPP is gross, not net: we should not subtract Rd from it. This is consistent with pmodel papers and Bonan's book
Bug fix: Farquhar works entirely with leaf level, we now pass APAR_canopy/LAI.
Bug fix: Pmodel works entirely with canopy level: no need to upscale to get GPP from A or when computing conductance of the canopy
Helper functions to get Vcmax25, Jmax, etc of _leaves_ from the pmodel - and rename variables to make it clear what is canopy vs leaf in downstream SIF and autotrophic respiration

----

### DIFF
--- a/src/standalone/Vegetation/photosynthesis.jl
+++ b/src/standalone/Vegetation/photosynthesis.jl
@@ -26,7 +26,7 @@ function update_photosynthesis!(
 )
     nothing
 end
-get_GPP_canopy(p, photosynthesis_model::AbstractPhotosynthesisModel) = nothing # used by diagnostics
+
 get_Vcmax_leaf(p, photosynthesis_model::AbstractPhotosynthesisModel) = nothing # used by solar induced fluorescence, diagnostics, autotrophic respiration
 get_J_over_Jmax(
     Y,
--- a/src/standalone/Vegetation/photosynthesis_farquhar.jl
+++ b/src/standalone/Vegetation/photosynthesis_farquhar.jl
@@ -94,9 +94,10 @@ function FarquharModel{FT}(
     return FarquharModel{eltype(parameters), typeof(parameters)}(parameters)
 end
 
-ClimaLand.auxiliary_vars(model::FarquharModel) = (:An, :Rd)
-ClimaLand.auxiliary_types(model::FarquharModel{FT}) where {FT} = (FT, FT)
-ClimaLand.auxiliary_domain_names(::FarquharModel) = (:surface, :surface)
+ClimaLand.auxiliary_vars(model::FarquharModel) = (:An, :Rd, :GPP)
+ClimaLand.auxiliary_types(model::FarquharModel{FT}) where {FT} = (FT, FT, FT)
+ClimaLand.auxiliary_domain_names(::FarquharModel) =
+    (:surface, :surface, :surface)
 
 """
     gross_leaf_photosynthesis_at_a_point_Farquhar
@@ -168,7 +169,8 @@ end
 )
 
 Computes the net leaf-level photosynthesis rate `An` (mol CO2/m^2/s) for the Farquhar 
-model, along with the dark leaf-level respiration `Rd` (mol CO2/m^2/s).
+model, along with the dark leaf-level respiration `Rd` (mol CO2/m^2/s), and
+canopy level gross photosynthesis (mol CO2/m^2/s).
 """
 function update_photosynthesis!(p, Y, model::FarquharModel, canopy)
     (;
@@ -202,6 +204,7 @@ function update_photosynthesis!(p, Y, model::FarquharModel, canopy)
     # unpack a bunch of stuff from p and params
     Rd = p.canopy.photosynthesis.Rd
     An = p.canopy.photosynthesis.An
+    GPP = p.canopy.photosynthesis.GPP
     T_canopy = canopy_temperature(canopy.energy, canopy, Y, p)
     f_abs = p.canopy.radiative_transfer.par.abs
     ψ = p.canopy.hydraulics.ψ
@@ -288,14 +291,15 @@ function update_photosynthesis!(p, Y, model::FarquharModel, canopy)
     ) # has moisture stress
     # Compute net assimilation:
     @. An = net_photosynthesis(A * β, Rd) # Rd has β accounted for already.
+    # GPP
+    @. GPP = A * β * LAI
 end
 Base.broadcastable(m::FarquharParameters) = tuple(m)
 
 get_Vcmax25_leaf(p, m::FarquharModel) = m.parameters.Vcmax25
 get_Rd_leaf(p, m::FarquharModel) = p.canopy.photosynthesis.Rd
 get_An_leaf(p, m::FarquharModel) = p.canopy.photosynthesis.An
-get_GPP_canopy(p, m::FarquharModel) =
-    @. lazy(p.canopy.photosynthesis.An * p.canopy.hydraulics.area_index.leaf)
+
 function get_J_over_Jmax(Y, p, canopy, m::FarquharModel)
     Jmax = compute_Jmax_leaf(Y, p, canopy, m) # lazy
     J = compute_J_leaf(Y, p, canopy, m) # lazy
--- a/src/diagnostics/Diagnostics.jl
+++ b/src/diagnostics/Diagnostics.jl
@@ -25,7 +25,6 @@ import ..Canopy:
     get_Vcmax25_leaf,
     get_An_leaf,
     get_Rd_leaf,
-    get_GPP_canopy,
     MedlynConductanceModel,
     PModelConductance,
     moisture_stress,
--- a/src/diagnostics/land_compute_methods.jl
+++ b/src/diagnostics/land_compute_methods.jl
@@ -356,11 +356,9 @@ end
 @diagnostic_compute "photosynthesis_gross_canopy" Union{
     SoilCanopyModel,
     LandModel,
-} get_GPP_canopy(p, land_model.canopy.photosynthesis)
-@diagnostic_compute "photosynthesis_gross_canopy" CanopyModel get_GPP_canopy(
-    p,
-    land_model.photosynthesis,
-)
+    CanopyModel,
+} p.canopy.photosynthesis.GPP
+
 @diagnostic_compute "photosynthesis_net_leaf" Union{SoilCanopyModel, LandModel} get_An_leaf(
     p,
     land_model.canopy.photosynthesis,


src/standalone/Vegetation/autotrophic_respiration.jl
```diff
@@ -191,3 +190,82 @@ function AutotrophicRespirationParameters(
     FT = CP.float_type(toml_dict)
     AutotrophicRespirationParameters{FT}(; parameters..., kwargs...)
 end
+
```
kmdeck, line 193: copied from canopy_parameterizations.jl

src/standalone/Vegetation/photosynthesis.jl
```diff
@@ -1,386 +1,298 @@
...
-        compute_Vcmax(is_c3, Vcmax25, T, R, To, ΔHVcmax, Q10, s1, s2, s3, s4)
-    Γstar = co2_compensation(Γstar25, ΔHΓstar, T, To, R)
-    ci = intercellular_co2(c_co2, Γstar, medlyn_factor)
-    Aj = light_assimilation(is_c3, J, ci, Γstar, APAR, E)
-    Kc = MM_Kc(Kc25, ΔHkc, T, To, R)
-    Ko = MM_Ko(Ko25, ΔHko, T, To, R)
-    Ac = rubisco_assimilation(is_c3, Vcmax, ci, Γstar, Kc, Ko, oi)
-    return net_photosynthesis(Ac, Aj, Rd, β)
+"""
+    c3_rubisco_assimilation(Vcmax::FT,
```
kmdeck, line 92: Both farquhar & pmodel use these. I would like to eventually move the dark_respiration and Vcmax computation here, but Im not sure if we can unify pmodel and farquhar for those. TBD when we add C4 to pmodel

src/standalone/Vegetation/photosynthesis.jl
```diff
@@ -1,386 +1,298 @@
...
+the leaf internal carbon dioxide partial pressure (`ci`),
+the CO2 compensation point (`Γstar`), and the effective Michaelis-Menten parameter
+`K_mm`.
+
+The units of ci, Γstar, and Kmm may be Pa or mol/mol, but they must be consistent. The
+units of Vcmax must be mol CO2/m^2.s.
+
+See Table 11.5 of G. Bonan's textbook, Climate Change and Terrestrial Ecosystem Modeling (2019).
+"""
+function c3_rubisco_assimilation(
```
kmdeck, line 109: I made it to farquhar computes Kmm and then it uses this function as well, instead of having a separate function for pmodel and farquhar (where farquhar computed Kmm internally, and the pmodel function took it as input)

src/standalone/Vegetation/photosynthesis_farquhar.jl
```diff
@@ -0,0 +1,716 @@
...
+        ΔHRd,
+        Q10,
+        s5,
+        s6,
+        fC4,
+    ) # has moisture stress
+    A = @. lazy(
+        gross_leaf_photosynthesis_at_a_point_Farquhar(
+            T_canopy,
+            compute_APAR_leaf_moles(
```
kmdeck, line 255: leaf level!

src/standalone/Vegetation/pmodel.jl
```diff
@@ -52,9 +52,9 @@ $(DocStringExtensions.FIELDS)
 Base.@kwdef struct PModelConstants{FT}
     """Gas constant (J mol^-1 K^-1)"""
     R::FT
-    """Michaelis-Menten parameter for carboxylation at 25°C (μmol mol^-1)"""
+    """Michaelis-Menten parameter for carboxylation at 25°C (Pa)"""
     Kc25::FT
-    """Michaelis-Menten parameter for oxygenation at 25°C (μmol mol^-1)"""
+    """Michaelis-Menten parameter for oxygenation at 25°C (Pa)"""
```
kmdeck, line 57: @ychnli please double check this

src/standalone/Vegetation/pmodel.jl
```diff
@@ -450,7 +457,7 @@ function update_optimal_EMA(
     VPD::FT,
     ca::FT,
     βm::FT,
-    APAR::FT,
+    APAR_canopy_moles::FT,
```
kmdeck, line 460: same value, renamed for clarity

src/standalone/Vegetation/pmodel.jl
```diff
@@ -911,44 +920,53 @@ function update_photosynthesis!(p, Y, model::PModel, canopy)
...
-                constants.To,
-                constants.Ha_Vcmax,
-                constants.Hd_Vcmax,
-                constants.aS_Vcmax,
-                constants.bS_Vcmax,
-                constants.R,
-            )
-        ) *
-        Vcmax
+        p.canopy.photosynthesis.OptVars.Vcmax25_opt *
```
kmdeck, line 928: I changed this because it seemed like we were multiplying Vcmax25_opt by the scaling to get Vcmax, and then dividing by the scaling here. It felt clearer to me to just use Vcmax25_opt.

src/standalone/Vegetation/photosynthesis_farquhar.jl
```diff
@@ -0,0 +1,716 @@
...
+See Ch 11, G. Bonan's textbook, Climate Change and Terrestrial Ecosystem Modeling (2019).
+"""
+function electron_transport_farquhar(
+    APAR::FT,
+    Jmax::FT,
+    θj::FT,
+    ϕ::FT,
+) where {FT}
+    # Light utilization of APAR
+    IPSII = ϕ * APAR / 2
```
AlexisRenchon, line 496: Same here, add a comment to say APAR can be leaf (farquhar) or canopy (pmodel), also explain in docstring

src/standalone/Vegetation/photosynthesis_farquhar.jl
```diff
@@ -0,0 +1,716 @@
...
+                          Ω::FT) where {FT}
+
+Computes the gross canopy photosynthesis (`GPP`) as a function of
+the gross carbon assimilation (`A`) at the leaf level, the extinction coefficient (`K`),
+leaf area index (`LAI`) and the clumping index (`Ω`).
+"""
+function GPP_from_leaf_level_A(A::FT, K::FT, LAI::FT, Ω::FT) where {FT}
+    GPP = A * (1 - exp(-K * LAI * Ω)) / (K * Ω)
+    return GPP
+end
```
AlexisRenchon, line 518: As discussed, I think this is incorrect for our case.

This is basically making a Beer-Lambert attenuation to account for APAR. 
But we already account for APAR (which is good, better than beer lambert), so that our A (leaf level) represents an average leaf --> so we can just do

GPP = A * LAI

AlexisRenchon (commented):
This is a start. 
1. let's clarify throughout that APAR may be leaf or canopy level, depending if we are using Farquhar or Pmodel (perhaps others later)

2. our GPP scaling from A is wrong. (it does a Beer Lambert attenuation). However we compute APAR already and separately, and use it to compute an average leaf A (using APAR_leaf). So GPP is simply A*LAI

src/standalone/Vegetation/photosynthesis_farquhar.jl
kmdeck: I fixed this. since it is such a simple computation, I removed GPP from the cache and introduced a helper function `get_GPP_canopy()` which computes `A*LAI as a diagnostic. in the pmodel it is a bit more complex, so GPP is still in the cache there `get_GPP_canopy()` returns `p....GPP`.

src/standalone/Vegetation/photosynthesis.jl
```diff
@@ -1,386 +1,298 @@
...
-    R = LP.gas_constant(canopy.parameters.earth_param_set)
-    return @. lazy(max_electron_transport(Vcmax25, ΔHJmax, T_canopy, To, R))
+Computes the electron transport limiting rate (`Aj`),
+in units of moles CO2/m^2/s.
+
+For C4 plants, this is a function of APAR and a efficiency parameter E, see Equation 11.70 of
+ G. Bonan's textbook, Climate Change and Terrestrial Ecosystem Modeling (2019).
+"""
+function c4_light_assimilation(::FT, ::FT, ::FT, APAR::FT, E::FT) where {FT}
+    Aj = APAR * E
```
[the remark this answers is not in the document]
kmdeck: "The returned assimilation rate
may be leaf-level (standard Farquhar) or canopy level (p-model), 
depending on the units of APAR (per unit leaf area, or per ground area)."

kmdeck:
thanks! I think I will merge this since both you and I have checked it, but still ask @ychnli and @braghiere to take a look when they can and we can make any changes required in a followup PR if needed

ychnli:
Thanks for making these fixes!

A few comments regarding future improvements:
- what's the reason that the Farquhar model is leaf-level but the P-model is canopy-level? I think a future improvement could be to unify them as both leaf-level models, and then implement a better leaf-to-canopy scaling than the big leaf. If both models are leaf-level, then implementing a sunlit+shaded big leaf could be feasible. I think this should also improve scientific clarity for users of the model
-  I think another improvement would be to unify the temperature scaling functions (e.g., for Vcmax and Jmax) we use for the P-model and Farquhar model. These should be common among the photosynthesis models. My understanding is that the ones used in the P-model are the most "up to date", but perhaps @braghiere can help verify this

kmdeck:
> Thanks for making these fixes!
> 
> A few comments regarding future improvements:

Thanks! Ill open an issue. these are both good points. I dont think we can make the regular Farquhar model canopy level (since Vcmax25 is a leaf level number provided), but it's a good idea to make the pmodel leaf level as well.

PR #1369, opened 2025-08-22
Fix horizontal_resolution_degrees jldocttest

### LINKED ISSUE #1297: jldoctest failures in docbuild
## Describe the bug
Building the docs results in doctest failures. The docbuild action does not show any issues unless the logs are inspected.

See [here](https://github.com/CliMA/ClimaLand.jl/actions/runs/16788900597)

ph-kev:
closes #1297 - There was a parsing issue and the result of the function change at some point.


imreddyTeja (approved):
Thank you!
Do you think it is worth doing something like [this](https://documenter.juliadocs.org/stable/man/doctests/#Filtering-Doctests) to do approximate testing.

PR #1384, opened 2025-08-27
New soil IC function + bug fix in surface soil moisture used in surface fluxes

kmdeck:
## Purpose 
- Add function which sets the soil initial conditions from a netcdf file with total water content and soil temperature.
- Fixes bug where the linearly interpolated soil moisture is below the residual fraction.
- Makes minor changes to other functions to clip values more consistently: this is primarily because in the function `effective_saturation` we clip values of nu_eff and theta_l, but some functions both call `effective_saturation` and use nu_eff and theta_l directly, and we werent clipping consistently.
- Removes comments about keeping S_l \in (0,1]. this is not true.

I tested this on clima using an initial condition from @costachris from ERA5 with soil temp and total water, and when forced with 2008 ERA5 data, it produced no NaNs in the first 8 weeks (I stopped after that).

I changed the IC by calling, in experiments/long_runs/soil.jl:
```
ic_path = "/net/sampo/data1/cchristo/clima/WeatherQuest/processing/data/era5_land_processed_20250701_1200.nc"
set_ic!(Y, p, t, model) =
    ClimaLand.Simulations.set_soil_initial_conditions_from_temperature_and_total_water!(
        Y,
        domain.space.subsurface,
        ic_path,
        model,
    )
```
and passing `set_ic!` as a kwarg to the `LandSimulation` constructor

## Content

----

### DIFF
--- a/src/simulations/initial_conditions.jl
+++ b/src/simulations/initial_conditions.jl
@@ -138,10 +138,10 @@ end
         enforce_porosity_constraint(ϑ_l::FT, θ_i::FT, ν::FT, θ_r::FT)
 
 Enforces the constraint that ϑ_l + θ_i <= ν, by clipping the ice content to be
-99% (ν-ϑ_l), or leaving it unchanged if the constraint is already satisfied.
+95% (ν-ϑ_l), or leaving it unchanged if the constraint is already satisfied.
 """
 function enforce_porosity_constraint(ϑ_l::FT, θ_i::FT, ν::FT) where {FT}
-    if ϑ_l + θ_i > ν # if we exceed porosity
+    if ϑ_l + θ_i > FT(0.95) * ν # if we exceed porosity
         return FT(0.95) * (ν - ϑ_l) # clip ice content to 95% of available pore space
     else
         return θ_i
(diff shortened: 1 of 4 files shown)


src/simulations/initial_conditions.jl
```diff
@@ -325,3 +325,68 @@ function make_set_initial_state_from_file(
...
+        soil_ic_path,
+        water_varname,
+        subsurface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc, interpolation_method),
+    )
+    raw_data = NCDataset(soil_ic_path)
+    raw_temp_data = raw_data[temp_varname][:]
+    FT = eltype(Y.soil.ϑ_l)
+    T_bounds = FT.(extrema(raw_temp_data[raw_temp_data .> 0]))
```
kmdeck, line 350: this may be specific to ERA5, which has T_soil = 0 over the ocean. Because our interpolation is not mask aware, and because the ClimaLand and ERA5 masks may not align exactly, we may end up with nonphysical values in `total_water_content` over locations that ClimaLand considers to be land. This clips temperature to the bounds over land in the data.

kmdeck: (but this could lead to some locations that ERA5 calls ocean, but ClimaLand calls land, being set to the minimum value of T in the data, when in reality that location is not in a cold area)

szy21: Maybe add a comment on this may result in too cold soil temperature in some areas?

src/simulations/initial_conditions.jl
```diff
@@ -325,3 +325,68 @@ function make_set_initial_state_from_file(
...
+        soil_ic_path,
+        temp_varname,
+        subsurface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc, interpolation_method),
+    )
+    temperature .= clip_to_bounds.(temperature, T_bounds[1], T_bounds[2])
+    (; θ_r, ν, ρc_ds, earth_param_set) = soil.parameters
+    function liquid_soil_water(twc, T, θ_r, ν)
+        if T > 273.15
```
szy21, line 362: Do you have T_freeze in the parameters? If so it would be better to use that.

szy21 (approved):
Thank you!

PR #1387, opened 2025-08-28
Add AbstractSoilMoistureStressModel, 3 methods

### LINKED ISSUE #1382: Add moisture stress component from Yuchen PR
New canopy component:

abstract type AbstractSoilMoistureStressModel{FT} <: AbstractCanopyComponent{FT} end

--> 2 methods, Tuzet and theta

Updates beta (moisture stress)

and Fr (root uptake of water and energy), (Fr = Transpiration for the case of b(theta), and prognostic for Tuzet)

AlexisRenchon:
closes #1382 

This commits adds a portion of Yuchen Li PR #1330, a new Canopy component: AbstractSoilMoistureStressModel, with 3 methods: the existing one (Tuzet) which was previously in PlantHydraulics, a Piecewise method using dry to wet soil moisture, and no moisture stress.

Long run: https://buildkite.com/clima/climaland-long-runs/builds/4212

Snowy land is unchanged, Snowy land + pmodel was changed to use new moisture stress. it looks worse overall compared to what we have in main (ILAMB ET and GPP, ERA5 LHF and SHF): GPP dropped by 2x, ET dropped but by much less.

Will revert change to use this is pmodel until we can calibrate it.
Moisture stress now like 0.3 instead of 0.7

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@ ClimaLand.jl Release Notes
 main
 -------
 
+- Add moisture stress component to the canopy model PR[#1387](https://github.com/CliMA/ClimaLand.jl/pull/1387)
+
 v0.20.1
 -------
 - Add C4 plant support to PModel PR[#1356](https://github.com/CliMA/ClimaLand.jl/pull/1356)
--- a/experiments/integrated/fluxnet/ozark_pmodel.jl
+++ b/experiments/integrated/fluxnet/ozark_pmodel.jl
@@ -180,6 +180,13 @@ conductance = PModelConductance{FT}()
 is_c3 = FT(1)
 photosynthesis = PModel{FT}(surface_domain; is_c3)
 
+# Set up soil moisture stress using soil retention parameters
+soil_moisture_stress = PiecewiseMoistureStressModel{FT}(
+    land_domain,
+    toml_dict;
+    soil_params = retention_parameters,
+)
+
 # Set up plant hydraulics
 # Read in LAI from MODIS data
 surface_space = land_domain.space.surface;
@@ -223,6 +230,7 @@ canopy = Canopy.CanopyModel{FT}(
     radiative_transfer,
     photosynthesis,
     conductance,
+    soil_moisture_stress,
     hydraulics,
     energy,
 )
@@ -266,7 +274,8 @@ simulation = LandSimulation(
     set_ic!,
     updateat,
     diagnostics = diags,
-);
+)
+
 @time solve!(simulation)
 
 comparison_data = FluxnetSimulations.get_comparison_data(site_ID, time_offset)
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -77,6 +77,19 @@ struct LandModel{
         @assert soil.parameters.earth_param_set ==
                 snow.parameters.earth_param_set
 
+        # Check that soil moisture stress parameters are consistent between canopy and soil
+        if canopy.soil_moisture_stress isa PiecewiseMoistureStressModel
+            # Note that these functions allocate. These checks should not occur except on initialization.
+            check_land_equality(
+                canopy.soil_moisture_stress.θ_high,
+                soil.parameters.ν,
+            )
+            check_land_equality(
+                canopy.soil_moisture_stress.θ_low,
+                soil.parameters.θ_r,
+            )
+        end
+
         # LandModel-specific checks
         # Runoff and sublimation are also automatically included in the soil model
         @assert RootExtraction{FT}() in soil.sources
@@ -89,8 +102,6 @@ struct LandModel{
         return new{FT, MM, SM, VM, SnM}(soilco2, soil, canopy, snow)
     end
 end
-
-
 
 """
     LandModel{FT}(;
@@ -185,6 +196,9 @@ function LandModel{FT}(;
     end
 
     canopy = Canopy.CanopyModel{FT}(;
+        soil_moisture_stress = canopy_component_types.soil_moisture_stress(
+            canopy_component_args.soil_moisture_stress...,
+        ),
         autotrophic_respiration = canopy_component_types.autotrophic_respiration(
             canopy_component_args.autotrophic_respiration...,
         ),
--- a/src/standalone/Vegetation/Canopy.jl
+++ b/src/standalone/Vegetation/Canopy.jl
@@ -34,7 +34,8 @@ import ClimaLand:
     get_model_callbacks,
     total_liq_water_vol_per_area!,
     total_energy_per_area!,
-    FrequencyBasedCallback
+    FrequencyBasedCallback,
+    Soil
 using ClimaLand: PrescribedGroundConditions, AbstractGroundConditions
 using ClimaLand.Domains: Point, Plane, SphericalSurface, get_long
 export SharedCanopyParameters, CanopyModel, set_canopy_prescribed_field!
@@ -42,6 +43,7 @@ include("./component_models.jl")
 include("./biomass.jl")
 include("./PlantHydraulics.jl")
 using .PlantHydraulics
+include("./soil_moisture_stress.jl")
 include("./stomatalconductance.jl")
 include("./photosynthesis.jl")
 include("./photosynthesis_farquhar.jl")
@@ -59,6 +61,60 @@ include("./spatially_varying_parameters.jl")
 ########################################################
 # Convenience constructors for Canopy model components
 ########################################################
+## Soil Moisture Stress
+"""
+    PiecewiseMoistureStressModel{FT}(
+    domain,
+    toml_dict::CP.ParamDict;
+    c::FT = toml_dict["moisture_stress_c"],
+    soil_params
+) where {FT <: AbstractFloat}
+
+Helper function to create `PiecewiseMoistureStressModel` by calculating field capacity (θ_high)
+and wilting point (θ_low) from the soil parameters.
+
+The low and high thresholds for the piecewise soil moisture stress function are given by
+the residual soil water content and the soil porosity, respectively.
+
+The soil parameters should be a named tuple with keys of `ν` and `θ_r` 
+(additional keys may be present but they will not be used). These may 
+be ClimaCore fields or floats, but must be consistently one or the other.
+
+Note that unlike other Canopy components, this model is defined on the subsurface domain.
+"""
+function PiecewiseMoistureStressModel{FT}(
+    domain,
+    toml_dict::CP.ParamDict;
+    c::FT = toml_dict["moisture_stress_c"],
+    soil_params = Soil.soil_vangenuchten_parameters(
+        domain.space.subsurface,
+        FT,
+    ),
+) where {FT <: AbstractFloat}
+    if c <= 0
+        throw(
+            ArgumentError("Curvature parameter `c` must be greater than zero"),
+        )
+    end
+    θ_high = soil_params.ν
+    θ_low = soil_params.θ_r
+
+    return PiecewiseMoistureStressModel{FT}(; θ_high, θ_low, c)
+end
+
+"""
+    TuzetMoistureStressModel{FT}(toml_dict; sc::FT = toml_dict["moisture_stress_sc"], pc::FT = toml_dict["moisture_stress_pc"]) where{FT}
+
+A constructor for TuzetMoistureStressModel which uses the toml_dict
+values, allowing optional overrides by keyword argument.
+"""
+function TuzetMoistureStressModel{FT}(
+    toml_dict;
+    sc::FT = toml_dict["moisture_stress_sc"],
+    pc::FT = toml_dict["moisture_stress_pc"],
+) where {FT}
+    return TuzetMoistureStressModel{FT}(; sc, pc)
+end
 
 ## Autotrophic respiration models
 """
@@ -110,21 +166,15 @@ The `photosynthesis_parameters` argument is a NamedTuple that contains
 - `Vcmax25`: a Float or Field representing the maximum carboxylation rate at 25C (mol m^-2 s^-1)
 By default, these parameters are set by the `clm_photosynthesis_parameters` function,
 which reads in CLM data onto the surface space as ClimaUtilities SpaceVaryingInputs.
-
-The following additional default parameters are used:
-- sc = 5e-6 (Pa^{-1}) - sensitivity to low water pressure in the moisture stress factor [Tuzet et al. (2003)]
-- pc = -2e6 (Pa) - reference water pressure for the moisture stress factor [Tuzet et al. (2003)]
 """
 function FarquharModel{FT}(
     domain;
     photosynthesis_parameters = clm_photosynthesis_parameters(
         domain.space.surface,
     ),
-    sc::FT = LP.get_default_parameter(FT, :low_water_pressure_sensitivity),
-    pc::FT = LP.get_default_parameter(FT, :moisture_stress_ref_water_pressure),
 ) where {FT <: AbstractFloat}
     (; is_c3, Vcmax25) = photosynthesis_parameters
-    parameters = FarquharParameters(FT, is_c3; Vcmax25, sc, pc)
+    parameters = FarquharParameters(FT, is_c3; Vcmax25)
     return FarquharModel{FT, typeof(parameters)}(parameters)
 end
 
@@ -143,8 +193,6 @@ end
         ϕa1_c4 = FT(0.022*0.087),
         ϕa2_c4 = FT(-0.00034*0.087),
         α = FT(0.933),
-        sc = LP.get_default_parameter(FT, :low_water_pressure_sensitivity),
-        pc = LP.get_default_parameter(FT, :moisture_stress_ref_water_pressure),
     ) where {FT <: AbstractFloat}
 
 Constructs a P-model (an optimality model for photosynthesis) using default parameters.
@@ -161,8 +209,6 @@ The following default parameters are used:
 - ϕa1_c4 = 0.022*0.087 (K^-1) - first order term in quadratic intrinsic quantum yield (Scott and Smith, 2022)
 - ϕa2_c4 = -0.00034*0.087 (K^-2) - second order term in quadratic intrinsic quantum yield (Scott and Smith, 2022)
 - α = 0.933 (unitless) - 1 - 1/T where T is the timescale of Vcmax, Jmax acclimation. Here T = 15 days. (Mengoli 2022)
-- sc = 5e-6 (Pa^{-1}) - sensitivity to low water pressure in the moisture stress factor [Tuzet et al. (2003)]
-- pc = -2e6 (Pa) - reference water pressure for the moisture stress factor [Tuzet et al. (2003)]
 """
 function PModel{FT}(
     domain;
@@ -179,8 +225,6 @@ function PModel{FT}(
     ϕa1_c4 = FT(0.022 * 0.087),
     ϕa2_c4 = FT(-0.00034 * 0.087),
     α = FT(0.933),
-    sc = LP.get_default_parameter(FT, :low_water_pressure_sensitivity),
-    pc = LP.get_default_parameter(FT, :moisture_stress_ref_water_pressure),
 ) where {FT <: AbstractFloat}
     parameters = ClimaLand.Canopy.PModelParameters(
         cstar,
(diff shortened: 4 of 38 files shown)


AlexisRenchon:
@juliasloan25 note that I formatted this branch with JuliaFormatter v1.0.62 (the latest v1), which added formatting to multiple file, including unrelated to this PR. I added them because we will have to eventually, if we use that formatting from now on. 

Or is this unexpected?

docs/src/APIs/soil_moisture_stress.md
```diff
@@ -0,0 +1,12 @@
...
+
+```@meta
+CurrentModule = ClimaLand.Canopy
+```
+
+## Models and Parameters
+
+```@docs
+ClimaLand.Canopy.TuzetMoistureStressParameters
+ClimaLand.Canopy.TuzetMoistureStressModel{FT}()
```
kmdeck, line 11: Are there constructors we should add here too?

src/integrated/land.jl
```diff
@@ -184,6 +184,17 @@ function LandModel{FT}(;
         energy_model = PrescribedCanopyTempModel{FT}()
     end
 
+    if :soil_moisture_stress in propertynames(canopy_component_types)
```
kmdeck, line 187: note to double check that this default matches the defaults elsewhere

src/standalone/Vegetation/Canopy.jl
```diff
@@ -521,6 +539,9 @@ function CanopyModel{FT}(;
     photosynthesis::AbstractPhotosynthesisModel{FT},
     conductance::AbstractStomatalConductanceModel{FT},
     hydraulics::AbstractPlantHydraulicsModel{FT},
+    soil_moisture_stress::AbstractSoilMoistureStressModel{FT} = TuzetMoistureStressModel{
```
kmdeck, line 542: e.g. here the default does not match the default in the LandModel constructor you changed.

kmdeck: I also think we should change the default to be the Piecewise approach, but that can be a next PR so this PR is not behavior changing

src/standalone/Vegetation/Canopy.jl
```diff
@@ -1216,4 +1250,41 @@ function set_historical_cache!(p, Y0, m::AbstractPhotosynthesisModel, canopy)
     return nothing
 end
 
+"""
+    required_model_callbacks(start_date, t0, dt, model::CanopyModel)
+For some canopy components (namely the P-model), we need a callback to ensure regular
+updates at a preset frequency. This method dispatches off of the type of the photosynthesis
+model.
+"""
+function required_model_callbacks(start_date, t0, dt, model::CanopyModel)
```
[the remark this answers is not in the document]
kmdeck: I think this is already handled - could we remove this change?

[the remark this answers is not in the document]
kmdeck: https://github.com/CliMA/ClimaLand.jl/pull/1381

src/standalone/Vegetation/Canopy.jl
```diff
@@ -879,13 +905,18 @@ function initialize_boundary_vars(model::CanopyModel{FT}, coords) where {FT}
...
 """
-     ClimaLand.make_update_aux(canopy::CanopyModel{FT,
-                                                  <:AutotrophicRespirationModel,
-                                                  <:AbstractRadiationModel,
-                                                  <:AbstractPhotosynthesisModel,
-                                                  <:AbstractStomatalConductanceModel,
-                                                  <:PlantHydraulicsModel,},
-                              ) where {FT}
+     ClimaLand.make_update_aux(
+        canopy::CanopyModel{
```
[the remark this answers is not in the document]
kmdeck: (not sure - maybe this will be helpful in the future. but if we change it here, we have a couple other places to change it also - e.g. boundary_fluxes)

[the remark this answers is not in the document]
juliasloan25: Yeah good idea! I don't think having the types here does anything, and it makes the code more verbose

src/standalone/Vegetation/Canopy.jl
```diff
@@ -1016,6 +1048,9 @@ function ClimaLand.make_update_aux(
         end
         # We update the fa[n_stem+n_leaf] element once we have computed transpiration
 
+        # Update soil moisture stress, used in photosynthesis and conductance
+        update_soil_moisture_stress!(p, Y, canopy.soil_moisture_stress, canopy)
```
kmdeck, line 1052: great! thanks for adding this comment. The order in which we update matters here!

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+end
+
+Base.eltype(::TuzetMoistureStressParameters{FT}) where {FT} = FT
+Base.broadcastable(x::TuzetMoistureStressParameters) = tuple(x)
+
+"""
+    TuzetMoistureStressModel{FT, TMSP <: TuzetMoistureStressParameters{FT}}
+    <: AbstractSoilMoistureStressModel{FT}
+
+An implementation of the Tuzet moisture stress function.
```
kmdeck, line 31: Could we provide the formula here as well?

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+ClimaLand.auxiliary_domain_names(::TuzetMoistureStressModel) = (:surface,)
+
+
+"""
+    compute_tuzet_moisture_stress(
+        parameters::TuzetMoistureStressParameters{FT},
+        p_leaf::FT
+    ) where {FT}
+
+This function computes the soil moisture stress factor using the leaf water potential (Pa) and two
```
kmdeck, line 72: "This pointwise function"

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+        θ_high,
+        θ_low,
+        c,
+        β0,
+        soil_zmin
+    ) where {FT <: AbstractFloat}
+
+An outer constructor for PiecewiseMoistureStressParameters. The parameters should fall within
+the following ranges:
+- θ_high, θ_low should be in [0,ν] where ν is the porosity of the soil
```
kmdeck, line 178: "in (\theta_r, \nu]"

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+- c should be positive
+- β should be in (0,1]
+- soil_zmin should be negative
```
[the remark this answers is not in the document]
kmdeck: (soil_zmin)

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+"""
+function PiecewiseMoistureStressParametersFromHydrology(
+    ::Type{FT},
+    hydrology_cm,
+    ν::Union{FT, ClimaCore.Fields.Field},
+    θ_r::Union{FT, ClimaCore.Fields.Field},
+    soil_zmin::Union{FT, ClimaCore.Fields.Field};
+    c::FT = FT(1.0),
+    β0::FT = FT(1.0),
+    ψ_fc::FT = FT(-3.36734694),    # -33 kPa, equivalent in m H2O
```
kmdeck, line 264: Not sure if we need psi_fc and psi_wp if we have the retention curve params? I believe we can compute these from the retention curve.

kmdeck: this is great though! This is what I mean by having a default

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+    if :soil in canopy.boundary_conditions.prognostic_land_components
+        # unpack soil moisture field, parameters
+        ϑ_l = Y.soil.ϑ_l
+        (; θ_high, θ_low, c, β0, soil_zmin) = model.parameters
+        z = ClimaCore.Fields.coordinate_field(axes(ϑ_l)).z
+
+        # normalized distribution for root density
+        norm = @. lazy(
+            Canopy.PlantHydraulics.root_distribution_CDF(
+                soil_zmin,
```
kmdeck, line 364: I think here it makes sense to use the minimum of the soil domain

kmdeck: because the zmin we use has to be the same as in the integrals below for the compute norm to be the actual norm

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+            ) / norm,
+        )
+
+        # compute the root zone-averaged βm
+        ClimaCore.Operators.column_integral_definite!(
+            p.canopy.soil_moisture_stress.βm,
+            βm_root_distribution,
+        )
+    else
+        (; θ_high, θ_low, c, β0, soil_zmin) = model.parameters
```
kmdeck, line 386: (and here zmin is not needed)

src/integrated/land.jl
kmdeck, line 187: Note to change the soil_canopy_model constructor as well

src/standalone/Vegetation/Canopy.jl
```diff
@@ -403,6 +396,24 @@ function PModelConductance{FT}(; Drel = FT(1.6)) where {FT <: AbstractFloat}
     return PModelConductance{FT}(cond_params)
 end
 
+"""
```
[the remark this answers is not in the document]
kmdeck: oh, i see that you have this! Maybe that could be moved here, and not have the new name?

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+
+function PiecewiseMoistureStressModel{FT}(
+    parameters::PiecewiseMoistureStressParameters{FT},
+) where {FT <: AbstractFloat}
+    return PiecewiseMoistureStressModel{eltype(parameters), typeof(parameters)}(
+        parameters,
+    )
+end
+
+ClimaLand.auxiliary_vars(model::PiecewiseMoistureStressModel) = (:βm, :ϑ_root)
```
kmdeck, line 223: I believe theta_root is not used

test/standalone/Vegetation/canopy_model.jl
```diff
@@ -1444,7 +1450,7 @@ end
                 :LW_d,
                 :cosθs,
                 :frac_diff,
-                :ψ,
+                :θ,
```
kmdeck, line 1453: I dont think this will work, we need to update https://github.com/CliMA/ClimaLand.jl/blob/1e7651d9a7e7b1aa928ee986ee7a7f913891484e/src/shared_utilities/drivers.jl#L1085 first.

(This was part of Yuchen's PR, but we can rebase his PR off of main once this merges, to get those changes in there? I think it's fine to do - it just means we cant run the prescribed soil+canopy with the new moisture stress parameterization until we add it in)

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+"""
+function compute_piecewise_moisture_stress(
+    θ_high::FT,
+    θ_low::FT,
+    c::FT,
+    β0::FT,
+    θ::FT,
+) where {FT}
+    # avoid taking e.g. sqrt of negative numbers for rational c
+    arg = max((θ - θ_low) / (θ_high - θ_low), FT(0))
```
kmdeck, line 333: I think we should be OK without this, at least for the progostic case, and in the prescribed case also.

if we keep this, we do not need to clip at zero in line 335, because arg is guaranteed to be >=0 already.

src/standalone/Vegetation/soil_moisture_stress.jl
ychnli: you're correct, sorry. This is leftover from a previous version where I was integrating \theta over the soil domain before computing \beta, as opposed to computing \beta throughout the soil domain and integrating. I chose the latter approach because the threshold parameters may depend on depth-dependent soil hydrology params and therefore be depth-dependent.

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,390 @@
...
+saturation at field capacity and wilting point, then converts to volumetric water content.
+"""
+function PiecewiseMoistureStressParametersFromHydrology(
+    ::Type{FT},
+    hydrology_cm,
+    ν::Union{FT, ClimaCore.Fields.Field},
+    θ_r::Union{FT, ClimaCore.Fields.Field},
+    soil_zmin::Union{FT, ClimaCore.Fields.Field};
+    c::FT = FT(1.0),
+    β0::FT = FT(1.0),
```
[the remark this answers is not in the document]
ychnli: I agree, it complicates the interpretation of what soil moisture stress really means if it is capped at a value less than 1. I originally added it because they do something like this in this paper https://onlinelibrary.wiley.com/doi/abs/10.1111/gcb.70098

experiments/integrated/fluxnet/ozark_pmodel.jl
```diff
@@ -179,6 +180,17 @@ conductance = PModelConductance{FT}()
 # Set up photosynthesis
 photosynthesis = PModel{FT}()
 
+# Set up soil moisture stress
+soil_moisture_stress_params = PiecewiseMoistureStressParameters(
+    FT;
+    θ_c = FT(0.60),
```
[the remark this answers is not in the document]
AlexisRenchon: I agree with Yuchen here - we can't really have default theta_c or theta_w, as this varies with porosity of the soil? 
I am not sure about c and beta_0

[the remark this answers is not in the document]
kmdeck: ah, we can compute them from the soil retention parameters, which vary in space. I will add that.

src/standalone/Vegetation/Canopy.jl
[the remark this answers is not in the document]
kmdeck: yes!

src/standalone/Vegetation/soil_moisture_stress.jl
kmdeck: ah, ok. I dont think this is a good idea (if it is known that these thresholds are not constant). I updated this so that we either use porosity and the residual as the high and low values, or we use the van genuchten curves to get the air entry and a "dry" location) based on the shape of the curve: 
Screenshot 2025-09-05 at 1 34 39 PM

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,383 @@
...
+- `soil_zmin`: Depth of the lowest soil cell face (m), must be negative
+- `c`: Curvature parameter (unitless), default 1.0
+- `β0`: Intrinsic moisture stress factor (unitless), default 1.0
+- `ψ_fc`: Field capacity matric potential (m H2O), default -3.36 m (-33 kPa)
+- `ψ_wp`: Wilting point matric potential (m H2O), default -152.9 m (-1500 kPa)
+
+The function uses the `inverse_matric_potential` function to calculate the effective
+saturation at field capacity and wilting point, then converts to volumetric water content.
+"""
+function PiecewiseMoistureStressParametersFromHydrology(
```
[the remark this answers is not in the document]
kmdeck: I fixed this

src/standalone/Vegetation/soil_moisture_stress.jl
kmdeck: ah, this is hard to obtain from "canopy" alone quantities. I changed this to compute the norm by doing the integral, even though it is analytically available, and constant in time.

src/standalone/Vegetation/Canopy.jl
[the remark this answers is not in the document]
kmdeck: doing in another PR, will try and merge that first

src/standalone/Vegetation/soil_moisture_stress.jl
```diff
@@ -0,0 +1,270 @@
...
+    model,
+    canopy,
+)
+    @error(
+        "You cannot use the PiecewiseSoilMoistureStress model with a prescribed soil yet."
+    )
+    #(; θ_high, θ_low, c,) = model.parameters
+    # Interpret p.drivers.θ as the root zone value.
+    #@. p.canopy.soil_moisture_stress.βm =
+    #    compute_piecewise_moisture_stress(θ_high, θ_low, c, p.drivers.θ)
```
kmdeck, line 269: currently, p.drivers.psi is stored. It is another big change to convert that. let's leave that for another PR. All that means is that we cant use this new soil moisture stress parameterization with a prescribed soil model yet.

src/standalone/Vegetation/Canopy.jl
```diff
@@ -59,6 +60,79 @@ include("./spatially_varying_parameters.jl")
...
+
+If porosity_residual == true, we use the porosity as θ_high, and the residual as θ_low; if false,
+we compute the air entry θ (θ_high) and use the intersection of the tangent of the van Genuchten curve at
+the inflection point with θ = 0 as θ_low.
+"""
+function PiecewiseMoistureStressModel{FT}(
+    toml_dict::CP.AbstractTOMLDict;
+    c::FT = toml_dict["moisture_stress_c"],
+    porosity_residual = true,
+    soil_params,
```
kmdeck, line 87: @juliasloan25 - I have a few questions here...
- it might be nice to experiment with setting theta_low = theta_r, and theta_high = \nu, in addition to the functions of van genuchten parameters. The same computation of moisture stress holds once we define theta_low and theta_high, just the meaning of those values change. Is the way I did it here complicated? In particular, it makes it hard to check if the parameters are consistent with the soil model, because I dont know which theta_low/high definition was used. I could add a field to the Model struct indicating?
- It might be nice for this function to take the soil model hydrology_cm directly, instead of a tuple of soil_params which has the same information (just without the vG struct). But then the canopy module will depend on the Soil module. Is this OK?
- While this convenience constructor is based off of soil parameters, the default and outer constructor are not. so someone can set whatever values they want for theta_high, _low. in this case, any check that they are consistent with the soil model will fail. Is this desirable or should we constrain the values to be consistent?

NEWS.md
```diff
@@ -3,6 +3,7 @@ ClimaLand.jl Release Notes
 
 main
 -------
+- ![breaking change][badge-💥breaking] Add moisture stress module to the canopy model PR[#1387](https://github.com/CliMA/ClimaLand.jl/pull/1387)
```
juliasloan25, line 6: Could you add a little more detail about why this is a breaking change and what users should do to update for the change?

src/integrated/land.jl
```diff
@@ -77,6 +77,38 @@ struct LandModel{
         @assert soil.parameters.earth_param_set ==
                 snow.parameters.earth_param_set
 
+        if canopy.soil_moisture_stress isa PiecewiseMoistureStressModel
```
juliasloan25, line 80: Could you add a comment here explaining the check being done?

src/standalone/Vegetation/Canopy.jl
```diff
@@ -59,6 +61,82 @@ include("./spatially_varying_parameters.jl")
...
+Helper function to create `PiecewiseMoistureStressModel` by calculating field capacity (θ_high)
+and wilting point (θ_low) from the soil parameters.
+
+The soil parameters are expected to include ν and θ_r (if porosity_residual
+== true) AND  the van Genuchten parameters α and n (if porosity_residual == false). These
+may be fields or floats, but must be consistently one or the other.
+
+If porosity_residual == true, we use the porosity as θ_high, and the residual as θ_low; if false,
+we compute the air entry θ (θ_high) and use the intersection of the tangent of the van Genuchten curve at
+the inflection point with θ = 0 as θ_low.
```
juliasloan25, line 82: I'm not sure I understand this - if porosity_residual is true, alpha and n aren't used here so they don't strictly need to be provided right?

Could you explain more in the docstring about what porosity_residual corresponds to physically? It seems like it indicates whether to use the combination of soil porosity and the residual water content for the moisture stress thresholds, and to otherwise use "air_entry" and "esp_wp", but some more detail would be helpful. Also can you add more description of what the functions `compute_air_entry` and `compute_esp_wp` are calculating?

PR #1394, opened 2025-09-04
tag v0.20.0; up deps

### LINKED ISSUE #1395: Error when running with CoupledAtmosphere
## Describe the bug
When we use a CoupledAtmosphere, `atmos.h` is initialized as a Field, rather than a Float when we read in forcing from Fluxnet or ERA5.
[These assertions](https://github.com/CliMA/ClimaLand.jl/blob/81835da851772262644c228ab724cf3cfaae80f8/src/simulations/Simulations.jl#L147-L156) fail because `atmos.h` is a Field and `compartment_surfaces` is a Float. Instead of a direct `>=` check, we need to convert the Field to an Array and broadcast the check.

## To Reproduce
Construct the CanopyModel (or an integrated model with canopy) with a CoupledAtmosphere, and then try to make a LandSimulation with it.

<details><summary>Project</summary>
<p>
If not using the `examples` project:
```
paste your Project.toml here.
```
```
paste your Manifest.toml here.
```

</p>
</details>

## System details

Any relevant system information:
- Julia version
- operating system
- modules loaded on cluster (`module list`)

## Related issues / PRs

Please add any relevant links.

juliasloan25:
## Purpose 
Release so we can use new IC function downstream in ClimaCoupler

Also includes a bugfix that closes #1395. Fix was tested manually from ClimaCoupler and works.

## Dependency updates
```
    Updating `~/clima/ClimaLand.jl/.buildkite/Project.toml`
  [d414da3d] ↑ ClimaCore v0.14.39 ⇒ v0.14.40
  [08f4d4ce] ~ ClimaLand v0.19.0 `..` ⇒ v0.20.0 `..`
  [90137ffa] ↑ StaticArrays v1.9.14 ⇒ v1.9.15
```

### DIFF
--- a/src/simulations/Simulations.jl
+++ b/src/simulations/Simulations.jl
@@ -146,13 +146,13 @@ function LandSimulation(
 )
     # Enforce `h_atmos >= h_canopy` for models with canopy
     if hasproperty(model, :canopy)
-        h_atmos = model.canopy.boundary_conditions.atmos.h
+        h_atmos_min = first(extrema(model.canopy.boundary_conditions.atmos.h))
         h_canopy = model.canopy.hydraulics.compartment_surfaces[end]
-        @assert h_atmos >= h_canopy "Atmospheric height must be greater than or equal to canopy height"
+        @assert h_atmos_min >= h_canopy "Atmospheric height must be greater than or equal to canopy height. Got min atmos height $h_atmos_min and canopy height $h_canopy"
     elseif model isa ClimaLand.Canopy.CanopyModel
-        h_atmos = model.boundary_conditions.atmos.h
+        h_atmos_min = first(extrema(model.boundary_conditions.atmos.h))
         h_canopy = model.hydraulics.compartment_surfaces[end]
-        @assert h_atmos >= h_canopy "Atmospheric height must be greater than or equal to canopy height"
+        @assert h_atmos_min >= h_canopy "Atmospheric height must be greater than or equal to canopy height. Got min atmos height $h_atmos_min and canopy height $h_canopy"
     end
 
     start_date = isnothing(t0.epoch) ? nothing : date(t0)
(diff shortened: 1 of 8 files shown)


src/simulations/Simulations.jl
```diff
@@ -146,13 +147,23 @@ function LandSimulation(
...
     if hasproperty(model, :canopy)
-        h_atmos = model.canopy.boundary_conditions.atmos.h
-        h_canopy = model.canopy.hydraulics.compartment_surfaces[end]
-        @assert h_atmos >= h_canopy "Atmospheric height must be greater than or equal to canopy height"
+        # `atmos.h` is not meaningful for a CoupledAtmosphere, so we skip the check
+        if !(
+            model.canopy.boundary_conditions.atmos isa
+            ClimaLand.CoupledAtmosphere
+        )
+            h_atmos_min =
```
kmdeck, line 155: is atmos.h a field? I guess in coupled sims it is

kmdeck: But then this check is not carried out in coupled sims?

src/simulations/Simulations.jl
```diff
@@ -2,6 +2,7 @@ module Simulations
 using ClimaTimeSteppers
 using ClimaComms
 import ClimaComms: context, device
+import ClimaCore.Fields: field2array
```
kmdeck, line 5: is this used?

src/simulations/Simulations.jl
juliasloan25: Yes that's right. Float in standalone runs, and Field in coupled runs

src/simulations/Simulations.jl
juliasloan25: oops no

kmdeck (approved):
looks good! thanks for fixing and making the release.

(maybe you havent uploaded all the changes, but we can still remove field2array)

PR #1398, opened 2025-09-05
Unify `update!` usage in canopy update aux

kmdeck:
## Purpose 
All canopy components but radiative transfer and hydraulics use `update_X(p,Y,component, canopy)` syntax in update aux to hide the update details. Plant hydraulics and radiation were the exception, mostly likely because we did not need to dispatch on the type of the component model. Even so, it is nice to use the same syntax for these, and it makes the canopy update aux function more readable. In the future, with different moisture stress parameterizations, we do not need to use the same plant hydraulics model and may introduce another type there also. This also sets us up for that.

## Content

----


src/standalone/Vegetation/Canopy.jl
```diff
@@ -920,101 +920,12 @@ function ClimaLand.make_update_aux(
...
-            ground_albedo_NIR(
-                Val(bc.prognostic_land_components),
-                bc.ground,
-                Y,
-                p,
-                t,
-            ),
-        )
+        # Update p.canopy.radiative_transfer.par, .nir, .ϵ, .par_d, .nir_d
+        update_radiative_transfer!(p, Y, t, canopy.radiative_transfer, canopy)
```
kmdeck, line 924: I dont think we need `t` here, but removing it would change more files. I can do this if we want! so the arguments are the same for all.

src/standalone/Vegetation/canopy_parameterizations.jl
```diff
@@ -1,397 +1,11 @@
 using ..ClimaLand.Canopy
-export canopy_sw_rt_beer_lambert, # Radiative transfer
```
kmdeck, line 2: moved this into radiation - eventually we can remove this file

AlexisRenchon (approved):

PR #1399, opened 2025-09-05
update toml with calibrated parameters

kmdeck:
## Purpose 
This has the results from @ph-kev 's calibration of SW_u using snow albedo parameters.

## Content

----

### DIFF
--- a/toml/default_parameters.toml
+++ b/toml/default_parameters.toml
@@ -79,31 +79,31 @@ tag = "PlantHydraulicsParameters"
 # ZenithAngleAlbedoModel parameters
 
 [alpha_0]
-value = 0.64
+value = 0.62
 type = "float"
 description = "Free parameter controlling the minimum snow albedo"
 tag = "ZenithAngleAlbedoModel"
 
 [delta_alpha]
-value = 0.06
+value = 0.21
 type = "float"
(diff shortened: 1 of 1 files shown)


ph-kev:
I am still skeptical if these parameters will improve the long run. I think a long run of 19 years should be done and checked with the ILAMB leaderboard, before merging this in.

ph-kev:
> > I am still skeptical if these parameters will improve the long run. I think a long run of 19 years should be done and checked with the ILAMB leaderboard before merging this in.
> 
> Im skeptical because they are constrained by the prior bound (the parameters cant be converged then...)
> 
> Is your skepticism coming from the fact that only a few years were used to calibrate these?

I am skeptical, because the minibatch size is one year and I think the calibration could benefit from using multiple years for a single minibatch.

kmdeck:
> I am skeptical, because the minibatch size is one year and I think the calibration could benefit from using multiple years for a single minibatch. Also, the means of the priors before calibration are not the same as the current parameter values, so I don't know how the two set of parameter values compared against each other.

is it possible to calibrate this more robustly, and adjust the priors? (I am thinking of how to make it more streamlined to move from a calibration run to an improved model in main)

PR #1401, opened 2025-09-05
remove parameteric types in canopy make functions

kmdeck:
## Purpose 
We do not dispatch off them; these are not needed.

They also can lead to bugs when new canopy components are introduced (if these are not all updated to reflect that)

## Content

----


juliasloan25 (approved):
Much more concise! Thank you

PR #1404, opened 2025-09-08
LandSimulation constructor for ITime dt + start and end dates

juliasloan25:
## Purpose 
The LandSimulation constructor that takes in `start_date` and `end_date` as DateTimes previously only accepted `dt` as a Float or Dates.Second. This PR allows `dt` to also be an `ITime`, and asserts that in this case `dt.epoch == start_date`. This is a case used by ClimaCoupler.

## To-do
- [x] allow ITime dt
- [x] tag patch release to use in ClimaCoupler

### DIFF
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -314,9 +314,9 @@ version = "0.5.0"
 
 [[deps.CFTime]]
 deps = ["Dates", "Printf"]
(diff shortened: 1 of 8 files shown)


imreddyTeja (approved):
It might be a good idea to ensure that this works with the coupler before making the release.

kmdeck:
@juliasloan25 I am closing this but please re-open if it is still a needed or nice to have change

PR #1405, opened 2025-09-09
follow UTC offset sign convention; fix Ozark timezone

juliasloan25:
## Purpose 
Up to now, our fluxnet sites have been using UTC offsets that are internally consistent but not following the standard convention. This PR fixes their signs and updates how they're used to remain consistent.

Also, the timezone for MOz was wrong and is now fixed.

## To do
- [ ] check sign of fluxnet2015 dataset utc offsets

## Content
- [x] fix timezone/UTC offset for MOz fluxnet site - previously we had 7 (mountain standard time) but this should be -6 (central standard time)
- [x] fix sign of UTC offsets - for North American sites these should be negative following the standard convention

----


kmdeck (approved):
this looks good! thank you!

PR #1407, opened 2025-09-09
add error metrics and fix freezing front tutorial

kmdeck:
## Purpose 
- Add error metrics to the tutorials since we include these tutorials in the paper.
- Since the freezing front was changed to save at every timestep, the indexing was incorrect to extract the solution at 12h, 24h, 50h. This also fixes that.

## Content

----

### DIFF
--- a/docs/src/tutorials/standalone/Soil/freezing_front.jl
+++ b/docs/src/tutorials/standalone/Soil/freezing_front.jl
@@ -79,6 +79,8 @@ using CairoMakie
 
 using ClimaCore
 import ClimaParams as CP
+using ClimaUtilities.Utils: linear_interpolation
+using Statistics
 
 using ClimaLand
 using ClimaLand.Domains: Column
@@ -198,18 +200,9 @@ end
 # We choose the initial and final simulation times:
 t0 = Float64(0)
 tf = Float64(60 * 60 * 50);
+dt = Float64(100);
 
-dt = Float64(100)
-
-timestepper = CTS.ARS111();
-ode_algo = CTS.IMEXAlgorithm(
-    timestepper,
-    CTS.NewtonsMethod(
-        max_iters = 3,
-        update_j = CTS.UpdateEvery(CTS.NewNewtonIteration),
-    ),
-);
-
+# Create the simulation and solve
 simulation = LandSimulation(
     t0,
     tf,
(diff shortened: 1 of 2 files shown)


docs/src/tutorials/standalone/Soil/freezing_front.jl
```diff
@@ -198,33 +200,21 @@ end
 # We choose the initial and final simulation times:
 t0 = Float64(0)
 tf = Float64(60 * 60 * 50);
+dt = Float64(100);
 
-dt = Float64(100)
-
-timestepper = CTS.ARS111();
```
kmdeck, line 204: this is the default

docs/src/tutorials/standalone/Soil/freezing_front.jl
```diff
@@ -274,42 +264,45 @@ limits!(ax3, 0.2, 0.5, -0.2, 0.0)
...
+sim_24h_interp = [
+    linear_interpolation(z, sim_24h, data_depth) for
+    data_depth in -depth[mask_24h]
+]
+@show mae(sim_24h_interp, vwc[mask_24h])
+sim_50h_interp = [
+    linear_interpolation(z, sim_50h, data_depth) for
+    data_depth in -depth[mask_50h]
+]
+@show mae(sim_50h_interp, vwc[mask_50h])
```
imreddyTeja, line 304:
```suggestion
mae(x, obs) = mean(abs.(x .- obs));
sim_12h_interp = [
    linear_interpolation(z, sim_12h, data_depth) for
    data_depth in -depth[mask_12h]
]
mae(sim_12h_interp, vwc[mask_12h])
#-
sim_24h_interp = [
    linear_interpolation(z, sim_24h, data_depth) for
    data_depth in -depth[mask_24h]
]
mae(sim_24h_interp, vwc[mask_24h])
#-
sim_50h_interp = [
    linear_interpolation(z, sim_50h, data_depth) for
    data_depth in -depth[mask_50h]
]
mae(sim_50h_interp, vwc[mask_50h])
```

I think this should fix it

imreddyTeja (changes requested):
This looks good, other than one comment I left about the rendering.

While looking at this, I noticed that both the edited docs have small issues with headings, but that existed before this PR.

docs/src/tutorials/standalone/Soil/freezing_front.jl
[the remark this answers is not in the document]
kmdeck: thank you! I committed your change and fixed the same issue in the other tutorial. Ill double check the rendered docs this time when it builds.

imreddyTeja (approved):

PR #1436, opened 2025-09-17
Precomopile CUDA RT

imreddyTeja:
## Purpose 
Reduce buildkite ci duration 

## Content
- increase number of precompiled files. I don't know if this helps, but this is what ClimaAtmos does
- precompile the CUDA runtime. The CUDA.jl docs say to do this when running on a cluster.

~~Buildkite on main takes an hour and 15 minutes. This changes it to take 1 hour.~~
It seems like this sometimes speeds up ci.

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -9,6 +9,7 @@ env:
   SLURM_KILL_BAD_EXIT: 1
   JULIA_DEPOT_PATH: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/depot/default"
   JULIA_CUDA_MEMORY_POOL: "cuda"
+  JULIA_MAX_NUM_PRECOMPILE_FILES: 100
 
 
 steps:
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -38,6 +38,7 @@ steps:
       slurm_ntasks_per_node: 8
     env:
       JULIA_NUM_PRECOMPILE_TASKS: 8
+      JULIA_MAX_NUM_PRECOMPILE_FILES: 50
```
juliasloan25, line 41: is there a default for this that we were using before?

imreddyTeja: This defaults to 10

imreddyTeja:
> This looks fine to me! Did you find that it consistently speeds up buildkite, or is it hard to tell?

It seems like it makes it ~30% faster on half the times I tried it. That might be because the other ci jobs share the same depot, and are deleting cache files because they use a lower limit. I don't think this should ever make it slower though

Issue #600, opened 2024-04-29
Canopy Biomass model


Espeer5:
**Is your feature request related to a problem? Please describe.**
For legacy reasons, the biomass parameters of the canopy (LAI, SAI, RAI) are stored inside of the `PlantHydraulics` model, but are used across multiple canopy submodules. 

This was acceptable when plant hydraulics was required, but with the new moisture stress functions we are adding, this is no longer the case. This requires moving the area index and rooting depth out of PlantHydraulics.

**Describe the solution you'd like**
The canopy should be refactored such that these biomass values live in a separate biomass submodule of the canopy.

PR #1442, opened 2025-09-17
sif parameters from a toml

kmdeck:
## Purpose 
Add the SIF parameters to the toml

## To-do
Rebase after #600 merges

## Content

----

### DIFF
--- a/src/standalone/Vegetation/Canopy.jl
+++ b/src/standalone/Vegetation/Canopy.jl
@@ -61,6 +61,28 @@ include("./spatially_varying_parameters.jl")
 ########################################################
 # Convenience constructors for Canopy model components
 ########################################################
+
+"""
+    Lee2015SIFModel{FT}(toml_dict::CP.ParamDict) where {FT}
+
+Constructs the `Lee2015SIFModel` from the `toml_dict` using the parameters
+in the `toml_dict`.
+"""
+function Lee2015SIFModel{FT}(toml_dict::CP.ParamDict) where {FT}
+    parameters = SIFParameters{FT}(;
+        kf = toml_dict["kf"],
+        kp = toml_dict["kp"],
+        kd_p1 = toml_dict["kd_p1"],
+        kd_p2 = toml_dict["kd_p2"],
+        min_kd = toml_dict["min_kd"],
+        kn_p1 = toml_dict["kn_p1"],
+        kn_p2 = toml_dict["kn_p2"],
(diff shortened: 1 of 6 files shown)


src/standalone/Vegetation/Canopy.jl
```diff
@@ -61,6 +61,28 @@ include("./spatially_varying_parameters.jl")
 ########################################################
 # Convenience constructors for Canopy model components
 ########################################################
+
+"""
+    Lee2015SIFModel{FT}(toml_dict::CP.ParamDict) where {FT}
+
+Constructs the Lee2015SIFModel from the toml_dict using the parameters
+in the toml_dict.
```
ph-kev, line 69:
```suggestion
Constructs the `Lee2015SIFModel` from the `toml_dict` using the parameters
in the `toml_dict`.
```

src/standalone/Vegetation/Canopy.jl
```diff
@@ -61,6 +61,28 @@ include("./spatially_varying_parameters.jl")
...
+"""
+function Lee2015SIFModel{FT}(toml_dict::CP.ParamDict) where {FT}
+    parameters = SIFParameters{FT}(;
```
ph-kev, line 72: I think you can remove the `FT` here, since you can get the float type of the `toml_dict` with `ClimaParams.float_type(toml_dict)`.

toml/default_parameters.toml
```diff
@@ -405,3 +405,58 @@ tag = "BucketModelParameters"
...
+value = 0.045
+type = "float"
+description = "Slope of line relating leaf-level fluorescence to spectrometer-observed fluorescence as a function of Vcmax 25 (leaf level). Lee et al 2015."
+tag = "SIFParameters"
+
+["kappa_p2"]
+value = 7.85
+type = "float"
+description = "Intercept of line relating leaf-level fluorescence to spectrometer-observed fluorescence as a function of Vcmax 25 (leaf level). Lee et al 2015."
+tag = "SIFParameters"
```
ph-kev, line 462: Can you add a new line to this file?

ph-kev (approved):

src/standalone/Vegetation/Canopy.jl
kmdeck: I think so too, it is a matter of how consistent we want to be? I think we usually use something like Struct{FT}(;....) type constructors...even if FT is not needed..

PR #1446, opened 2025-09-18
Set compression level to 1 for NetCDF writers

ph-kev:
This PR sets the compression level to 1 for the NetCDF writers.


kmdeck:
Compression doesnt help (@ph-kev)

PR #1447, opened 2025-09-18
Add GHA for doc preview cleanup

imreddyTeja:
## Content

----

### DIFF
--- a/.github/workflows/DocCleanup.yml
+++ b/.github/workflows/DocCleanup.yml
@@ -1,0 +1,33 @@
+name: Doc Preview Cleanup
+
+on:
+  pull_request:
+    types: [closed]
+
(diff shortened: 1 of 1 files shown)


.github/workflows/DocCleanup.yml
```diff
@@ -0,0 +1,33 @@
+name: Doc Preview Cleanup
+
+on:
+  pull_request:
+    types: [closed]
+
+# Ensure that only one "Doc Preview Cleanup" workflow is force pushing at a time
+concurrency:
+  group: doc-preview-cleanup
+  cancel-in-progress: false
```
juliasloan25, line 10: Do we want `cancel-in-progress: true` so that we don't have multiple of this GHA running at the same time? We have this set to true for our other GHAs

juliasloan25 (approved):
thank you!!

PR #1451, opened 2025-09-19
Set Julia depot cache key manually to reduce the number of cache items

petebachant:
This PR sets the cache key for Julia dependencies so it's more likely to be reused, and doesn't create a new item for every single PR/workflow/attempt. With the old caching setup, I believe the cache would only be hit if rerunning a failed workflow job.

This will probably need to be propagated to other workflows, but it would be good to check if this will stop GitHub Actions from creating so many caches.

## TODO

- [x] Push a trivial change to this branch to ensure the cache is hit and reused
- [x] Check if there are more files we should be caching, e.g., precompiled Julia code?


juliasloan25:
> These caches end up being quite a bit smaller, so maybe some important files are missing from them?
> 
> image

The registries (which I saw you just added) are something we were also saving before, but I don't think would be very large. Julia-actions/cache also saves the logs by default which it says "Helps auto-Pkg.gc() keep the cache small."

petebachant:
> This is what the settings look like for that run. The cache name doesn't include anything about the julia version, os, or manifest - maybe we get a cache hit because the cache name is the same in subsequent runs, but the environment still needs to be re-built because there's no guarantee it's actually the same between runs

That was confusing me too. `docbuild` is getting cache hits, but `ci` sometimes [does](https://github.com/CliMA/ClimaLand.jl/actions/runs/17865342222/job/50806175365) and sometimes [doesn't](https://github.com/CliMA/ClimaLand.jl/actions/runs/17864707982/job/50804128293). I can't see a pattern yet.

juliasloan25:
I'm concerned that since we don't check in our test/Manifest.toml, the cache may not be used correctly. The Manifest gets regenerated for each GHA, potentially using different package versions. I think the question boils down to: if we re-generate a Manifest.toml for a given environment with the same packages, will the manifest hash be the same as the previous time it was generated? If not, I don't know that the cache will work without a checked-in Manifest

juliasloan25 (approved):
Thank you! Could you squash your commits into 1 before merging please?

PR #1452, opened 2025-09-19
Construct earth_param_set internally for drivers

### LINKED ISSUE #1424: Construct `LandParameters` internally
In PR #1385, it was discussed that  `LandParameters` should be constructed internally. As of now, we do something like

```julia
earth_param_set = LandParameters(toml_dict)
```

and pass `earth_param_set` to the various drivers. Instead of constructing `earth_param_set`, we should be able to construct a `earth_param_set` internally and pass a `toml_dict` instead.

ph-kev:
closes #1424 - This PR adds logging for parameters for long runs and make some functions take in `toml_dict` instead of `earth_param_set`.

### DIFF
--- a/experiments/long_runs/bucket.jl
+++ b/experiments/long_runs/bucket.jl
@@ -131,6 +131,7 @@ simulation = LandSimulation(
 @info "Timestep: $Δt s"
 @info "Start Date: $start_date"
 @info "Stop Date: $stop_date"
+CP.log_parameter_information(toml_dict, joinpath(root_path, "parameters.toml"))
 solve!(simulation);
 
 short_names = ["tsfc", "lhf", "shf", "wsoil"]
--- a/experiments/long_runs/bucket.jl
+++ b/experiments/long_runs/bucket.jl
@@ -48,7 +48,7 @@ diagnostics_outdir = joinpath(root_path, "global_diagnostics")
 outdir =
     ClimaUtilities.OutputPathGenerator.generate_output_path(diagnostics_outdir)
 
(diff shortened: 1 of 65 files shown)


experiments/long_runs/bucket.jl
```diff
@@ -131,6 +130,7 @@ simulation = LandSimulation(
 @info "Timestep: $Δt s"
 @info "Start Date: $start_date"
 @info "Stop Date: $stop_date"
+CP.log_parameter_information(toml_dict, joinpath(root_path, "parameters.toml"))
```
juliasloan25, line 133: Could you add a note in the docs page `docs/src/APIs/parameters.md` explaining that this function can be used to log the params used in a simulation?

juliasloan25 (approved):
Thanks Kevin, this is a nice improvement!

There are still some places in docs/ where we create an `earth_param_set` to get either:
- `thermo_params` to set up the atmosphere TVIs, or 
- `insol_params` to compute the zenith angle

I think this okay, especially for the `thermo_params` since in those cases the user is specifying a custom atmosphere. For the `insol_params` cases, we could think about how to isolate more setup of a default zenith angle function (maybe with a function that takes in `toml_dict`)

One comment on a file that doesn't show up in the diff:
- docs/src/tutorials/standalone/Soil/soil_energy_hydrology.jl - don't need to define `earth_param_set` anymore

experiments/long_runs/bucket.jl
ph-kev: I think this is already in the docs. I am not sure what else to add to the existing docs.

kmdeck (approved):
This looks good to me. thank you!

PR #1453, opened 2025-09-19
Steady state hydraulics option

kmdeck:
## Purpose 
Add a steady state plant hydraulics model. in this case the root water extraction is set to transpiration in integrated models (the soil needs it) but the canopy does not need this water flux because by definition root_water_flux = transpiration and we do not solve for the water content in the canopy. this means we still need p.root_extraction, but not p.canopy.fa_roots

## To-do
add restriction - cannot run with steady state unless you pick piecewise soil moisture stress
test long runs

## Content

----

### DIFF
--- a/ext/fluxnet_simulations/initial_conditions.jl
+++ b/ext/fluxnet_simulations/initial_conditions.jl
@@ -186,26 +186,28 @@ function set_fluxnet_ic!(
     )
 
     Y.canopy.energy.T .= T_air_0
-    FT = eltype(Y.canopy.energy.T)
-    ψ_stem_0 = FT(-1e5 / 9800) # pressure in the leaf divided by rho_liquid*gravitational acceleration [m]
-    ψ_leaf_0 = FT(-2e5 / 9800)
     hydraulics = model.hydraulics
-    n_stem = hydraulics.n_stem
-    n_leaf = hydraulics.n_leaf
-    ψ_comps = n_stem > 0 ? [ψ_stem_0, ψ_leaf_0] : ψ_leaf_0
-    S_l_ini =
-        ClimaLand.Canopy.PlantHydraulics.inverse_water_retention_curve.(
-            hydraulics.parameters.retention_model,
-            ψ_comps,
-            hydraulics.parameters.ν,
-            hydraulics.parameters.S_s,
-        )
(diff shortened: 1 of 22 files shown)


ext/fluxnet_simulations/initial_conditions.jl
```diff
@@ -176,26 +176,28 @@ function set_fluxnet_ic!(
...
-        ClimaLand.Canopy.PlantHydraulics.inverse_water_retention_curve.(
-            hydraulics.parameters.retention_model,
-            ψ_comps,
-            hydraulics.parameters.ν,
-            hydraulics.parameters.S_s,
-        )
-    for i in 1:(n_stem + n_leaf)
-        Y.canopy.hydraulics.ϑ_l.:($i) .=
-            ClimaLand.Canopy.PlantHydraulics.augmented_liquid_fraction.(
+    if hydraulics isa ClimaLand.Canopy.PlantHydraulicsModel
```
kmdeck, line 180: I realized that our set_ic! methods are quite brittle (e.g. they assume temperature is prognostic, and also something about plant hydraulics prior to this change). both here and in src/simulations/initial_conditions.jl

src/integrated/land.jl
```diff
@@ -713,13 +716,12 @@ function make_set_initial_cache(model::Union{LandModel, SoilCanopyModel})
...
         # Make sure that the hydraulics scheme and the biomass scheme are compatible
         hydraulics = canopy.hydraulics
-        n_stem = hydraulics.n_stem
-        n_leaf = hydraulics.n_leaf
-        Canopy.lai_consistency_check.(
-            n_stem,
-            n_leaf,
-            p.canopy.biomass.area_index,
-        )
+        if hydraulics isa PlantHydraulics.PlantHydraulicsModel
```
kmdeck, line 719: this pattern appears twice in the code now

src/diagnostics/default_diagnostics.jl
```diff
@@ -393,8 +402,6 @@ function get_possible_diagnostics(model::CanopyModel)
         "trans",
         "clhf",
         "cshf",
-        "lwp",
-        # "fa", # return a Tuple
         "far",
```
kmdeck, line 405: delete this

PR #1457, opened 2025-09-22
New doc structure

### LINKED ISSUE #1462: Reorganize documentation
As we discussed briefly during the last meeting, particularly following MC3 feedbacks, I am reorganizing ClimaLand documentation a bit.
I think it is good for a new user to see what they expect to see from a land model at a glance in the menu. See image 1st attempt.
I am thinking further (the order is important and TBD):
- Home (landing page, broad intro to CliMA and ClimaLand)
- Fundamental concepts (standalone and integrated models, prognostic and diagnostic variables, domain, parameters, boundary conditions, initialization, forcings, etc.)
- Running you first simulation (a simple example to run a model, which gives a concrete application of the above fundamental concepts)
- Tutorials (more examples, more in details of both the physics and the code, for all models - this is a big subsection)
note that "burried" in tutorials are currently intro to ClimaLand Domains and Intro to timestepping which may be good to expose
- Equations (this explains the equations of each component, e.g., canopy photosynthesis)
- Calibration (single site, global, explain how we do it with EKP.jl and ClimaCalibrate.jl)
- Benchmark (this contains our leaderboard docs, I prefer naming it benchmark which is a more familiar term for the community?)
- Outputs (this is also a renaming of diagnostics. I think many land scientists are not familiar with the term diagnostics. But we can use either.)
- Running on GPU or with MPI
- Additional resources (contains some info about Julia, physical units, repo structure, restarting a simulation, Itime type)
- API
- Contributor guides (some tips about CI, formatting, etc.)
- References

AlexisRenchon:
closes #1462 

This new documentation structure is primarily targeted to scientists who use land surface models, and secondly to developers who may not be familiar with ClimaLand yet.
The goal in this commit is to have sections expected for a land model (run the model, model outputs, calibration, benchmark, etc.)

https://clima.github.io/ClimaLand.jl/previews/PR1457/

TODO:

From Julia:
- [x] "Model Outputs" could be more specific - maybe "Accessing Model Outputs" or "Analyzing Model Outputs"?
- [x] It might be helpful to have the "Julia Background" page visible at the top level. It can be near the bottom, but I think many people wouldn't expect we necessarily have a page like that, so more visibility would be useful
- [x] We could make the "Model Calibration" page a section instead, and move the calibration tutorials there so everything's in one place (credit to Kat for this suggestion!)
- [x] We could also rename to "Calibrating Model Parameters", unless "Model Calibration" is widely used and understood to refer to parameters
- [x] We should also expose the "Parameters" docs page to users more because it has a lot of useful information and right now it's hidden in the API docs. I would propose we copy everything above "API" on that page to a new docs page "Model Parameters" (or a similar name) which is available at the top level

From Kat:
- [x] Benchmark could be a subsection of calibration

### DIFF
--- a/docs/list_of_apis.jl
+++ b/docs/list_of_apis.jl
@@ -32,5 +32,4 @@ apis = [
         "Simulations" => "APIs/Simulations.md",
         "Solver Functions" => "APIs/SolverFunctions.md",
     ],
-    "Parameters" => "APIs/parameters.md",
 ]
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -11,7 +11,25 @@ using Pkg
 Pkg.add(ClimaLand)
 ```
 
-## Running a simple soil model
+## Steps to run a model
+
+To run a model, you will always need to do these steps:
+- Import dependencies. You need some libraries to run the model.
+- Choose floating point precision: Float32 or Float64.
+- Import [parameters via .toml files](@ref parameters_page). You need some constants and parameters values to run the model.
+- Define your domain. This specifies where you are solving the equations: globally (on the sphere),
+in a region (in a box), or at a site (in a column). See the [tutorial to ClimaLand domains!](@ref "Domain Tutorial")
+- Set your simulation period: start date, end date, timestep.
+- Set the forcings (for example, air temperature, precipitations).
+- Set the initial conditions. Differential equations required initial conditions to solve.
+For some simulations (e.g., global land simualtions), we provide default initial condition ("ic") functions,
+but in many cases you need to create your own `set_ic!` function.
+- Set how output is saved. For some simulations (e.g., global land simulations), we use the default diagnostic output,
+corresponding to monthly averages of specific variables saved to netcdf files.
+For your use case, you may wish to change these defaults.
+See [writing and accessing outputs](@ref "Using ClimaLand Diagnostics to save simulation output").
+
+## Example: running a simple soil model
 
 Now that we have Julia installed, let's set up and run a simple ClimaLand simulation!
 You can follow along by copying the following code segments into your REPL.
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -66,27 +66,28 @@ include("list_standalone_models.jl")
 include("list_diagnostics.jl")
 pages = Any[
     "Home" => "index.md",
+    "Fundamental Concepts" => "fundamental_concepts.md",
     "Running your first simulation" => "getting_started.md",
+    "Available Models and Parameterizations" => "available_models.md",
+    "Parameters" => "parameters.md",
     "Tutorials" => tutorials,
+    "Writing and accessing outputs" => diagnostics,
     "Model Equations" => standalone_models,
+    "Calibration and benchmark" => [
+        "Calibrating model parameters" => "calibration.md",
+        "Model Benchmark" => "leaderboard/leaderboard.md",
+    ],
+    "Running on GPU or with MPI" => "architectures.md",
     "Additional resources" => [
-        "Model structure" => "model_structure.md",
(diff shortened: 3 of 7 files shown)


docs/list_of_apis.jl
```diff
@@ -29,5 +29,4 @@ apis = [
         "Drivers" => "APIs/Drivers.md",
         "Callbacks" => "APIs/Callbacks.md",
     ],
-    "Parameters" => "APIs/parameters.md",
```
kmdeck, line 32: did we mean to delete this?

docs/src/getting_started.md
```diff
@@ -11,7 +11,18 @@ using Pkg
 Pkg.add(ClimaLand)
 ```
 
-## Running a simple soil model
+## Steps to run a model
+
+To run a model, you will always need to do these steps:
+- Import dependencies. You need some libraries to run the model.
+- Chose Float32 or Float64.
+- Import parameters. You need some constants and parameters values to run the model.
```
kmdeck, line 19: we can note that parameters are specified via toml files, see page X for details

docs/src/getting_started.md
```diff
@@ -11,7 +11,18 @@ using Pkg
...
+- Define your domains. You need a domain to solve PDEs. See the [tutorial to ClimaLand domains!](@ref "Domain Tutorial")
+- Set your simulation period: start date, end date, timestep.
+- Import the forcings (for example, air temperature, precipitations).
```
kmdeck, line 22: maybe "Set up the forcing data" since "Import" has a specific meaning, and the forcings are not a module the user imports.

docs/src/getting_started.md
```diff
@@ -11,7 +11,18 @@ using Pkg
...
+- Set the initial conditions. PDEs are solved within a domain, advancing in time from initial conditions. (For example, initial soil moisture).
+
+## Example: running a simple soil model
```
kmdeck, line 25: I think we might also include:
- "Set how output is saved. For some simulations (e.g., global land simulations), we use the default diagnostic output, corresponding to monthly averages of specific variables saved to netcdf files. For your use case, you may wish to change these defaults. See X for details."

docs/src/getting_started.md
```diff
@@ -11,7 +11,18 @@ using Pkg
 Pkg.add(ClimaLand)
 ```
 
-## Running a simple soil model
+## Steps to run a model
+
+To run a model, you will always need to do these steps:
+- Import dependencies. You need some libraries to run the model.
+- Chose Float32 or Float64.
```
kmdeck, line 18: "Choose floating point precision: Float32 or Float64"

juliasloan25 (approved):
This looks great, thank you! I would suggest moving the "Tutorials" section to be right after "Running your first simulation" so it doesn't get lost among the more informational pages/tabs

docs/src/getting_started.md
```diff
@@ -11,7 +11,25 @@ using Pkg
...
+To run a model, you will always need to do these steps:
+- Import dependencies. You need some libraries to run the model.
+- Choose floating point precision: Float32 or Float64.
+- Import [parameters via .toml files](@ref parameters_page). You need some constants and parameters values to run the model.
+- Define your domain. This specifies where you are solving the equations: globally (on the sphere),
+in a region (in a box), or at a site (in a column). See the [tutorial to ClimaLand domains!](@ref "Domain Tutorial")
+- Set your simulation period: start date, end date, timestep.
+- Set the forcings (for example, air temperature, precipitations).
+- Set the initial conditions. Differential equations required initial conditions to solve.
+For some simulations (e.g., global land simualtions), we provide default initial condition ("ic") functions,
```
kmdeck, line 25: simulations (typo)

PR #1471, opened 2025-09-26
Update calibration pipeline

ph-kev:
This PR updates the calibration pipeline by

- generating the observations as part of the slurm/pbs scripts,
- enabling the diagnostics that can be calibrated right now,
- update the model interface,
- not enabling the `NaNCallback`,
- and logging parameters before solving the simulation.

TODO

- [ ] Test calibration pipeline on either GCP or Derecho


experiments/calibration/generate_observations.jl
```diff
@@ -161,6 +161,7 @@ function preprocess_single_era5_var(var::OutputVar, short_name, nelements)
 end
```
kmdeck, line 161: A note that I got an error running this as part of the job on GPU - it said it I needed to call the clima comms function import required backends.

nefrathenrici (approved):
Thanks for cleaning this up. It would be good to use the `GEnsembleBuilder` once we make a new release of ClimaCalibrate as well.

PR #1481, opened 2025-10-02
Add tests and buildkite for calibration pipeline

### LINKED ISSUE #1370: Make a buildkite CI to ensure the calibration is maintained
It would be good to set up a buildkite for a perfect model calibration using simulation data as the observations. In addition, there should be a GitHub label created for running the buildkite for the calibration.

ph-kev:
closes #1370

This PR adds a new buildkite for testing a calibration which can be found here: https://buildkite.com/clima/climaland-calibrate-e2e

The calibration is meant to be short and exercise every single line of the calibration code rather than test a realistic calibration like a perfect model calibration.

### DIFF
--- a/.buildkite/calibration/pipeline.yml
+++ b/.buildkite/calibration/pipeline.yml
@@ -1,0 +1,54 @@
+agents:
+  queue: clima
+  slurm_mem: 8G
+  modules: climacommon
+
+steps:
+  - label: "init :GPU:"
+    command:
+      - echo "--- Instantiate"
+
+      - "julia --project -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
+      - "julia --project -e 'using Pkg; Pkg.status()'"
+
+      - julia --project=.buildkite -e 'using Pkg; Pkg.instantiate(;verbose=true)'
+      - julia --project=.buildkite -e 'using Pkg; Pkg.precompile()'
+      - julia --project=.buildkite -e 'using CUDA; CUDA.precompile_runtime()'
+      - julia --project=.buildkite -e 'using Pkg; Pkg.status()'
+
+    agents:
+      slurm_gpus: 1
+      slurm_cpus_per_task: 8
+    env:
+      JULIA_NUM_PRECOMPILE_TASKS: 8
+      JULIA_MAX_NUM_PRECOMPILE_FILES: 50
+
+  - wait
+
+  - label: "Generate observations"
+    command:
+      - julia --color=yes --project=.buildkite/ experiments/calibration/generate_observations.jl
+    env:
+      # Generating observations doesn't require a CUDA, but these environment
+      # variables may be loaded in the slurm script
+      CLIMACOMMS_DEVICE: "CUDA"
+      CLIMACOMMS_CONTEXT: "SINGLETON"
+
+  - wait
+
+  - label: "Calibration"
+    command:
+      - julia --color=yes --project=.buildkite/ experiments/calibration/run_calibration.jl
+    agents:
+      slurm_time: 2:00:00
+      slurm_gpus: 1
+    env:
+      CLIMACOMMS_DEVICE: "CUDA"
+      CLIMACOMMS_CONTEXT: "SINGLETON"
+    artifact_paths: "experiments/calibration/land_model/**/*.png"
+
+  - wait
+
+  - label: "Test calibration results"
+    command:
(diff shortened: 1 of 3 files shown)


.buildkite/calibration/pipeline.yml
```diff
@@ -0,0 +1,57 @@
+agents:
+  queue: clima
+  slurm_mem: 8G
+  modules: climacommon
+
+env:
+  JULIA_CUDA_MEMORY_POOL: "cuda"
```
nefrathenrici, line 7: Is this needed?

experiments/calibration/test/test_calibration.jl
```diff
@@ -0,0 +1,54 @@
...
+
+    (; output_dir, n_iterations) = CALIBRATE_CONFIG
+    ekp = JLD2.load_object(ClimaCalibrate.ekp_path(output_dir, n_iterations))
+
+    @test EKP.get_N_iterations(ekp) == n_iterations
+
+    # Test if covariance weighted loss decreases
+    # If there is only one iteration, then there is only one value for the error
+    if n_iterations > 1
+        @test first(ekp.error_metrics["loss"]) > last(ekp.error_metrics["loss"])
```
nefrathenrici, line 45: We should check that the error has decreased by a certain fraction (0.5, 0.2 etc).

experiments/calibration/test/test_calibration.jl
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    # Spread should decrease
+    ekp_u = EKP.get_u(ekp)
+    spreads = map(std, ekp_u)
+    @test first(spreads) > last(spreads)
```
nefrathenrici, line 51: Same here - we should expect significant convergence if the calibration has succeeded.

nefrathenrici (commented):
This looks good to me.

experiments/calibration/test/test_calibration.jl
ph-kev: I don't know how well this calibration does, since only one iteration is done. Should I remove this test?

.buildkite/calibration/pipeline.yml
ph-kev: I removed this and I don't think it made a difference.

experiments/calibration/test/test_calibration.jl
kmdeck: I think you can remove this test. sometimes the error increases on the first iteration before decreasing, or it can bounce around if the model error is really big or if the covariance is far too large.

experiments/calibration/test/test_calibration.jl
```diff
@@ -0,0 +1,54 @@
...
+    # Test if covariance weighted loss decreases
+    # If there is only one iteration, then there is only one value for the error
+    if n_iterations > 1
+        @test first(ekp.error_metrics["loss"]) > last(ekp.error_metrics["loss"])
+    end
+
+    # Spread should decrease
+    ekp_u = EKP.get_u(ekp)
+    spreads = map(std, ekp_u)
+    @test 0.1 * first(spreads) > last(spreads)
```
kmdeck, line 51: what does this check?

ph-kev: This checks the parameter values in unconstrained space converge from the first and last iteration.

experiments/calibration/test/test_calibration.jl
```diff
@@ -0,0 +1,54 @@
...
+
+@testset "Calibration results" begin
+    isdir(CALIBRATE_CONFIG.output_dir) || error(
+        "Cannot find output directory of calibration. These tests will not work without first running a test calibration",
+    )
+
+    (; output_dir, n_iterations) = CALIBRATE_CONFIG
+    ekp = JLD2.load_object(ClimaCalibrate.ekp_path(output_dir, n_iterations))
+
+    @test EKP.get_N_iterations(ekp) == n_iterations
```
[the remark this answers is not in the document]
ph-kev: I don't think that test makes sense to added since it depends on the process that is being used from `EKP`.

experiments/calibration/test/test_calibration.jl
kmdeck: this seems specific to the calibration run but i think it's fine to keep.

(for example, the prior might be too narrow to start with)

kmdeck (approved):

nefrathenrici (approved):

PR #1497, opened 2025-10-17
Add bucket as an option for calibration

ph-kev:
This PR sets up the infrastructure for running a calibration with the bucket model. Because there are limited options for observations and diagnostics available for the bucket model, the bucket model can only be calibrated with `shf`.

### DIFF
--- a/docs/src/calibration.md
+++ b/docs/src/calibration.md
@@ -285,6 +285,7 @@ CalibrateConfig(;
     nelements = (101, 15),
     output_dir = "experiments/calibration/land_model",
     rng_seed = 42,
+    model_type = ClimaLand.LandModel,
 )
 ```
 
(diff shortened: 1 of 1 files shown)


experiments/calibration/models/snowy_land.jl
```diff
@@ -0,0 +1,193 @@
...
+# Number of spatial elements: 101 in horizontal, 15 in vertical
+# Soil depth: 50 m
+# Simulation duration: 730 d
+# Timestep: 450 s
+# Timestepper: ARS111
+# Fixed number of iterations: 3
+# Jacobian update: every new Newton iteration
+# Atmos forcing update: every 3 hours
+
+function setup_model(
```
kmdeck: I recently ran into a compiler bug on Clima and locally with this function. It had to be changed to 
function setup_model(::Type{FT},...) where {FT}

experiments/calibration/observation_map.jl
```diff
@@ -154,6 +154,8 @@ function ClimaCalibrate.analyze_iteration(
     plot_output_path = ClimaCalibrate.path_to_iteration(output_dir, iteration)
     plot_constrained_params_and_errors(plot_output_path, ekp, prior)
 
+    config = CALIBRATE_CONFIG.configuration
+    config == Val{:snowy_land} || return nothing
```
kmdeck: What is this?

experiments/calibration/models/snowy_land.jl
kmdeck: Dennis can better explain, but it was something about this type being undefined when the function is being compiled and so something downstream inside the function broke when it used {FT} in its constructor

juliasloan25: This constructor is duplicated between the longrun and here. It would be nice if we moved it to a shared location so we don't have to update both every time anything changes.

It also looks like `forward_model` is mostly agnostic to the model type being used, except for passing through the model type to `setup_model` but we should be able to name that variable to avoid hardcoding the Val. The other things would be diagnostics and logging some info about the simulation - maybe we could pull these out into functions that dispatch on the model type.

What do you think? These don't have to be done in this PR (especially the `setup_model` change), but could be a good followup

ph-kev: Yes, I think this would a good idea to do

kmdeck (approved):
I think this looks good! Thanks!

PR #1502, opened 2025-10-22
Spatially varying canopy height

braghiere:
## Purpose 
Add support for spatially-varying canopy height in ClimaLand vegetation model, replacing the fixed 1m default with realistic height data from CLM artifacts (allowing for a cap).

### Core Implementation
- **Canopy height parameterization**: Added `clm_canopy_height()` function to read spatially-varying canopy height from CLM artifact data (`src/standalone/Vegetation/spatially_varying_parameters.jl`), optionally capping it
- **Type system updates**: Extended `PrescribedBiomassModel` to support spatially-varying height through type parameter `HTH` (canopy height type)

### Testing and Validation
- **Unit tests**: Comprehensive test suite covering spatially-varying height functionality, including Field operations, capping behavior, and integration with canopy model

## Impact Assessment

image

*Figure: Sensible heat flux comparison showing Run 1 (default 1m height), Run 2 (spatially-varying height), absolute difference, and percent change.*

The spatially-varying canopy height primarily affects:
- **Turbulent fluxes**: Changes in aerodynamic resistance modify sensible and latent heat exchange with atmosphere
- **Regional patterns**: Strongest impacts in forested regions (tropical/temperate) where height exceeds 1m default
- **Minimal impact**: Grasslands and croplands show small changes (CLM heights ~1m already)

## To-do
- [ ] Add NEWS.md entry after reviewer feedback

---

### DIFF
--- a/src/standalone/Vegetation/Canopy.jl
+++ b/src/standalone/Vegetation/Canopy.jl
@@ -293,7 +293,7 @@ end
         n_stem::Int = 0,
         n_leaf::Int = 1,
         h_stem::FT = FT(0),
-        h_leaf::FT = FT(1),
+        h_leaf::FT = toml_dict["canopy_height"],
         ν::FT = FT(1.44e-4),
         S_s::FT = FT(1e-2 * 0.0098), # m3/m3/MPa to m3/m3/m
         conductivity_model = Weibull{FT}(


src/standalone/Vegetation/Canopy.jl
```diff
@@ -693,9 +694,18 @@ function CanopyModel{FT}(;
...
+            hydraulics.compartment_surfaces[end] -
+            hydraulics.compartment_surfaces[1]
+        # For scalar height: direct comparison
+        # For Field height: check all values match (with tolerance for floating point)
+        if biomass.height isa ClimaCore.Fields.Field
+            @assert all(
+                isapprox.(biomass.height, hydraulics_height, rtol = 1e-10),
+            ) "Biomass height Field must match hydraulics height"
+        else
+            @assert biomass.height ≈ hydraulics_height "Biomass height must match hydraulics height"
```
kmdeck: I dont quite understand how this works, because the hydraulics_height is a scalar. How can this pass when biomass.height is spatialy varying?

In general we need to go through src and make sure these are consistent, because in some places we use the 'hydraulics" height as the height

experiments/long_runs/snowy_land_pmodel.jl
```diff
@@ -106,6 +108,21 @@ function setup_model(FT, start_date, stop_date, Δt, domain, toml_dict)
...
 
+    # Read spatially-varying canopy height from CLM data
+    raw_canopy_height = clm_canopy_height(surface_space)
+    # Cap height to stay below atmospheric reference height (ERA5 at 10m)
+    # Use 8m max to leave 2m buffer for atmospheric coupling
+    canopy_height =
+        effective_canopy_height(raw_canopy_height, FT(10.0); buffer = FT(2.0))
+
+    # Create biomass model with spatially-varying height
+    biomass = PrescribedBiomassModel{FT}(
```
kmdeck: I think it would be good to add lines 112-116 into the PrescribedBiomassModel constructor, so it is not something that has to be updated each time (if we decide to include this as the default)

src/standalone/Vegetation/spatially_varying_parameters.jl
```diff
@@ -305,3 +305,118 @@ function clm_medlyn_g1(
...
+    context = ClimaComms.context(surface_space)
+    clm_artifact_path = Artifacts.clm_data_folder_path(; context, lowres)
+
+    canopy_height = SpaceVaryingInput(
+        joinpath(clm_artifact_path, "vegetation_properties_map.nc"),
+        "z_top",
+        surface_space;
+        regridder_type,
+        regridder_kwargs = (; extrapolation_bc, interpolation_method),
+    )
```
juliasloan25: We could provide the `effective_canopy_height` as a preprocessing function instead, so we don't have to think about it wherever we access the canopy height data. This would be safer because we'll avoid users forgetting to apply the clipping.
```
        file_reader_kwargs = (; preprocess_func = (data) -> effective_canopy_height(data, z_atm; buffer),),
```
And then we take `z_atm` and `buffer` as inputs to this function

experiments/long_runs/run2_spatially_varying_only.jl
```diff
@@ -0,0 +1,187 @@
+# RUN 2 ONLY: Spatially-Varying Canopy Height
```
juliasloan25: If this is going to be merged into main it would be good to rename to something like `snowy_land_varying_height.jl`

src/standalone/Vegetation/Canopy.jl
braghiere: You're right - the hydraulics compartments are always scalar (1m). When biomass.height varies spatially, only the aerodynamic calculations use those varying heights. The hydraulics structure stays uniform everywhere.

This is a current limitation - we can extend hydraulics to support spatial variation in a follow-up PR.

Project.toml
```diff
@@ -72,3 +72,10 @@ StatsBase = "0.34"
 SurfaceFluxes = "0.12.3, 0.13"
 Thermodynamics = "0.14.1, 0.15"
 julia = "1.10"
+JuliaFormatter = "2.1.7"
```
AlexisRenchon: I don't recall why exactly, but we still use JuliaFormatter 1. (not 2.) 

JuliaFormatter shouldn't be a dependency of ClimaLand, you can keep it in your default environment 

do: 
]add JuliaFormatter@1

this will install the latest v1 

to reset this to main, do: 
git checkout origin/main Project.toml

braghiere:
> let's try running this and see how the SHF, LHF, etc change compared to observations
> 
> Moving forwards, I think there are two ways this can be extended:
> 
> * make a canopy roughness model type. right now we are using physical quantities (height) to compute roughness and displacement height (using physical equations someone has come up with). Something more flexible would be to predict roughness, displacement height, from inputs like LAI, canopy height, etc using a data driven model. Having a  type for "canopy roughness" would let us switch between the two easily. (I am separately adding this already in another PR)
> * in Biomass - create a type of "height model": this would be prescribed (constant or spatially varying) for now, PrescribedCanopyHeight. Then in the future we can also easily make different parameterizations for canopy height (e.g. prognostic, predicting it from input/time varying data). but this seems very far in the future compared to the first point

Largest changes are observed: 

Sensible Heat (ERA5): Bias ↓ (8.27 → 7.89 W m⁻²) ~5% improvement, RMSE same.
[here](https://braghiere.github.io/ILAMB_CliMA_test/HydrologyCycle/SensibleHeat/ERA5/ERA5.html)

Evaporative Fraction (ERA5): Bias slightly closer to 0 (−0.0626 → −0.0601) ~4% improvement, Phase same. (No RMSE.) 
[here](https://braghiere.github.io/ILAMB_CliMA_test/HydrologyCycle/EvaporativeFraction/ERA5/ERA5.html)

Surface Upward LW (CERES): Bias closer to 0 (−5.85 → −5.52 W m⁻²) ~6% improvement, RMSE ~ +0.1, Phase ≈ same. 
[here](https://braghiere.github.io/ILAMB_CliMA_test/RadiationandEnergyCycle/SurfaceUpwardLWRadiation/CERESed4.2/CERESed4.2.html)

All the other variables have minimal changes.

src/standalone/Vegetation/Canopy.jl
```diff
@@ -692,10 +693,37 @@ function CanopyModel{FT}(;
         @assert typeof(photosynthesis) <: PModel{FT} "When using PModelConductance for stomatal conductance, you must also use PModel for photosynthesis"
     end
 
+    # Height consistency check between biomass and hydraulics:
```
juliasloan25: Could you move this comment into the docstring? That way it'll be visible in the docs too

src/standalone/Vegetation/Canopy.jl
```diff
@@ -692,10 +693,37 @@ function CanopyModel{FT}(;
...
+    #
+    # biomass.height:
+    #   - Used for: Aerodynamic calculations (displacement height, roughness length)
```
juliasloan25: I see that the biomass.height is used in `displacement_height`, but it looks like the roughness lengths still independently come from the toml dict [here](https://github.com/CliMA/ClimaLand.jl/blob/7787892f6787c32f326cf06e5011c5ef69772dea/src/standalone/Vegetation/Canopy.jl#L799-L800)

src/standalone/Vegetation/spatially_varying_parameters.jl
```diff
@@ -305,3 +306,124 @@ function clm_medlyn_g1(
...
+height = effective_canopy_height(raw_height, z_atm)
+```
+
+# Notes
+The function will log a warning message indicating how many grid cells had their
+canopy height capped and provide statistics about the original height distribution.
+This typically affects ~7% of global cells, primarily in regions with tall forests
+(e.g., tropical rainforests, temperate forests in Patagonia and New Zealand).
+"""
+function effective_canopy_height(
```
kmdeck: I think this OK for now! In the future, we can adjust it. because, what actually matters (at least in standard Monin-Obukhov, I think) is not quite z_atm - h_canopy, but z_atmos - displacement height. so I think this will transform to be 
```
d = displacement_height(canopy_height..., LAI, maxLAI,....)
effective_d = min(d, z_atm - buffer)
```
and perhaps we even compute this when we compute the surface fluxes, not as an a-priori step (esp if displacement height changes in time)

src/standalone/Vegetation/Canopy.jl
kmdeck: Yes, I think this PR just adjusts displacement height. The roughness lengths are either constant (set in the toml, the default) or computed  like so: https://github.com/CliMA/ClimaLand.jl/blob/7787892f6787c32f326cf06e5011c5ef69772dea/experiments/long_runs/snowy_land_pmodel.jl#L97-L99

but they are required to be scalar.

I am adding a PR which allows these to be spatially varying, among other things.

kmdeck (commented):
I think this looks good, but I dont love that we now have multiple definitions of height, and have some pause since we arent sure we want to predict canopy height in the future.

I wonder, since these additions affect the turbulent boundary fluxes only, and because the height only affects the roughness lengths and displacment height but is not used directly itself, if we should move `height` out of the biomass model and only compute displacement and roughness length fields using the height once, and store those, and not the height.

This isnt really possible in this PR, but after PR #1505 merges, I think we can do something like this.

src/standalone/Vegetation/spatially_varying_parameters.jl
kmdeck: but I can change it to be the preprocessing function instead

PR #1505, opened 2025-10-22
Canopy roughness parameterization modular

### LINKED ISSUE #1351: Roughness lengths for canopy
The canopy roughness and displacement heights are examples of parameterizations. E.g. right now we support them being constant globally, and there are entries in our parameter toml for roughness lengths. However, in many examples, these constants are chosen using a canopy height and the defaults are overwritten:
```
z_0m = 0.1 h_canopy
z_0b = 0.01 h_canopy
d = 0.67 h_canopy
```
This is undesirable for a few reasons:
- Our default is often overwritten, so maybe the other case should be the default
- We dont make it easy to change roughness/displacement parameterizations
- roughness, displacement, should eventually support being a function of space (fields and not scalars)

To do:
- make the roughness length and displacement height parameterizations modular
- make it so that one choice of parameterizations is the height based approach above, and use this as the default
- remove roughness lengths from the parameter toml
- make it possible for roughness lengths and displacement heights to be fields

kmdeck:
## Purpose 
Our default sets roughness lengths for the canopy as parameters using the toml. In almost all our example, though, we override these by using a constant times the canopy height. This PR updates the default so that it does what we want

This also creates a roughness parameterization struct which we can dispatch off of when computing surface fluxes. Since the shared canopy parameter struct then only has the earth param set, we can remove that from our code; this also enables us to change roughness paramterizations in the future

Closes #1351 

The science content is in src/standalone/Vegetation/canopy_boundary_fluxes.jl and src/standalone/Vegetation/canopy_surface_fluxes.jl; other files are mostly software changes only
## To-do
run long runs before merging

## Content
- Creates a MonionObukhovCanopyFluxes object which specifies how we compute the turbulent fluxes. This contains the roughness lengths, displacement height, and another constant which had been hardcoded in our SF parameterization.
- Creates a constructor for this type which sets roughness lengths and displacement height using the canopy height. Sets this to be the default, which simplifies our model setup.
- Gets rid of the SharedCanopyParameters struct which only contained roughness lengths and the earth_param_set
- Updates tests, experiments, docs

----

### DIFF
--- a/src/standalone/Vegetation/canopy_boundary_fluxes.jl
+++ b/src/standalone/Vegetation/canopy_boundary_fluxes.jl
@@ -3,14 +3,12 @@ using Thermodynamics
 using StaticArrays
 import SurfaceFluxes.Parameters as SFP
 
-import ClimaLand:
-    surface_temperature,
-    surface_evaporative_scaling,
-    surface_height,
-    surface_resistance,
-    displacement_height,
-    turbulent_fluxes!,
-    AbstractBC
+import ClimaLand: turbulent_fluxes!, AbstractBC, get_earth_param_set
+
+function get_earth_param_set(model::CanopyModel)
+    return model.earth_param_set
+end
+
 
 """
     AbstractCanopyBC <: ClimaLand.AbstractBC
@@ -23,6 +21,7 @@ abstract type AbstractCanopyBC <: ClimaLand.AbstractBC end
         A <: AbstractAtmosphericDrivers,
         B <: AbstractRadiativeDrivers,
         G <: AbstractGroundConditions,
+        R <: AbstractCanopyFluxParameterization,
         C::Tuple
     } <: AbstractCanopyBC
 
@@ -39,6 +38,7 @@ struct AtmosDrivenCanopyBC{
     A <: AbstractAtmosphericDrivers,
     B <: AbstractRadiativeDrivers,
     G <: AbstractGroundConditions,
+    R <: AbstractCanopyFluxParameterization,
     C <: Tuple,
 } <: AbstractCanopyBC
     "The atmospheric conditions driving the model"
@@ -47,6 +47,8 @@ struct AtmosDrivenCanopyBC{
     radiation::B
     "Ground conditions"
     ground::G
+    "Turbulent flux (latent, sensible, vapor, and momentum) parameterization"
+    turbulent_flux_parameterization::R
     "Prognostic land components present"
     prognostic_land_components::C
 end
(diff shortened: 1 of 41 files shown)


src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -47,6 +58,8 @@ struct AtmosDrivenCanopyBC{
     radiation::B
     "Ground conditions"
     ground::G
+    "Surface flux parameterization"
```
kmdeck: Maybe this can be clearer: "Surface flux (latent and sensible heat) parameterization"

src/standalone/Vegetation/canopy_surface_fluxes.jl
```diff
@@ -0,0 +1,61 @@
...
+    "Constant used to compute the resistance for fluxes between the leaves and canopy airspace (unitless)"
+    η::FT
+end
+
+"""
+    MoninObukhovHeightBased(toml_dict, height)
+
+A constructor for a MoninObukhovCanopyFluxes surface flux theory,
+ specifying how to compute latent and sensible heat
+fluxes between the atmosphere and the canopy based on Monin-Obukhov
```
juliasloan25: also turbulent moisture flux and momentum fluxes right? Maybe we could just say "specifying how to compute turbulent fluxes between the atmosphere and canopy ..."

toml/default_parameters.toml
```diff
@@ -466,3 +454,34 @@ value = 7.85
 type = "float"
 description = "Intercept of line relating leaf-level fluorescence to spectrometer-observed fluorescence as a function of Vcmax 25 (leaf level). Lee et al 2015."
 tag = "SIFParameters"
+
+# Canopy roughness
+["canopy_z_0min"]
+value = 0.013
+type = "float"
+description = "Minimum roughness"
+tag = "ConstantRoughnessModel"
```
juliasloan25: I don't see `ConstantRoughnessModel` defined in the src code, should this be `MoninObukhovCanopyFluxes`?

src/standalone/Vegetation/canopy_boundary_fluxes.jl
```diff
@@ -3,14 +3,23 @@ using Thermodynamics
...
-import ClimaLand:
-    surface_temperature,
-    surface_evaporative_scaling,
-    surface_height,
-    surface_resistance,
-    displacement_height,
-    turbulent_fluxes!,
-    AbstractBC
+import ClimaLand: turbulent_fluxes!, AbstractBC, surface_air_density
+function surface_air_density(
```
juliasloan25: Can the canopy use the default method defined in src/shared_utilities/drivers.jl? If the only difference is how we access `earth_param_set`, it might be cleaner to define a function `get_earth_param_set(model::...)`

kmdeck: Ill make this change!

AlexisRenchon (approved):
This looks good to me. It is nice to simplify the experiments!

szy21:
I wonder if this PR breaks the coupler: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/631#019ab8be-4b08-4fa2-9708-6c9ea702de40/174-296? But maybe we are going to change it anyway so it would be ok.

Issue #1476, opened 2025-09-30
remaining pieces from Yuchen's PR


juliasloan25:
There are a couple small pieces of work left from Yuchen's PR #1330 that haven't been added into main yet:

- [ ] - "Add an option to construct a soil driver for Fluxnet, so that we can run standalone canopy at Fluxnet sites. This is added as a kwarg for get_data_dates and prescribed_forcing_fluxnet, false by default." --> construct_prescribed_soil flag
- [X] - "Previously, atmospheric CO2 concentration was hardcoded to be a constant in fluxnet. Now, we read in the measured values at each site." --> c_co2 from TVI in ext/fluxnet_simulations/forcing.jl
- [X] - "Some fluxnet2015 sites have non-integer time offsets (e.g., in Australia), which can now be converted to a Dates.Period " --> hour_offset_to_period function in ext/fluxnet_simulations/data_processing.jl

PR #1509, opened 2025-10-23
fluxnet improvements: CO2, non-integer UTC offsets

juliasloan25:
## Purpose 
Addresses the first two points in #1476 
Originally authored by @ychnli 

## Content
- [x] read in CO2 from fluxnet data (read from TVI in ext/fluxnet_simulations/forcing.jl)
- [x] handle non-integer UTC offsets (hour_offset_to_period function in ext/fluxnet_simulations/data_processing.jl)

----

### DIFF
--- a/ext/fluxnet_simulations/data_processing.jl
+++ b/ext/fluxnet_simulations/data_processing.jl
@@ -1,11 +1,24 @@
 """
     var_missing(x; val = -9999)
 
-A function that checks if the value of `x` is equal to 
--9999, which is the value that Fluxnet uses for missing 
+A function that checks if the value of `x` is equal to
+-9999, which is the value that Fluxnet uses for missing
 data. Returns true if x == -9999.
 """
 var_missing(x; val = -9999) = x == val
+
+"""
+    hour_offset_to_period(hour_offset_from_UTC)
+
+Convert a numerical hour offset (which may be fractional) to a
+Dates.Period that can be added to DateTime objects.
+Supports both integer and fractional time zones (e.g., 5.5 hours).
+"""
+function hour_offset_to_period(hour_offset_from_UTC)
+    hours = floor(Int, hour_offset_from_UTC)
+    minutes = round(Int, (hour_offset_from_UTC - hours) * 60)
+    return Dates.Hour(hours) + Dates.Minute(minutes)
+end
 
 """
    mask_data(t, v; val = -9999)
@@ -48,10 +61,11 @@ should map between varname and column id, and the `time_in_seconds`
 should be the timestamp in seconds relative to the start_date of the
 simulation corresponding to each row of data.
 
-If you need to preprocess the data (e.g., unit conversion), you must pass 
+If you need to preprocess the data (e.g., unit conversion), you must pass
 a pointwise function preprocess_func(var) as a keyword argument.
 
-Note that this function handles missing data by removing it (assuming it is marked by missing by a given value equal to `val`), because the 
+Note that this function handles missing data by removing it (assuming it is
+marked by missing by a given value equal to `val`), because the
 TimeVaryingInput object is an interpolating object (in time).
 """
 function time_varying_input_from_data(
@@ -81,7 +95,7 @@ end
 
 Returns a TimeVaryingInput object which is computing using
 `preprocess_func` as a pointwise function of -in order - the columns in
-`data` specified by `varnames`. 
+`data` specified by `varnames`.
 
 For example, if you wish to compute specific humidity from temperature,
 pressure, and vpd, you would do:
@@ -98,7 +112,7 @@ should map between varname and column id, and the `time_in_seconds`
 should be the timestamp in seconds relative to the start_date of the
 simulation corresponding to each row of data.
 
-Note that this function handles missing data by removing it (assuming it is marked by missing by a given value equal to `val`), because the 
+Note that this function handles missing data by removing it (assuming it is marked by missing by a given value equal to `val`), because the
 TimeVaryingInput object is an interpolating object (in time).
 """
 function time_varying_input_from_data(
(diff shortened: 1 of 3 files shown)


ext/fluxnet_simulations/data_processing.jl
```diff
@@ -143,6 +157,32 @@ function get_data_at_start_date(
...
+Returns (data, columns, local_datetime, UTC_datetime).
+
+If `hour_offset_from_UTC` is nothing, the UTC datetime is not returned.
+"""
+function read_fluxnet_data(site_ID; hour_offset_from_UTC = nothing)
+    fluxnet_csv_path = ClimaLand.Artifacts.experiment_fluxnet_data_path(site_ID)
+    (data, columns) = readdlm(fluxnet_csv_path, ','; header = true)
+
+    # Convert timestamps
+    local_datetime = DateTime.(string.(Int.(data[:, 1])), "yyyymmddHHMM")
```
kmdeck: it might be safer to extract the column with the label TIMESTAMP instead of indexing with 1

we should also note that this is what the function does (since we sometimes want to return a different timestamp)

ext/fluxnet_simulations/data_processing.jl
```diff
@@ -143,6 +157,32 @@ function get_data_at_start_date(
...
+
+    # Convert to UTC if hour offset is given
+    if isnothing(hour_offset_from_UTC)
+        UTC_datetime = nothing
+    else
+        UTC_datetime =
+            local_datetime .+ hour_offset_to_period(hour_offset_from_UTC)
+    end
+
+    return (data, columns, local_datetime, UTC_datetime)
```
kmdeck: do we need both local and UTC date time? Shouldnt we do everything in UTC, always? Im not sure if the zenith angle computation will be correct if the simulation time is not in UTC.

ext/fluxnet_simulations/forcing.jl
```diff
@@ -436,8 +443,9 @@ function FluxnetSimulations.get_comparison_data(
     val = -9999,
     timestamp_end = true,
 )
-    fluxnet_csv_path = ClimaLand.Artifacts.experiment_fluxnet_data_path(site_ID)
-    (data, columns) = readdlm(fluxnet_csv_path, ','; header = true)
+    # Read data and process timestamps
+    (data, columns, local_datetime, UTC_datetime) =
+        read_fluxnet_data(site_ID; hour_offset_from_UTC)
```
kmdeck: does the same thing happen here? we overwrite local and UTC datetime later on (to use the TIMESTAMP_END)

ext/fluxnet_simulations/data_processing.jl
```diff
@@ -143,6 +157,32 @@ function get_data_at_start_date(
...
 
+"""
+    read_fluxnet_data(site_ID; hour_offset_from_UTC = nothing)
+
+Reads Fluxnet CSV data and processes timestamps, converting local time to UTC.
+Returns (data, columns, local_datetime, UTC_datetime).
+
+If `hour_offset_from_UTC` is nothing, the UTC datetime is not returned.
+"""
+function read_fluxnet_data(site_ID; hour_offset_from_UTC = nothing)
```
kmdeck: It kind of seems like this function does two things in one function (reads the data, and returns specific columns and transformations of a column also), but often we only need it to do one of those things (how it used later on, where we overwrite the time vectors).

Maybe it should be split into two:
(data, columns) = read_fluxnet_data(site_ID)
UTC_datetimes = get_UTC_datetimes(data, columns; timestamp = "start")
UTC_datetimes = get_UTC_datetimes(data, columns; timestamp = "end")
UTC_datetimes = get_UTC_datetimes(data, columns; timestamp = "average")

and local datetimes shouldnt be returned?

ext/fluxnet_simulations/data_processing.jl
```diff
@@ -1,12 +1,25 @@
...
 var_missing(x; val = -9999) = x == val
 
+"""
+    hour_offset_to_period(hour_offset_from_UTC)
+
+Convert a hour offset (which may be fractional) to a Dates.Period
+that can be added to DateTime objects.
+Supports both integer and half-hour time zones (e.g., 5.5 hours).
+"""
+function hour_offset_to_period(hour_offset_from_UTC)
```
kmdeck: This should be used throughout whenever we use hour_offset_from_UTC. there are examples where we still use the `hour_offset_from_UTC` directly and dont convert the fractional part

ext/fluxnet_simulations/forcing.jl
```diff
@@ -50,13 +50,13 @@ function FluxnetSimulations.prescribed_forcing_fluxnet(
...
     gustiness = 1,
-    c_co2 = TimeVaryingInput((t) -> 4.2e-4),
 )
     earth_param_set = LP.LandParameters(toml_dict)
     thermo_params = LP.thermodynamic_parameters(earth_param_set)
 
-    fluxnet_csv_path = ClimaLand.Artifacts.experiment_fluxnet_data_path(site_ID)
-    (data, columns) = readdlm(fluxnet_csv_path, ','; header = true)
+    # Read data and process timestamps
+    (data, columns, local_datetime, UTC_datetime) =
```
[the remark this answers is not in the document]
kmdeck: (and we want to use the timestamp average, the value returned by the function is just TIMESTAMP)

ext/fluxnet_simulations/forcing.jl
```diff
@@ -269,13 +268,26 @@ A helper function to get the difference in time between observations;
...
 function FluxnetSimulations.get_data_dt(site_ID)
-    fluxnet_csv_path = ClimaLand.Artifacts.experiment_fluxnet_data_path(site_ID)
-    (data, columns) = readdlm(fluxnet_csv_path, ','; header = true)
-    local_datetime = DateTime.(string.(Int.(data[:, 1])), "yyyymmddHHMM")
-    # Convert to seconds
+    (data, columns) = read_fluxnet_data(site_ID)
+
+    varnames = ("TIMESTAMP_START",)
+    column_name_map =
+        get_column_name_map(varnames, columns; error_on_missing = false)
```
kmdeck: shouldnt this error on missing?

kmdeck (approved):
looks good, and it's cleaner now. Thanks!

ext/fluxnet_simulations/forcing.jl
juliasloan25: yes, good catch!

PR #1526, opened 2025-11-03
Improve diagnostics test speed

imreddyTeja:
The diagnostics tests can take a very long
time (8 min on gha). This commit makes two
changes to speed this test up.

1. Override some constructors inside the @testset

The diagnostics tests do not need a real jacobian, and they do not need to perform a real step. This commit adds constructors for the implicit and explicit tendency updates, cache update, boundary fluxes update, and jacobian compuation which do nothing. It also uses a fake
field matrix and solver.

2. Change define_diagnsotics! and default_diagnostics to only add valid diagnostics. Constructing a DiagnosticVariable is expensive, and this reduces the number of useless ones created.

The speedup I get from this PR is much larger when I run locally. It makes gha ci a few minutes faster, and the buildkite gpu tests ~8 min faster.

----

### DIFF
--- a/src/diagnostics/diagnostic.jl
+++ b/src/diagnostics/diagnostic.jl
@@ -200,8 +200,8 @@ include("land_compute_methods.jl")
 # Default diagnostics and higher level interfaces
 include("default_diagnostics.jl")
 
-# define_diagnostics.jl contains the list of all the diagnostics
-include("define_diagnostics.jl")
+# construct_diagnostics.jl contains the list of all the diagnostics
+include("construct_diagnostics.jl")
 
 if pkgversion(ClimaDiagnostics) < v"0.2.13"
     # Default diagnostic resolution given a Space (approximately one point per
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@ ClimaLand.jl Release Notes
 main
 -------
 - Switch to use the Vast filesystem on central
+- `define_diagnostics!` renamed to `construct_diagnostics`
+  PR[#1526](https://github.com/CliMA/ClimaLand.jl/pull/1526)
 
 v1.0.0
 ------
--- a/docs/src/diagnostics/developers_diagnostics.md
+++ b/docs/src/diagnostics/developers_diagnostics.md
@@ -10,10 +10,10 @@ We want to provide users with default options, but also the possibility to defin
 Internally, this is done by using the [`ClimaDiagnostics.jl`](https://github.com/CliMA/ClimaDiagnostics.jl) package, that provides the functionality to produce a
 [`ClimaLand.Diagnostics`](https://github.com/CliMA/ClimaLand.jl/tree/main/src/Diagnostics/Diagnostics.jl) module in the src/Diagnostics.jl folder. In this folder,
  - `Diagnostics.jl` defines the module,
- - `diagnostic.jl` defines `ALL_DIAGNOSTICS`, a Dict containing all diagnostics variables defined in `define_diagnostics.jl`, it also defines the function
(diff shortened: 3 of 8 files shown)


imreddyTeja:
> The speed up is great!
> 
> Conceptually I'm a little confused about how defining diagnostics should depend on requested diagnostics. Requested diagnostics seems like more of what you want to get out of a simulation, while the definitions should be independent of that?
> 
> E.g. here we request only two diagnostics as output:
> 
> https://github.com/CliMA/ClimaLand.jl/blob/8fa13339aa083adf4ad7247dc5ed659bd3a6508b/docs/src/tutorials/standalone/Canopy/default_canopy.jl#L93-L99
> 
> Is this doing something different?

Requested diagnostics is not a great name for it. When `default_diagnostics` is used, requested_diagnostics would always be equal to `output_vars`. I'll rename it to `output_vars` if that makes more sense.

I think the name of `define_diagnostics` is a bit misleading. The diagnostic compute methods are always defined. `define_diagnostics` creates diagnostic variables and puts them in the `ALL_DIAGNOSTICS` dict. I'm not sure why, but creating diagnostic variables seems to take a very long time (maybe because the type is parameterized by the compute function, which is a closure). The current setup creates diagnostic variables that are not relevant for the model (for example the sif diagnostic variable is created for the bucket). If a user doesn't want to use every possible diagnostic, the diagnostic variable for it does not need to be created. Maybe a better solution would be to create diagnostic variables for all possible diagnostics, rather than for every single diagnostic (done currently), or only for the output vars (what this pr does)

src/diagnostics/diagnostic.jl
```diff
@@ -49,24 +51,29 @@ Keyword arguments
...
 )
-    ALL_DIAGNOSTICS[short_name] = DiagnosticVariable(;
-        short_name,
-        long_name,
-        standard_name,
-        units,
-        comments,
-        compute!,
-    )
+    if isnothing(requested_diags) || short_name in requested_diags
```
juliasloan25: Instead of using `requested_diags == nothing`, we can use `requested_diags = get_possible_diagnostics(model)` to get the entire set of diagnostics available for a model. Then we can just have:
```suggestion
    if short_name in requested_diags
```
and this will have the same effect of not restricting the DiagnosticVariables that are constructed

src/diagnostics/define_diagnostics.jl
```diff
@@ -1,12 +1,14 @@
 """
-    define_diagnostics!(land_model)
+    define_diagnostics!(land_model; requested_diags = nothing)
 
 Calls `add_diagnostic_variable!` for all available variables specializing the
-compute function for `land_model`.
+compute function for `land_model`. If `requested_diags` is not `nothing`, diagnostics are
+only added if their `short_name` is in `requested_diags`.
```
juliasloan25:
```suggestion
compute function for `land_model`. Diagnostics are
only constructed if their `short_name` is in `requested_diags`.
```

kmdeck (approved):
Thanks for explaining more.

Renaming `define_diagnostics!` to `construct_diagnostics!` is fine - or even `construct_requested_diagnostics!`? and we can keep the name `requested_diags`, and I like Julia's idea of the default being get_possible_diagnostics(land_model).

I stumbled a bit seeing that `define_diagnostics!` takes the model as input, but then does things internally that apply to *all* models - like, as you wrote, define_diagnostics!(bucket_model) adds canopy diagnostics, etc. so i almost wonder if a better solution might be to have different methods for different models, but this is more work for the same result, so I think what you have here is good!

imreddyTeja:
superseded by #1460

PR #1534, opened 2025-11-06
Add ILAMB leaderboard to 20 years long run

ph-kev:
This PR adds a minimal set up for generating the ILAMB leaderboard at the end of every 20 year long run. For now, this is uploaded as an artifact, but this should be uploaded to GitHub pages.

The infrastructure for generating the ILAMB leaderboard can be found at `/net/sampo/data1/ilamb` on `clima`.

To manually run a longer run and generate the ILAMB leaderboard, you can pass `LONGER_RUN=` and `ILAMB_RUN=` as environment variables on Buildkite when starting a new build.

For each long run, the name that will appear in ILAMB is `$(CURR_DATE)_COMMIT_ID`. For example, it could look like `2025-11-05_8fa9e7d`.

### DIFF
--- a/.buildkite/longruns_gpu/pipeline.yml
+++ b/.buildkite/longruns_gpu/pipeline.yml
@@ -147,3 +147,42 @@ steps:
         env:
           CLIMACOMMS_DEVICE: "CUDA"
           LONGER_RUN: ""
+
+  - wait
+
+  # To check if the ILAMB leaderboard should be generated or not, we check for
+  # the existence of `LONGER_RUN` and `ILAMB_RUN`. If both environment variables
+  # exist, then this group step is ran and otherwise, this group step does not
+  # run. If you want to generate the ILAMB leaderboard for a twenty year long
+  # run, then pass both `LONGER_RUN=""` and `ILAMB_RUN=""` as environment
+  # variables on Buildkite when starting a new build.
+
+  - group: "ILAMB leaderboard"
+    if: build.env("LONGER_RUN") != null && build.env("ILAMB_RUN") != null
+    steps:
+    - label: "Set up ILAMB leaderboard"
+      command:
+        - julia --color=yes --project=.buildkite/ experiments/ilamb/tests/test_ilamb_setup.jl
+        - julia --color=yes --project=.buildkite/ experiments/misc/ilamb_setup.jl $$ILAMB_DIAG_DIR $$ILAMB_BUILDKITE_DIR $BUILDKITE_BUILD_NUMBER $BUILDKITE_COMMIT
+      env:
+        ILAMB_DIAG_DIR: "snowy_land_longrun_gpu/global_diagnostics/ILAMB_diagnostics"
+        ILAMB_BUILDKITE_DIR: "/net/sampo/data1/ilamb/buildkite"
+
+    - wait
+
+    - label: "Run ILAMB leaderboard"
+      command:
+        - export PATH="$${ILAMB_ROOT}/ilamb_env/bin:$PATH"
+        - ilamb-run --config $$ILAMB_ROOT/buildkite/buildkite_ilamb.cfg --model_root $$ILAMB_ROOT/buildkite/MODELS/ --regions global --build_dir $$ILAMB_ROOT/buildkite/_build
+        - tar czf $$ILAMB_ROOT/buildkite/ilamb_leaderboard.tar.gz -C $$ILAMB_ROOT/buildkite _build
+      env:
+        # This cannot be renamed as the ILAMB software checks for the existence
+        # of the `ILAMB_ROOT` environment variable.
+        ILAMB_ROOT: "/net/sampo/data1/ilamb"
+      artifact_paths:
+        - "/net/sampo/data1/ilamb/buildkite/ilamb_leaderboard.tar.gz"
+      agents:
+        - slurm_mem: 32GB
+        - slurm_time: "2:00:00"
+      concurrency: 1
+      concurrency_group: "ilamb-run"


experiments/misc/tests/test_ilamb_setup.jl
```diff
@@ -0,0 +1,83 @@
+include(joinpath(@__DIR__, "..", "ilamb_setup.jl"))
+
+using Test
+
```
ph-kev: I wasn't too sure if this should be included in `test`.

experiments/misc/ilamb_setup.jl
```diff
@@ -0,0 +1,146 @@
...
+
+    create_symlinks(ilamb_diagnostics_dir, model_dir)
+    validate_symlinks(MODELS_dir)
+end
+
+"""
+    create_model_name(commit_id)
+
+Create a model name from the date and commit id.
+"""
```
ph-kev: I wasn't sure if there is any other information that should be included in the name of the long run.

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -147,3 +147,33 @@ steps:
...
+  - wait
+
+  - group: "ILAMB leaderboard"
+    if: build.env("LONGER_RUN") != null && build.env("ILAMB_RUN") != null
+    steps:
+    - label: "Set up ILAMB leaderboard"
+      command:
+        - julia --color=yes --project=.buildkite/ experiments/misc/ilamb_setup.jl $$ILAMB_DIAG_DIR $$ILAMB_BUILDKITE_DIR $BUILDKITE_COMMIT
+      env:
+        ILAMB_DIAG_DIR: "/home/kphan2/worktree/ClimaLand.jl/ilamb-script/ilamb_files"
```
kmdeck: Is there any concern with this being tied to a user or branch?

experiments/misc/ilamb_setup.jl
kmdeck: What else do we have access to? I think this is sufficient, in any case!

experiments/misc/tests/test_ilamb_setup.jl
kmdeck: What if we also changed the weekly longrun call to make the ILAMB output? and then if it breaks we know something affected it in the last week

Maybe the weekly longrun call can also execute this test, first?

experiments/misc/ilamb_setup.jl
kmdeck: oh yes! that would be good to add

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -147,3 +147,33 @@ steps:
         env:
           CLIMACOMMS_DEVICE: "CUDA"
           LONGER_RUN: ""
+
+  - wait
+
+  - group: "ILAMB leaderboard"
+    if: build.env("LONGER_RUN") != null && build.env("ILAMB_RUN") != null
```
juliasloan25: Could you add some documentation somewhere explaining how/when to set each of these? I think LONGER_RUN is pretty self-explanatory, but it's still good to note the difference it makes. and I think ILAMB_RUN is a little less clear - does setting that only determine whether the leaderboard gets generated?

experiments/misc/ilamb_setup.jl
```diff
@@ -0,0 +1,146 @@
+import Dates
```
juliasloan25: maybe we can rename the folder `experiments/misc/` to `experiments/ilamb/` since everything in that folder is for ilamb

experiments/misc/ilamb_setup.jl
```diff
@@ -0,0 +1,146 @@
...
+        error("$ilamb_diagnostics_dir is not a directory")
+    isdir(ilamb_models_parent_dir) || error("$ilamb_root is not a directory")
+
+    # Make the MODELS directory
+    MODELS_dir = joinpath(ilamb_models_parent_dir, "MODELS")
+    if !isdir(MODELS_dir)
+        # This should not be called that often, because the MODELS directory
+        # should persist from run to run
+        mkdir(MODELS_dir)
+    end
```
juliasloan25:
```suggestion
    # This should not be called that often, because the MODELS directory
    # should persist from run to run
    isdir(MODELS_dir) || mkdir(MODELS_dir)
```

experiments/misc/ilamb_setup.jl
```diff
@@ -0,0 +1,146 @@
...
+"""
+function create_symlinks(ilamb_diagnostics_dir::String, model_dir::String)
+    nc_filepaths = String[]
+    # The directory should be flat, but we iterate recursively in case this
+    # change for whatever reason
+    for (root, _, files) in walkdir(ilamb_diagnostics_dir)
+        for file in files
+            if endswith(file, ".nc")
+                push!(nc_filepaths, joinpath(root, file))
+            end
```
juliasloan25:
```suggestion
            endswith(file, ".nc") && push!(nc_filepaths, joinpath(root, file))
```

experiments/ilamb/tests/test_ilamb_setup.jl
```diff
@@ -0,0 +1,83 @@
+include(joinpath(@__DIR__, "..", "ilamb_setup.jl"))
+
+using Test
+
```
juliasloan25: I think it's fine to put this either in the regular tests (as long as it's relatively short) or in the longrun pipeline

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -147,3 +147,44 @@ steps:
...
+      env:
+        ILAMB_DIAG_DIR: "snowy_land_longrun_gpu/global_diagnostics/ILAMB_diagnostics"
+        ILAMB_BUILDKITE_DIR: "/net/sampo/data1/ilamb/buildkite"
+
+    - wait
+
+    - label: "Run ILAMB leaderboard"
+      command:
+        - export PATH="$${ILAMB_ROOT}/ilamb_env/bin:$PATH"
+        - export ILAMB_ROOT=$$ILAMB_ROOT
```
juliasloan25: I believe this is redundant now

juliasloan25 (approved):

experiments/ilamb/tests/test_ilamb_setup.jl
juliasloan25: It would be good to have the test called from somewhere automatically before merging

PR #1537, opened 2025-11-07
minimize optimization of tests

juliasloan25:
## Purpose 
Use the minimal optimization flag `-O0` to reduce optimization of tests run in buildkite. This reduces precompilation time at the expense of increased runtime. This is only for the tests though so it should overall speed up the tests.

This got the GPU runtests to run in [32 minutes](https://buildkite.com/clima/climaland-ci/builds/9517#019a6071-e165-42a9-bf86-3bd4607b234c) as compared to [55 minutes on main](https://buildkite.com/clima/climaland-ci/builds/9516#019a6045-eea9-4346-9113-22c0553c7cb9). I expect there's some variability, but this seems like a big enough improvement to merge in. The documentation GHA also decreased from [50.5 minutes](https://github.com/CliMA/ClimaLand.jl/actions/runs/19178352163) to [43.5 minutes](https://github.com/CliMA/ClimaLand.jl/actions/runs/19178352163)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -185,7 +185,7 @@ steps:
   - group: "GPU: unit tests and global bucket"
     steps:
       - label: "GPU runtests"
-        command: "julia --color=yes --project=test test/runtests.jl"
+        command: "julia -O0 --color=yes --project=test test/runtests.jl"
         agents:
           slurm_ntasks: 1
           slurm_gpus: 1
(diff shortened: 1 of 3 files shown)


imreddyTeja (approved):
Wow, that is so much quicker.

.github/workflows/ci.yml
```diff
@@ -45,7 +45,11 @@ jobs:
             ${{ matrix.os }}-julia-${{ matrix.version }}-
             ${{ matrix.os }}-julia-
       - uses: julia-actions/julia-buildpkg@v1
+        with:
+          flags: -O0
       - uses: julia-actions/julia-runtest@v1
+        with:
+          flags: -O0
```
imreddyTeja: I'm not sure that setting the flag here does anything. See
https://github.com/julia-actions/julia-runtest/issues/139

It also looks like the GitHub actions tests are not sped up by this.

PR #1538, opened 2025-11-10
compute fluxes when using CoupledAtmosphere

juliasloan25:
## Purpose 
The integrated land has been unstable in coupled simulations, and tests point to the cause being our explicit treatment of turbulent fluxes. This PR changes turbulent fluxes to be computed in the land model's step with both prescribed and coupled atmospheres, rather than leaving this to the coupler in coupled simulations.

See https://github.com/CliMA/ClimaCoupler.jl/issues/1557 for more details.

## To-do
- [x] rename `coupler_compute_turbulent_fluxes!` to `turbulent_fluxes!`
- [x] update dispatch of `turbulent_fluxes!` so that it only does nothing for a bucket run with a coupled atmosphere

cc @kmdeck


kmdeck (commented):
I think this looks correct! My only thought might be to create a helper function
`h_atmos, u_atmos, atmos_ts, gustiness = unpack_atmos_state(atmos, Y, p)` or something, and then possibly we could collapse the two turbulent_flux methods into 1 for each model (except Bucket)? 

We could wait on this until we feel confident this is the timestepping approach we want to continue with, though

PR #1545, opened 2025-11-17
Make soil organic carbon and soil o2 prognostic

### LINKED ISSUE #1551: Prognostic SOC and O2
Make SOC and O2 prognostic in existing DAMM model

AlexisRenchon:
closes #1551 

docs https://clima.github.io/ClimaLand.jl/previews/PR1545/

biogeochem_timeseries_all_top1m

biogeochem_profiles_top1m

### DIFF
--- a/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
+++ b/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
@@ -31,6 +31,7 @@ export SoilCO2ModelParameters,
     MicrobeProduction,
     SoilCO2FluxBC,
     AtmosCO2StateBC,
+    AtmosO2StateBC,
     SoilCO2StateBC,
     AbstractSoilDriver,
     SoilDrivers
@@ -57,17 +58,21 @@ Base.@kwdef struct SoilCO2ModelParameters{FT <: AbstractFloat, PSE}
     kM_sx::FT
     "Michaelis constant for O2 (m3 m-3)"
     kM_o2::FT
-    "Volumetric fraction of O₂ in the soil air, dimensionless"
-    O2_a::FT
+    "Volumetric fraction of O₂ in atmospheric air (reference value for boundary condition), dimensionless"
+    O2_f_atm::FT
     "Diffusion coefficient of oxygen in air, dimensionless"
     D_oa::FT
     "Fraction of soil carbon that is considered soluble, dimensionless"
     p_sx::FT
+    "Molar mass of carbon (kg/mol)"
+    M_C::FT
+    "Molar mass of oxygen (kg/mol)"
+    M_O2::FT
     "Physical constants used Clima-wide"
     earth_param_set::PSE
 end
 
-## For interfacting with ClimaParams
+## For interfacing with ClimaParams
 
 
 """
@@ -86,9 +91,11 @@ function SoilCO2ModelParameters(
         :soilCO2_activation_energy => :Ea_sx,
         :michaelis_constant => :kM_sx,
         :O2_michaelis_constant => :kM_o2,
-        :O2_volume_fraction => :O2_a,
+        :O2_volume_fraction => :O2_f_atm,
         :oxygen_diffusion_coefficient => :D_oa,
         :soluble_soil_carbon_fraction => :p_sx,
+        :molar_mass_carbon => :M_C,
+        :molar_mass_oxygen => :M_O2,
     )
     parameters = CP.get_parameter_values(toml_dict, name_map, "Land")
     FT = CP.float_type(toml_dict)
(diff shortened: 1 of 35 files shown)


src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -307,7 +428,48 @@ function ClimaLand.source!(
...
+    #   θ_a * P * M_O2 / (R * T) * dO2_a/dt = -(M_O2/M_C) * Sm
+    #
+    # Step 3: Solve for dO2_a/dt:
+    #   dO2_a/dt = -(M_O2/M_C) * Sm / (θ_a * P * M_O2 / (R * T))
+    #            = -(M_O2/M_C) * (R * T) / (θ_a * P * M_O2) * Sm
+    #            = -(R * T) / (M_C * θ_a * P) * Sm
+    #
+    # This is the physically correct conversion from C respiration to O2_a consumption.
+
+    M_C = eltype(p.soilco2.Sm)(12.0e-3)   # kg/mol (molar mass of carbon)
```
kmdeck: we need to obtain these from parameter structs instead of hardcoding

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -141,85 +148,137 @@ SoilCO2Model{FT}(
...
-    boundary_conditions::BC = (
-        top = AtmosCO2StateBC(),
-        bottom = SoilCO2FluxBC((p, t) -> 0.0), # no flux
-    ),
+    boundary_conditions = nothing,
     sources::Tuple = (MicrobeProduction{FT}(),),
-) where {FT, BC, DT}
+) where {FT, DT}
+    # Set default boundary conditions if not provided
+    if boundary_conditions === nothing
```
kmdeck: What if we made this the default, instead of `nothing`? I think it achieves the same purpose
```
parameters::SoilCO2ModelParameters{FT} = SoilCO2ModelParameters(toml_dict),
boundary_conditions = (
            top = (
                co2 = AtmosCO2StateBC(),
                o2 = AtmosO2StateBC(
                    parameters.earth_param_set,
                    parameters.M_O2,
                    parameters.O2_a,
                ),
            ),
            bottom = (
                co2 = SoilCO2FluxBC((p, t) -> 0.0),
                o2 = SoilCO2FluxBC((p, t) -> 0.0),
            ),
        )
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -141,85 +148,137 @@ SoilCO2Model{FT}(
...
-ClimaLand.prognostic_domain_names(::SoilCO2Model) = (:subsurface,)
+ClimaLand.prognostic_vars(::SoilCO2Model) = (:C, :O2_a, :SOC)
+ClimaLand.prognostic_types(::SoilCO2Model{FT}) where {FT} = (FT, FT, FT)
+ClimaLand.prognostic_domain_names(::SoilCO2Model) =
+    (:subsurface, :subsurface, :subsurface)
 
 ClimaLand.auxiliary_vars(model::SoilCO2Model) = (
     :D,
+    :D_o2,
+    :O2,
```
kmdeck: What is the difference between O2_a, O2, O2_avail?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -245,26 +324,77 @@ end
...
+        divf2c_O2 = Operators.DivergenceF2C(
+            top = Operators.SetValue(p.soilco2.top_bc_o2_wvec),
+            bottom = Operators.SetValue(p.soilco2.bottom_bc_o2_wvec),
+        ) # O2 diffusion with atmospheric O2 (0.21) at top boundary
+
+        # CO2 diffusion
         @. dY.soilco2.C =
             -divf2c_C(-interpc2f(p.soilco2.D) * gradc2f_C(Y.soilco2.C))
 
+        # O2 diffusion: Apply diffusion to O2 mass concentration in air (p.soilco2.O2)
```
kmdeck: I think we can remove these comments?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -245,26 +324,77 @@ end
...
+        # Convert to O2_a tendency: divide by (θ_a * P * M_O2 / (R * T))
+        # Protect against division by zero when θ_a = 0
+        conversion_factor =
+            @. max(p.soilco2.θ_a * P_sfc * M_O2 / (R * T_soil), eps(FT))
+        @. dY.soilco2.O2_a =
+            -divf2c_O2(
+                -interpc2f(p.soilco2.D_o2 * p.soilco2.θ_a) *
+                gradc2f_O2(p.soilco2.O2),
+            )
+        @. dY.soilco2.O2_a = dY.soilco2.O2_a / conversion_factor
```
kmdeck:
```suggestion
        @. dY.soilco2.O2_a = dY.soilco2.O2_a / max(p.soilco2.θ_a * P_sfc * M_O2 / (R * T_soil), eps(FT))
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -307,7 +444,48 @@ function ClimaLand.source!(
...
+    T_soil = p.soilco2.T  # soil temperature (K)
+    P_sfc = p.drivers.P   # atmospheric pressure (Pa)
+
+    # Get θ_a from auxiliary variables (already computed in update_aux!)
+    θ_a = p.soilco2.θ_a
+
+    # Apply stoichiometric O2 consumption with proper unit conversion
+    # Factor: (R * T) / (M_C * θ_a * P) converts from kg C/(m³ soil · s) to O2_a fraction/s
+    # Protect against division by zero when θ_a = 0
+    dY.soilco2.O2_a .-=
```
kmdeck:
```suggestion
    @. dY.soilco2.O2_a -= (R * T_soil) / max(M_C * θ_a * P_sfc, eps(FT)) *
           p.soilco2.Sm
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -435,20 +610,49 @@ This has been written so as to work with Differential Equations.jl.
 """
 function ClimaLand.make_update_aux(model::SoilCO2Model)
     function update_aux!(p, Y, t)
+        # Enforce positivity of prognostic variables to prevent unphysical values
+        # This acts as a safety net after the ODE integrator updates Y
+        FT = eltype(Y.soilco2.C)
+        Y.soilco2.C .= max.(Y.soilco2.C, FT(0))
```
kmdeck: Enforcing positivity should not happen here, I think..

Optimally, the PDEs themselves would enforce it. Even so, small numerical error can cause them to become negative, though. But even in this case, we should not adjust `Y` directly (this would break conservation. If some carbon flux has led to negative value for C, and we zero it out, we have just added carbon to the system for free).

What we do in the soil is let Y evolve but when the values are used in diagnostic functions, we clip them. Then we never encounter errors because of e.g. negative water content, but we also havent added water to the system. 

This works as long as the error is not self-reinforcing. What I mean by that is that the flux should go to zero if there is no carbon, no O2. If not, the solver might keep trying to remove carbon from a layer that has no carbon (making Y more and more negative)

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -435,20 +610,49 @@ This has been written so as to work with Differential Equations.jl.
...
+        @. p.soilco2.O2 =
+            o2_concentration(Y.soilco2.O2_a, T_soil, P_sfc, params)
+
+        # Compute O2 availability (dimensionless) for microbial kinetics (tortuosity-based)
+        (; D_oa) = params
+        @. p.soilco2.O2_avail =
+            o2_availability(Y.soilco2.O2_a, p.soilco2.θ_a, D_oa)
+
+        # Compute microbial source using O2 availability
+        p.soilco2.Sm .=
```
kmdeck:
```suggestion
        @. p.soilco2.Sm =microbe_source(T_soil, θ_l, Csom, p.soilco2.O2_avail, params)
```

and elsewhere...

this saves us a kernel launch (every .=, .*, etc is a kernel launch, the @. fuses them into a single one.

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -600,8 +804,83 @@ function ClimaLand.boundary_flux!(
...
+    T_soil_top = ClimaLand.Domains.top_center_to_surface(p.soilco2.T)
+    P_sfc = p.drivers.P
+
+    # Get parameters from BC struct (from ClimaParams and model parameters)
+    R = bc.R
+    M_O2 = bc.M_O2
+    O2_a_atm = bc.O2_a_atm
+
+    # O2 mass concentration in air at the boundary (kg O2/m³ air)
+    O2_bc = @. O2_a_atm * P_sfc * M_O2 / (R * T_soil_top)
```
kmdeck: These are all allocating, but we can fuse them below

kmdeck (commented):
We can also remove any reference to prescribed carbon from src/shared_utilities/drivers!

PR #1550, opened 2025-11-19
Keep a fixed number of models on the ILAMB leaderboard

### LINKED ISSUE #1544: Automatically clear old model runs in ILAMB leaderboard
For the ILAMB leaderboard, it would be nice to automatically clear out old model runs.

This might need more work than just deleting the directory containing the NetCDF files, because old results might still be cached by ILAMB.

ph-kev:
closes #1544 

This PR adds functionality to clear out old runs from the ILAMB leaderboard while still preserving the caching functionality when running the ILAMB leaderboard with the same models as before.

Since ENSEMBLE CMIP model is always include, there can be at most 4 runs produced by the ClimaLand simulation at once on the leaderboard.

### DIFF
--- a/.buildkite/longruns_gpu/pipeline.yml
+++ b/.buildkite/longruns_gpu/pipeline.yml
@@ -163,7 +163,7 @@ steps:
     - label: "Set up ILAMB leaderboard"
       command:
         - julia --color=yes --project=.buildkite/ experiments/ilamb/tests/test_ilamb_setup.jl
-        - julia --color=yes --project=.buildkite/ experiments/misc/ilamb_setup.jl $$ILAMB_DIAG_DIR $$ILAMB_BUILDKITE_DIR $BUILDKITE_BUILD_NUMBER $BUILDKITE_COMMIT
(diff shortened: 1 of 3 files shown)


experiments/ilamb/ilamb_setup.jl
```diff
@@ -14,12 +18,17 @@ This involves making the `MODELS` directory in `ilamb_models_parent_dir` if it
 does not exist, creating a model directory in the `MODELS` directory for this
 run, and creating symbolic links in the model directory pointing to NetCDF files
 in `ilamb_diagnostics_dir`.
+
+If there are more models present than `num_models_to_keep`, models are
```
kmdeck: Do you think it is possible to denote a model in ILAMB as "latest main" and not just buildkite number? I think latest main will always be our default model (which can evolve in time).

ph-kev: From our discussion offline, I think we can always keep the latest 20 years long run from main, ENSEMBLE-CMIP6, and whatever other models that are being tested that will eventually be cycled out. I think this can be done in a different PR though. I would like to see a successful ILAMB leaderboard artifact upload first.

kmdeck (approved):
Thanks, Kevin!

PR #1555, opened 2025-11-21
No flux BC; change default IC for soil

kmdeck:
## Purpose 
- Add saturated initial conditions as (new default) option
- Change to no flux boundary conditions for soil. However, our old saved IC wont be spun up for these boundary conditions for the soil...We may need to change the IC file or else add an option to change the BC more easily.
- make IC setup work even if a CoupledAtmosphere is used

## Content

----

### DIFF
--- a/src/simulations/initial_conditions.jl
+++ b/src/simulations/initial_conditions.jl
@@ -91,52 +91,6 @@ end
 
 Sets the soil initial conditions, stored in `Y.soil.ϑ_l`, `Y.soil.θ_i`,
 `Y.soil.ρe_int`, and defined on the `subsurface_space` corresponding to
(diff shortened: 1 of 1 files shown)


ph-kev (commented):
If the file path is going to change, can you update this line?

https://github.com/CliMA/ClimaLand.jl/blob/3fe2ded75ee4b2ec4bba914ead62c8670b99b87c/.buildkite/longruns_gpu/pipeline.yml#L168

kmdeck:
> If the file path is going to change, can you update this line?
> 
> https://github.com/CliMA/ClimaLand.jl/blob/3fe2ded75ee4b2ec4bba914ead62c8670b99b87c/.buildkite/longruns_gpu/pipeline.yml#L168

@AlexisRenchon can you change the path in your PR?

PR #1558, opened 2025-11-25
Serve longrun web dashboard on CI

AlexisRenchon:
This will serve the latest longrun every time they are run on CI at https://clima.westus3.cloudapp.azure.com/jsserve/land_longrun (currently doesn't exist)

Edit: I realize this PR is not good. Sending the data is OK. It's a command from buildkite. 

The other command (sudo) is a command to run on the Azure VM, not on buikdkite.

### DIFF
--- a/.buildkite/longruns_gpu/pipeline.yml
+++ b/.buildkite/longruns_gpu/pipeline.yml
@@ -109,7 +109,7 @@ steps:
         env:
           CLIMACOMMS_DEVICE: "CUDA"
           PERTURBED_RUN: ""
- 
+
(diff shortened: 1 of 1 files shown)


.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -137,6 +137,11 @@ steps:
           CLIMACOMMS_DEVICE: "CUDA"
           LONGER_RUN: ""
 
+      - label: "Send simulation output to Azure VM for visualization"
```
kmdeck: The path looks fine! But since this is running when LONGER_RUN = true, we should either make the longer run the p-model (my preference) or else this path should be to snowy_land_longrun_gpu

PR #1559, opened 2025-11-26
Implicit cache update

kmdeck:
## Purpose 
This PR looks at reducing the number of cache variable updates that we carry out each step. Currently, we update the cache like:
- update the entire auxiliary state for each component model
- update all fluxes for each component model (which depend on Y and the auxiliary state as well)

This occurs in the explicit tendency, the implicit tendency, and the jacobian update (prior to evaluating these functions, the cache is updated, and then these functions use only precomputed quantities). 

However, much of these variables and fluxes do not need to be updated during the implicit solve. 

This PR looks at one way of creating an implicit cache update which only updates a subset of the entire cache variables. The explicit tendency still updates the entire cache, but the implicit tendency and jacobian only update a smaller subset in the implicit cache update.

If this approach works (i.e. no loss of stability), we could do a few things:
- try and use the cache! and cache_imp! of ClimaTimesteppers rather than reinventing it here (is it doing anything different?)
- return to the conservation problem (where different fluxes were implicitly updated vs explicitly, making it hard to track conservation)
- rebenchmark without soil co2 and see where time is spent

## Results
- for soil alone (global), SYPD increases by 2x (to 200)
- for snowy land pmodel, SYPD increased to ~86 (about 2x also)

I ran the 20 year run to assess stability changes:
https://buildkite.com/clima/climaland-long-runs/builds/4446
It looks good.

## Content
- Define implicit cache = update implicit aux state + update implicit fluxes
- the default is to call the full update aux and update fluxes
- the default land model options is to call update implicit aux for each component, followed by update fluxes for each component.
- full cache = what we called update cache does not update drivers anymore
- update initial cache does so explicitly since those are updated by a callback

----


src/integrated/land.jl
```diff
@@ -389,6 +382,42 @@ function make_update_boundary_fluxes(
     return update_boundary_fluxes!
 end
 
+function make_update_implicit_cache(
```
imreddyTeja: Is this the same behavior as `make_update_implicit_cache` for `SoilCanopyModel`?

imreddyTeja (commented):
This looks good overall to me. I agree that we should use `cache!` and `cache_imp!` from CTS. I think that this would also help with making sure the fluxes are at the correct state when conservation checks are called.

src/integrated/land.jl
kmdeck: I think it should be, because all the implicitly treated quantities are in soil or snow. These functions are the same in the code. I will make them one method for Union{LandModel,SoilCanopyModel}

kmdeck: (leaving it as is for now)

kmdeck:
> Is the buildkite failure expected?

I have to fix that still.

src/ClimaLand.jl
```diff
@@ -208,6 +204,30 @@ function make_update_boundary_fluxes(land::AbstractLandModel)
     return update_boundary_fluxes!
 end
 
+function make_update_implicit_aux(land::AbstractLandModel)
```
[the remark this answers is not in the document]
kmdeck: I will open an issue

PR #1561, opened 2025-12-01
remove PrescribedTranspiration

kmdeck:
## Purpose 
Removes the option to prescribe transpiration, which was only used in a single test, and is not required. 

## Content
-Remove Prescribed vs Diagnostic Transpiration
- Transpiration is always computed now using turbulent fluxes computation
- fix test to reflect this
- fix bug in harmonic mean 

----

### DIFF
--- a/test/standalone/Vegetation/plant_hydraulics_test.jl
+++ b/test/standalone/Vegetation/plant_hydraulics_test.jl
@@ -14,7 +14,6 @@ import ClimaLand.Parameters as LP
 import Insolation
 using Dates
 
-
 for FT in (Float32, Float64)
     @testset "Plant hydraulics parameterizations, FT = $FT" begin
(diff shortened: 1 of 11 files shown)


test/standalone/Vegetation/plant_hydraulics_test.jl
```diff
@@ -280,11 +272,11 @@ for FT in (Float32, Float64)
         tendecy of the model also results in a steady state. This check is repeated using
         the plant hydraulics model directly.
         =======================#
-
+        ftol = FT == Float32 ? FT(1e-10) : FT(1e-16)
         soln = nlsolve(
```
ph-kev: Is there any particular reason why these particular tolerances were chosen?

ph-kev (approved):

PR #1562, opened 2025-12-01
Update ClimaDiagnostics to v0.3.0

ph-kev:
This PR updates ClimaDiagnostics to v0.3.0 and makes the necessary changes for the breaking release of ClimaDiagnostics. In particular, the dates of any time reductions is now at the start of the period rather than at the end. This means that shifting dates is not necessary for postprocessing.

TODO
- [x] Merge https://github.com/CliMA/ClimaDiagnostics.jl/pull/139
- [x] Update manifests


kmdeck (approved):
Looks good, but I think we should update the Manifest in docs also (both 1.10 and 1.11 for both)

thank you!

PR #1563, opened 2025-12-02
Update SurfaceFluxes compat (0.14.1)

akshaysridhar:
- Updates SurfaceFluxes compat to 0.14.1 (which removes neutral stability catches and changes the solver to use fixed point iterations, introduces interface support for iterative solution of aerodynamic roughnesses). `kwarg` separators are used for clarity. 
- Fixes an issue with the snow depth cap (currently, for instance, the `snow + soil` test sees 5m snow depth, whereas the cap is expected to be 10cm) 
- Fixes an issue with the `energy_from_q_l_and_swe` function in the integrated `snow + soil` test case. (Previously this resulted in unphysical temperature, pressure values in the surface state)


src/integrated/soil_snow_model.jl
```diff
@@ -228,7 +228,7 @@ function update_soil_snow_ground_heat_flux!(
     κ_soil = ClimaLand.Domains.top_center_to_surface(p.soil.κ)
 
     # Depths
-    Δz_snow = @. lazy(max(p.snow.z_snow, FT(0.1)))
+    Δz_snow = @. lazy(min(p.snow.z_snow, FT(0.1)))
```
juliasloan25: why does this change as part of the SF update?

juliasloan25: same for the change in test/

PR #1575, opened 2025-12-09
revert harmonic mean fix

kmdeck:
## Purpose 
By running the long runs on PRs 1563 (no NaNs), 1566 (no NaNs), 1530 (no NaNs), I determined that the NaNs are introduced in PR #1561. This should not be behavior changing, except we did change the formula for the harmonic mean to be the correct formula (previously it was off by a factor of 1/2).

Changing this back to the incorrect formula fixes the NaNs in the long runs. So, this PR reverts that change, and I opened an issue to debug this. https://github.com/CliMA/ClimaLand.jl/issues/1576

The test had to update to use the `harmonic_mean` function (as it is used in src) when computing `fa` in steady state. Previously it was not using the harmonic_mean function, but the harmonic mean was the same as each of the area indices since they were all the same. Since the harmonic_mean function has an error, we need to match that error in the test by using the function

## Content

----

### DIFF
--- a/src/standalone/Vegetation/PlantHydraulics.jl
+++ b/src/standalone/Vegetation/PlantHydraulics.jl
@@ -251,11 +251,10 @@ end
 """
     harmonic_mean(x::FT,y::FT) where {FT}
 
-Computes the harmonic mean of x >=0 and y >=0; returns zero if both
-x and y are zero.
-
+Computes the harmonic mean of x >=0 and y >=0; returns zero if either
+x or y are zero.
 """
-harmonic_mean(x::FT, y::FT) where {FT} = 2 * x * y / max(x + y, eps(FT))
+harmonic_mean(x::FT, y::FT) where {FT} = x * y / max(x + y, eps(FT))
 
 """
     water_flux(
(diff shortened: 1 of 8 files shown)


src/standalone/Vegetation/PlantHydraulics.jl
```diff
@@ -251,11 +251,10 @@ end
 """
     harmonic_mean(x::FT,y::FT) where {FT}
 
-Computes the harmonic mean of x >=0 and y >=0; returns zero if both
+Computes the harmonic mean of x >=0 and y >=0; returns zero if either
 x and y are zero.
```
juliasloan25:
```suggestion
x or y are zero.
```

kmdeck:
> It looks like after PR #1563 was merged, we merged:
> 
>     * #1562
> 
>     * #1566 (only changing a buildkite path)
> 
>     * #1530
> 
>     * #1561
> 
>     * #1545
> 
> 
> Are you sure that #1561 is the PR that introduced the NaNs, not one of the other PRs merged after 1563? If so this looks good to me.
> 
> It would be good to also tag a bugfix release (1.2.1) on this branch, so the coupler doesn't start using a version of ClimaLand with NaNs

I think so - I tried starting the long runs from the 1566 and 1530 commits, and they were OK. I then tried from the 1561 commit, saw NaNs and then tried without the factor of 2, and then it was OK. Also, in this PR, running the long runs, I dont see NaNs now.

and good call! I'll release 1.2.1

trontrytel (approved):
Thank you!

PR #1579, opened 2025-12-10
Tr/nn gpu

imreddyTeja:
## Purpose 

Make snow depth nn model gpu compatible. 

The cpu and gpu plots look different, especially for phase change flux and water runoff. See #1590

## Content
- Uses Adapt to make nn static. Fix type allocations in nn to make gpu compatible.
- Model can no longer be trained in ClimaLand because weights are static. This could be supported if needed.
- Model can no longer use matrix inputs for batch evaluation. 

----

### DIFF
--- a/test/standalone/Snow/tool_tests.jl
+++ b/test/standalone/Snow/tool_tests.jl
@@ -217,15 +217,15 @@ if !isnothing(DataToolsExt)
         z_idx = 1
         swe_idx = 2
         p_idx = 7
-        model = ModelTools.make_model(nfeatures, n, z_idx, p_idx)
+        model = ModelTools.make_model(Float32, nfeatures, n, z_idx, p_idx)
         ps = ModelTools.get_model_ps(model)
         for item in ps
             item[:] .= Float32(1.0)
         end
 
-        test_input = Matrix{Float32}(ones(nfeatures, 8))
+        test_input = Matrix{Float32}(ones(nfeatures, 1))
         @test model(test_input)[1] == 1968
-        @test size(model(test_input)) == (1, 8)
+        @test size(model(test_input)) == (1,)
         @test sum(length, ps) == nfeatures * (n * (2 * nfeatures + 1) + 2) + 1
         ModelTools.setoutscale!(model, 0.5)
         @test model[:final_scale].weight[3, 3] == 0.5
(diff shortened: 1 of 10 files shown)


test/standalone/Snow/tool_tests.jl
```diff
@@ -395,67 +368,56 @@ if !isnothing(DataToolsExt)
         Y.snow.S .= FT(0.1)
         # The snow module is broken when using CUDA because the density model is stored in
```
a-charbon: Are we at a state to get rid of the comments at 369-370 with these fixes?

experiments/standalone/Snow/snowmip_simulation_gpu.jl
```diff
@@ -0,0 +1,452 @@
...
+    # @error("Please provide a site name as command line argument")
+    SITE_NAME = "cdp"
+else
+    SITE_NAME = ARGS[1]
+end
+climaland_dir = pkgdir(ClimaLand)
+
+FT = Float32
+context = ClimaComms.context()
+ClimaComms.init(context)
```
kmdeck: maybe we could grab the device in the original snowmip simulation script, and if device == cpu make the plots, and if GPU, dont make them. Is it worth checking that the output is the same between the two runs?

test/standalone/Snow/tool_tests.jl
```diff
@@ -217,15 +217,15 @@ if !isnothing(DataToolsExt)
...
         p_idx = 7
-        model = ModelTools.make_model(nfeatures, n, z_idx, p_idx)
+        model = ModelTools.make_model(Float32, nfeatures, n, z_idx, p_idx)
         ps = ModelTools.get_model_ps(model)
         for item in ps
             item[:] .= Float32(1.0)
         end
 
-        test_input = Matrix{Float32}(ones(nfeatures, 8))
+        test_input = Matrix{Float32}(ones(nfeatures, 1))
```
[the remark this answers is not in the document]
imreddyTeja: This is because after the changes for gpu support, "batch" evaluation (using a matrix as an input) does not work. It could be added back, but it makes type stability easier to only support vectors. You can see this limitation in the layers that explicitly index into vectors and return svectors.

PR #1580, opened 2025-12-10
Tr/new integrated

imreddyTeja:
## Purpose 
Add support for LandModel without a soilco2 model.

## To-do
- Add documentation explaining behavior of nothing components
- Add unit tests

## Content
Allow the soilco2 model to be `nothing` in the `LandModel`

----

### DIFF
--- a/experiments/long_runs/snowy_land_pmodel.jl
+++ b/experiments/long_runs/snowy_land_pmodel.jl
@@ -89,6 +89,7 @@ function setup_model(
 
     ground = ClimaLand.PrognosticGroundConditions{FT}()
     canopy_forcing = (; atmos, radiation, ground)
+    prognostic_land_components = (:canopy, :snow, :soil, :soilco2)
 
     # Construct the P model manually since it is not a default
     photosynthesis = PModel{FT}(domain, toml_dict)
@@ -102,7 +103,7 @@ function setup_model(
         canopy_forcing,
         LAI,
         toml_dict;
-        prognostic_land_components = (:canopy, :snow, :soil, :soilco2),
+        prognostic_land_components,
         photosynthesis,
         conductance,
         soil_moisture_stress,
@@ -120,13 +121,22 @@ function setup_model(
         forcing,
         toml_dict,
         Δt;
-        prognostic_land_components = (:canopy, :snow, :soil, :soilco2),
+        prognostic_land_components,
         α_snow,
         scf,
     )
 
     # Construct the land model with all default components except for snow
-    land = LandModel{FT}(forcing, LAI, toml_dict, domain, Δt; snow, canopy)
+    land = LandModel{FT}(
+        forcing,
+        LAI,
+        toml_dict,
+        domain,
+        Δt;
+        prognostic_land_components,
+        snow,
+        canopy,
+    )
     return land
 end
 
--- a/src/diagnostics/land_compute_methods.jl
+++ b/src/diagnostics/land_compute_methods.jl
@@ -753,12 +753,23 @@ function compute_total_respiration!(
     if isnothing(out)
         out = zeros(land_model.canopy.domain.space.surface) # Allocates
         fill!(field_values(out), NaN) # fill with NaNs, even over the ocean
-        @. out =
-            p.soilco2.top_bc * FT(83.26) + p.canopy.autotrophic_respiration.Ra # [3.664 kg CO2/ kg C] x [10^3 g CO2/ kg CO2] x [1 mol CO2/44.009 g CO2] = 83.26 mol CO2/kg C
+        if isnothing(land_model.soilco2)
+            @. out = p.canopy.autotrophic_respiration.Ra
+        else
+            @. out =
+                p.soilco2.top_bc * FT(83.26) +
+                p.canopy.autotrophic_respiration.Ra # [3.664 kg CO2/ kg C] x [10^3 g CO2/ kg CO2] x [1 mol CO2/44.009 g CO2] = 83.26 mol CO2/kg C
+        end
         return out
     else
-        out .=
-            p.soilco2.top_bc .* FT(83.26) .+ p.canopy.autotrophic_respiration.Ra
+        if isnothing(land_model.soilco2)
+            @. out = p.canopy.autotrophic_respiration.Ra
+
+        else
+            out .=
+                p.soilco2.top_bc .* FT(83.26) .+
+                p.canopy.autotrophic_respiration.Ra
+        end
     end
 end
 @diagnostic_compute "total_respiration" CanopyModel p.canopy.autotrophic_respiration.Ra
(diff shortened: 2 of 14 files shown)


src/diagnostics/land_compute_methods.jl
```diff
@@ -753,12 +753,23 @@ function compute_total_respiration!(
     if isnothing(out)
         out = zeros(land_model.canopy.domain.space.surface) # Allocates
         fill!(field_values(out), NaN) # fill with NaNs, even over the ocean
-        @. out =
-            p.soilco2.top_bc * FT(83.26) + p.canopy.autotrophic_respiration.Ra # [3.664 kg CO2/ kg C] x [10^3 g CO2/ kg CO2] x [1 mol CO2/44.009 g CO2] = 83.26 mol CO2/kg C
+        if land_model.soilco2 isa ClimaLand.NothingModel
+            @. out = p.canopy.autotrophic_respiration.Ra
```
imreddyTeja: Is this correct if there is no soilco2 model?

test/shared_utilities/variable_types.jl
```diff
@@ -60,6 +60,53 @@ FT = Float32
...
+@testset "Nothing model, FT = $FT" begin
+    nm = ClimaLand.NothingModel{FT}()
+    @test ClimaLand.prognostic_vars(nm) == ()
+    @test ClimaLand.prognostic_types(nm) == ()
+    @test ClimaLand.prognostic_domain_names(nm) == ()
+    @test ClimaLand.auxiliary_vars(nm) == ()
+    @test ClimaLand.auxiliary_types(nm) == ()
+    @test ClimaLand.auxiliary_domain_names(nm) == ()
+    # ClimaLand.name(::DefaultModel) = :default
+    # dm = DefaultModel{FT}()
```
imreddyTeja: remove this

juliasloan25:
Instead of using `NothingModel`, could we just have `soilco2 = nothing` to run without that model? We can have a `function has_soil_model(land) = !isnothing(land.soilco2)` and use this e.g. to build the right `prognostic_land_components`

kmdeck:
> Instead of using `NothingModel`, could we just have `soilco2 = nothing` to run without that model? We can have a `function has_soil_model(land) = !isnothing(land.soilco2)` and use this e.g. to build the right `prognostic_land_components`

I think this could work, because we do things like loop over the output of this function  `prognostic_land_components(model)` which will call things like `prognostic_vars(model.component)`, etc. so as long as :soilco2 is not listed there, it should be OK. But we have some things hard coded (we call update_soilco2_bf! = make_update_boundary_fluxes(model.soilco2), which expects model.soilco2 to be a type of AbstractModel, and then call update_soilco2_bf! later on). But maybe those hard coded things can be generalized.

kmdeck:
FWIW, I had originally though the NothingModel might be a type of SoilCO2 model, not a general AbstractModel, but I like this too, because it maybe will help us generalize other AbstractLandModels into a single type. We might still need special methods for different combinations of type of component models, but can dispatch off of the types of the component models, rather than the type of the land model

src/integrated/land.jl
```diff
@@ -98,16 +102,17 @@ struct LandModel{
...
-        @assert soilco2.drivers.met isa PrognosticMet
+        isnothing(soilco2) || @assert soilco2.drivers.met isa PrognosticMet
         comparison = PrognosticMet(soil.parameters)
         # check_land_equality allocates, and should only be used in initialization
-        for property in propertynames(soilco2.drivers.met)
-            check_land_equality(
-                getproperty(soilco2.drivers.met, property),
-                getproperty(comparison, property),
-            )
+        if !isnothing(soilco2)
```
kmdeck: we could group all the checks into one if ~isnothing() block

src/integrated/land.jl
```diff
@@ -572,7 +595,10 @@ the same method as the soil+snow integrated model.
 function Soil.compute_liquid_influx(
     p,
     model,
-    prognostic_land_components::Val{(:canopy, :snow, :soil, :soilco2)},
+    prognostic_land_components::Union{
+        Val{(:canopy, :snow, :soil, :soilco2)},
+        Val{(:canopy, :snow, :soil)},
+    },
```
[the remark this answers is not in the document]
kmdeck: hmm good question. I think it is OK right now to be explicit.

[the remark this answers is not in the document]
kmdeck: Let's address it in the future! E.g. when/if we eventually remove SoilCanopyModel and SoilSnow model types and make use of the component fields being nothing instead.

src/integrated/land.jl
```diff
@@ -226,6 +239,8 @@ the components returned by `land_components!`.
 This needs to be fixed.
 """
 ClimaLand.land_components(land::LandModel) = (:soil, :snow, :soilco2, :canopy)
+ClimaLand.land_components(land::LandModel{FT, Nothing}) where {FT} =
```
kmdeck: maybe it would be better to have this be like
```
function ClimaLand.land_components(land::LandModel)
    land.soil_co2 isa Nothing ? (:soil, :snow, :canopy) : (:soil, :snow, :soilco2, :canopy)
end
```

like Julia suggested... is this slightly more readable? Up to you. Maybe it doesnt take advantage of the types in the same way

experiments/long_runs/snowy_land_pmodel.jl
```diff
@@ -120,13 +121,22 @@ function setup_model(
...
         Δt;
-        prognostic_land_components = (:canopy, :snow, :soil, :soilco2),
+        prognostic_land_components,
         α_snow,
         scf,
     )
 
     # Construct the land model with all default components except for snow
-    land = LandModel{FT}(forcing, LAI, toml_dict, domain, Δt; snow, canopy)
+    land = LandModel{FT}(
```
kmdeck: similar changes for LandModel run regionally. Do we need to update any of the benchmark scripts, calibration script also? 

I think it is OK for those all to be the "default" without soilco2, if that amounts to less changes, but keeping the current behavior in snowy_land and snowy land_pmodel is good for now

test/integrated/full_land.jl
```diff
@@ -78,6 +78,56 @@ for FT in (Float32, Float64)
...
+        )
+        soilco2 = nothing
+        model = LandModel{FT}(canopy, snow, soil, soilco2)
+
+        # The constructor has many asserts that check the model
+        # components, so we don't need to check them again here.
+        @test model.soil == soil
+        @test isnothing(model.soilco2)
+        @test model.canopy == canopy
+        @test model.snow == snow
```
kmdeck: we could also do `Y, p, cds = initialize(model)` and make sure Y and p dont have soilco2 as a key? and check `land_components(model)`

experiments/long_runs/snowy_land_pmodel.jl
kmdeck: For calibration though I think we can use no soilco2 - it doesnt affect what we are calibrating right now, and slows it down!

juliasloan25 (approved):
Thanks Teja!

PR #1584, opened 2025-12-15
Expand IC set

kmdeck:
## Purpose 
This PR does a lot of small changes (many of which came up when calibrating the model with the new IC, with the global box domain, etc):
- adds the saturated land IC set. This differs from our current IC set in a number of ways: it was generated with zero flux at the bottom of the domain, the soil depth is 15 instead of 50 m, and it containts the leaf water potential.
- Users can change which IC set they want by passing in the correct path. By creating the `enforce_constraints` argument, the current behavior is preserved if they pass the old path, set enforce_constraints to true, change the soil BC (now exposed via kwarg), and change the domain depth to 50 m.
- Fixed a bug in the HybridBox where the Plane created did not receive the correct context.
- make it so we do not assume atmos.T is a PrescribedAtmosphere when setting IC
- saturated IC are now the default in Simulations
- Fixes a bug with the calibration scripts to make it work with spherical domains (two element nelements Tuple) and boxes (three element nelements Tuple)
- Makes the default land model in calibration the pmodel

I'm not sure we really need to preserve all of the old IC machinery, but for now it makes sense, I think, in case we want to switch between them

## To-do
Run long runs before merging

## Content

----

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -7,13 +7,55 @@ import ClimaUtilities.ClimaArtifacts: @clima_artifact
 import LazyArtifacts
 
 """
+     saturated_land_ic_path(; context = nothing)
+
+Returns the path to the spun-up land initial conditions created with zero
+flux at the bottom of the soil domain, which extends 15 m in vertical 
+extent; these boundary conditions allow for a water table to form.
+
+This file contains the following variables:
+- soil ϑ_l
+- soil θ_i
+- soil ρe_int
+- leaf water potential
+- snow water equivalent
+
+Spun-up initial conditions similar to these can be created by 
+running the snowy_land_pmodel.jl file for 20 years (LONGER_RUN set),
+using the default global_box_domain, with a soil model with no flux
+at the bottom boundary,
+and then running the ClimaArtifacts script 
+`saturated_soil_ic/create_artifacts.jl` with a command line argument
+pointing to the correct path, as documented in that file.
+"""
+function saturated_land_ic_path(; context = nothing)
+    dir = @clima_artifact("saturated_land_ic", context)
+    return joinpath(dir, "saturated_land_ic.nc")
+end
+
+"""
     soil_ic_2008_50m_path(; context)
 
 Return the path to the file that contains the spun-up soil and snow initial
-conditions for Jan 1, 2008.
+conditions for Jan 1, 2008. A free drainage boundary condition was used.
+The soil domain has a depth of 50m.
 
-The soil domain has a depth of 50m; we have ensured that surface properties
-and fluxes are spun-up, but the deep soil water may not be.
+Initial conditions corresponding to a no flux boundary condition, and for
+a shallower domain, can be obtained using `satured_land_ic_path`.
+
+This file contains the following variables:
+- soil ϑ_l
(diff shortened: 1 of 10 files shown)


src/Artifacts.jl
```diff
@@ -6,14 +6,27 @@ import ClimaUtilities.ClimaArtifacts: @clima_artifact
 
 import LazyArtifacts
 
+"""
+     saturated_land_ic_path(; context = nothing)
+
+Returns the path to the spun-up land initial conditions created with zero
+flux at the bottom of a 15m soil column; these boundary conditions allow
+for a water table to form.
```
juliasloan25: I wonder if it would be possible/helpful to include in the repo a setup that is compatible with running from or generating these ICs

test/integrated/full_land.jl
```diff
@@ -210,6 +211,57 @@ land = LandModel{FT}(forcing, LAI, toml_dict, domain, Δt);
 @test ClimaComms.context(land) == ClimaComms.context()
 @test ClimaComms.device(land) == ClimaComms.device()
 @test ClimaLand.land_components(land) == (:soil, :snow, :soilco2, :canopy)
+@testset "Initial condition functions" begin
+    Y, p, cds = initialize(land)
+    t0 = ITime(0, Second(1), start_date)
+    ic_path = ClimaLand.Artifacts.soil_ic_2008_50m_path(; context)
```
juliasloan25: should we add a test for each IC file?

src/simulations/initial_conditions.jl
```diff
@@ -338,24 +437,38 @@ The returned function is a closure for `ic_path`. It could also be for `model`,
 many other ClimaLand functions are, but we wish to preserve the argument `model`
 in `set_ic!` for users who wish to define their own initial condition function,
 which may require parameters, etc, stored in `model`.
+
+If `enforce_constraints = true`, we ensure the soil water content is between porosity and the
+residual value, and that the temperature is bounded to be within the extrema of the air temperature
+at the surface.
+
+It is assumed that in CoupledAtmosphere simulations that `p.drivers.T` has 
+been updated already.
```
juliasloan25: With the current initialization in the coupler this won't be the case - we initialize all models with some ICs, then perform an initial exchange (which is where p.drivers gets filled)

src/simulations/initial_conditions.jl
```diff
@@ -338,24 +437,38 @@ The returned function is a closure for `ic_path`. It could also be for `model`,
...
-        evaluate!(p.drivers.T, atmos.T, t0)
-        T_bounds = extrema(p.drivers.T)
-
+        if atmos isa ClimaLand.PrescribedAtmosphere
+            evaluate!(p.drivers.T, atmos.T, t0)
+        end
+        if enforce_constraints
+            # get only the values over land
+            foo = Array(parent(p.drivers.T))[:]
+            T_bounds = extrema(foo[foo .> sqrt(eps(FT))])
```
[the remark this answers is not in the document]
juliasloan25: same for the 2 other functions in this file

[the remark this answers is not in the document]
kmdeck: good call!

PR #1593, opened 2026-01-09
Addition of Snow Surface Temperature Parameterization

a-charbon:
## Purpose 

This PR serves to add the Snow Surface Temperature parameterization to the Snow Module, to better constrain the surface energy balance. This includes the following:
- adding the RootSolvers library to ClimaLand
- including the surface temperature parameterization, and adding a residual surface flux variable to carry over additional energy to add to the bulk
- included unit tests
- adding SnowSurfaceTemperatureModel abstract type to allow user to pick between possible surface temperature parameterizations.

## To-do

- handle left comments

## Content

----
- [ X] I have read and checked the items on the review checklist.


src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -523,3 +564,260 @@ function update_density_prog!(
 )
     return nothing
 end
+
```
kmdeck: extra new lines here and elsewhere

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -523,3 +564,260 @@ function update_density_prog!(
...
+"""
+function flux_balance(
+    T_sfc_guess::FT,
+    T_bulk::FT,
+    z::FT,
+    κ::FT,
+    ρ_snow::FT,
+    α_snow::FT,
+    ϵ_snow::FT,
+    SW_d::FT,
```
kmdeck: Eventually this should be passed SW_n instead (because we compute SW_up later, and this introduces a place where the two computations could agree)

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -539,3 +555,263 @@ function update_density_prog!(
...
+
+    #get surf_temp values, even if they are > T_freeze:
+    p.snow.T_sfc .=
+        solve_for_surface_temp_at_a_point.(
+            p.snow.T,
+            p.snow.z_snow,
+            p.snow.κ,
+            p.snow.ρ_snow,
+            model.parameters.ϵ_snow,
+            p.snow.R_n, #The net SW_down
```
kmdeck: net SW, right?

src/standalone/Snow/Snow.jl
```diff
@@ -377,12 +440,14 @@ function SnowParameters{FT}(
     κ_ice,
     ΔS = FT(0.1),
     scf::SCFM,
+    surf_temp::STM = EquilibriumGradientTemperatureModel(FT(0.1)), #BulkSurfaceTemperatureModel{FT}(),
```
kmdeck: here also

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -554,3 +570,330 @@ function update_density_prog!(
...
+
+    sol = RootSolvers.find_zero(
+        T -> flux_balance_closure(T),
+        method,
+        CompactSolution(),
+        ResidualTolerance(tol),
+        N_iters,
+    )
+
+    #condition = (sol.root <= eps(FT)) || (abs(flux_balance_closure(sol.root)) > tol) #if using SolutionTolerance(tol)
```
kmdeck: can remove comments.

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -554,3 +570,330 @@ function update_density_prog!(
...
+    d_safe = max(surface_temp_scaling_length(κ, ρ, z, parameters), eps(FT))
+    return T_sfc_root > _T_freeze ? FT(-κ * (T_sfc_root - _T_freeze) / d_safe) :
+           FT(0)
+end
+
+"""
+    update_surf_temp!(model::SnowModel, surf_temp::EquilibriumGradientTemperatureModel, SW_net, LW_down, Y, p, t)
+
+Updates the surface temperature variable, storing residual energy flux in p.snow.surf_residual_flux to be added to the bulk.
+Note, this update function is not called for the integrated land model - the update to the surface temperature happens via a different
```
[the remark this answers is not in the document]
kmdeck: In other cases, e.g. snow_boundary_fluxes! we use this `prognostic_land_components::Val{(:snow,)},` to indicate standalone vs not.

Does it make sense to be consistent and use that as an argument here too?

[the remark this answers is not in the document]
kmdeck: Actually, it seems like this function is called as is in the integrated models?

src/standalone/Snow/Snow.jl
```diff
@@ -556,7 +626,7 @@ prognostic_domain_names(m::SnowModel) =
     (:surface, :surface, :surface, density_prog_names(m.parameters.density)...)
 
 """
-    density_prog_vars(::AbstractDensityModel)
+    density_prog_names(::AbstractDensityModel)
```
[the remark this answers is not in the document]
a-charbon: yes, some of the documentation names were inconsistent against the function they defined - I'm fixing the ones I find as I go, the function below it is `density_prog_names` but the doc was titled as `density_prog_vars`. Something similar is found on `density_prog_types`

src/standalone/Snow/boundary_fluxes.jl
```diff
@@ -104,6 +104,17 @@ function snow_boundary_fluxes!(
...
+        model.parameters.surf_temp,
+        SW_net,
+        p.drivers.LW_d,
+        Y,
+        p,
+        t,
+    )
+
     turbulent_fluxes!(p.snow.turbulent_fluxes, bc.atmos, model, Y, p, t)
     net_radiation!(p.snow.R_n, bc.radiation, model, Y, p, t)
```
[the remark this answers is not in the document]
a-charbon: Let's chat about this one - I am ambivalent and could be convinced either way

src/standalone/Snow/Snow.jl
```diff
@@ -249,6 +248,63 @@ function WuWuSnowCoverFractionModel(
...
+    N_iters::Int
+    "Allowable tolerance in the root-solve algorithm"
+    tol::FT
+end
+
+function EquilibriumGradientTemperatureModel(
+    tol::FT;
+    N_iters::Int = 8,
+) where {FT}
+    return EquilibriumGradientTemperatureModel{FT}(N_iters, tol)
```
juliasloan25: I find this constructor a little strange - it would be more straightforward to choose one ordering of N_iters/tol and use it consistently rather than having a constructor that swaps their order

a-charbon: good point - this point is now handled!

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -554,3 +570,330 @@ function update_density_prog!(
...
+
+    # This circumvents the scenario where for certain non-ideal combinations of initial conditions
+    # and snowpack state, the root-solver iterates to a T_sfc_guess <= eps(FT) Kelvin, which breaks SurfaceFluxes:
+    # (this gets the root-solver back on course, and lhf/shf are otherwise 0 in the near-0 Kelvin regime)
+    # Note - BrentsMethod/bounded iterative methods will not exhibit such an issue, but it takes noticeably more compute on average
+    # Note - Placing the conditional return here, instead of using safe_T_sfc_guess = max(T_sfc_guess, FT(1)), skips unnecessary
+    #        and costly computation of q_sfc and latent/sensible heat flux (the calculation is expensive enough that the branching
+    #        on GPU should be less costly than using `ifelse()`, which forces computation of both cases).
+    if T_sfc_guess <= eps(FT) #|| isinf(T_sfc_guess) || isnan(T_sfc_guess) #last two are needed for a manual root-finding implementation
+        return d * (SW_net + lw_net) + κ * (T_sfc_guess - T_bulk)
```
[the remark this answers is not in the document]
kmdeck:
```
# For some secant initial guesses, a negative T_sfc can be reached, but the algorithm may recover with the branch
below. Negative T_sfc cause domain errors in SurfaceFluxes, which would prevent any recovery.
```

PR #1594, opened 2026-01-13
Removes thermodynamic states in favor of functional TD calls

### LINKED ISSUE #1603: Use consistent thermodynamics in ClimaLand

There are several places where ClimaLand uses hardcoded empirical approximations rather than the consistent thermodynamics formulation provided by Thermodynamics.jl. 

For example, [`caopy_energy.jl`](https://github.com/CliMA/ClimaLand.jl/blob/d2b2b918f30b9894204e027fc720a635cd4dfb28/src/standalone/Vegetation/canopy_energy.jl#L220) use empirical derivative approximations for saturation specific humidity, where exact derivatives are available (and simpler). And FluxNet [`forcing.jl`](https://github.com/CliMA/ClimaLand.jl/blob/d2b2b918f30b9894204e027fc720a635cd4dfb28/ext/fluxnet_simulations/forcing.jl#L400) likewise uses hardcoded constants in equations where functions to compute what is needed are available in Thermodynamics.jl.

As a rule, we should avoid hardcoded physical constants that appear in more than one place. 

It's a relatively simple refactor to update this in ClimaLand.

tapios:
## Purpose 
* Removes thermodynamic states and PhasePartition objects in favor of the functional TD API
* Leaves state constructors where SurfaceFluxes is called. These should be updated after SF update. 

## Content
Closes #1603 and most of #1596 

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@ ClimaLand.jl Release Notes
 
 main
 ------
+- ![breaking change][badge-💥breaking] Remove thermal state from ClimaLand, use latest Thermodynamics package [#1594](https://github.com/CliMA/ClimaLand.jl/pull/1594)
 
 v1.4.0
 ------
--- a/src/shared_utilities/utils.jl
+++ b/src/shared_utilities/utils.jl
@@ -32,7 +32,7 @@ function partial_q_sat_partial_T(
         Thermodynamics.latent_heat_vapor(thermo_params, T),
         Thermodynamics.latent_heat_sublim(thermo_params, T),
     )
-    return Thermodynamics.∂q_vap_sat_∂T(thermo_params, nothing, T, q_sat, L)
+    return Thermodynamics.∂q_vap_sat_∂T_from_L(thermo_params, q_sat, L, T)
 end
 
 """


ext/neural_snow/NeuralSnow.jl
```diff
@@ -234,8 +234,12 @@ function update_density_prog!(
     dY.snow.Qrel_avg .=
         density.α .* (
             Thermodynamics.relative_humidity.(
-                model.boundary_conditions.atmos.thermo_params,
-                p.drivers.thermal_state,
+                Ref(model.boundary_conditions.atmos.thermo_params),
```
kmdeck: I believe we do not need the Ref here (it looks like it was running before without it)

ext/neural_snow/NeuralSnow.jl
```diff
@@ -234,8 +234,12 @@ function update_density_prog!(
...
+                p.drivers.T,
+                p.drivers.P,
+                p.drivers.q,
+                Ref(zero(eltype(p.drivers.q))),  # TODO: Remove after TD update
```
kmdeck: I think also here it might be ok without the Ref
```
julia> function f(x,y)
       return x+y
       end
f (generic function with 1 method)

julia> f.([1.0,2.0,3.0],zero(Float64))
3-element Vector{Float64}:
 1.0
 2.0
 3.0
```

docs/make.jl
```diff
@@ -56,7 +56,8 @@ tutorials_jl = flatten_to_array_of_strings(get_second(tutorials))
 println("Building literate tutorials...")
 tutorials_dir = joinpath(@__DIR__, "src", "tutorials")
 tutorials_jl = map(x -> joinpath(tutorials_dir, x), tutorials_jl)
-pmap(t -> generate_tutorial(tutorials_dir, t), tutorials_jl)
+# pmap(t -> generate_tutorial(tutorials_dir, t), tutorials_jl)
+map(t -> generate_tutorial(tutorials_dir, t), tutorials_jl)
```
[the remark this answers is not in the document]
kmdeck: Kevin had changed this because our docs take quite some time to build. For example, without parallel mapping, this takes 86 minutes to build. In the PR I just merged into main, it took 47 minutes to build.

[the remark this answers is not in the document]
tapios: Ok. I revert it. But I think the real issue is that the docbuild is too complicated.

test/bucket_restart/output_active
```diff
@@ -0,0 +1 @@
+output_0000
```
kmdeck: can remove this file

PR #1613, opened 2026-01-26
Recalibrate land model

kmdeck:
## Purpose 
Recalibrate the land model post:
- surface flux changes (canopy: improved SHF and LW biases by 5x, soil: increase SH bias but decrease LW bias)
- thermodynamics changes
- implicit cache update changes (halved biases in SHF and LW)
- fixed quantum yield for C4 photosynthesis, small change to C3. see Equations 4 and 5 here: https://iopscience.iop.org/article/10.1088/1748-9326/abc64e/pdf
- removal of clipping of ustar in g_land computation
- fix bug in Rd computation (Equation 2 for Rd temp scaling: https://nph.onlinelibrary.wiley.com/doi/pdf/10.1111/nph.19355)

## Content

----

### DIFF
--- a/experiments/calibration/PBS_calibration.pbs
+++ b/experiments/calibration/PBS_calibration.pbs
@@ -17,5 +17,5 @@ export CLIMACOMMS_DEVICE="CUDA"
 export CLIMACOMMS_CONTEXT="SINGLETON"
 
 julia --project=.buildkite -e 'using Pkg; Pkg.instantiate(;verbose=true)'
-#julia --project=.buildkite/ experiments/calibration/generate_observations.jl
+julia --project=.buildkite/ experiments/calibration/generate_observations.jl
 julia --project=.buildkite/ experiments/calibration/run_calibration.jl
--- a/experiments/calibration/run_calibration.jl
+++ b/experiments/calibration/run_calibration.jl
@@ -11,16 +11,14 @@ import JLD2
 include(joinpath(pkgdir(ClimaLand), "experiments/calibration/api.jl"))
 
 const CALIBRATE_CONFIG = CalibrateConfig(;
-    short_names = ["lwu", "shf", "lhf"],
+    short_names = ["lwu"],
     minibatch_size = 1,
-    n_iterations = 10,
-    sample_date_ranges = [
-        ("$(2000 + 2*i)-12-1", "$(2002 + 2*i)-9-1") for i in 0:9
-    ],
+    n_iterations = 1,
+    sample_date_ranges = [("2007-12-1", "2007-12-1")],
     extend = Dates.Month(3),
-    spinup = Dates.Month(3),
+    spinup = Dates.Month(0),
     nelements = (180, 360, 15),
-    output_dir = "/glade/derecho/scratch/kdeck/recalibrate_saturated_K_lw",
+    output_dir = "experiments/calibration/land_model",
     rng_seed = 42,
     obs_vec_filepath = "experiments/calibration/land_observation_vector.jld2",
     model_type = ClimaLand.LandModel,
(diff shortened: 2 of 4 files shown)


ph-kev (approved):

AlexisRenchon:
This is an AI review (it might not be helpful, but maybe?):

  - Rd temperature scaling now uses T^2 in Kelvin. The coefficients (aRd=0.1012, bRd=-0.0005) are the Heskel (2016) values that are typically fit with T in °C, so squaring Kelvin will strongly change the response (and the docstring still cites Heskel). If Eq. 2 uses °C, this is a bug; if it uses K, the coefficients likely need re‑fitting or conversion. [here](src/standalone/Vegetation/pmodel.jl:1458-1459), [and here](toml/default_parameters.toml:324-334)
  - quadratic_intrinsic_quantum_yield now returns 0 for T < 0 °C, introducing a discontinuity not mentioned in the docstring. Please confirm Eq. 4–5 actually specify this behavior; otherwise, keep the polynomial + clamp only. [here](src/standalone/Vegetation/pmodel.jl:1120-1126)
  - The updated quantum yield coefficients change ϕ noticeably: at 25 °C, C3 goes from ~0.060 to ~0.086 and C4 to ~0.050. That seems plausible, but it’s not a “small” change in magnitude. Also, the constant path (pmodel_ϕ0_c3/ϕ0_c4) remains Skillman (2008); if temperature‑dependent yield is intended by default, fine, but if not, the constants may now be inconsistent. [here](toml/default_parameters.toml:179-224)
  - Removing the max(u_star, 1) floor can make g_leaf and g_h both → 0 when u_star → 0, yielding g_land/g_h as 0/0 in several places. If gustiness or your surface‑flux scheme guarantees u_star > 0, consider documenting that or keep a small floor to prevent NaNs. [1](src/standalone/Vegetation/canopy_boundary_fluxes.jl:300-324), [2](src/standalone/Vegetation/canopy_boundary_fluxes.jl:375-382), [3](src/standalone/Vegetation/canopy_boundary_fluxes.jl:415-418), [4](src/standalone/Vegetation/canopy_boundary_fluxes.jl:454-455)
  - Roughness/displacement defaults changed by orders of magnitude (e.g., canopy_d_coeff 0.67 → 0.007). If these are intended global defaults, they will strongly affect canopy‑atmosphere exchange; please confirm or scope to calibration configs instead of default_parameters.toml. [here](toml/default_parameters.toml:470-498)

kmdeck:
>     * Rd temperature scaling now uses T^2 in Kelvin. The coefficients (aRd=0.1012, bRd=-0.0005) are the Heskel (2016) values that are typically fit with T in °C, so squaring Kelvin will strongly change the response (and the docstring still cites Heskel). If Eq. 2 uses °C, this is a bug; if it uses K, the coefficients likely need re‑fitting or conversion. [here](src/standalone/Vegetation/pmodel.jl:1458-1459), [and here](toml/default_parameters.toml:324-334)
Yes! I had made a mistake. I believe it is OK now if you want to take a look @AlexisRenchon 
>     * quadratic_intrinsic_quantum_yield now returns 0 for T < 0 °C, introducing a discontinuity not mentioned in the docstring. Please confirm Eq. 4–5 actually specify this behavior; otherwise, keep the polynomial + clamp only. [here](src/standalone/Vegetation/pmodel.jl:1120-1126)
I feel like this is OK?
>     * The updated quantum yield coefficients change ϕ noticeably: at 25 °C, C3 goes from ~0.060 to ~0.086 and C4 to ~0.050. That seems plausible, but it’s not a “small” change in magnitude. Also, the constant path (pmodel_ϕ0_c3/ϕ0_c4) remains Skillman (2008); if temperature‑dependent yield is intended by default, fine, but if not, the constants may now be inconsistent. [here](toml/default_parameters.toml:179-224)
The constant values need to be double checked but that is a separate issue
>     * Removing the max(u_star, 1) floor can make g_leaf and g_h both → 0 when u_star → 0, yielding g_land/g_h as 0/0 in several places. If gustiness or your surface‑flux scheme guarantees u_star > 0, consider documenting that or keep a small floor to prevent NaNs. [1](src/standalone/Vegetation/canopy_boundary_fluxes.jl:300-324), [2](src/standalone/Vegetation/canopy_boundary_fluxes.jl:375-382), [3](src/standalone/Vegetation/canopy_boundary_fluxes.jl:415-418), [4](src/standalone/Vegetation/canopy_boundary_fluxes.jl:454-455)
g_h is actually the drag coefficient times air speed, and gustiness will prevent this from being zero. This is handed in SurfaceFluxes, but we can add a small floor to be safe.
>     * Roughness/displacement defaults changed by orders of magnitude (e.g., canopy_d_coeff 0.67 → 0.007). If these are intended global defaults, they will strongly affect canopy‑atmosphere exchange; please confirm or scope to calibration configs instead of default_parameters.toml. [here](toml/default_parameters.toml:470-498)
The canopy height now varies in space, so these coefficients changing is not very surprising, as they multiple canopy height. I think it is important to recalibrate these as we improve the model to ensure they are physically realistic and not compensating for something else.

PR #1616, opened 2026-01-28
return volumetric liquid content instead of augmented liquid water

kmdeck:
## Purpose 
Our prognostic variable for soil water is the augmented liquid fraction. This can be larger than porosity, in which case the retention curve relationship specifies a corresponding (positive) pressure. This positive pressure would drive flow away from the saturated regions to bring the augmented liquid fraction back to porosity.

Whenever the true volumetric liquid fraction is required, we should use \theta_l (in `p`) since this is limited to be less than porosity. Above porosity, our prognostic variable reflects a pressure and shouldnt really be interpreted as a volume of water larger than the available pore space (except for when conserving mass). 

This PR changes the diagnostic we compare to data for surface soil moisture in the first 10cm to use theta_l, but we still define our "soil water content" diagnostic (depth resolved) to be the state Y.soil.\vartheta_l.

## Content

----

### DIFF
--- a/src/diagnostics/land_compute_methods.jl
+++ b/src/diagnostics/land_compute_methods.jl
@@ -951,7 +951,7 @@ end
     SoilCanopyModel,
     LandModel,
     EnergyHydrology,
-} p.soil.θ_l
+} Y.soil.ϑ_l
 @diagnostic_compute "soil_ice_content" Union{
     SoilCanopyModel,
     LandModel,


src/diagnostics/land_compute_methods.jl
```diff
@@ -947,19 +947,22 @@ end
...
 @diagnostic_compute "soc" Union{SoilCanopyModel, LandModel} Y.soilco2.SOC
-@diagnostic_compute "soil_water_content" Union{SoilCanopyModel, LandModel} Y.soil.ϑ_l
-# @diagnostic_compute "plant_water_content" Union{SoilCanopyModel, LandModel} Y.canopy.hydraulics.ϑ_l # return a Tuple
-@diagnostic_compute "soil_ice_content" Union{SoilCanopyModel, LandModel} Y.soil.θ_i
-@diagnostic_compute "soil_internal_energy" Union{SoilCanopyModel, LandModel} Y.soil.ρe_int
+@diagnostic_compute "soil_water_content" Union{
+    SoilCanopyModel,
+    LandModel,
+    EnergyHydrology,
+} Y.soil.ϑ_l
```
kmdeck: I think it is useful for debugging to still output this Y, but we can discuss

AlexisRenchon (approved):

PR #1625, opened 2026-02-03
Add prognostic optimal LAI model (Zhou et al. 2025)

AlexisRenchon:
This PR implements ZhouOptimalLAIModel, a new concrete type of AbstractBiomassModel,  that computes LAI dynamically from energy and water constraints, following Zhou et al. (2025). 

- [Zhou et al., 2025 manuscript](https://onlinelibrary.wiley.com/doi/pdf/10.1111/gcb.70125) 
- [Documentation](https://clima.github.io/ClimaLand.jl/previews/PR1625/standalone/pages/vegetation/canopy_structure/optimal_lai/)
- [Dashboard (select lai)](https://dashboard.cupoftea.earth/land_longrun)

Issues for future PR:
#1634 - Storing time-integrated variables in Y 
#1636 - Dynamic C3/C4 vegetation distribution
#1637 - Calibrate GPP and LAI
#1638 - Refactor biomass model to decouple LAI from other fields 
#1639 - Use canopy temperature instead of air temperature for potential GPP (maybe it doesn't make sense though - how do we know canopy temperature if there's no canopy...)

NOTE: As discussed with Kat, this no longer has C3 C4 partitioning (will be a separate PR)

NOTE (Alexis): I believe this PR is ready to be merged. Further progress will be made in future PRs (see issues). The model works well - except LAI is overestimated in locations where GPP is overestimated - but this is related to GPP, not the LAI model. See [ILAMB GPP](https://dashboard.cupoftea.earth/ilamb/EcosystemandCarbonCycle/GrossPrimaryProductivity/FLUXCOM/FLUXCOM.html).

### DIFF
--- a/experiments/integrated/fluxnet/ozark_pmodel.jl
+++ b/experiments/integrated/fluxnet/ozark_pmodel.jl
@@ -181,12 +181,8 @@ soil_moisture_stress = PiecewiseMoistureStressModel{FT}(
 )
 
 # Set up plant hydraulics
-# Read in LAI from MODIS data
-surface_space = land_domain.space.surface;
-LAI =
-    ClimaLand.Canopy.prescribed_lai_modis(surface_space, start_date, stop_date)
 # Get the maximum LAI at this site over the first year of the simulation
-maxLAI = FluxnetSimulations.get_maxLAI_at_site(start_date, lat, long);
+maxLAI = FluxnetSimulations.get_maxLAI_at_site(start_date, lat, long)
 RAI = maxLAI * f_root_to_shoot
 hydraulics = Canopy.PlantHydraulicsModel{FT}(
     surface_domain,
@@ -201,8 +197,14 @@ hydraulics = Canopy.PlantHydraulicsModel{FT}(
     retention_model,
 )
 height = h_stem + h_leaf
-biomass =
-    Canopy.PrescribedBiomassModel{FT}(; LAI, SAI, RAI, rooting_depth, height)
+biomass = Canopy.ZhouOptimalLAIModel{FT}(
+    surface_domain,
+    toml_dict;
+    SAI,
+    RAI,
+    rooting_depth,
+    height,
+)
 
 # Set up the energy model
 energy = Canopy.BigLeafEnergyModel{FT}(toml_dict; ac_canopy)
@@ -214,7 +216,6 @@ canopy_forcing = (; atmos, radiation, ground)
 canopy = Canopy.CanopyModel{FT}(
     surface_domain,
     canopy_forcing,
-    LAI,
     toml_dict;
     prognostic_land_components,
     radiative_transfer,
@@ -258,6 +259,7 @@ output_vars = [
     "so2",
     "soc",
     "scms",
+    "lai",
 ]
 diags = ClimaLand.default_diagnostics(
     land,
@@ -300,7 +302,7 @@ LandSimVis.make_timeseries(
     diags,
     start_date;
     savedir,
-    short_names = ["swc", "tsoil", "swe"],
+    short_names = ["swc", "tsoil", "swe", "lai"],
     spinup_date = start_date + Day(20),
     comparison_data,
 )
--- a/src/diagnostics/default_diagnostics.jl
+++ b/src/diagnostics/default_diagnostics.jl
@@ -439,6 +439,14 @@ function add_diagnostics!(
     append!(diagnostics, ["ws"])
     return nothing
 end
+function add_diagnostics!(
+    diagnostics,
+    model::CanopyModel,
+    subcomponent::ZhouOptimalLAIModel,
+)
+    append!(diagnostics, ["a0d", "a0a"])
+    return nothing
+end
 
 ## Possible diagnostics for standalone models
 """
@@ -516,6 +524,8 @@ function get_possible_diagnostics(model::CanopyModel)
 
     # Add conditional diagnostics based on atmosphere type
     add_diagnostics!(diagnostics, model, model.boundary_conditions.atmos)
+    # Add conditional diagnostics based on biomass model type
+    add_diagnostics!(diagnostics, model, model.biomass)
     return diagnostics
 end
 function get_possible_diagnostics(model::SnowModel)
@@ -611,7 +621,9 @@ function get_short_diagnostics(model::SoilCO2Model)
     return ["sco2"]
 end
 function get_short_diagnostics(model::CanopyModel)
-    return ["gpp", "ct", "lai", "trans", "er", "sif"]
+    diagnostics = ["gpp", "ct", "lai", "trans", "er", "sif"]
+    add_diagnostics!(diagnostics, model, model.biomass)
+    return diagnostics
 end
 function get_short_diagnostics(model::SnowModel)
     return get_possible_diagnostics(model)
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -228,6 +228,79 @@ function LandModel{FT}(
 end
 
 """
+    LandModel{FT}(
+        forcing,
+        toml_dict::CP.ParamDict,
+        domain::Union{
+            ClimaLand.Domains.Column,
+            ClimaLand.Domains.SphericalShell,
+            ClimaLand.Domains.HybridBox,
+        },
+        Δt;
+        prognostic_land_components = (:canopy, :snow, :soil),
+        soil = ...,
+        soilco2 = ...,
+        canopy = Canopy.CanopyModel{FT}(..., toml_dict; ...),
+        snow = ...,
+    ) where {FT}
+
+A convenience constructor for setting up the default LandModel using prognostic (modeled) LAI
+via the `ZhouOptimalLAIModel`. This is identical to the LAI-prescribing constructor, except
+that `LAI` is not a required argument and the default canopy model uses `ZhouOptimalLAIModel`
+for the biomass component instead of `PrescribedBiomassModel`.
+
+See also the constructor `LandModel{FT}(forcing, LAI, toml_dict, domain, Δt; ...)` for
+the version that uses prescribed LAI from a `TimeVaryingInput`.
+"""
+function LandModel{FT}(
+    forcing,
+    toml_dict::CP.ParamDict,
+    domain::Union{
+        ClimaLand.Domains.Column,
+        ClimaLand.Domains.SphericalShell,
+        ClimaLand.Domains.HybridBox,
+    },
+    Δt;
+    prognostic_land_components = (:canopy, :snow, :soil),
+    soil = Soil.EnergyHydrology{FT}(
+        domain,
+        forcing,
+        toml_dict;
+        prognostic_land_components,
+        additional_sources = (ClimaLand.RootExtraction{FT}(),),
+    ),
+    soilco2 = :soilco2 in prognostic_land_components ?
+              Soil.Biogeochemistry.SoilCO2Model{FT}(
+        domain,
+        Soil.Biogeochemistry.SoilDrivers(
+            PrognosticMet(soil.parameters),
+            forcing.atmos,
+        ),
+        toml_dict,
+    ) : nothing,
+    canopy = Canopy.CanopyModel{FT}(
+        Domains.obtain_surface_domain(domain),
+        (;
+            atmos = forcing.atmos,
+            radiation = forcing.radiation,
+            ground = ClimaLand.PrognosticGroundConditions{FT}(),
+        ),
+        toml_dict;
+        prognostic_land_components,
+    ),
+    snow = Snow.SnowModel(
+        FT,
+        ClimaLand.Domains.obtain_surface_domain(domain),
+        forcing,
+        toml_dict,
+        Δt;
+        prognostic_land_components,
+    ),
+) where {FT}
+    return LandModel{FT}(canopy, snow, soil, soilco2)
+end
+
+"""
     ClimaLand.land_components(land::LandModel)
 
 Returns the components of the `LandModel`.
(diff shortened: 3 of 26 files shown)


experiments/integrated/fluxnet/ozark_pmodel.jl
```diff
@@ -278,7 +274,19 @@ simulation = LandSimulation(
     diagnostics = diags,
 )
 
-@time solve!(simulation)
+using Logging
```
kmdeck: Remove or keep?

experiments/long_runs/snowy_land_pmodel.jl
```diff
@@ -80,13 +80,6 @@ function setup_model(
     )
     forcing = (; atmos, radiation)
 
-    # Read in LAI from MODIS data
```
kmdeck: I think we should keep the 20 year run running with prescribed LAI for now.

src/diagnostics/default_diagnostics.jl
```diff
@@ -610,7 +620,11 @@ function get_short_diagnostics(model::SoilCO2Model)
     return ["sco2"]
 end
 function get_short_diagnostics(model::CanopyModel)
-    return ["gpp", "ct", "lai", "trans", "er", "sif"]
+    diagnostics = ["gpp", "ct", "lai", "trans", "er", "sif"]
+    if model.biomass isa ZhouOptimalLAIModel
```
kmdeck: can we instead use this function already defined?
```
function add_diagnostics!(
    diagnostics,
    model::CanopyModel,
    subcomponent::ZhouOptimalLAIModel,
)
    append!(diagnostics, ["a0d", "a0a"])
    return nothing
end
```

src/standalone/Vegetation/Canopy.jl
```diff
@@ -419,6 +420,65 @@ function PrescribedBiomassModel{FT}(
     )
 end
 
+"""
+    ZhouOptimalLAIModel{FT}(
+        domain,
+        toml_dict::CP.ParamDict;
+        optimal_lai_inputs = optimal_lai_initial_conditions(domain.space.surface),
```
kmdeck: nice! this matches nicely with other spatially varying input read from maps

src/standalone/Vegetation/biomass.jl
```diff
@@ -317,3 +320,282 @@ where `rooting_depth` replaces (-1)/(100ln(β)) and z is expected to be negative
...
+- vpd_gs is the mean VPD during growing season (Pa), used in the WUE factor for water limitation.
+- lai_init comes from MODIS first timestep, providing spatially-varying initial conditions.
+- f0 is the spatially varying fraction of precipitation for transpiration from Zhou et al.
+"""
+function set_historical_cache!(
+    p,
+    Y0,
+    model::ZhouOptimalLAIModel,
+    canopy;
+    A0_daily = eltype(model.parameters)(0.5),
```
kmdeck: should this be in the parameter toml?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+- `pmodel_constants`: P-model constants (including Gamma_star25, Kc25, Ko25, etc.)
+- `T::FT`: Temperature (K)
+- `P_air::FT`: Atmospheric pressure (Pa)
+- `VPD::FT`: Vapor pressure deficit (Pa)
+- `ca::FT`: Ambient CO2 mixing ratio (mol/mol)
+
+# Returns
+- `chi::FT`: Optimal ci/ca ratio (dimensionless), typically 0.7-0.85 for C3, ~0.4 for C4
+"""
+function compute_chi(
```
kmdeck: I think this code should live in the pmodel file

and the blending of c3 and c4 can be removed

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+) where {FT}
+    # Equation 20: m = (sigma * GSL * LAI_max) / (A0_sum * fAPAR_max)
+    fAPAR_max = fAPAR_max_fun(k, LAI_max)
+
+    # Handle edge case: when LAI_max ~ 0 (very low productivity regions like poles),
+    # fAPAR_max = 1 - exp(-k*0) = 0, causing division by zero -> NaN.
+    # In this case, return m = 0 since there's no vegetation to support.
+    # Use ifelse for GPU compatibility (no branching).
+    denominator = Ao_annual * fAPAR_max
+    m = ifelse(
```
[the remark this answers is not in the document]
kmdeck: Mathematically it also seems like it should be ok to do
`m = (sigma * GSL * LAI_max) / (max(fAPAR_max, eps(FT))*A_0_annual)`
because if LAI_max is zero, this returns something that is zero.

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+# Device Compatibility
+This implementation is designed to work on both CPU and GPU:
+- All operations are scalar and supported on CUDA.jl
+- No array allocations or dynamic loops
+- Type-generic over AbstractFloat (Float32, Float64)
+
+# References
+Corless et al. (1996) "On the Lambert W function"
+"""
+@inline function lambertw0(x::T; maxiter::Int = 16) where {T <: AbstractFloat}
```
kmdeck: Do we need 16 iterations? How many iterations does this usually take to converge?

Since it only has one input, we can test at many locations (including near the critical points ID-ed in the branches) and test how many it takes to converge. Does it fail to converge ever? how often?

do we have some unit tests for these functions? solving the equation and then verifying that the root was found correctly

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+    if !(isfinite(x)) || x < T(MINARG)
```
kmdeck: If I understand correctly, MINARG is -1/e. Should x ever be negative? Our initial guess never should be, I presume, but maybe it can somehow go there during the solve?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+        # Using our custom lambertw0 function (W0 is the principal branch)
+        w_val = lambertw0(arg)
+        L_s = mu + (FT(1) / k) * w_val
+
+        # Take minimum with LAI_max (Equation 15)
+        L_s = min(L_s, LAI_max)
+    end
+
+    # Ensure non-negative (should be guaranteed mathematically, but enforce for numerical stability)
+    L_s = max(zero(FT), L_s)
```
kmdeck: does this happen in practice?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+  achieved if fAPAR = 1, calculated from LUE * PPFD.
+- `m::FT`: Parameter relating steady-state LAI to steady-state GPP (dimensionless), from
+  `compute_m()`
+- `k::FT`: Light extinction coefficient (dimensionless), typically 0.5
+- `LAI_max::FT`: Seasonal maximum LAI constraint (m^2 m^-2, dimensionless)
+
+# Returns
+- `L_steady::FT`: Steady-state leaf area index (m^2 m^-2, dimensionless). Always >= 0.
+
+# Example values
```
kmdeck: I personally dont find these example values very useful and they make the docstring a little cluttered

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+            P_air,
+            VPD,
+            ca,
+            PPFD,
+            βm,  # moisture stress factor beta
+        ),
+    )
+
+    # Compute instantaneous A0_annual with actual moisture stress factor beta (kg C m^-2 s^-1)
+    A0_annual_inst = @. lazy(
```
kmdeck: Why is A0_annual the same as A0_daily? computing the same thing 2x?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+    dt_seconds = FT(float(dt))
+    Mc = pmodel_constants.Mc
+    A0_daily_per_timestep = @. lazy(A0_daily_inst * dt_seconds / Mc)
+    A0_annual_per_timestep = @. lazy(A0_annual_inst * dt_seconds / Mc)
+
+    # Accumulate to daily A0 (with moisture stress factor beta) - drives L_steady
+    @. p.canopy.biomass.A0_daily_acc += A0_daily_per_timestep
+
+    # Accumulate to daily A0_annual (with actual beta) - will be added to annual accumulator at noon
+    @. p.canopy.biomass.A0_annual_daily_acc += A0_annual_per_timestep
```
kmdeck: as far as I can tell, this is the same as A0_daily_per_timestep at this point?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+            chi,
+            p.canopy.biomass.vpd_gs,
+        ),
+        p.canopy.biomass.area_index.leaf,
+    )
+
+    # Update A0_annual if year reset (365 days accumulated)
+    @. p.canopy.biomass.A0_annual = ifelse(
+        local_noon_mask == FT(1) &&
+        p.canopy.biomass.days_since_reset >= FT(365),
```
kmdeck: What about leap years?

kmdeck: Perhaps better is to ask if the time corresponds to Dec 31?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,908 @@
...
+    @. p.canopy.biomass.A0_annual_daily_acc = ifelse(
+        local_noon_mask == FT(1),
+        FT(0),
+        p.canopy.biomass.A0_annual_daily_acc,
+    )
+
+    # Update days_since_reset at noon
+    # On year reset: start counting from 1 (today is day 1 of new accumulation period)
+    # Otherwise: increment by 1
+    @. p.canopy.biomass.days_since_reset = ifelse(
```
kmdeck: I wonder if we actually want a second callback for computing annual quantities, that executes once per year

experiments/integrated/fluxnet/ozark_pmodel.jl
AlexisRenchon: Amended, thanks!

AlexisRenchon: note I use Logging to write to a file, so if it crashes, it is easy to see the top of the error... but this is just for developing. shouldn't be merged.

src/diagnostics/default_diagnostics.jl
AlexisRenchon: amended

src/standalone/Vegetation/optimal_lai.jl
AlexisRenchon: using what we had (same chi for c3 and c4) in the pmodel.

AlexisRenchon:
@kmdeck I did my best to address your comments in the latest commit!

src/integrated/land.jl
```diff
@@ -227,6 +227,79 @@ function LandModel{FT}(
     return LandModel{FT}(canopy, snow, soil, soilco2)
 end
 
+"""
+    LandModel{FT}(
```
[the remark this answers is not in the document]
AlexisRenchon: I think if LAI is modeled, it doesn't have to be an argument to pass. But if it is measurements, maybe it has to? (that is how it is set up on main)

I think it comes from https://github.com/CliMA/ClimaLand.jl/blob/main/src/standalone/Vegetation/Canopy.jl#L757

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,707 @@
...
+    # Water-limited fAPAR (Equation 11, second term)
+    # fAPAR_water = f0 * P / A0 * (ca(1-chi)) / (1.6*D)
+    # The iWUE factor (ca(1-chi))/(1.6*D) converts water flux to carbon flux
+    # Guard against zero VPD
+    vpd_safe = ifelse(vpd_gs < eps(FT), eps(FT), vpd_gs)
+    iWUE_factor = (ca_pa * (FT(1) - chi)) / (FT(1.6) * vpd_safe)
+    fAPAR_water = f0 * precip_annual / Ao_annual_safe * iWUE_factor
+
+    # fAPAR_max is the minimum of energy and water constraints (Equation 11)
+    fAPAR_max = min(fAPAR_energy, fAPAR_water)
```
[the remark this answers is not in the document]
kmdeck: maybe once merged we can try running without this second water limitation and see how it looks?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -0,0 +1,707 @@
...
+            pmodel_constants.planck_h,
+            pmodel_constants.N_a,
+        ),
+    )
+
+    # Compute instantaneous potential GPP (kg C m^-2 s^-1)
+    # A single computation serves both daily and annual accumulation
+    dt_seconds = FT(float(dt))
+    Mc = pmodel_constants.Mc
+    @. p.canopy.biomass.A0_daily_acc += compute_A0_daily(
```
kmdeck: This relies on this function only being called once per dt, but this wont be the case if e.g. we use a higher order integrator. We might consider computing these sums in a way that is independent of the time stepping scheme

src/standalone/Vegetation/optimal_lai.jl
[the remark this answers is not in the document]
AlexisRenchon: As discussed, I have tried, it didn't work. I will send an email to the authors to discuss further.

tapios (approved):
LGTM. Unfortunately, I did not have time to do justice to understanding the code. But it looks clean and clear (thank you!). 

For the future, I think we should find a way to represent the effect of GSL locally in time (if needed by an ODE that averages in time). It's odd to have temporally nonlocal (and hence non-causal) parameters like this.

PR #1629, opened 2026-02-09
Checking conservation with constant LAI

### LINKED ISSUE #777: Consistency in fluxes and setting radiative properties of land

kmdeck:
## Background 
A previous PR made it possible to compute certain fluxes using the values of the state at the current step (explicit) and others with fluxes computed using the state (of implicit variables) at the next step.

The soil boundary fluxes should be explicit
The snow boundary fluxes should be explicit
All interaction fluxes should be explicit (root extraction, eg)
Canopy SHF, LHF, LW_n needs to be implicit (evaluated with T_canopy at next step).

Note that soil prognostic vars are implicit, so the implicit state also includes T_soil computed from rhoe_int.

Our goal is to make it so the fluxes stored in the cache are consistent with the fluxes being used as boundary conditions for soil/canopy/snow. This should enable enforcing conservation.
## Purpose 
(1) Split radiation updates into LW and SW pieces. This is because LW pieces require being part of the implicit tendency for the canopy, and require knowledge about other components in the land model, while SW can be explicit for all components. 
(2) Split radiation updates into explicit and implicit updates
(3) Unify sign convention for radiation: LW_n and SW_n are positive if upwards. This required a change in the canopy and in the canopy.soil model

If we want to change the sign convention later on, we can, but for now this makes all land components consistent

## Content

----


src/integrated/land.jl
```diff
@@ -438,119 +437,162 @@ function make_update_implicit_cache(
...
+    LW_u_soil = p.scratch2
+    LW_u_snow = p.scratch3
+    LW_net_canopy = p.canopy.radiative_transfer.LW_n
+    LW_u = p.LW_u
+
+    ϵ_canopy = p.canopy.radiative_transfer.ϵ # this takes into account LAI/SAI
+    T_canopy = ClimaLand.Canopy.canopy_temperature(canopy.energy, canopy, Y, p)
+
+    ϵ_soil = land.soil.parameters.emissivity
+    T_soil = ClimaLand.Domains.top_center_to_surface(p.soil.T)
```
kmdeck: This is technically T_soil at the next step

src/integrated/land.jl
```diff
@@ -438,119 +437,162 @@ function make_update_implicit_cache(
...
+
+    @. LW_u_soil = ϵ_soil * _σ * T_soil^4 + (1 - ϵ_soil) * LW_d_canopy # double checked
+    @. LW_u_snow = ϵ_snow * _σ * T_snow^4 + (1 - ϵ_snow) * LW_d_canopy # identical to soil, checked
+    @. LW_net_canopy = -(
+        ϵ_canopy * LW_d - 2 * ϵ_canopy * _σ * T_canopy^4 +
+        ϵ_canopy * LW_u_soil * (1 - p.snow.snow_cover_fraction) +
+        ϵ_canopy * LW_u_snow * p.snow.snow_cover_fraction
+    ) # area weighted by snow cover fraction, OK
 
+    @. LW_u =
```
kmdeck: As long as this is defined this way, I think we should have conservation with atmos

kmdeck:
closed by #1771

PR #1644, opened 2026-02-13
new set ic methods

kmdeck:
## Purpose 
Move "set bucket ic from file" code to the Land model from Coupler repo. this streamlines coupler code between bucket and integrated land and makes NCDatasets not a dependency.

Note that the land model does not have an ic file like for the bucket. this is used in Weather quest/coupled runs only. so I dont have a test for the `maket_set_from_file` function in land.

I also created another bucket and land ic function which matches what is in the coupler. this streamlines coupler code between bucket and integrated land and offlines runs
## Content

----

### DIFF
--- a/src/simulations/initial_conditions.jl
+++ b/src/simulations/initial_conditions.jl
@@ -566,3 +566,154 @@ function set_soil_initial_conditions_from_temperature_and_total_water!(
         )
     return nothing
 end
+
+"""
+    make_set_initial_state_from_file(ic_path, model::ClimaLand.Bucket.BucketModel)
+
+Returns a function which takes (Y,p,t0,model) as arguments, and updates
+the state Y of the `BucketModel` in place with initial conditions from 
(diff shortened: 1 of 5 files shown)


src/simulations/initial_conditions.jl
```diff
@@ -566,3 +566,146 @@ function set_soil_initial_conditions_from_temperature_and_total_water!(
...
+atmospheric temperature. 
+
+It is assumed that in CoupledAtmosphere simulations that `p.drivers.T` has 
+been updated already.
+"""
+function make_set_initial_state_from_atmos_and_parameters(
+    land::LandModel{FT},
+) where {FT}
+    function set_ic!(Y, p, t0, land)
+        atmos = land.soil.boundary_conditions.top.atmos
```
juliasloan25:
```suggestion
        atmos = ClimaLand.get_drivers(land)[1]
```
I'm not sure which is better

PR #1645, opened 2026-02-13
Fix EnergyHydrology conservation diagnostics

imreddyTeja:
Follow up to eb886570645581e0a629c956efcfc5318ebe5dbd . After that commit, only requested diagnostics get created and added to `ALL_DIAGNOSTICS`.
When conservation diagnostics are requested, they were not getting created, leading to an error.

This commit defines the conservation
diagnostics when they are requested.

----


kmdeck (approved):
thanks!

PR #1657, opened 2026-02-27
change CairoMakie compat

kmdeck:
## Purpose 
Version 0.12 CairoMakie does not work so I updated the compat

## Content

----


juliasloan25 (approved):
thank you! Surprising that the downgrade tests didn't catch the incompatibility 🤔

PR #1591, opened 2026-01-07
Update to new surface fluxes and refactor `turbulent_fluxes!` function

kmdeck:
## Purpose 
Upgrade to the new surface fluxes interface.

Since this let's us provide parameterization functions for q_sfc, T_sfc, we no longer need special methods of `turbulent_fluxes!` for each land model component, and can instead have methods for functions like `component_temperature`, etc, which are called by `turbulent_fluxes!`.
See:
```
function turbulent_fluxes!(
    dest,
    atmos::AbstractAtmosphericDrivers,
    model::AbstractModel,
    Y,
    p,
    t,
)

    T_sfc = component_temperature(model, Y, p) # guess
    q_sfc = component_specific_humidity(model, Y, p) # guess
    roughness_model = surface_roughness_model(model, Y, p)
    update_T_sfc = get_update_surface_temperature_function(model, Y, p)
    update_q_sfc = get_update_surface_humidity_function(model, Y, p)
    h_sfc = surface_height(model, Y, p)
    displ = surface_displacement_height(model, Y, p)
    update_∂T_sfc∂T = get_∂T_sfc∂T_function(model, Y, p)
    update_∂q_sfc∂T = get_∂q_sfc∂T_function(model, Y, p)
    earth_param_set = get_earth_param_set(model)
    momentum_fluxes = Val(return_momentum_fluxes(atmos))
    dest .=
        turbulent_fluxes_at_a_point.(
            momentum_fluxes, # return_extra_fluxes
            p.drivers.P,
            p.drivers.T,
            p.drivers.q, # q_tot
            p.drivers.u,
            atmos.h,
            T_sfc,
            q_sfc,
            roughness_model,
            update_T_sfc,
            update_q_sfc,
            h_sfc,
            displ,
            update_∂T_sfc∂T,
            update_∂q_sfc∂T,
            earth_param_set,
        )
    return nothing
end
```
the exception is for the soil model - I will update that in a later PR. (It is using the latest SF.jl, but not the common function as canopy, snow, bucket)

This function: `turbulent_fluxes_at_a_point` uses the new SF.jl
## To-do
new unit tests or documentation pages? (doc strings are updated in this PR)

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #1659, opened 2026-03-01
add in NeuralAlbedo functionality, prepare for global runs

### LINKED ISSUE #1169: RadiativeFluxes driver update functions allocate
This applies to both `PrescribedRadiativeFluxes` and `CoupledRadiativeFluxes`:
When the driver has the necessary fields to update the cosine zenith angle, the update function is called in a way that allocates an intermediate surface field.  The driver should be able to update cosine zenith without the allocations.

a-charbon:
## Purpose 

This PR exists to flesh out the actual interaction of the NeuralAlbedoModel within ClimaLand.Snow, to be used in global model runs. Fills out methods for the type and adds tests.

With two neural models inside SnowParameters, the size of the SnowParameters object is now too large to be passed as an argument considering the size limits of a GPU thread. Snow.jl and affiliated parameterizations have been edited accordingly to deal with this size issue.

Deals with broadcasting of solar zenith angle, which resolves #1169 

also removes code in both pipeline.yml files from #1591 

## Content

----
- [x ] I have read and checked the items on the review checklist.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -26,12 +26,10 @@ steps:
 
       - echo "--- Instantiate experiments"
       - "julia --project=.buildkite -e 'using Pkg; Pkg.develop(;path=\".\"); Pkg.instantiate(;verbose=true)'"
-      - "julia --project=.buildkite -e 'using Pkg; Pkg.add(Pkg.PackageSpec(;name=\"SurfaceFluxes\", rev=\"main\"))'"
       - "julia --project=.buildkite -e 'using Pkg; Pkg.precompile(;strict=true); using CUDA; CUDA.precompile_runtime(); Pkg.status()'"
 
       - echo "--- Instantiate test"
       - "julia --project=test -e 'using Pkg; Pkg.develop(;path=\".\"); Pkg.add(\"MPI\"); Pkg.add(\"CUDA\"); Pkg.instantiate(;verbose=true)'"
-      - "julia --project=.buildkite -e 'using Pkg; Pkg.add(Pkg.PackageSpec(;name=\"SurfaceFluxes\", rev=\"main\"))'"
       - "julia --project=test -e 'using Pkg; Pkg.precompile(;strict=true); using CUDA; CUDA.precompile_runtime(); Pkg.status()'"
 
     agents:


experiments/standalone/Snow/snowmip_simulation.jl
```diff
@@ -47,19 +47,34 @@ include(
...
 t0 = FT(0.0)
 tf = FT(seconds[end])
 ndays = (tf - t0) / 3600 / 24
 Δt = FT(60 * 60)
 
-domain = ClimaLand.Domains.Point(; z_sfc = FT(0))
+domain = ClimaLand.Domains.Point(; z_sfc = FT(0), longlat = FT.((long, lat)))
 
-density = NeuralSnow.NeuralDepthModel(FT)
+surf_temp = Snow.EquilibriumGradientTemperatureModel{FT}()
```
kmdeck: could we set these based on flags defined at the. very beginning of the script? i.e. not rely on commenting something out

src/simulations/initial_conditions.jl
```diff
@@ -191,9 +191,52 @@ function set_snow_initial_conditions!(
...
+            p.snow.T,
+            params.ΔS,
+            params.earth_param_set,
+        )
+    if :Z in propertynames(Y.snow)
+        FT = eltype(Y.snow.S)
+        #no depth field in spin-up file: start with reasonable guess (snow density 333 kg/m^3)
+        Y.snow.Z .= FT(3) .* Y.snow.S
+        if :P_avg in propertynames(Y.snow) #EMA is present, set all the vars
+            #drivers are not set yet - how to initialize these? Should we at all?
```
kmdeck: How long do they take to spinup?

src/simulations/initial_conditions.jl
```diff
@@ -191,9 +191,52 @@ function set_snow_initial_conditions!(
...
+    end
     return nothing
 end
+
+"""
+    initial_albedo_constraint(S::FT)
+
+Enforces the constraint that A must be snowy only if snow exists:
+"""
+function initial_albedo_constraint(S::FT) where {FT}
```
kmdeck: slightly confusing if A is meant to be the true snow albedo?

src/standalone/Snow/Snow.jl
```diff
@@ -572,16 +572,21 @@ For this model, we track the snow water equivalent S in meters (liquid
...
+    :S,
+    :S_l,
+    :U,
+    extra_prog_vars(m.parameters.density)..., #Note - this/below would need to change if model choices share variables
+    extra_prog_vars(m.parameters.α_snow)...,
+)
 
 """
-    density_prog_vars(::AbstractDensityModel)
+    extra_prog_vars(::AbstractDensityModel, ::AbstractAlbedoModel)
```
kmdeck: update to be the Union, with only a single argument

ext/constrained_nn/NeuralSnow.jl
```diff
@@ -21,7 +22,48 @@ using Flux: relu
...
+be combined with the prognostic update dynamics of a NeuralAlbedoModel.
+
+An instance of this type must have at least:
+    1) one functor method defined, (::T)(model, Y, p)
+       for T <: AlbedoInitialValueModel that returns the initial albedo value,
+       which will be called in `albedo_reset_rate()`
+    2) a float field "reset_limit" (m/s) of the critical liquid-equivalent snowfall preciptiation rate,
+       which will also get used in resetting the snow albedo.
+"""
+abstract type AlbedoInitialValueModel{FT <: AbstractFloat} end
```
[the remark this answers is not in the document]
kmdeck: this is slightly misleading though, because we dont alter `Y` directly, but instead create a tendency..

ext/constrained_nn/NeuralSnow.jl
```diff
@@ -344,29 +426,189 @@ function eval_nn(
...
+method (::T)(model, Y, p) where T is the desired type of AlbedoInitialValueModel.
+"""
+function albedo_reset_rate(
+    init_alb::AlbedoInitialValueModel,
+    model,
+    Δt::FT,
+    Y,
+    p,
+) where {FT <: AbstractFloat}
+    return (init_alb(model, Y, p) .- Y.snow.A) ./ Δt
```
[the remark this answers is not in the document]
kmdeck: I think just unpacking it here is clearer than making a function which returns init_alpha. If we need different behavior, we can specialize on the type of init_alb in this `albedo_reset_rate` function

or, if you prefer the flexibility (in case someone in the future wants to add another way of coming up with init_alpha), perhaps defining a function is better
```
function get_initial_albedo(init_alb, Y, p)
```
instead of `init_alb(Y,p)`

I like this better because there are other things stored in init_alb than the initial value or directions for computing the initial albedo. But making the action of `init_alb` be the initial value ignores that

ext/constrained_nn/NeuralSnow.jl
```diff
@@ -344,29 +426,189 @@ function eval_nn(
...
+    return w * new + (1 - w) * old
+end
+
+"""
+    snow_surf_SW_down(model_type::Val{(:snow,)}, p)
+
+Obtains the correct downward shortwave radiation magnitude for a snow model without an
+overlying canopy (standalone or just coupled with soil).
+"""
+@inline snow_surf_SW_down(
```
kmdeck: could this just be "snow_SW_down"?

ext/constrained_nn/NeuralSnow.jl
```diff
@@ -344,29 +426,189 @@ function eval_nn(
...
+        eval_z_nn.(
+            Ref(density.z_model),
             Y.snow.Z,
             Y.snow.S,
-            Y.snow.P_avg,
-            Y.snow.T_avg,
-            Y.snow.R_avg,
-            Y.snow.Qrel_avg,
-            Y.snow.u_avg,
+            ema_value.(density.w, abs.(p.drivers.P_snow), Y.snow.P_avg),
```
kmdeck: why is this changing?

ext/constrained_nn/NeuralSnow.jl
```diff
@@ -381,62 +623,88 @@ so that depth does not become negative. Additionally, if the
...
+    Y,
+    p,
+    t,
+)
+    update_dαdt!(dY.snow.A, albedo, model, Y, p, t)
 end
 
 end
+
+# note to self: sodankyla latitude is wrong in the snowmip artifact
```
kmdeck: can remove

ext/constrained_nn/NeuralSnow.jl
```diff
@@ -21,7 +22,48 @@ using Flux: relu
...
+Default values are a new albedo of 0.8 for 3 mm liquid-equivalent of snow precipitation in 24 hours.
+"""
+struct ConstantInitialAlbedo{FT} <: AlbedoInitialValueModel{FT}
+    "Initial albedo constant (unitless 0-1)"
+    init_α::FT
+    "Threshold of liquid-equiavlent snow precipitation for a reset (m/s)"
+    reset_limit::FT
+end
+ConstantInitialAlbedo(::Type{FT}) where {FT <: AbstractFloat} =
+    ConstantInitialAlbedo(FT(0.8), FT(0.003 / 86400))
```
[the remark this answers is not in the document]
a-charbon: This makes sense to add regardless, but doing it this way means I cannot specify a default version of the parameterization when no toml is provided, like the "ConstantInitialAlbedo(FT)" used as a default optional keyword arg in the constructor of NeuralAlbedoModel. Do we want to make the user have to specify this choice (when no other types are available) just to build a NeuralAlbedoModel? Or should I keep the (FT) version and the (toml_dict) version you have here?

ext/constrained_nn/NeuralSnow.jl
a-charbon: see private slack messages over our discussion on this. I have maintained the original (using `ema_value`) behavior for now.

kmdeck: Easier to discuss in person i think. can we set a meeting time?

kmdeck (approved):
thanks! Im excited to see how this changes the global sims

.buildkite/Manifest-v1.12.toml
```diff
@@ -3161,9 +3167,9 @@ version = "7.8.3+2"
 
 [[deps.SurfaceFluxes]]
 deps = ["RootSolvers", "Thermodynamics"]
-git-tree-sha1 = "830e95e01f1fc11ddbe4c4fcbabb017736c290aa"
+git-tree-sha1 = "f63a8a9eddd9d2efb229b421adda86437c8e230a"
 uuid = "49b00bb7-8bd4-4f2b-b78c-51cd0450215f"
-version = "0.15.1"
+version = "1.0.0"
```
kmdeck: @a-charbon here

PR #1670, opened 2026-03-13
fix pmodel gpp for c4

AlexisRenchon:
This PR makes changes to our pmodel implementation to match [pyrealm](https://github.com/ImperialCollegeLondon/pyrealm) implementation which is used, for example, by the Zhou optimal LAI model code

To do about Amazon / c4 grassland bias in south America (LHF and GPP too low in Amazon, GPP too high in c4): 
- run single column at coordinates of Amazon, South America grassland c4. Check what is suspicious: does PAR look good? APAR? Root depth? Moisture stress? LAI? Are things distinct by c3 / c4 (e.g., root depth, and therefore, moisture stress for c3 and c4...), is RAI = 1 reasonable (or should we try RAI = f*maxLAI)?  

- try calibrating GPP + LHF on the first commit of this PR (params adjusted). need to add a dataset to calibrate on (for gpp) on derecho. It seems GPP is overestimated everywhere, so possibly calibration can help (just offset globally)

### DIFF
--- a/src/standalone/Vegetation/Canopy.jl
+++ b/src/standalone/Vegetation/Canopy.jl
@@ -222,7 +222,8 @@ end
         toml_dict::CP.ParamDict;
         is_c3 = clm_photosynthesis_parameters(domain.space.surface).is_c3,
         cstar = toml_dict["pmodel_cstar"],
-        β = toml_dict["pmodel_β"],
+        β_c3 = toml_dict["pmodel_β_c3"],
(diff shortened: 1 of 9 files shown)


AlexisRenchon:
> @AlexisRenchon it looks like GPP is still biased high everywhere, but if this is what you see in their model, let's make the change anyways?
> 
> what values does this model code use for quantum yield (temperature dependent)?

I'll keep working on this. This is what they do in pyrealm, the package the optimal lai model used

PR #1672, opened 2026-03-13
inland waters slab

AlexisRenchon:
# Inland water slab

[DOCS](https://clima.github.io/ClimaLand.jl/previews/PR1672/standalone/pages/inland_waters/slab_lake/)

This PR adds a fixed-depth slab lake representation for inland water grid points (lakes, rivers, reservoirs). A prognostic `SlabLakeModel` (in its own `InlandWater` module) evolves lake internal energy (`U_lake`) at the surface while the underlying soil column acts as a thermal sediment substrate.

## Key features

- `SlabLakeModel` as a standalone component with `SlabLakeParameters` (configurable depth, conductance, albedo liquid/ice, emissivity, roughness lengths)
- Prognostic lake internal energy with phase change (liquid fraction `q_l` determines ice/liquid state)
- Lake–sediment heat exchange via a single tunable conductance parameter (series-resistance model with `Δz_top` read from the soil domain at runtime)
- Fixed-depth assumption: net surface water flux (P − E) balanced by signed runoff residual
- Energy conservation: surface energy balance + sediment heat flux + runoff enthalpy = dU_lake/dt
- Continuous blending of soil and lake fluxes via `lake_fraction` mask
- Lake diagnostics: `lkie` (lake internal energy), `tlake` (temperature), `qlake` (liquid fraction)
- Inland water mask derived from [IMERG land-sea mask](https://gpm.nasa.gov/data/directory/imerg-land-sea-mask-netcdf) (water fraction > 80%, latitude ±60°), located in `InlandWater.inland_water_mask()`

## What changed at lake grid points

- Surface fluxes drive the slab lake energy equation, blended with soil via lake fraction
- Soil receives heat only through the lake–sediment interface (no direct atmospheric forcing at lake points)
- Soil sublimation source masked at lake points
- LAI, SAI, and RAI zeroed at lake points (no canopy over lakes)
- Ground albedo and LW upwelling area-weighted across soil, snow, and lake fractions
- Surface albedo and specific humidity vary with ice fraction; roughness uses lake values

## Architecture

- Lake component uses dispatch (`Nothing` vs `SlabLakeModel`) throughout — no runtime `if !isnothing` checks in per-step code
- `soil_boundary_fluxes!` dispatches on `Val(prognostic_land_components)` for lake-aware tuples
- Radiation dispatch (`ground_albedo_PAR/NIR`, `_ground_lw_upwelling`, `_update_lake_radiative_fluxes!`) uses separate methods for with-lake and without-lake component tuples
- Sediment heat flux computation lives in `src/integrated/soil_lake_model.jl`, following the same pattern as `update_soil_snow_ground_heat_flux!` in `soil_snow_model.jl`
- `lake_fraction` is a static cache field (set once at initialization from `inland_water_mask`) needed by canopy ground albedo dispatch methods

## Follow-up work

- Ice-dependent roughness lengths (currently constant for liquid and frozen states)
- Standalone lake mode (currently lake only runs in integrated mode with a soil model)

image

image

Note: adjusting the mask water fraction threshold (currently using above 80% water) might reduce RMSE further a bit. Not tested yet.

Note2, for discussion: mask is currently binary. Do we want to ensure the code will work in case in the future we use a non-binary lake fraction f?

### DIFF
--- a/experiments/calibration/models/snowy_land.jl
+++ b/experiments/calibration/models/snowy_land.jl
@@ -69,15 +69,8 @@ function setup_model(
     conductance = PModelConductance{FT}(toml_dict)
     soil_moisture_stress =
         ClimaLand.Canopy.PiecewiseMoistureStressModel{FT}(domain, toml_dict)
-    biomass = ClimaLand.Canopy.PrescribedBiomassModel{FT}(
-        domain,
-        LAI,
-        toml_dict;
-        height = ClimaLand.Canopy.clm_canopy_height(
-            surface_space;
-            max_height = atmos.h * FT(0.9),
-        ),
-    )
+    biomass =
+        ClimaLand.Canopy.PrescribedBiomassModel{FT}(domain, LAI, toml_dict)
     canopy = ClimaLand.Canopy.CanopyModel{FT}(
         surface_domain,
         canopy_forcing,
--- a/experiments/long_runs/snowy_land_pmodel.jl
+++ b/experiments/long_runs/snowy_land_pmodel.jl
@@ -101,15 +101,8 @@ function setup_model(
     # Use the soil moisture stress function based on soil moisture only
     soil_moisture_stress =
         ClimaLand.Canopy.PiecewiseMoistureStressModel{FT}(domain, toml_dict)
-    biomass = ClimaLand.Canopy.PrescribedBiomassModel{FT}(
-        domain,
-        LAI,
-        toml_dict;
-        height = ClimaLand.Canopy.clm_canopy_height(
-            surface_space;
-            max_height = atmos.h * FT(0.9),
-        ),
-    )
+    biomass =
+        ClimaLand.Canopy.PrescribedBiomassModel{FT}(domain, LAI, toml_dict)
     canopy = ClimaLand.Canopy.CanopyModel{FT}(
         surface_domain,
         canopy_forcing,
--- a/test/standalone/Vegetation/canopy_model.jl
+++ b/test/standalone/Vegetation/canopy_model.jl
@@ -519,7 +519,7 @@ end
         @test abs.(
             Array(parent(dY_2.canopy.energy.T .- dY.canopy.energy.T))[1] ./ ΔT -
             ∂Ṫ∂T,
-        ) / abs.(∂Ṫ∂T) < 0.4 # Error propagates here from ∂LHF∂T
+        ) / abs.(∂Ṫ∂T) < 0.5 # Error propagates here from ∂LHF∂T
     end
 end
 
--- a/toml/default_parameters.toml
+++ b/toml/default_parameters.toml
@@ -539,19 +539,19 @@ description = "Minimum roughness"
 tag = "MoninObukhovCanopyFluxes"
 
 ["canopy_z_0m_coeff"]
-value = 0.02
+value = 0.0175
 type = "float"
 description = "Scalar multipling canopy height to get z_0m"
 tag = "MoninObukhovCanopyFluxes"
 
 ["canopy_z_0b_coeff"]
-value = 0.0007
+value = 0.0175
 type = "float"
 description = "Scalar multipling canopy height to get z_0b"
 tag = "MoninObukhovCanopyFluxes"
 
 ["canopy_d_coeff"]
-value = 0.007
+value = 0.16
 type = "float"
 description = "Scalar multipling canopy height to get displacement height"
 tag = "MoninObukhovCanopyFluxes"
--- a/toml/uncalibrated_parameters.toml
+++ b/toml/uncalibrated_parameters.toml
@@ -513,19 +513,19 @@ description = "Minimum roughness"
 tag = "MoninObukhovCanopyFluxes"
 
 ["canopy_z_0m_coeff"]
-value = 0.13
+value = 0.0175
 type = "float"
 description = "Scalar multipling canopy height to get z_0m"
 tag = "MoninObukhovCanopyFluxes"
 
 ["canopy_z_0b_coeff"]
-value = 0.013
+value = 0.0175
 type = "float"
 description = "Scalar multipling canopy height to get z_0b"
 tag = "MoninObukhovCanopyFluxes"
 
 ["canopy_d_coeff"]
-value = 0.67
+value = 0.16
 type = "float"
 description = "Scalar multipling canopy height to get displacement height"
 tag = "MoninObukhovCanopyFluxes"


docs/src/standalone/pages/inland_waters/slab_lake.md
```diff
@@ -0,0 +1,220 @@
...
+- $Q_{\rm sfc}$ is the net surface energy flux into the atmosphere (W m$^{-2}$)
+- $Q_{\rm runoff}$ is the energy carried away by runoff (W m$^{-2}$)
+- $Q_{\rm sed}$ is the heat flux from the lake to the underlying sediment (W m$^{-2}$)
+
+## Surface Energy Flux
+
+The surface energy balance includes radiative, turbulent, and precipitation terms:
+
+```math
+Q_{\rm sfc} = R_n + \text{LHF} + \text{SHF} + P_{\rm liq} \cdot \rho \hat{e}(T_{\rm lake}, q_l)
```
kmdeck: I believe the energy from precipitation should have the air temperature, not the lake temperature, and not depend on q_l.

`P_liq *rho_l * cp_l *(T_air - T0)`

(In coupled simulations we need to get this from the atmosphere instead, so eventually we need to modify our code to dispatch off of `PrescribedAtmosphere` vs `CoupledAtmosphere` to do this)

docs/src/standalone/pages/inland_waters/slab_lake.md
```diff
@@ -0,0 +1,220 @@
...
+## Boundary Conditions at the Sediment
+
+At lake grid points, the soil beneath the lake receives only conductive heat from the lake-sediment interface. No water infiltrates into the sediment:
+
+- **Heat flux (top boundary)**: $Q_{\rm sed}$ (the lake-sediment heat flux)
+- **Water flux (top boundary)**: 0 (no infiltration)
+- **Surface/subsurface runoff**: 0 (runoff is handled by the lake)
+- **Hydraulic conductivity**: set to 0 throughout the soil column (prevents water movement)
+
+Soil sublimation is also suppressed at lake points, since the ice sublimation flux is handled by the lake surface energy balance.
```
kmdeck: All soil fluxes with the atmosphere are zero, not just sublimation

docs/src/standalone/pages/inland_waters/slab_lake.md
```diff
@@ -0,0 +1,220 @@
...
+
+When coupled with the snow model via `SoilSnowModel`, snow can accumulate on the lake surface. Snow-on-lake coupling is handled through the same fraction-based blending as the soil–lake interface. Detailed snow–lake partitioning of surface fluxes (e.g., separate treatment of exposed vs. snow-covered lake fractions) is not yet implemented and is planned for future work.
+
+## Model Assumptions
+
+1. **Well-mixed column**: The lake has no vertical temperature gradient. This is a reasonable approximation for shallow or wind-mixed lakes.
+2. **Fixed depth**: Lake volume does not change. Excess water is removed as runoff; water deficits are implicitly supplied.
+3. **No horizontal transport**: Each lake grid point is independent.
+4. **Phase-dependent albedo**: Albedo transitions smoothly between open water and ice.
+5. **No vegetation**: Canopy processes are not active at lake points.
```
kmdeck: do we need to enforce that LAI = 0 if the lake fraction is nonzero?

docs/src/standalone/pages/inland_waters/slab_lake.md
```diff
@@ -0,0 +1,202 @@
...
+### Fully liquid ($q_l \geq 1$)
+
+```math
+T_{\rm lake} = T_0 + \frac{U}{\rho_l \cdot d \cdot c_{p,l}}
+```
+
+where:
+- $T_0$ is the reference temperature (K)
+- $\rho_l$ is the density of liquid water (kg m$^{-3}$)
+- $d$ is the lake depth (m)
```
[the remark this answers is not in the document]
kmdeck: It only appears in the conversion from U to T when multiplied by cp, right?

src/diagnostics/land_compute_methods.jl
```diff
@@ -819,16 +797,21 @@ function compute_evapotranspiration!(
...
         @. out =
             (
-                (1 - p.snow.snow_cover_fraction) *
-                p.soil.turbulent_fluxes.vapor_flux_liq +
-                (1 - p.snow.snow_cover_fraction) *
-                p.soil.turbulent_fluxes.vapor_flux_ice +
-                p.canopy.turbulent_fluxes.vapor_flux +
-                p.snow.snow_cover_fraction * p.snow.turbulent_fluxes.vapor_flux
+                (1 - f_lake) * (
+                    (1 - p.snow.snow_cover_fraction) *
```
kmdeck: I think this should be `(1-f_lake -p.snow.snow_cover_fraction)`

src/integrated/land_radiation.jl
```diff
@@ -115,9 +115,18 @@ function Canopy.ground_albedo_PAR(
...
-    @. p.α_ground.PAR =
-        (1 - p.snow.snow_cover_fraction) * p.soil.PAR_albedo +
-        p.snow.snow_cover_fraction * p.snow.α_snow
+    snow_frac = p.snow.snow_cover_fraction
+    α_soil = p.soil.PAR_albedo
+    α_snow = p.snow.α_snow
+    if hasproperty(p, :lake)
+        f_lake = p.lake.lake_fraction
+        @. p.α_ground.PAR =
+            (1 - f_lake) * ((1 - snow_frac) * α_soil + snow_frac * α_snow) +
```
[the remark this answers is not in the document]
kmdeck: same comment about defining f_lake using an if statement and then using the same formula for both (not going to comment on every instance of it)

IMERG_land_sea_mask.nc
kmdeck: can move to clima artifacts

src/standalone/InlandWater/InlandWater.jl
```diff
@@ -0,0 +1,596 @@
...
+    params::SlabLakeParameters{FT},
+    earth_param_set,
+) where {FT}
+    _ρ_l = FT(LP.ρ_cloud_liq(earth_param_set))
+    _T_ref = FT(LP.T_0(earth_param_set))
+    _T_freeze = FT(LP.T_freeze(earth_param_set))
+    _LH_f0 = FT(LP.LH_f0(earth_param_set))
+    _cp_i = FT(LP.cp_i(earth_param_set))
+    _cp_l = FT(LP.cp_l(earth_param_set))
+    depth = max(params.depth, sqrt(eps(FT)))
```
kmdeck: Is this needed? Depth is a parameter and should be much much bigger than sqrt(eps(FT))

src/standalone/InlandWater/InlandWater.jl
```diff
@@ -0,0 +1,596 @@
...
+        FT, # T
+        FT, # q_l
+        FT, # sediment_heat_flux
+        FT, # surface_water_flux
+        FT, # surface_energy_flux
+        FT, # runoff_energy_flux
+        FT, # runoff
+        FT, # lake_fraction
+        FT, # lake_albedo
+        NamedTuple{(:lhf, :shf, :vapor_flux, :∂lhf∂T, :∂shf∂T), Tuple{FT, FT, FT, FT, FT}}, # turbulent_fluxes
```
kmdeck: For coupled runs, we also need to store momentum fluxes - see the other components for an example. We also need to talk with Julia about adding this new component to the fluxes atmos sees

src/standalone/InlandWater/InlandWater.jl
```diff
@@ -0,0 +1,596 @@
...
+    model::SlabLakeModel{FT},
+    Y,
+    p,
+    t,
+) where {FT}
+    # Compute turbulent fluxes using the generic interface
+    turbulent_fluxes!(p.lake.turbulent_fluxes, bc.atmos, model, Y, p, t)
+
+    # Compute net radiation in standalone mode
+    ClimaLand.net_radiation!(p.lake.R_n, bc.radiation, model, Y, p, t)
```
kmdeck: this is not correct in simulations with multiple components...

docs/src/standalone/pages/inland_waters/slab_lake.md
```diff
@@ -0,0 +1,213 @@
...
+## Governing Equation
+
+The prognostic variable is the lake internal energy per unit ground area, $U$ (J m$^{-2}$):
+
+```math
+\frac{dU}{dt} = -Q_{\rm sfc} - Q_{\rm runoff} + Q_{\rm sed}
+```
+
+where:
+- $Q_{\rm sfc}$ is the net surface energy flux into the atmosphere (W m$^{-2}$)
```
tapios: For some reason, the formatting does not work if you start a bullet item with a math symbol (it centers the symbol on the line). I haven't figured out how to work around this. Instead, I suggest you start the sentence after the bullet/dash with a work, e.g., 
```suggestion
- the net surface energy flux into the atmosphere, $Q_{\rm sfc}$ (W m$^{-2}$)
```
(and similarly for the rest).

docs/src/standalone/pages/inland_waters/slab_lake.md
```diff
@@ -0,0 +1,213 @@
...
+where $P_{\rm liq}$ is the liquid precipitation rate, $E_{\rm liq}$ is the liquid evaporation rate, and $E_{\rm ice}$ is the ice sublimation rate. To enforce constant depth, the runoff is defined as:
+
+```math
+R_{\rm lake} = -F_{\rm water}
+```
+
+Positive runoff corresponds to net precipitation (water drains away), while negative runoff corresponds to net evaporation (water must be supplied to maintain depth). The associated energy flux is:
+
+```math
+Q_{\rm runoff} = R_{\rm lake} \cdot \rho \hat{e}(T_{\rm lake}, q_l)
```
tapios:
```suggestion
Q_{\rm runoff} = R_{\rm lake} \cdot \rho \widehat{e}(T_{\rm lake}, q_l)
```
(\hat and \tilde do not render correctly. Use \widehat and \widetilde instead). 

Re the subtance and @kmdeck's comment: I think it's ok to assume that the runoff balancing precipitation has the lake temperature, even if that leaves a time tendency in the energy balance from precipitation. I think this is how it is in reality: Precipitation falls at the surface, runoff (effluent) leaves the lake somewhere else, at the equilibrated temperature. We should just make this explicit. 

Evaporation is more complicated. Assuming that the water enters with the lake temperature is probably not very accurate, but it's perhaps ok, and we have no clear alternative. If we want to stick with it, we should make it explicit.

docs/src/standalone/pages/inland_waters/slab_lake.md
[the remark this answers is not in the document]
tapios: Agree with @kmdeck here. We should express G in terms of an effective d.

test/integrated/inland_water_slab.jl
```diff
@@ -0,0 +1,106 @@
...
+    toml_dict = LP.create_toml_dict(FT)
+    earth_param_set = LP.LandParameters(toml_dict)
+    params = SlabLakeParameters{FT}(; earth_param_set)
+
+    @test params.depth == FT(10)
+    @test params.conductance == FT(0.1)
+    @test params.liquid_albedo == FT(0.08)
+    @test params.ice_albedo == FT(0.6)
+    @test params.emissivity == FT(0.97)
+end
```
tapios: Include energy and mass conservation tests for the various scenarios (precip onto lake, evap out of lake etc.)

src/integrated/land.jl
```diff
@@ -468,6 +544,27 @@ function make_update_boundary_fluxes(
...
+                (1 - f) * p.soil.top_bc.heat + f * p.lake.sediment_heat_flux
+            # Zero out infiltration and runoff at lake points
+            if haskey(p.soil, :infiltration)
+                @. p.soil.infiltration = (1 - f) * p.soil.infiltration
+            end
+            if haskey(p.soil, :R_s)
+                @. p.soil.R_s = (1 - f) * p.soil.R_s
+            end
+            if haskey(p.soil, :R_ss)
+                @. p.soil.R_ss = (1 - f) * p.soil.R_ss
```
kmdeck: I am not sure this one should have this weighting

src/standalone/InlandWater/InlandWater.jl
```diff
@@ -0,0 +1,596 @@
...
+# ─── Update aux ───────────────────────────────────────────────────────────────
+
+function ClimaLand.make_update_aux(model::SlabLakeModel{FT}) where {FT}
+    function update_aux!(p, Y, t)
+        params = model.parameters
+        earth_param_set = params.earth_param_set
+        mask = model.inland_water_mask
+
+        # Lake fraction from mask
+        @. p.lake.lake_fraction = mask
```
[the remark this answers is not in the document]
kmdeck: I think the model is always available to grab it from, because all of our functions are made using a closure:
function make_update_X(model)
function update_X(Y,p,t)
          # can access model here, dont need to access in `p`
end
return update_X
end

For example, spatially varying soil parameters are stored in the `model`, and access from the model each step. We do not update cache variables each step with the static parameter values

src/integrated/land.jl
```diff
@@ -223,99 +258,29 @@ function LandModel{FT}(
...
-`update_aux!` steps. So those must occur first (as controlled by the order of
-the components returned by `land_components!`.
-
-This needs to be fixed.
+Returns the prognostic components of the `LandModel`.
 """
 function ClimaLand.land_components(land::LM) where {LM <: LandModel}
-    isnothing(land.soilco2) ? (:soil, :snow, :canopy) :
-    (:soil, :snow, :soilco2, :canopy)
+    possible_land_components = Tuple(sort([propertynames(land)...])) # in Julia1.10, Tuples are immutable so we cannot sort the tuple returned by propertynames directly
```
kmdeck: Always alphabetical now

PR #1678, opened 2026-03-17
rm SciMLBase; use new CTS interface

juliasloan25:
## Purpose 
Update to use new ClimaTimeSteppers interface implemented in https://github.com/CliMA/ClimaTimeSteppers.jl/tree/ts/cts-refactor

## Content
- [x] remove SciMLBase dependency, use ClimaTimeSteppers throughout
- [x] add a function `convert_cb` so we can accommodate `ClimaDiagnostics.DiagnosticsCallback`s, which are still `SciMLBase.DiscreteCallback`s (this function will be removed after we make a release of ClimaDiagnostics switching to use `CTS.DiscreteCallback`)
- [x] bump the minimum allowed Thermodynamics version see [this PR](https://github.com/CliMA/Thermodynamics.jl/pull/322) for details why

## Compat changes
It looks like with SciMLBase (and its compat) removed there are more possible combinations of packages versions, so the downgrade tests are having trouble. To fix this I added the `forcedeps` option to the downgrade test, which more closely replicates the v1 behavior (see [this discussion](https://github.com/julia-actions/julia-downgrade-compat/issues/22)). I also had to update the compat entries for some packages (and made the Thermodynamics backport release linked above)

----


.buildkite/Manifest-v1.12.toml
```diff
@@ -528,14 +507,26 @@ uuid = "5c42b081-d73a-476f-9059-fd94b934656c"
...
+    ClimaTimeSteppersDiffEqExt = ["DiffEqBase", "SciMLBase"]
+    ClimaTimeSteppersSciMLExt = "SciMLBase"
+
+    [deps.ClimaTimeSteppers.weakdeps]
+    BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+    CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+    DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
+    OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
+    PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
+    SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
```
tapios: Same: shouldn't need this.

ph-kev (approved):

PR #1679, opened 2026-03-17
Removes type restrictions for CTS compatibility

tapios:
## Purpose 

* Removes unnecessary type restrictions that prevent compatibility with ClimaTimeStepper refactor (and its SciML wrappers). 
* Bumps patch release

## Content

----


tapios:
> I also have this PR which drops SciMLBase altogether and switches to use CTS only #1678. But it makes sense to merge this first so the downstream tests in CTS can pass

Oh, that's great!

PR #1680, opened 2026-03-17
Support fractional C3/C4 for PModel

ph-kev:
This PR adds support for fractional C3 and C4 for the ~~`FarquharModel`~~ `PModel`.

The name `is_c3` is changed to `fractional_c3` to better reflect this. Support for fractional C3/C4 for the `PModel` will be done in ~~another~~ this PR.


ph-kev:
I think there is still some more work to do with removing code duplication and potentially adding another test for the `PModel`, but I think it is good for an initial review.

PR #1700, opened 2026-04-02
GPU vs CPU differences in snow

kmdeck:
## Purpose 
Previously we saw that the GPU and CPU col de porte simulations looked very similar in all outputs except the phase change flux. See linked issue. This flux converts ice to liquid and vice versa in the snow, but does not change the overall snow water content.

On GPU, the flux cumulatively grew to large (positive) numbers. On CPU we saw it cumulatively grow to negative numbers, but with a much smaller magnitude overall. For context, a positive number corresponds to refreezing, so this indicates that there was much more proposed "refreezing" on the GPU. 

The if statement for refreezing compared q_l to 0. I noticed on a run between CPU and GPU that q_l could be e.g. 1e-40 and the snow would "refreeze", but on CPU this was q_l=0. Not sure if 1e-40 is considered "roundoff" or why errors like this might be more frequent on GPU.

I think that roundoff differences in the flux near q_l = 0 are only "felt" if they are positive, because we clip S_l (or the liquid mass fraction q_l) to be positive - so roundoff errors leading to q_l< 0 are masked, but q_l>0 are not. And somehow on GPU this roundoff or whatever occurs more often, leading to more refreezing. Since the flux is clipped in the tendency, the downstream behavior looks the same in the prognostic state Y between the two.

This PR "fixes" this by using a larger threshold on q_l to indicate refreezing..

@imreddyTeja does this make sense?
## Content

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -51,7 +51,7 @@ steps:
 
       - label: "Snow Col de Porte"
         command: "julia --color=yes --project=.buildkite experiments/standalone/Snow/snowmip_simulation.jl cdp"
-        artifact_paths: "experiments/standalone/Snow/cpu/cdp/output_active/*png"
+        artifact_paths: "experiments/standalone/Snow/cpu/cdp_default_gradtemp/output_active/*png"
 
       - label: "Varying LAI, no stem compartment"
         command: "julia --color=yes --project=.buildkite experiments/standalone/Vegetation/varying_lai.jl"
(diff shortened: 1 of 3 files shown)


imreddyTeja (approved):
Documenting what I found to be the difference between cpu and gpu here:

The gpu compiler uses FMA more aggressively than the normal compiler. FMA should compute results more accurately than a non fused multiply and add. This results in the explicit `S_l` update in ClimaTimeSteppers having different results on gpu and cpu. When I ran snowmip with FMA in ClimaTimeSteppers on cpu, the results and plots were similar to the gpu results

kmdeck:
closed by #1868

PR #1705, opened 2026-04-06
calibrate gpp

AlexisRenchon:
## Summary

Adds GPP calibration (against ILAMB FLUXCOM) to the global calibration pipeline and restructures `experiments/calibration/` so different calibration experiments can be selected without editing the driver script.

### What changed

- **Per-experiment config files** under `experiments/calibration/configs/`:
  - `energy_fluxes.jl` — prior default (1 param, `lwu`/`shf`/`lhf`)
  - `gpp.jl` — 5-param P-model + moisture-stress calibration vs `gpp`
  - `gpp_lhf.jl` — same 5 params vs `gpp` + `lhf`
  - `test.jl` — single-iteration smoke test (loaded when `TEST_CALIBRATION` is set)
  
  Each config defines `CALIBRATE_CONFIG`, `NOISE_SCALARS`, and `get_calibration_prior()`. Select one at runtime via `CALIBRATION_CONFIG=gpp_lhf.jl bash run_calibration.sh`.

- **Per-variable noise scalars**: `generate_observations.jl` now builds a separate `EKP.Observation` per variable with its own `ScalarCovariance`, then combines them block-diagonally. This lets GPP (scalar 2, g C m⁻² day⁻¹) and LHF (scalar 20, W m⁻²) coexist with appropriate weighting.

- **ILAMB support**: `make_era5_observation_vector` → `make_observation_vector`; new `get_calibration_obs_var_dict` in `ext/land_sim_vis/leaderboard/data_sources.jl` merges ERA5 energy fluxes with ILAMB GPP. Preprocessing now derives the windowing date range from the data itself (instead of the hardcoded 1979–2024 ERA5 range) and propagates NaNs through resampling rather than erroring. `analyze_iteration` now also computes ILAMB monthly/seasonal leaderboards.

- **Runtime plumbing**:
  - `run_calibration.sh` accepts an optional `OUTPUT_DIR` positional arg (useful for Derecho scratch) and sets `HDF5_USE_FILE_LOCKING=FALSE` to avoid NC_EHDFERR on Lustre.
  - `module_load_string` for both `ClimaGPUBackend` and `DerechoBackend` now reads the `climacommon/YYYY_MM_DD` version from the driver's `LOADEDMODULES`, so forward-model jobs inherit whichever version was loaded before invoking the script (no more hardcoded version in `run_calibration.jl`).
  - Forward-model diagnostics extended to include `gpp` and `et` so leaderboards render.

## Results

Calibration on GPP + LHF (noise scalars 2 and 20, respectively — see `experiments/calibration/configs/gpp_lhf.jl`):

gpp_lhf_ilamb

gpp_lhf_era5

gpp_lhf_errors

**RMSE = 0.9, Bias = 0.1**. For reference, [other ILAMB models](https://www.ilamb.org/land-hist/EcosystemandCarbonCycle/GrossPrimaryProductivity/FLUXCOM/FLUXCOM.html) land around RMSE ≈ 1.8, Bias ≈ 0.5 — so we're comfortably ahead with a simpler parameterisation (P-model rather than PFT-specific). Caveat: we prescribe LAI, and I haven't confirmed whether the ILAMB baselines do.

### DIFF
--- a/experiments/calibration/configs/energy_fluxes.jl
+++ b/experiments/calibration/configs/energy_fluxes.jl
@@ -1,0 +1,41 @@
+# Calibration configuration for a single soil-emissivity parameter against
+# energy fluxes (lwu, shf, lhf). This matches the production default that
+# lived in run_calibration.jl on main prior to the config-split refactor.
+
+"""Noise scalars for the covariance matrix of each observed variable.
+
+Units follow the observational dataset:
+- `lwu`: W m^-2 (ERA5)
+- `shf`: W m^-2 (ERA5)
+- `lhf`: W m^-2 (ERA5)
+"""
+const NOISE_SCALARS = Dict("lwu" => 1.0, "shf" => 1.0, "lhf" => 1.0)
+
+"""
+    get_calibration_prior()
+
+Return the combined prior distribution for the calibration parameters.
+
+Calibrates one parameter: soil emissivity. True value is near 0.96.
+"""
+function get_calibration_prior()
+    priors =
+        [EKP.constrained_gaussian("emissivity_bare_soil", 0.82, 0.12, 0.0, 2.0)]
+    return EKP.combine_distributions(priors)
+end
+
+const CALIBRATE_CONFIG = CalibrateConfig(;
+    short_names = ["lwu", "shf", "lhf"],
+    minibatch_size = 1,
+    n_iterations = 10,
+    sample_date_ranges = [
+        ("$(2000 + 2*i)-12-1", "$(2002 + 2*i)-9-1") for i in 0:9
+    ], # 2000 to 2020
+    extend = Dates.Month(3),
+    spinup = Dates.Month(3),
+    nelements = (180, 360, 15),
+    output_dir = OUTPUT_DIR,
+    rng_seed = 42,
+    obs_vec_filepath = "experiments/calibration/land_observation_vector.jld2",
+    model_type = ClimaLand.LandModel,
+)
--- a/experiments/calibration/configs/gpp.jl
+++ b/experiments/calibration/configs/gpp.jl
@@ -1,0 +1,59 @@
+# Calibration configuration for GPP only
+#
+# This file is included by run_calibration.jl and defines:
+#   - CALIBRATE_CONFIG (CalibrateConfig)
+#   - NOISE_SCALARS (per-variable covariance scaling)
+#   - get_calibration_prior() (combined prior distribution)
+#
+# To use this config, set the CALIBRATION_CONFIG env var when invoking
+# run_calibration.sh, e.g.
+#   CALIBRATION_CONFIG=gpp.jl bash experiments/calibration/run_calibration.sh
+
+"""Noise scalars for the covariance matrix of each observed variable.
+
+Units follow the observational dataset:
+- `gpp`: g C m^-2 day^-1 (ILAMB FLUXCOM)
+"""
+const NOISE_SCALARS = Dict("gpp" => 2.0)
+
+"""
+    get_calibration_prior()
+
+Return the combined prior distribution for the calibration parameters.
+
+Calibrates 5 parameters: 4 P-model + 1 soil moisture stress.
+Ensemble size for TransformUnscented: 5 * 2 + 1 = 11 members.
+
+Note: ϕ0_c3/ϕ0_c4 are not calibrated because temperature_dep_yield = true
+uses the quadratic coefficients (ϕa0, ϕa1, ϕa2) instead.
+"""
+function get_calibration_prior()
+    priors = [
+        EKP.constrained_gaussian("pmodel_cstar", 0.41, 0.05, 0.2, 0.7),
+        EKP.constrained_gaussian("pmodel_β_c3", 146.0, 40.0, 50.0, 300.0),
+        EKP.constrained_gaussian("pmodel_β_c4", 16.222, 5.0, 5.0, 40.0),
+        EKP.constrained_gaussian("pmodel_α", 0.933, 0.02, 0.85, 0.999),
+        EKP.constrained_gaussian("moisture_stress_c", 0.27, 0.15, 0.05, 1.0),
+    ]
+    return EKP.combine_distributions(priors)
+end
+
+const CALIBRATE_CONFIG = CalibrateConfig(;
+    short_names = ["gpp"],
+    minibatch_size = 2,
+    n_iterations = 10,
+    # 10 yearly samples: each covers DJF, MAM, JJA (Dec 1 -> Sep 1)
+    # Ending at Sep 1 avoids DJF over-representation across consecutive
+    # samples that would occur with Dec-to-Dec ranges.
+    # with extend = Month(3), simulation runs through Nov 30 of year+1
+    sample_date_ranges = [
+        ("$(year)-12-1", "$(year+1)-9-1") for year in 2001:2010
+    ],
+    extend = Dates.Month(3),
+    spinup = Dates.Month(3),
+    nelements = (180, 360, 15),
+    output_dir = OUTPUT_DIR,
+    rng_seed = 42,
+    obs_vec_filepath = "experiments/calibration/land_observation_vector.jld2",
+    model_type = ClimaLand.LandModel,
+)
--- a/experiments/calibration/configs/gpp_lhf.jl
+++ b/experiments/calibration/configs/gpp_lhf.jl
@@ -1,0 +1,59 @@
+# Calibration configuration for GPP and LHF
+#
+# This file is included by run_calibration.jl and defines:
+#   - CALIBRATE_CONFIG (CalibrateConfig)
+#   - NOISE_SCALARS (per-variable covariance scaling)
+#   - get_calibration_prior() (combined prior distribution)
+#
+# To create a new calibration experiment, copy this file and adjust the
+# settings below.
+
+"""Noise scalars for the covariance matrix of each observed variable.
+
+Units follow the observational dataset:
+- `gpp`: g C m^-2 day^-1 (ILAMB FLUXCOM)
+- `lhf`: W m^-2 (ERA5)
+"""
+const NOISE_SCALARS = Dict("gpp" => 2.0, "lhf" => 20.0)
+
+"""
+    get_calibration_prior()
+
+Return the combined prior distribution for the calibration parameters.
+
+Calibrates 5 parameters: 4 P-model + 1 soil moisture stress.
+Ensemble size for TransformUnscented: 5 * 2 + 1 = 11 members.
+
+Note: ϕ0_c3/ϕ0_c4 are not calibrated because temperature_dep_yield = true
+uses the quadratic coefficients (ϕa0, ϕa1, ϕa2) instead.
+"""
+function get_calibration_prior()
+    priors = [
+        EKP.constrained_gaussian("pmodel_cstar", 0.41, 0.05, 0.2, 0.7),
+        EKP.constrained_gaussian("pmodel_β_c3", 146.0, 40.0, 50.0, 300.0),
+        EKP.constrained_gaussian("pmodel_β_c4", 16.222, 5.0, 5.0, 40.0),
+        EKP.constrained_gaussian("pmodel_α", 0.933, 0.02, 0.85, 0.999),
+        EKP.constrained_gaussian("moisture_stress_c", 0.27, 0.15, 0.05, 1.0),
+    ]
+    return EKP.combine_distributions(priors)
+end
+
+const CALIBRATE_CONFIG = CalibrateConfig(;
+    short_names = ["gpp", "lhf"],
+    minibatch_size = 2,
+    n_iterations = 10,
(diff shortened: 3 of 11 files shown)


experiments/calibration/models/snowy_land.jl
```diff
@@ -160,7 +160,7 @@ function ClimaCalibrate.forward_model(
     # Need to include "lhf", "shf", "lwu", "swu" because plotting the
     # leaderboard requires these diagnostics
     short_names = CALIBRATE_CONFIG.short_names
-    short_names = unique!([short_names; ["lhf", "shf", "lwu", "swu"]])
+    short_names = unique!([short_names; ["lhf", "shf", "lwu", "swu", "gpp", "et"]])
```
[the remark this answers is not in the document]
AlexisRenchon: the bucket does not have those variables, so I think we shouldn't add them?

experiments/calibration/configs/gpp.jl
```diff
@@ -0,0 +1,71 @@
...
+        n_iterations = 10,
+        # 10 yearly samples: each covers DJF, MAM, JJA (Dec 1 -> Sep 1)
+        # Ending at Sep 1 avoids DJF over-representation across consecutive
+        # samples that would occur with Dec-to-Dec ranges.
+        # with extend = Month(3), simulation runs through Nov 30 of year+1
+        sample_date_ranges = [
+            ("$(year)-12-1", "$(year+1)-9-1") for year in 2001:2010
+        ],
+        extend = Dates.Month(3),
+        spinup = Dates.Year(1),
```
kmdeck: I think we could spinup for only 1 season, and then use 2 years in the minibatch (i think this is what weve done in the calibration runs before)

experiments/calibration/configs/gpp.jl
```diff
@@ -0,0 +1,71 @@
...
+        nelements = (180, 360, 15),
+        output_dir = "/glade/derecho/scratch/arenchon/calibration_gpp",
```
kmdeck: the test run should have a path on clima (or central) - wherever it is run.

experiments/calibration/configs/gpp.jl
```diff
@@ -0,0 +1,71 @@
...
+        EKP.constrained_gaussian("pmodel_cstar", 0.41, 0.05, 0.2, 0.7),
+        EKP.constrained_gaussian("pmodel_β_c3", 146.0, 40.0, 50.0, 300.0),
+        EKP.constrained_gaussian("pmodel_β_c4", 16.222, 5.0, 5.0, 40.0),
+        EKP.constrained_gaussian("pmodel_α", 0.933, 0.02, 0.85, 0.999),
+        EKP.constrained_gaussian("moisture_stress_c", 0.27, 0.15, 0.05, 1.0),
+    ]
+    return EKP.combine_distributions(priors)
+end
+
+if !TEST_CALIBRATION
```
kmdeck: Im not sure we need to test each config - is it enought to test 1 of them?

kmdeck: How does our current test calibration (of lwu with one soil parameter) fit into this new setup?

experiments/calibration/configs/gpp.jl
```diff
@@ -0,0 +1,71 @@
...
+    @info "Using calibration config for test calibration"
+    const CALIBRATE_CONFIG = CalibrateConfig(;
+        short_names = ["gpp"],
+        minibatch_size = 1,
+        n_iterations = 1,
+        sample_date_ranges = [("2007-12-1", "2007-12-1")],
+        extend = Dates.Month(3),
+        spinup = Dates.Month(0),
+        nelements = (180, 360, 15),
+        output_dir = "/glade/derecho/scratch/arenchon/calibration_gpp",
```
kmdeck: should this be provided by command line arg?

experiments/calibration/configs/gpp_lhf.jl
```diff
@@ -0,0 +1,71 @@
...
+# This file is included by run_calibration.jl and defines:
+#   - CALIBRATE_CONFIG (CalibrateConfig)
+#   - NOISE_SCALARS (per-variable covariance scaling)
+#   - get_calibration_prior() (combined prior distribution)
+#
+# To create a new calibration experiment, copy this file and adjust the
+# settings below.
+
+"""Noise scalars for the covariance matrix of each observed variable."""
+const NOISE_SCALARS = Dict("gpp" => 2.0, "lhf" => 20.0)
```
kmdeck, line 17: Perhaps we should give units here? and in the gpp only file

experiments/calibration/configs/gpp_lhf.jl
```diff
@@ -0,0 +1,71 @@
...
+        EKP.constrained_gaussian("pmodel_cstar", 0.41, 0.05, 0.2, 0.7),
+        EKP.constrained_gaussian("pmodel_β_c3", 146.0, 40.0, 50.0, 300.0),
+        EKP.constrained_gaussian("pmodel_β_c4", 16.222, 5.0, 5.0, 40.0),
+        EKP.constrained_gaussian("pmodel_α", 0.933, 0.02, 0.85, 0.999),
+        EKP.constrained_gaussian("moisture_stress_c", 0.27, 0.15, 0.05, 1.0),
+    ]
+    return EKP.combine_distributions(priors)
+end
+
+if !TEST_CALIBRATION
```
kmdeck: same comments re: testing, test path, minibatch size, spinup time, and real calibration run path

experiments/calibration/models/snowy_land.jl
```diff
@@ -153,7 +153,7 @@ function ClimaCalibrate.forward_model(
...
```
kmdeck: Should ET be here? (I think we want LHF (and all the energy fluxes) from ERA5, and just GPP from fluxcomm)

experiments/calibration/run_calibration.jl
```diff
@@ -34,58 +34,40 @@ model_interface = joinpath(
...
-        spinup = Dates.Month(3),
-        nelements = (180, 360, 15),
-        output_dir = "experiments/calibration/land_model",
-        rng_seed = 42,
-        obs_vec_filepath = "experiments/calibration/land_observation_vector.jld2",
-        model_type = ClimaLand.LandModel,
-    )
-else
-    @info "Using calibration config for test calibration"
-    const CALIBRATE_CONFIG = CalibrateConfig(;
```
kmdeck, line 55: should this test calibration be the one we retain as a test (no need to test GPP/GPP+LHF configs also?)

experiments/calibration/run_calibration.jl
```diff
@@ -34,58 +34,40 @@ model_interface = joinpath(
...
-if !TEST_CALIBRATION
-    const CALIBRATE_CONFIG = CalibrateConfig(;
-        short_names = ["lwu", "shf", "lhf"],
-        minibatch_size = 1,
-        n_iterations = 10,
-        sample_date_ranges = [
-            ("$(2000 + 2*i)-12-1", "$(2002 + 2*i)-9-1") for i in 0:9
-        ], # 2000 to 2020
-        extend = Dates.Month(3),
-        spinup = Dates.Month(3),
```
kmdeck, line 46: ^^ this spinup and extend was what I was thinking of. Also - the minibatch size is 1 and not 4.

experiments/calibration/models/snowy_land.jl
AlexisRenchon: this is to get the ILAMB plot which needs ET

experiments/calibration/run_calibration.sh
```diff
@@ -2,7 +2,12 @@
 
 # See the README.md for more information on how to run a calibration
 
-module load climacommon
+module load climacommon/2025_02_25
```
kmdeck: Should this be looked up and obtained from the ClimaCalibrate.module_load_string(::ClimaCalibrate.DerechoBackend) function?

experiments/calibration/configs/gpp.jl
ph-kev: This should be made into another config file and in `run_calibration.jl`, you can check if the environment variable `TEST_CALIBRATION` is defined and include this config if it is.

experiments/calibration/generate_observations.jl
```diff
@@ -22,120 +22,132 @@ include(
...
-    # TODO: The start_date argument can be removed by making data_sources.jl not
-    # dependent on a start date.
-
     # The start date doesn't matter since we never resample along the
     # time dimension, so we grab the first date in sample_date_ranges
     start_date = first(first(sample_date_ranges))
-    era5_vars = preprocess_era5_vars(short_names, start_date, nelements)
+    obs_vars = preprocess_obs_vars(short_names, start_date, nelements)
+
+    # Build per-variable covariance estimators
```
ph-kev, line 57: I think this is okay for now, but if we ever want to switch to using the `SVDplusDCovariance`, this would need to change.

experiments/calibration/generate_observations.jl
```diff
@@ -22,120 +22,132 @@ include(
...
+    # interpolating good observations with zeros. Some valid points near
+    # NaN regions may be lost, but this is preferred over corrupting them.
+
+    # Window to ensure that each season contains all three months.
+    # Use the data's own time range, clamped to full seasons.
+    times = ClimaAnalysis.times(var)
+    first_time = first(times)
+    last_time = last(times)
+    t_min = Dates.DateTime(Dates.year(first_time), 3)
+    t_max = Dates.DateTime(Dates.year(last_time), 8)
```
ph-kev: Can you check if this works? I'm surprised it does. I think it should be

 
```suggestion
    var_dates = ClimaAnalysis.dates(var)
    first_date = first(var_dates)
    last_date = last(var_dates)
    date_min = Dates.DateTime(Dates.year(first_date), 3)
    date_max = Dates.DateTime(Dates.year(last_date), 8)
```

experiments/calibration/run_calibration.sh
ph-kev: If you want to have support for automatically finding the right climacommon version, I think it is easier to use `get(ENV, "LOADEDMODULES", "")` and look for the climacommon version that was loaded.

experiments/calibration/configs/gpp.jl
AlexisRenchon: that's a good idea

experiments/calibration/configs/gpp.jl
AlexisRenchon: we could make a config for it, and include it by default

experiments/calibration/configs/gpp.jl
```diff
@@ -0,0 +1,58 @@
+# Calibration configuration for GPP only
+#
+# This file is included by run_calibration.jl and defines:
+#   - CALIBRATE_CONFIG (CalibrateConfig)
+#   - NOISE_SCALARS (per-variable covariance scaling)
+#   - get_calibration_prior() (combined prior distribution)
+#
+# To use this config, change the include path in run_calibration.jl to
+# point to this file instead of gpp_lhf.jl.
```
ph-kev: Can you also include information about how to change the config being ran in `run_calibration.sh`?

experiments/calibration/run_calibration.jl
```diff
@@ -34,58 +34,65 @@ model_interface = joinpath(
...
+Falls back to `"climacommon"` (default version) if none is found.
+
+This lets forward-model jobs pick up whichever `climacommon` the driver script
+loaded, instead of hardcoding a version per backend.
+"""
+function _loaded_climacommon()
+    loaded = get(ENV, "LOADEDMODULES", "")
+    for mod in split(loaded, ':')
+        startswith(mod, "climacommon/") && return String(mod)
+    end
```
ph-kev: If you load multiple versions of `climacommon`, this can load the wrong one. You should use the last `climacommon` version found in the string.

It would also be good to sanitize it and check that it is of the form `climacommon/YYYY_MM_DD`.

ph-kev (approved):
Can you run the test calibration? I already added the label so I think it should just work if you push again.

experiments/calibration/configs/energy_fluxes.jl
```diff
@@ -0,0 +1,41 @@
...
+# lived in run_calibration.jl on main prior to the config-split refactor.
+
+"""Noise scalars for the covariance matrix of each observed variable.
+
+Units follow the observational dataset:
+- `lwu`: W m^-2 (ERA5)
+- `shf`: W m^-2 (ERA5)
+- `lhf`: W m^-2 (ERA5)
+"""
+const NOISE_SCALARS = Dict("lwu" => 1.0, "shf" => 1.0, "lhf" => 1.0)
```
kmdeck, line 12: maybe could rename file to "energy_fluxes"?

kmdeck (approved):
thank you!

AlexisRenchon:
@kmdeck @ph-kev thank you both! 

I'll add your recommended change and run the test, and also retry gpp_lhf on derecho, see that it works as intended, and that results are okay with 3 months spinup and 1 minibatch. I won't merge before all that seems good.

PR #1707, opened 2026-04-06
Add max LAI artifact and use this to determine SAI/RAI

kmdeck:
## Purpose 
Explore having nonzero SAI, spatially varying RAI

We should not add this purely to make the model for realistic. It should be to address a shortcoming of the model that is important.
For example:
1. plants have thermal mass from trunks/stems, but we dont model it now [I do not think this is important, because plants reach equilbrium/steady state very quickly, so the extra thermal mass "doesnt matter"]. The same reason why most models have canopy temperature diagnostic and not prognostic.
2. trunks/stems can emit long wave radiation and sensible heat [this could matter in the winter when some trees have lost their leaves?]
3. trunks/stems can absorb some SW radiation. [this seems hard to model and we would need to keep track of what the leaves absorb on their own as well to get photosynthesis correct]
4. Larger RAI in principle would suggest more water extraction from the soil, but the bottleneck is transpiration, which is dictated by leaves. so larger RAI at the same LAI wouldnt lead to more extraction of water from soil.
5. Both RAI and SAI matter for autotrophic respiration

## Content

----

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -1,3 +1,10 @@
+[modis_max_lai]
+git-tree-sha1 = "3977938f17914088ff69224a3c74e4526cc91bb1"
+
+    [[modis_max_lai.download]]
+    sha256 = "b128f8f5a6f7be5ef68ddfb6f03080dd2daf9f787ef2aaa265bca3c016f541d1"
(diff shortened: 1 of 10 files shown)


AlexisRenchon (approved):

kmdeck:
Added as part of https://github.com/CliMA/ClimaLand.jl/pull/1767

PR #1709, opened 2026-04-07
Refactor plant hydraulics

kmdeck:
## Purpose 
- Make single compartment. We never use the multi-compartment scheme. In fact we never use the plant hydraulics currently! But even if we did, the single compartment scheme is more in line with the big leaf approach we are taking. The main limitation now is that we cannot specify different root/stem/leaf properties, and simply model the plant hydraulics system as a single entity.
- Put plant hydraulics on the same footing as every other canopy component - not a module of its own
- Making these changes + getting the height from the biomass component removes the question/problem about having multiple heights stored in the Canopy model, so that is a nice bonus.

## Content

----


src/standalone/Vegetation/plant_hydraulics.jl
```diff
@@ -194,59 +137,30 @@ prognostic_vars(model::PlantHydraulicsModel) = (:ϑ_l,)
 
 A function which returns the names of the auxiliary
 variables of the `PlantHydraulicsModel`,
-the water potential `ψ` (m), the volume flux\\*\\cross section `fa` (1/s),
-and the volume flux\\*root cross section in the roots `fa_roots` (1/s),
+the water potential `ψ` (m) and the 
+volume flux\\*root cross section in the roots `fa_roots` (1/s),
```
AlexisRenchon: 1/s or m/s?

AlexisRenchon (approved):
Actually made a small PR with suggested changes, mostly typos, have a look!

PR #1714, opened 2026-04-10
Respiration update

AlexisRenchon:
## Summary

Overhauls the soil biogeochemistry / ecosystem-respiration pathway: fixes NaNs in global SoilCO2 runs, switches to Henry's-law air–water partitioning, initialize SOC with SoilGrids data, recalibrates against GPP + ER + NEE + LHF, and adds ER and NEE to the leaderboard.

## Highlights

- **No more NaNs** — `O2_f` clamped to [0, 0.21] each step + CFL-aware tendency limiter when `Δt > 0`. `Δt` is now a required positional arg on `SoilCO2Model{FT}` so the limiter can't be silently skipped.
- **Henry's-law air–water partitioning** (Sander 2015) for CO2 and O2: `θ_eff = max(θ_a + β·θ_l, θ_eff,min)` stays bounded as the soil saturates.
- **DAMM Arrhenius in centered form** — `V_ref` and `Ea` are orthogonal under calibration; T response reduces to a slope (~Q10) + intercept.
- **SOC from SoilGrids** instead of uniform constant; pool held fixed in time but kept prognostic for future dynamic SOC.
- **CO2 IC** in equilibrium with atmospheric `c_co2` via Henry's law; **O2 IC** exponential profile (0.21 surface → 0.15 at -1 m).
- **New calibration config** `gpp_er_nee_lhf.jl` (11 parameters). Defaults in `toml/default_parameters.toml` updated to the latest 8th-iteration posterior mean.
- **Leaderboard** now includes ILAMB FLUXCOM ER and NEE.
- **[Updated docs](https://clima.github.io/ClimaLand.jl/previews/PR1714/standalone/pages/soil/biogeochemistry/DAMM_model/)**.

## Next PR

Calibrate against a NEE dataset reproducing atmospheric CO2 (inversion dataset?), with a loss capturing (1) seasonality, (2) inter-annual variability, and (3) the decadal trend (land sink keeping up with rising emissions). Relies on human emissions and ocean sinks being known.

---

cfluxes
er_energy_2
cfluxes_errors

### DIFF
--- a/docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
+++ b/docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
@@ -1,29 +1,46 @@
 # Soil Biogeochemistry: CO₂ Production and O₂ Consumption
 
-This section describes the coupled soil biogeochemistry model implemented in ClimaLand, which simulates the production and diffusion of CO₂, consumption of O₂, and decomposition of soil organic carbon (SOC) through microbial respiration. The model combines the Dual Arrhenius and Michaelis-Menten (DAMM) kinetics framework for microbial respiration with gas diffusion equations that account for soil structure, temperature, and moisture dependencies.
+This section describes the coupled soil biogeochemistry model implemented in ClimaLand, which simulates the production and diffusion of CO₂, consumption of O₂, and microbial respiration sourced by a fixed pool of soil organic carbon ($Y_{C_{som}}$). The model combines the Dual Arrhenius and Michaelis-Menten (DAMM) kinetics framework for microbial respiration with gas diffusion equations that account for soil structure, temperature, and moisture, and with Henry's-law partitioning of dissolved CO₂ and O₂ between the gas and aqueous phases of the pore space.
 
 ## Model Overview
 
-The biogeochemistry model tracks three prognostic variables that evolve in space (with soil depth) and time:
-- CO₂ concentration in soil air ($ρ_{CO_2}$, kg C m⁻³)
-- Volumetric O₂ fraction in soil air ($O_{2f}$, dimensionless)
-- Soil organic carbon ($C_{som}$, kg C m⁻³)
+The biogeochemistry model carries three prognostic fields that evolve in space (with soil depth) and time. The variables, listed in the units used internally, are:
 
-The model captures the fundamental processes of aerobic decomposition: microbes consume soil organic carbon and oxygen to produce carbon dioxide. These gases diffuse through the soil pore space, with diffusion rates controlled by soil structure (porosity, tortuosity), moisture content, and temperature.
+- the total CO₂ stored per unit soil volume (gas phase plus dissolved), $Y_{CO_2}$ (kg C m⁻³ soil), with $Y_{CO_2} = \theta_{\rm eff}\, c_g$ where $c_g$ is the air-phase carbon mass concentration (kg C m⁻³ air) and $\theta_{\rm eff}$ is the effective porosity, defined as the sum of the gas-filled pore volume and the dissolved-equivalent volume of the liquid pore water (see Effective Porosity below);
+- the volumetric O₂ fraction in soil air, $Y_{\theta_{O_2}}$ (dimensionless);
+- the soil organic carbon density, $Y_{C_{som}}$ (kg C m⁻³), held fixed at its initial value (no tendency, see below).
+
+The model captures the fundamental processes of aerobic decomposition: microbes consume soluble carbon and oxygen to produce carbon dioxide. CO₂ and O₂ diffuse through the soil pore space, with diffusion rates controlled by soil structure (porosity, tortuosity), moisture, temperature, and pressure.
+
+$Y_{C_{som}}$ is initialized from the SoilGrids organic carbon density product and is treated as a stationary spatial parameter: we assume that the input of organic matter (e.g., from litter and root turnover) is equal to the rate of decomposition, so the actual $Y_{C_{som}}$ is constant in time. The model thus captures geographic heterogeneity of carbon stocks but does not integrate carbon depletion on simulation timescales — a deliberate simplification appropriate for the multi-decade runs targeted by ClimaLand. In the future, we may model the rate of decomposition explicitly and track $Y_{C_{som}}$ over time; expressing $Y_{C_{som}}$ as a prognostic variable now allows that change to be made easily.
 
 ## Governing Equations
 
 ### CO₂ Transport and Production
 
-The evolution of CO₂ concentration in the soil follows a reaction-diffusion equation:
+The prognostic variable $Y_{CO_2}$ is the **total** carbon mass stored per unit soil volume, summing the gas phase (occupying the air-filled pore volume $\theta_a$ at concentration $c_g$) and the dissolved phase (held in the liquid pore water $\theta_l$ at concentration $c_{aq}$, related to $c_g$ by Henry's law as $c_{aq} = \beta\, c_g$). The total is therefore
 
 ```math
 \begin{equation}
-\frac{\partial ρ_{CO_2}}{\partial t} = -\nabla \cdot \left[-D \nabla ρ_{CO_2}\right] + S_m
+Y_{CO_2} = \theta_a\, c_g + \theta_l\, c_{aq} = (\theta_a + \beta\, \theta_l)\, c_g \equiv \theta_{\rm eff}\, c_g,
 \end{equation}
 ```
 
-where $ρ_{CO_2}$ is the CO₂ concentration (kg C m⁻³), $D$ is the effective diffusivity of CO₂ in soil (m² s⁻¹), and $S_m$ is the microbial CO₂ production rate (kg C m⁻³ s⁻¹). The diffusive flux is $-D\nabla ρ_{CO_2}$, directed from regions of high to low concentration.
+where $\theta_{\rm eff} \equiv \theta_a + \beta\, \theta_l$ is the **effective porosity** — the equivalent volume that would hold $Y_{CO_2}$ at concentration $c_g$ if all the carbon were in the gas phase. The gas-phase concentration is recovered as $c_g = Y_{CO_2} / \theta_{\rm eff}$. Storing $Y_{CO_2}$ rather than $c_g$ as the prognostic keeps the formulation well-behaved as the soil saturates ($\theta_a \to 0$): dissolved storage keeps $\theta_{\rm eff} \geq \beta\, \theta_l > 0$ as long as there is any liquid water. Diffusion is driven by gas-phase gradients:
+
+```math
+\begin{equation}
+\frac{\partial Y_{CO_2}}{\partial t} = \nabla \cdot \left[D \, \nabla c_g\right] + S_m,
+\qquad c_g = Y_{CO_2} / \theta_{\rm eff}
+\end{equation}
+```
+
+where:
+
+- the effective diffusivity of CO₂ in soil, $D$ (m² s⁻¹);
+- the microbial CO₂ production rate, $S_m$ (kg C m⁻³ s⁻¹).
+
+Here $\theta_a = \nu - \theta_l - \theta_i$ is the volumetric air content, i.e. the fraction of the soil volume occupied by the air phase. Note that we do not currently model fluxes of dissolved CO₂ due to fluxes of the soil water; only gas-phase diffusion is represented.
 
 ### O₂ Transport and Consumption
 
@@ -31,77 +48,91 @@ Oxygen transport and consumption in soil is described by:
 
 ```math
 \begin{equation}
-\frac{\partial O_{2f}}{\partial t} = \frac{R T}{\theta_a P M_{O_2}} \nabla \cdot \left[D_{O_2} \theta_a \nabla \rho_{O_2}\right] - \frac{R T}{M_C \theta_a P} S_m
+\frac{\partial Y_{\theta_{O_2}}}{\partial t}
+= \frac{R\, T}{\theta_{\rm eff,O_2}\, P\, M_{O_2}}\,
+\nabla \cdot \left[D_{O_2}\, \nabla c_{O_2}\right]
+- \frac{R\, T}{\theta_{\rm eff,O_2}\, P\, M_{C}}\, S_m
 \end{equation}
 ```
 
 where:
-- $O_{2f}$ is the volumetric O₂ fraction in air (dimensionless, ~0.21 in atmosphere)
-- $\theta_a$ is the volumetric air content (m³ air m⁻³ soil)
-- $\rho_{O_2}$ is the O₂ mass concentration in air (kg O₂ m⁻³ air)
-- $D_{O_2}$ is the effective diffusivity of O₂ in soil (m² s⁻¹)
-- $R$ is the universal gas constant (8.314 J mol⁻¹ K⁻¹)
-- $T$ is soil temperature (K)
-- $P$ is atmospheric pressure (Pa)
-- $M_{O_2}$ is the molar mass of O₂ (0.032 kg mol⁻¹)
-- $M_C$ is the molar mass of carbon (0.012 kg mol⁻¹)
 
-The second term represents O₂ consumption by microbial respiration, following the stoichiometry C + O₂ → CO₂, where each 12 g of carbon respired consumes 32 g of oxygen.
+- the effective porosity for O₂, $\theta_{\rm eff,O_2} = \theta_a + \beta_{O_2}\, \theta_l$ (m³ m⁻³ soil) — same construction as the CO₂ effective porosity defined above, but with the O₂-specific Henry's-law factor. Because O₂ is much less soluble than CO₂ ($\beta_{O_2} \approx 0.032$ vs $\beta_{CO_2} \approx 0.84$ at 298 K), the dissolved-equivalent term contributes much less for O₂;
+- the O₂ mass concentration in air, $c_{O_2}$ (kg O₂ m⁻³ air), recovered from $Y_{\theta_{O_2}}$ via the ideal gas law;
+- the effective diffusivity of O₂ in soil, $D_{O_2}$ (m² s⁻¹);
+- the universal gas constant, $R$ (8.314 J mol⁻¹ K⁻¹);
+- the soil temperature, $T$ (K);
+- the atmospheric surface pressure, $P$ (Pa);
+- the molar mass of O₂, $M_{O_2}$ (0.032 kg mol⁻¹);
+- the molar mass of carbon, $M_C$ (0.012 kg mol⁻¹).
 
-### Soil Organic Carbon Dynamics
+The second term represents O₂ consumption by microbial respiration. The stoichiometry follows C + O₂ → CO₂: each 12 g of carbon respired consumes 32 g of O₂. The conversion factor $R T / (\theta_{\rm eff,O_2}\, P\, M_C)$ translates the carbon mass source $S_m$ (kg C m⁻³ s⁻¹ soil) into a tendency on the volumetric fraction $Y_{\theta_{O_2}}$.
 
-The soil organic carbon pool is depleted by microbial decomposition:
+### Soil Organic Carbon
+
+The $Y_{C_{som}}$ pool is held fixed at its initial value:
 
 ```math
 \begin{equation}
-\frac{\partial C_{som}}{\partial t} = -S_m
+\frac{\partial Y_{C_{som}}}{\partial t} = 0
 \end{equation}
 ```
 
-where $C_{som}$ is the soil organic carbon content (kg C m⁻³). This equation enforces carbon mass conservation: the rate of SOC consumption equals the rate of CO₂ production.
+The pool $Y_{C_{som}}$ enters the DAMM substrate kinetics (below) but is not depleted by respiration. Its value is set from the SoilGrids organic carbon density at the start of the simulation, providing the spatial structure of carbon availability.
 
 ## Microbial Respiration: DAMM Model
 
-The microbial source term $S_m$ represents heterotrophic respiration and is computed using the Dual Arrhenius and Michaelis-Menten (DAMM) kinetics model [Davidson2012](@citet):
+The microbial source term $S_m$ is computed using the Dual Arrhenius and Michaelis-Menten (DAMM) kinetics model [Davidson2012](@citet):
 
 ```math
 \begin{equation}
-S_m = V_{max} \cdot MM_{sx} \cdot MM_{O_2}
-\end{equation}
-```
-
-where $V_{max}$ is the maximum potential rate of respiration (temperature-dependent), $MM_{sx}$ represents substrate availability (0-1, dimensionless), and $MM_{O_2}$ is the oxygen limitation factor (0-1, dimensionless).
-
-### Maximum Respiration Rate
-
-The maximum potential respiration rate follows Arrhenius kinetics:
-
-```math
-\begin{equation}
-V_{max} = \alpha_{sx} \exp\left(\frac{-Ea_{sx}}{RT}\right)
-\end{equation}
-```
-
-where $\alpha_{sx}$ is the pre-exponential factor (kg C m⁻³ s⁻¹), $Ea_{sx}$ is the activation energy (J mol⁻¹), $R$ is the gas constant, and $T$ is soil temperature (K). This formulation captures the exponential increase in microbial activity with temperature.
-
-### Substrate Availability
-
-The concentration of soluble carbon substrates available to microbes depends on diffusion through soil water films:
-
-```math
-\begin{equation}
-[S_x] = p_{sx} \cdot C_{som} \cdot D_{liq} \cdot \theta_l^3
+S_m = V_{\max} \cdot MM_{sx} \cdot MM_{O_2}
 \end{equation}
 ```
 
 where:
-- $[S_x]$ is the concentration of soluble substrate (kg C m⁻³)
-- $p_{sx}$ is the fraction of SOC that is soluble (dimensionless)
-- $C_{som}$ is the total soil organic carbon (kg C m⁻³)
-- $D_{liq}$ is the diffusion coefficient of soluble carbon (dimensionless)
-- $\theta_l$ is the volumetric liquid water content (m³ m⁻³)
 
-The cubic dependence on moisture ($\theta_l^3$) reflects the strong constraint that water films place on substrate diffusion to microbes.
+- the maximum potential rate of respiration (temperature-dependent), $V_{\max}$ (kg C m⁻³ s⁻¹);
+- the substrate availability factor, $MM_{sx}$ (dimensionless, 0–1);
+- the oxygen limitation factor, $MM_{O_2}$ (dimensionless, 0–1).
+
+### Maximum Respiration Rate
+
+The maximum respiration rate follows Arrhenius kinetics, written here in centered form:
+
+```math
+\begin{equation}
+V_{\max} = V_{{\rm ref},sx} \, \exp\!\left[-\frac{E_{a,sx}}{R}\!\left(\frac{1}{T} - \frac{1}{T_{{\rm ref},sx}}\right)\right]
+\end{equation}
+```
+
+where:
+
+- the maximum respiration rate at the reference temperature, $V_{{\rm ref},sx}$ (kg C m⁻³ s⁻¹);
+- the reference temperature for the centered Arrhenius form, $T_{{\rm ref},sx}$ (K);
+- the activation energy, $E_{a,sx}$ (J mol⁻¹).
+
+This is algebraically equivalent to the uncentered form $V_{\max} = \alpha_{sx}\, \exp[-E_{a,sx}/(RT)]$ with $\alpha_{sx} = V_{{\rm ref},sx}\, \exp[E_{a,sx}/(R\, T_{{\rm ref},sx})]$. The centered form is preferred for calibration because $V_{{\rm ref},sx}$ (the rate at $T_{{\rm ref},sx}$) and $E_{a,sx}$ (the temperature sensitivity) are nearly orthogonal under parameter estimation, while $\alpha_{sx}$ and $E_{a,sx}$ are strongly correlated.
+
+### Substrate Availability
+
+The concentration of soluble carbon substrate available to microbes depends on diffusion through soil water films:
+
+```math
+\begin{equation}
+[S_x] = p_{sx} \cdot Y_{C_{som}} \cdot D_{liq} \cdot \theta_l^{3}
+\end{equation}
+```
+
+where:
+
+- the concentration of soluble substrate, $[S_x]$ (kg C m⁻³);
+- the soluble fraction of $Y_{C_{som}}$, $p_{sx}$ (dimensionless);
+- the soil organic carbon density, $Y_{C_{som}}$ (kg C m⁻³);
+- the dimensionless diffusion coefficient of soluble carbon in water, $D_{liq}$;
+- the volumetric liquid water content, $\theta_l$ (m³ m⁻³).
+
+The cubic dependence on moisture reflects the strong constraint that water films place on substrate diffusion to microbial sites.
 
 The substrate limitation factor follows Michaelis-Menten kinetics:
 
@@ -111,23 +142,24 @@ MM_{sx} = \frac{[S_x]}{kM_{sx} + [S_x]}
 \end{equation}
 ```
 
-where $kM_{sx}$ is the Michaelis constant for substrate (kg C m⁻³).
+where $kM_{sx}$ is the substrate Michaelis constant (kg C m⁻³).
 
 ### Oxygen Availability
 
-The oxygen availability for microbial respiration accounts for diffusion limitations in porous media using a tortuosity model:
+The oxygen availability for microbial respiration accounts for diffusion limitations in the partially saturated pore space using a Millington–Quirk tortuosity model:
 
 ```math
 \begin{equation}
-O_{2,avail} = D_{oa} \cdot O_{2f} \cdot \theta_a^{4/3}
+O_{2,{\rm avail}} = D_{oa} \cdot Y_{\theta_{O_2}} \cdot \theta_a^{4/3}
 \end{equation}
 ```
 
 where:
-- $O_{2,avail}$ is the dimensionless O₂ availability metric
-- $D_{oa}$ is the oxygen diffusion coefficient in air (dimensionless)
-- $O_{2f}$ is the volumetric O₂ fraction in air
-- $\theta_a^{4/3}$ is the Millington-Quirk tortuosity factor
+
+- the dimensionless O₂ availability metric, $O_{2,{\rm avail}}$;
+- the dimensionless O₂ diffusion coefficient used in the availability scaling, $D_{oa}$;
+- the volumetric O₂ fraction in air, $Y_{\theta_{O_2}}$ (dimensionless);
+- the Millington–Quirk tortuosity factor, $\theta_a^{4/3}$.
 
 The exponent 4/3 captures how tortuous diffusion pathways limit gas transport in partially saturated porous media.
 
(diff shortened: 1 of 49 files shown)


experiments/calibration/configs/er.jl
```diff
@@ -0,0 +1,116 @@
...
+#   term in Rpm is gated off by heaviside(SAI), so Rpm = Rd · (β + μr·RAI/LAI)
+#   and reducing μr monotonically reduces Ra. This prevents the optimizer
+#   from compensating a shrunk Hr with an inflated Ra. If SAI is ever turned
+#   on (via the SAI kwarg of PrescribedBiomassModel or the SAI entry in
+#   toml_dict), this guarantee holds only in expectation, not strictly.
+
+"""Noise scalars for the covariance matrix of each observed variable.
+
+Units follow the observational dataset:
+- `er`: g C m^-2 day^-1 (ILAMB FLUXCOM reco)
```
kmdeck: this is just a note - I believe we pass in a variance, not a standard deviation, in which case the units of these (and all) would be ()^2. but not 100% sure.

experiments/calibration/configs/er.jl
```diff
@@ -0,0 +1,116 @@
...
+#   - get_calibration_prior() (combined prior distribution)
+#
+# To use this config, set the CALIBRATION_CONFIG env var when invoking
+# run_calibration.sh, e.g.
+#   CALIBRATION_CONFIG=er.jl bash experiments/calibration/run_calibration.sh
+#
+# Observational target: ILAMB FLUXCOM reco (reco_FLUXCOM_reco.nc), in g C m^-2 day^-1.
+#
+# Autotrophic respiration one-sided constraint:
+#   root_leaf_nitrogen_ratio (μr) and relative_contribution_factor (Rel) have
```
kmdeck: what should we do so that this is modeled correctly and we don't need this big caveat? :)

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -1,3 +1,52 @@
...
+
+Build a `more_kwargs` dictionary for `ClimaAnalysis.Visualize.contour2D_on_globe!`
+that clips the colormap to `[quantile(data, low_q), quantile(data, high_q)]`
+and lets `Makie.contourf!` draw extend-arrows for values outside that range.
+
+Returns an empty `Dict` (i.e. fall back to the plotting defaults) when the
+field is all-NaN, has fewer than two finite samples, or has a degenerate
+percentile range.
+"""
+function _percentile_contour_kwargs(var, low_q, high_q; nlevels = 11)
```
kmdeck: Do you want to make this so 
```
_percentile_contour_kwargs(var; low_q = 0.1, high_q = 0.9, nlevels = 11)
```

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -1,3 +1,52 @@
...
+    vals = filter(isfinite, vec(var.data))
+    length(vals) < 2 && return Dict{Symbol, Any}()
```
kmdeck: FWIW, it looks like this case can be included in the one on line 17:
```
julia> quantile([0,], 0.1)
0.0

julia> quantile([0,], 0.9)
0.0
```

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -1,3 +1,52 @@
...
+"""
+    _nee_diverging_contour_kwargs(var, q; nlevels = 21)
+
+Build `more_kwargs` for `contour2D_on_globe!` that uses a diverging
+green → white → red colormap centered at zero, so NEE ≈ 0 reads as white,
+strong sinks (NEE ≪ 0) as saturated green, and strong sources (NEE ≫ 0)
+as saturated red. The half-range is set to the `q`-th percentile of
+`|NEE|` so single-cell outliers don't dominate.
+"""
+function _nee_diverging_contour_kwargs(var, q; nlevels = 21)
```
kmdeck, line 39: is this specific to NEE for some reason? do you want to adopt a different color map for all the variables?

kmdeck: I also wonder about passing in the color map as a kwarg, and then using the same function for both:
_percentile_contour_kwargs

kmdeck: not sure, the abs. is unique to this!

src/diagnostics/land_compute_methods.jl
```diff
@@ -721,9 +749,18 @@ function compute_total_runoff!(
         out = zeros(soil.domain.space.surface) # Allocates
         fill!(field_values(out), NaN) # fill with NaNs, even over the ocean
     end
-    out .=
-        Runoff.get_surface_runoff(soil.boundary_conditions.top.runoff, Y, p) .+
-        Runoff.get_subsurface_runoff(soil.boundary_conditions.top.runoff, Y, p)
+    runoff = soil.boundary_conditions.top.runoff
```
kmdeck, line 752: was this a problem for sites where subsurface runoff wasn't modeled?

thanks for catching!

src/integrated/land.jl
```diff
@@ -359,7 +360,8 @@ function LandModel{FT}(
             PrognosticMet(soil.parameters),
             forcing.atmos,
         ),
-        toml_dict,
+        toml_dict;
+        Δt,
```
kmdeck, line 366: Maybe we can follow the snow model here - dt is required

kmdeck: (not a kwarg)

src/simulations/initial_conditions.jl
```diff
@@ -97,6 +97,92 @@ function set_soil_initial_conditions!(
     return nothing
 end
 
+"""
+    set_soilco2_initial_conditions!(Y, p, land)
+
+Initialize soil CO2 state to be consistent with atmospheric CO2 and soil state.
+CO2 is stored as carbon mass per soil volume (kg C m⁻³ soil).
```
kmdeck, line 104: "This assumes that the soil moisture state (Y.soil.\vartheta_l and Y.soil.\theta_i) have already been set.

src/simulations/initial_conditions.jl
```diff
@@ -97,6 +97,92 @@ function set_soil_initial_conditions!(
...
+function set_soilco2_initial_conditions!(Y, p, land)
+    FT = eltype(Y.soilco2.CO2)
+    soil = land.soil
+    soilco2 = land.soilco2
+    params = soilco2.parameters
+
+    θ_l = Y.soil.ϑ_l
+    θ_i = Y.soil.θ_i
+    θ_w = θ_l .+ θ_i
+    ν = soil.parameters.ν
```
kmdeck, line 116: Do you want to clip \vartheta_l such that \vartheta_l < \nu - \theta_i?

src/simulations/initial_conditions.jl
```diff
@@ -97,6 +97,92 @@ function set_soil_initial_conditions!(
...
+    T_soil =
+        ClimaLand.Soil.temperature_from_ρe_int.(
+            Y.soil.ρe_int,
+            θ_i,
+            ρc_s,
+            soil.parameters.earth_param_set,
+        )
+
+    θ_a = ClimaLand.Soil.Biogeochemistry.volumetric_air_content.(θ_w, ν)
+    R = FT(ClimaLand.Parameters.gas_constant(params.earth_param_set))
```
kmdeck: This and the line below don't need the FT (they are already the correct float type)

src/simulations/initial_conditions.jl
```diff
@@ -97,6 +97,92 @@ function set_soil_initial_conditions!(
...
+        params.K_H_co2_298,
+        params.dln_K_H_co2_dT,
+        T_soil,
+    )
+    β = @. ClimaLand.Soil.Biogeochemistry.beta_gas(K_H, R, T_soil)
+    θ_eff = @. ClimaLand.Soil.Biogeochemistry.effective_porosity(θ_a, θ_l, β)
+
+    @. Y.soilco2.CO2 =
+        θ_eff * p.drivers.c_co2 * p.drivers.P * M_C / (R * T_soil)
+    # Exponential O2 profile: 0.21 at z=0, 0.15 at z=-1 m (k = ln(0.21/0.15)).
```
kmdeck, line 152: Any citation for this or further explanation?

src/simulations/initial_conditions.jl
```diff
@@ -97,6 +97,92 @@ function set_soil_initial_conditions!(
...
+    z = soil.domain.fields.z
+    k_o2 = FT(log(FT(0.21) / FT(0.15)))
```
kmdeck:
```suggestion
    k_o2 = FT(log(0.21 / 0.15))
```

src/simulations/initial_conditions.jl
```diff
@@ -97,6 +97,92 @@ function set_soil_initial_conditions!(
...
+
+"""
+    set_soilco2_SOC_from_soilgrids!(SOC_field)
+
+Initialize the prognostic soil organic carbon field `SOC_field` (kgC m⁻³)
+using the SoilGrids organic carbon density dataset.
+
+The data is regridded onto the 3D subsurface space of `SOC_field`. Below
+the deepest data level (-1.5 m) and above the shallowest (-0.025 m) values
+are extrapolated flat in depth, matching the convention used for the
```
kmdeck: Did you want to add in exponential decay with depth below 1.5m? I thought Eva found that important.

src/simulations/initial_conditions.jl
```diff
@@ -97,6 +97,92 @@ function set_soil_initial_conditions!(
...
+SoilGrids-based soil composition parameters. Ocean / missing-data grid
+points are zero in the source file and therefore come through as zero.
+Negative values produced by horizontal interpolation near coastlines are
```
kmdeck: Do we get negative values? If ocean values are zero, and land values are positive...

src/simulations/initial_conditions.jl
```diff
@@ -417,6 +498,9 @@ function make_set_initial_state_from_file(
             enforce_constraints,
         )
 
+        # SoilCO2 IC (requires soil state)
```
kmdeck, line 519: Does this need a "        if !isnothing(land.soilco2)
" block?

src/simulations/initial_conditions.jl
```diff
@@ -548,10 +632,15 @@ function make_set_subseasonal_initial_conditions(
...
+        # Set initial conditions that aren't read in from file.
+        # Y.soilco2.CO2 stores total CO₂ as carbon mass per soil volume
+        # (kg C m⁻³ soil) = θ_eff · c_g; the value below is a representative
+        # atmospheric-equilibrium initial guess (~412 ppm at 1 atm, 283 K,
+        # θ_eff ≈ 0.3) and is rapidly replaced by the diffusive boundary
+        # condition on the first time step.
+        Y.soilco2.CO2 .= FT(6e-5)
         Y.soilco2.O2_f .= FT(0.21)    # atmospheric O2 volumetric fraction
-        Y.soilco2.SOC .= FT(5.0)      # default SOC concentration (kg C/m³)
+        set_soilco2_SOC_from_soilgrids!(Y.soilco2.SOC)
```
kmdeck: likewise here 

does this need a         if !isnothing(land.soilco2)
block?

experiments/calibration/configs/er.jl
AlexisRenchon: For the calibration with all C fluxes (GPP, ER, NEE) + LHF, I let this be two-sided (autotrophic can go up or down), and it did okay - I was worried the calibration would set either heterotrophic or autotrophic very small, and compensate with the other... it did not happen

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: why not! I believe I changed it to 0.05 and 0.95 and I prefer it... the max and min area on the map seemed too large at .1 and .9

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: That is specific to NEE, because NEE goes above 0 (net source of C) or below 0 (net sink of C).

src/simulations/initial_conditions.jl
AlexisRenchon: I could add a comment - oxygen decreases with soil depth... I imagine it reaches some steady state quite fast so the IC doesn't matter that much, I did that in one of my attempt to fix instability. I had a look at some profiles of O2 in soil to get an idea of what it should look like in general. of course it depends on soil porosity etc. I think it's a more reasonable (realistic) IC that at atm concentration

src/simulations/initial_conditions.jl
AlexisRenchon: I was about to say maybe this is okay... but looking at my individual sites, like Australia outback, it seems co2 production is still quite high at 1.5 m depth, so you're right, we should probably add exponential decay. It might help for Australia positive bias

experiments/calibration/configs/er.jl
AlexisRenchon: (I am changing this to two sided since it worked)

src/integrated/land.jl
AlexisRenchon: amended!

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
```diff
@@ -1,107 +1,130 @@
...
 ## Model Overview
 
-The biogeochemistry model tracks three prognostic variables that evolve in space (with soil depth) and time:
-- CO₂ concentration in soil air ($ρ_{CO_2}$, kg C m⁻³)
-- Volumetric O₂ fraction in soil air ($O_{2f}$, dimensionless)
-- Soil organic carbon ($C_{som}$, kg C m⁻³)
+The biogeochemistry model carries three prognostic fields that evolve in space (with soil depth) and time. The variables, listed in the units used internally, are:
 
-The model captures the fundamental processes of aerobic decomposition: microbes consume soil organic carbon and oxygen to produce carbon dioxide. These gases diffuse through the soil pore space, with diffusion rates controlled by soil structure (porosity, tortuosity), moisture content, and temperature.
+- the total CO₂ stored per unit soil volume (gas phase plus dissolved), $Y_{CO_2}$ (kg C m⁻³ soil), with $Y_{CO_2} = \theta_{\rm eff}\, c_g$ where $c_g$ is the air-phase carbon mass concentration (kg C m⁻³ air);
```
kmdeck: and $\theta_{\rm eff}$ is the..

For consistency, should it also be Y_{O_{2f}}? and likewise for Y_{C_{som}}

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
```diff
@@ -1,107 +1,130 @@
...
+- the volumetric O₂ fraction in soil air, $O_{2f}$ (dimensionless);
+- the soil organic carbon density, $C_{som}$ (kg C m⁻³), held fixed at its initial value (no tendency, see below).
+
+The model captures the fundamental processes of aerobic decomposition: microbes consume soluble carbon and oxygen to produce carbon dioxide. CO₂ and O₂ diffuse through the soil pore space, with diffusion rates controlled by soil structure (porosity, tortuosity), moisture, temperature, and pressure.
+
+SOC is initialized from the SoilGrids organic carbon density product and is treated as a stationary spatial parameter. This means the model captures geographic heterogeneity of carbon stocks but does not integrate carbon depletion on simulation timescales — a deliberate simplification appropriate for the multi-decade runs targeted by ClimaLand.
```
kmdeck: SOC = C_{som}? we can stick with one notation, maybe?

kmdeck: Maybe we can also say:

We assume that the input of organic matter is equal to the rate of decomposition, such that the actual C_{som} is constant in time. In the future, we may model the rate of decomposition explicitly and track C_{som} over time; expressing C_{som} as a prognostic variable now allows that change to be more easily made in the future.

(otherwise, shouldnt we move C_{som} out of `Y`?)

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
```diff
@@ -1,107 +1,130 @@
...
 \end{equation}
 ```
 
-where $ρ_{CO_2}$ is the CO₂ concentration (kg C m⁻³), $D$ is the effective diffusivity of CO₂ in soil (m² s⁻¹), and $S_m$ is the microbial CO₂ production rate (kg C m⁻³ s⁻¹). The diffusive flux is $-D\nabla ρ_{CO_2}$, directed from regions of high to low concentration.
+where:
+
+- the effective diffusivity of CO₂ in soil, $D$ (m² s⁻¹);
+- the microbial CO₂ production rate, $S_m$ (kg C m⁻³ s⁻¹).
+
+This form is well-behaved as $\theta_a \to 0$: the gas-phase concentration $c_g$ remains finite because the dissolved phase keeps $\theta_{\rm eff}$ bounded away from zero (saturated soil still stores CO₂ in the aqueous phase).
```
kmdeck: Can define theta_a

kmdeck: We can also note that we do not currently model fluxes of dissolved CO2 due to fluxes of the soil water

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
kmdeck: Maybe better to define theta_eff here because it is used a couple times before definition

src/simulations/initial_conditions.jl
AlexisRenchon: you're right!

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
```diff
@@ -1,107 +1,130 @@
...
 ### O₂ Transport and Consumption
 
 Oxygen transport and consumption in soil is described by:
 
 ```math
 \begin{equation}
-\frac{\partial O_{2f}}{\partial t} = \frac{R T}{\theta_a P M_{O_2}} \nabla \cdot \left[D_{O_2} \theta_a \nabla \rho_{O_2}\right] - \frac{R T}{M_C \theta_a P} S_m
+\frac{\partial O_{2f}}{\partial t}
+= \frac{R\, T}{\theta_{\rm eff,O_2}\, P\, M_{O_2}}\,
+\nabla \cdot \left[D_{O_2}\, \nabla \rho_{O_2}\right]
```
kmdeck: we use rho_o2 for the gas concentration of O2 in air, but c_g for carbon. better to be consistent. 

We also use O_2f for a volumetric concentration, but maybe using theta_ instead would be better? since we use theta for volumetric fractions in soil otherwise

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
```diff
@@ -1,107 +1,130 @@
...
-- $\rho_{O_2}$ is the O₂ mass concentration in air (kg O₂ m⁻³ air)
-- $D_{O_2}$ is the effective diffusivity of O₂ in soil (m² s⁻¹)
-- $R$ is the universal gas constant (8.314 J mol⁻¹ K⁻¹)
-- $T$ is soil temperature (K)
-- $P$ is atmospheric pressure (Pa)
-- $M_{O_2}$ is the molar mass of O₂ (0.032 kg mol⁻¹)
-- $M_C$ is the molar mass of carbon (0.012 kg mol⁻¹)
 
-The second term represents O₂ consumption by microbial respiration, following the stoichiometry C + O₂ → CO₂, where each 12 g of carbon respired consumes 32 g of oxygen.
+- the effective porosity for O₂, $\theta_{\rm eff,O_2}$ (m³ m⁻³ soil), computed with the O₂-specific Henry's-law factor;
```
kmdeck: What does this mean? What is the effective porosity, and why is it different for O2?

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
```diff
@@ -1,107 +1,130 @@
...
 
 ```math
 \begin{equation}
-\frac{\partial C_{som}}{\partial t} = -S_m
+\frac{\partial C_{som}}{\partial t} = 0
 \end{equation}
 ```
 
-where $C_{som}$ is the soil organic carbon content (kg C m⁻³). This equation enforces carbon mass conservation: the rate of SOC consumption equals the rate of CO₂ production.
+$C_{som}$ enters the DAMM substrate kinetics (below) but is not depleted by respiration. The value of $C_{som}$ is set from the SoilGrids organic carbon density at the start of the simulation, providing the spatial structure of carbon availability.
```
kmdeck: this first $C_{som}$ rendered a bit oddly in the docs

AlexisRenchon:
@kmdeck I addressed your first round of review (from yesterday) in a new commit. Note many files are modified just to add dt

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
```diff
@@ -111,75 +134,130 @@ MM_{sx} = \frac{[S_x]}{kM_{sx} + [S_x]}
...
+- the gas-specific reference diffusivity at standard conditions, $D_{{\rm ref},{\rm gas}}$ (m² s⁻¹), evaluated as $D_{\rm ref}$ for CO₂ and $D_{{\rm ref},O_2}$ for O₂;
+- the standard reference temperature, $T_{\rm ref} = 273$ K;
+- the standard reference pressure, $P_{\rm ref} = 101325$ Pa.
+
+The temperature exponent 1.75 reflects increased molecular kinetic energy, while the pressure dependence accounts for the change in gas number density with surface pressure.
 
 ### Volumetric Air Content
 
-The volumetric air content is computed as:
+The volumetric air content uses the total water content (liquid plus ice), so frozen water displaces air space:
```
kmdeck, line 255: Just out of curiosity - do you see theta_w > \nu if you use p.soil.theta_l and not Y.soil.\vartheta_l?

p.soil.theta_l should respect this constrain already.

kmdeck: I might leave out this `max` in the docs because I dont think we need it from physics perspective, it might confuse the reader (why is theta_l + theta_i > nu ever?)

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
```diff
@@ -189,96 +267,120 @@ The O₂ mass concentration in air is computed from the volumetric fraction usin
...
 
-where $\rho_{O_2,atm}$ is computed from $O_{2f} = 0.21$ using the ideal gas law.
+At the bottom boundary, a zero-flux condition is applied: $F_{O_2,{\rm bottom}} = 0$.
+
+## Numerical Safeguards
+
+The continuous PDE above is integrated with an explicit scheme. Several pragmatic safeguards keep the discrete update well-behaved in extreme regimes (e.g. very dry hot columns or saturated ice-rich columns), without changing the equations in their physical regime:
 
-At the bottom boundary, a no-flux condition is typically applied: $F_{O_2,bottom} = 0$.
+- the floor on the effective porosity, $\theta_{\rm eff,min} = 10^{-4}$, prevents a singular $c_g = Y_{CO_2}/\theta_{\rm eff}$ when both $\theta_a$ and $\theta_l$ are vanishingly small;
```
kmdeck:
```suggestion
- the floor on the effective porosity, $\theta_{\rm eff,min} = 10^{-4}$, prevents a singular $c_g = Y_{CO_2}/\theta_{\rm eff}$ when both $\theta_a$ and $\theta_l$ are vanishingly small (saturated, frozen soil);
```

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
```diff
@@ -189,96 +267,120 @@ The O₂ mass concentration in air is computed from the volumetric fraction usin
...
+- the diffusivity ratio is capped at $\theta_a / \theta_{a100} \le 5$, preventing the Ryan/Moldrup expression from producing unphysically large $D$ in soils much drier than their reference state ($\theta_a \gg \theta_{a100}$);
```
kmdeck, line 336: This one I am curious about - is it needed?

I ask because theta_a is bounded between theta_residual and \nu - \theta_residual, and both are between 0 and 1. So the only way this factor becomes very big is if theta_{a,100} is very small. Since this is produced using the soil parameters, we can look and see if there are regions where this happens. It may be that in some areas we set the residual to 0 and then theta_a,100 is near zero

src/standalone/Soil/Biogeochemistry/co2_parameterizations.jl
```diff
@@ -136,6 +147,67 @@ function o2_availability(O2_f::FT, θ_a::FT, D_oa::FT) where {FT}
...
+
+The temperature dependence follows:
+    K_H(T) = K_H(T_ref) * exp[dln_K_H_dT * (1/T - 1/T_ref)]
+
+where T_ref = 298.15 K and dln_K_H_dT is the temperature coefficient.
+
+Reference: Sander (2015), Atmos. Chem. Phys., 15, 4399-4981.
+"""
+function henry_constant(K_H_298::FT, dln_K_H_dT::FT, T::FT) where {FT}
+    T_ref = FT(298.15)
```
kmdeck: I think this should be stored in the parameters somewhere. We already have T0 stored as 25C in the toml, I believe

src/standalone/Soil/Biogeochemistry/co2_parameterizations.jl
```diff
@@ -144,12 +216,56 @@ end
 Computes the volumetric air content (`θ_a`) in the soil,
 which is related to the total soil porosity (`ν`) and
 volumetric soil water content (`θ_w = θ_l+θ_i`).
+
+Note: The effective_porosity function provides numerical stability
+when θ_a approaches zero by accounting for dissolved gas in liquid water.
```
kmdeck, line 226: by preventing small negative numbers for theta_a.

src/standalone/Soil/Biogeochemistry/co2_parameterizations.jl
```diff
@@ -144,12 +216,56 @@ end
...
+
+Effective diffusivity of a gas within partially saturated soil (m² s⁻¹).
+
+Internal helper used by [`co2_diffusivity`](@ref) and [`o2_diffusivity`](@ref).
+The reference diffusivity `D_ref` is the gas's diffusion coefficient in free
+air at standard temperature and pressure (273 K, 101325 Pa); the rest of the
+expression follows Ryan et al., GMD 11, 1909-1928, 2018,
+https://doi.org/10.5194/gmd-11-1909-2018.
+"""
+function _gas_diffusivity_in_soil(
```
kmdeck:
```suggestion
function gas_diffusivity_in_soil(
```

kmdeck: (I think we should be careful about using the _ notation for function names. Whatever we choose, it should be consistent. so it seems odd to use it here but not for other functions in this file)

src/standalone/Soil/Biogeochemistry/co2_parameterizations.jl
```diff
@@ -144,12 +216,56 @@ end
...
+    θ_a100::FT,
+    b::FT,
+    ν::FT,
+    D_ref::FT,
+    earth_param_set,
+) where {FT}
+    T_ref = FT(LP.T_0(earth_param_set))
+    P_ref = FT(LP.P_ref(earth_param_set))
+    θ_a = volumetric_air_content(θ_w, ν)
+    D0 = D_ref * max((T_soil / T_ref), 0)^FT(1.75) * (P_ref / P_sfc)
```
kmdeck: Maybe this 1.75 should also be in the parameter toml

src/standalone/Soil/Biogeochemistry/co2_parameterizations.jl
```diff
@@ -182,14 +298,52 @@ function co2_diffusivity(
...
-        (θ_a / θ_a100)^(FT(2) + FT(3) / b)
-    return D
+    return _gas_diffusivity_in_soil(
+        T_soil,
+        θ_w,
+        P_sfc,
+        θ_a100,
+        b,
+        ν,
+        FT(params.D_ref),
```
kmdeck: this is already of type FT, we dont need to have this here

src/standalone/Soil/Biogeochemistry/co2_parameterizations.jl
```diff
@@ -182,14 +298,52 @@ function co2_diffusivity(
...
+    params::SoilCO2ModelParameters{FT},
+) where {FT}
+    return _gas_diffusivity_in_soil(
+        T_soil,
+        θ_w,
+        P_sfc,
+        θ_a100,
+        b,
+        ν,
+        FT(params.D_ref_o2),
```
kmdeck: likewise here

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -140,14 +161,17 @@ struct SoilCO2Model{FT, PS, D, BC, S, DT} <:
     sources::S
     "Drivers"
     drivers::DT
+    "Model timestep in seconds (used for the O2_f tendency limiter); pass FT(0) to disable the hard cap"
```
kmdeck, line 171: Suggestion to make this a required argument 

- will match snow model setup
- the model will produce NaNs in global run without this, so we really should require the user to pass a value here

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -379,6 +414,25 @@ function ClimaLand.make_compute_exp_tendency(model::SoilCO2Model)
...
+        # in hot/dry columns with fine near-surface layers), the unlimited
+        # tendency oscillates wildly in both directions. The lower cap
+        # -max(Y, 0)/Δt guarantees Y_new ≥ 0; the upper cap
+        # max(O2_f_atm - Y, 0)/Δt guarantees Y_new ≤ O2_f_atm (the soil air
+        # cannot physically exceed the atmospheric O2 fraction since soil is
+        # an O2 sink). For multi-stage RK integrators the bounds are slightly
+        # conservative (per-stage step is a fraction of Δt) but still safe.
+        if model.Δt > 0
+            inv_Δt = FT(1) / FT(model.Δt)
+            O2_f_atm = FT(model.parameters.O2_f_atm)
```
kmdeck: this should be FT already, no need to convert it again

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -379,6 +414,25 @@ function ClimaLand.make_compute_exp_tendency(model::SoilCO2Model)
...
+        # O2_f outside [0, O2_f_atm]. With CFL-violating diffusivities (common
+        # in hot/dry columns with fine near-surface layers), the unlimited
+        # tendency oscillates wildly in both directions. The lower cap
+        # -max(Y, 0)/Δt guarantees Y_new ≥ 0; the upper cap
+        # max(O2_f_atm - Y, 0)/Δt guarantees Y_new ≤ O2_f_atm (the soil air
+        # cannot physically exceed the atmospheric O2 fraction since soil is
+        # an O2 sink). For multi-stage RK integrators the bounds are slightly
+        # conservative (per-stage step is a fraction of Δt) but still safe.
+        if model.Δt > 0
+            inv_Δt = FT(1) / FT(model.Δt)
```
kmdeck:
```suggestion
            inv_Δt = 1 / model.Δt # model.dt is already type FT
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -379,6 +414,25 @@ function ClimaLand.make_compute_exp_tendency(model::SoilCO2Model)
...
+            O2_f_atm = FT(model.parameters.O2_f_atm)
+            @. dY.soilco2.O2_f = clamp(
+                dY.soilco2.O2_f,
+                -max(Y.soilco2.O2_f, FT(0)) * inv_Δt,
```
kmdeck: I also think you do not need the FT(0), and can pass 0 directly

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -551,6 +612,16 @@ function soil_moisture(driver::PrescribedMet, p, Y, t, z)
...
 end
 
+"""
+    soil_ice(driver::PrescribedMet, p, Y, t, z)
+
+Returns zero ice content for prescribed soil case (standalone mode has no ice).
+"""
+function soil_ice(driver::PrescribedMet, p, Y, t, z)
+    FT = eltype(z)
+    return FT(0) .* z  # Return field of zeros matching z
```
kmdeck: This will allocate on every step. Is it possible to just return FT(0) here?

or maybe @. lazy( 0 * z)

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
         z = model.domain.fields.z
         T_soil = soil_temperature(model.drivers.met, p, Y, t, z)
         θ_l = soil_moisture(model.drivers.met, p, Y, t, z)
-        Csom = Y.soilco2.SOC  # Now using prognostic SOC
+        θ_i = soil_ice(model.drivers.met, p, Y, t, z)
+        Csom = max.(Y.soilco2.SOC, FT(0))  # Clamp SOC to non-negative
```
kmdeck: This will allocate. Maybe we can put
`max(Y.soilco2.SOC, 0)` in place of Csom in every broadcasted expression in this function instead?

kmdeck: or do Csom = @. lazy(max(Y.soilco2.SOC, 0))

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
...
         P_sfc = p.drivers.P
-        θ_w = θ_l
         ν = model.drivers.met.ν
         θ_a100 = model.drivers.met.θ_a100
         b = model.drivers.met.b
 
         @. p.soilco2.T = T_soil
 
+        # Compute θ_a using total water (liquid + ice)
+        θ_w = θ_l .+ θ_i
```
kmdeck: This allocates on each step, should remove

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
...
+        @. p.soilco2.θ_a = volumetric_air_content(θ_w, ν)
```
kmdeck:
```suggestion
        @. p.soilco2.θ_a = volumetric_air_content(θ_l + θ_i, ν)
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
         z = model.domain.fields.z
         T_soil = soil_temperature(model.drivers.met, p, Y, t, z)
         θ_l = soil_moisture(model.drivers.met, p, Y, t, z)
```
kmdeck, line 647: does this return p.soil.theta_l? or Y.soil.\vartheta_l

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
...
         b = model.drivers.met.b
 
         @. p.soilco2.T = T_soil
 
+        # Compute θ_a using total water (liquid + ice)
+        θ_w = θ_l .+ θ_i
+        @. p.soilco2.θ_a = volumetric_air_content(θ_w, ν)
+
         @. p.soilco2.D =
             co2_diffusivity.(T_soil, θ_w, P_sfc, θ_a100, b, ν, params)
```
kmdeck:
```suggestion
            co2_diffusivity(T_soil, θ_l + θ_i, P_sfc, θ_a100, b, ν, params)
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
...
-        @. p.soilco2.D_o2 .=
-            co2_diffusivity.(T_soil, θ_w, P_sfc, θ_a100, b, ν, params)
+        @. p.soilco2.D_o2 =
+            o2_diffusivity.(T_soil, θ_w, P_sfc, θ_a100, b, ν, params)
```
kmdeck:
```suggestion
            o2_diffusivity(T_soil, θ_l + θ_i, P_sfc, θ_a100, b, ν, params)
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
...
+        # Compute air-equivalent CO2 concentration for diffusion
+        M_C = FT(params.M_C)
+        # Clamp negative CO2 to zero (can occur from numerical diffusion)
+        @. Y.soilco2.CO2 = max(Y.soilco2.CO2, FT(0))
+        @. p.soilco2.CO2_air_eq = Y.soilco2.CO2 / max(p.soilco2.θ_eff, eps(FT))
+
+        # Safety guard: cap CO2_air_eq at 100 million ppm equivalent.
+        # Values above this indicate numerical blow-up, not physical conditions.
+        # 100M ppm = 100 (mol/mol) → CO2_air_eq_max = 100 * M_C * P / (R * T)
+        CO2_air_eq_max = @. FT(100) * M_C * P_sfc / (R * T_soil)
```
kmdeck, line 707: This line allocates

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
...
-        @. p.soilco2.θ_a = max(ν - θ_l, FT(0))
+        # Compute air-equivalent CO2 concentration for diffusion
+        M_C = FT(params.M_C)
+        # Clamp negative CO2 to zero (can occur from numerical diffusion)
+        @. Y.soilco2.CO2 = max(Y.soilco2.CO2, FT(0))
+        @. p.soilco2.CO2_air_eq = Y.soilco2.CO2 / max(p.soilco2.θ_eff, eps(FT))
+
+        # Safety guard: cap CO2_air_eq at 100 million ppm equivalent.
+        # Values above this indicate numerical blow-up, not physical conditions.
+        # 100M ppm = 100 (mol/mol) → CO2_air_eq_max = 100 * M_C * P / (R * T)
```
kmdeck, line 705: Do you know where this blow up occurs/how it happens?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
kmdeck: also no need for FT here

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
...
+        CO2_air_eq_max = @. FT(100) * M_C * P_sfc / (R * T_soil)
+        @. p.soilco2.CO2_air_eq = ifelse(
+            p.soilco2.CO2_air_eq > CO2_air_eq_max,
```
kmdeck, line 709:
```suggestion
            p.soilco2.CO2_air_eq > 100 * M_C * P / (R * T),
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -567,22 +638,73 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
...
+            FT(NaN),
+            p.soilco2.CO2_air_eq,
+        )
+
+        # Clamp O2_f to physical range [0, 1]; explicit diffusion can
+        # overshoot in hot/dry columns where effective diffusivity D_o2/θ_eff_o2
+        # exceeds the CFL limit. Clamping keeps the simulation alive with
+        # physically meaningful values — the overshoot is a numerical artifact,
+        # not a real state.
+        @. Y.soilco2.O2_f = clamp(Y.soilco2.O2_f, FT(0), FT(1))
```
kmdeck:
```suggestion
        @. Y.soilco2.O2_f = clamp(Y.soilco2.O2_f, 0, 1)
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -713,8 +841,26 @@ end
...
+    M_C::FT
+end
+
+"""
+    AtmosCO2StateBC(earth_param_set, M_C::FT) where {FT}
+
+Constructor for AtmosCO2StateBC that gets parameters from LandParameters and model parameters.
+"""
+function AtmosCO2StateBC(earth_param_set, M_C::FT) where {FT}
+    R = FT(LP.gas_constant(earth_param_set))
```
kmdeck: do not need FT here, everything in earth_param_set already has the right type

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -739,8 +886,12 @@ function ClimaLand.boundary_flux!(
...
 )
     D_c = ClimaLand.Domains.top_center_to_surface(p.soilco2.D)
-    C_c = ClimaLand.Domains.top_center_to_surface(Y.soilco2.CO2)
-    C_bc = p.drivers.c_co2
+    C_c = ClimaLand.Domains.top_center_to_surface(p.soilco2.CO2_air_eq)
+    T_soil_top = ClimaLand.Domains.top_center_to_surface(p.soilco2.T)
+    P_sfc = p.drivers.P
+    R = bc.R
+    M_C = bc.M_C
+    C_bc = @. p.drivers.c_co2 * P_sfc * M_C / (R * T_soil_top)
```
kmdeck:
```suggestion
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -739,8 +886,12 @@ function ClimaLand.boundary_flux!(
...
     @. bc_field = ClimaLand.diffusive_flux(D_c, C_bc, C_c, Δz)
```
kmdeck:
```suggestion
    @. bc_field = ClimaLand.diffusive_flux(D_c, p.drivers.c_co2 * P_sfc * M_C / (R * T_soil_top), C_c, Δz)
```

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
AlexisRenchon: I think it used to happen before I forced O2 to be within [0-0.21], O2 would explode (due to solver instability) and then CO2 would follow. I think it isn't an issue anymore - maybe this threshold for co2 can be removed

toml/default_parameters.toml
```diff
@@ -223,19 +223,19 @@ tag = "ConstantResetAlbedo"
...
+value = 68.2139
 type = "float"
 description = "Unit cost ratio of Vcmax to transpiration for C3 plants (Stocker 2020)"
 tag = "PModelParameters"
 
 ["pmodel_β_c4"]
-value = 51
+value = 39.575
 type = "float"
 description = "Unit cost ratio of Vcmax to transpiration for C4 plants (146/9, pyrealm)"
```
kmdeck:
```suggestion
description = "Unit cost ratio of Vcmax to transpiration for C4 plants"
```

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
AlexisRenchon: good call to point that out

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
AlexisRenchon: it is explained in the docs text now! I think it is quite clear. In short, effective is due to split into gas and liquid phase, and converting to equivalent if there was only gas

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
AlexisRenchon: Amended (we can't start bullet points with symbols in the docs for some reason it doesn't render well)

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
AlexisRenchon: yeah I think this is one of the remnants of my attempt to fix the NaNs... probably not needed

docs/src/standalone/pages/soil/biogeochemistry/DAMM_model.md
AlexisRenchon: maybe it is no longer needed... another one of my attempts to fix NaNs

AlexisRenchon: (I will leave it for now)

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
AlexisRenchon: done already in previous commit! (addressing your first round of review)

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
AlexisRenchon: p.soil.theta_l

experiments/calibration/configs/energy_fluxes.jl
```diff
@@ -4,10 +4,11 @@
...
-- `lwu`: W m^-2 (ERA5)
-- `shf`: W m^-2 (ERA5)
-- `lhf`: W m^-2 (ERA5)
+These multiply the identity in `ScalarCovariance`, so they are variances.
+Units are the square of the observational dataset units:
+- `lwu`: (W m^-2)^2 (ERA5)
+- `shf`: (W m^-2)^2 (ERA5)
+- `lhf`: (W m^-2)^2 (ERA5)
 """
 const NOISE_SCALARS = Dict("lwu" => 1.0, "shf" => 1.0, "lhf" => 1.0)
```
kmdeck, line 13: not your PR, but should these be bigger?

src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -425,18 +485,26 @@ NVTX.@annotate function ClimaLand.source!(
 )
     dY.soilco2.CO2 .+= p.soilco2.Sm
 
-    M_C = eltype(p.soilco2.Sm)(params.M_C)
-    R = eltype(p.soilco2.Sm)(LP.gas_constant(params.earth_param_set))
+    FT = eltype(p.soilco2.Sm)
+    M_C = FT(params.M_C)
+    R = LP.gas_constant(params.earth_param_set)
     T_soil = p.soilco2.T  # soil temperature (K)
```
kmdeck, line 491: not for this PR, but are we storing soil temp 2x? we can access p.soil.T here and not save it in p.soilco2.T, I think.

kmdeck (approved):
thanks!

PR #1719, opened 2026-04-23
revert to main autotrophic respiration, update heterotrophic

AlexisRenchon:
- fix winter NEE for November December, before calibration
- revert to main autotrophic respiration
- update heterotrophic resp, add 1 DAMM param to calibrate (now intercept + slope for T response) 

low NEE in January-February: possibly ice or snow (saturated or icy soils would prevent co2 diffusion and or co2 production by microbes)

alexis_test_seasonality


AlexisRenchon:
@braghiere merging this branch would merge it to rb/callmip_phase1a

PR #1726, opened 2026-04-24
Update the default land model

kmdeck:
## Purpose 
**First commit:** Updates the default land model to use the defaults from snowy_land_pmodel.jl:
- Pmodel conductance and photosynthesis
- piecewise soil moisture stress
- snow albedo depends on zenith angle
- snow cover fraction with resolution read from domain

**Second commit:**
After making the first change^, the unit tests on GPU on central did not pass because of a parameter memory bug. This was because of the pmodel `update_photosynthesis!` call. After talking with @dennisYatunin, we determined it was the large number of stacked lazy operations. To address this, I created a pointwise function which does all of the computation for `update_photosynthesis!`. At the same time, it was convenient to make the cache variable a field of tuples, rather than 4 fields separately. Therefore, we stored GPP, An, Rd, and ci differently between the pmodel vs Farquhar photosynthesis. In one case, it is p.canopy.photosynthesis.InstVars, in the other, the 4 separately. Because of this, it was helpful to make additional "get" functions for these quantities. 

This seems to improve performance: the canopy model test on GPU went from 4 minutes to 2.5 minutes, and ozark_pmodel went from taking 199 seconds to 165 seconds. 

Note that Farquhar photosynthesis already calls a single pointwise function.


experiments/long_runs/snowy_land_pmodel.jl
```diff
@@ -91,58 +91,10 @@ function setup_model(
...
-        toml_dict,
-        domain,
-        Δt;
-        prognostic_land_components,
-        snow,
-        canopy,
-    )
+    # Construct the land model with all default components
+    prognostic_land_components =
+        (:canopy, :lake, :snow, :soil), land =
```
kmdeck: need to add new line

experiments/calibration/models/snowy_land.jl
```diff
@@ -46,52 +46,13 @@ function setup_model(
         stop_date,
     )
     prognostic_land_components = (:canopy, :snow, :soil)
```
kmdeck: this should probably have lake

src/standalone/Snow/Snow.jl
```diff
@@ -512,7 +512,7 @@ end
         ϵ_snow = toml_dict["snow_emissivity"],
         α_snow = ConstantAlbedoModel(toml_dict["snow_albedo"]),
```
kmdeck, line 513: need to update

src/standalone/Vegetation/Canopy.jl
```diff
@@ -729,8 +729,8 @@ end
         prognostic_land_components = (:canopy,),
         autotrophic_respiration = AutotrophicRespirationModel{FT}(toml_dict),
         radiative_transfer = TwoStreamModel{FT}(domain, toml_dict),
-        photosynthesis = FarquharModel{FT}(domain, toml_dict),
-        conductance = MedlynConductanceModel{FT}(domain, toml_dict),
+        photosynthesis = PModel{FT}(domain, toml_dict),
+        conductance = PModelConductance{FT}(toml_dict),
         soil_moisture_stress = TuzetMoistureStressModel{FT}(toml_dict),
```
kmdeck, line 734: This is hard to update to Piecewise, because we don't prescribed theta(t), nu, theta_r, for prescribed Soil. 

Maybe better to do NoMoistureStress...

experiments/long_runs/snowy_land_pmodel.jl
```diff
@@ -91,58 +91,9 @@ function setup_model(
...
-        LAI,
-        toml_dict,
-        domain,
-        Δt;
-        prognostic_land_components,
-        snow,
-        canopy,
-    )
+    # Construct the land model with all default components
+    prognostic_land_components = (:canopy, :lake, :snow, :soil)
```
kmdeck: @AlexisRenchon should we always run soilco2 by default?

AlexisRenchon: I think so! (unless it causes instability or slow things)

kmdeck: I added it back!

AlexisRenchon (approved):
It looks good to me! Thanks Kat!

Note that it changed autotrophic respiration substantially, but I think in a good direction. This is before this PR:

image

This is after this PR:

image

Before, Amazon respiration was too low, now it's a bit too high. But I think calibration would actually yield better results. No need to calibrate now though, as I plan to change what we calibrate C fluxes on (inversion and sif-based gpp instead of fluxcom)

EDIT: the new respiration oddly look very similar to GPP...

AlexisRenchon (changes requested):
Actually there seem to be a problem with respiration. I do not see HR anymore in the [outputs ](https://buildkite.com/clima/climaland-long-runs/builds/4798/canvas?sid=019df9c7-0dda-4bc9-8b74-5ffb31a1a79c&tab=artifacts), and ER = RA (instead of RA + HR)

Possibly because there is no :soilco2 in the new default? (as you mentioned), we should add it. RA is probably too high then (but we can calibrate later)

experiments/long_runs/snowy_land_pmodel.jl
```diff
@@ -91,58 +91,9 @@ function setup_model(
...
-        forcing,
-        LAI,
-        toml_dict,
-        domain,
-        Δt;
-        prognostic_land_components,
-        snow,
-        canopy,
-    )
+    land = LandModel{FT}(forcing, LAI, toml_dict, domain, Δt)
```
AlexisRenchon:
```suggestion
    land = LandModel{FT}(forcing, LAI, toml_dict, domain, Δt; prognostic_land_components)
```

AlexisRenchon (dismissed):
I think my edit would add back soilco2

AlexisRenchon (approved):

PR #1730, opened 2026-04-24
remove `convert_cb` function

### LINKED ISSUE #1729: remove `convert_cb` function
This was a temporary fix as we switched from SciMLBase to ClimaTimeSteppers. Now that we have updated to use ClimaDiagnostics v0.3.4+, we can remove this function. We will also need to set the ClimaDiagnostics compat to `0.3.4`.

juliasloan25:
## Purpose 
now that we use ClimaDiagnostics v0.3.4 we don't need `convert_cb`, which was added in https://github.com/CliMA/ClimaLand.jl/pull/1678.

closes #1729

## Content

----


Project.toml
```diff
@@ -49,8 +49,8 @@ Adapt = "4.5.2"
 CairoMakie = "0.13, 0.14, 0.15"
 ClimaAnalysis = "0.5.17"
 ClimaComms = "0.6.2"
-ClimaCore = "0.14.46"
-ClimaDiagnostics = "0.3.0"
+ClimaCore = "0.14.50"
```
ph-kev, line 52: Why did ClimaCore need to be updated to 0.14.50?

ph-kev (approved):

Project.toml
juliasloan25: When I kept the previous compat the downgrade tests failed because the compats didn't match the versions that the packages got resolved to. This is because ClimaDiagnostics 0.3.4 sets ClimaCore 0.14.50 lower bound ([here](https://github.com/CliMA/ClimaDiagnostics.jl/blob/a6326a72be1d1ca3642f2f7e8c512b9f5505db4f/Project.toml#L22)).

It seems odd to me that in a downstream package we need to set our compat to be the same as what the most restrictive upstream package has. Maybe this is because we're using downgrade v2? I thought it would replicate the v1 behavior since we set `mode: 'forcedeps'`, but it does seem a little different.

PR #1731, opened 2026-04-27
Fix plotting bug with annual timeseries

ph-kev:
The function assumes that the start date of the simulation is January, which is not necessarily true.

~~To fix this, the PR discards the first initial year. The other alternative that I haven't try is to prepend and append `NaN`s so every vector in `var_global_average` is of length 12.~~

I decided to prepend `NaN`s for the first year instead of discarding it.

The plot produced now look like this.

sr_annual_timeseries

### DIFF
--- a/ext/land_sim_vis/plotting_utils.jl
+++ b/ext/land_sim_vis/plotting_utils.jl
@@ -235,6 +235,13 @@ function make_ocean_masked_annual_timeseries(
         var = get(simdir; short_name)
         kwarg_z = ClimaAnalysis.has_altitude(var) ? Dict(:z => 1) : Dict() # if has altitude, take first layer
         var_sliced = ClimaAnalysis.slice(var; kwarg_z...)
+
+        # Get the first and last years of the simulation for windowing
+        first_date = first(ClimaAnalysis.dates(var_sliced))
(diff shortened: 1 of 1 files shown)


ext/land_sim_vis/plotting_utils.jl
```diff
@@ -235,6 +235,15 @@ function make_ocean_masked_annual_timeseries(
         var = get(simdir; short_name)
         kwarg_z = ClimaAnalysis.has_altitude(var) ? Dict(:z => 1) : Dict() # if has altitude, take first layer
         var_sliced = ClimaAnalysis.slice(var; kwarg_z...)
+
+        # Compute the first year with all 12 months
+        # For example, if the start date of the simulation is 2000-03-01, then discard the
+        # year 2000
```
kmdeck: Are we discarding the first year, or appending the missing values with NaNs? I think the latter?

If so, can update this comment

kmdeck (commented):
thanks so much!

ext/land_sim_vis/plotting_utils.jl
ph-kev: Appending the missing values with NaNs. I updated the comment.

kmdeck (approved):

PR #1734, opened 2026-04-28
Check for monthly data in annual time series

ph-kev:
This commit also updates the ClimaAnalysis version, because `make_ocean_masked_annual_timeseries` uses functionality that is not in ClimaAnalysis v0.5.17.

### DIFF
--- a/ext/land_sim_vis/plotting_utils.jl
+++ b/ext/land_sim_vis/plotting_utils.jl
@@ -233,6 +233,11 @@ function make_ocean_masked_annual_timeseries(
     simdir = ClimaAnalysis.SimDir(diagdir)
     for short_name in short_names
         var = get(simdir; short_name)
+
+        # Can only plot monthly data
+        # This assumes EveryCalendarDtSchedule is used with a period of 1 month
+        occursin("1 Month", ClimaAnalysis.long_name(var)) || continue
+
         kwarg_z = ClimaAnalysis.has_altitude(var) ? Dict(:z => 1) : Dict() # if has altitude, take first layer
         var_sliced = ClimaAnalysis.slice(var; kwarg_z...)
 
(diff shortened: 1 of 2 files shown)


ext/land_sim_vis/plotting_utils.jl
```diff
@@ -233,6 +233,10 @@ function make_ocean_masked_annual_timeseries(
     simdir = ClimaAnalysis.SimDir(diagdir)
     for short_name in short_names
         var = get(simdir; short_name)
+
+        # Can only plot monthly data
+        occursin("1 Month", ClimaAnalysis.long_name(var)) || continue
```
ph-kev, line 239: This relies on the diagnostics being produced with the `EveryCalendarDtSchedule`.

ext/land_sim_vis/plotting_utils.jl
ph-kev: This doesn't work if there is less than a year of simulation.

ext/land_sim_vis/plotting_utils.jl
kmdeck: It seems odd to only be able to make an annual timeseries if the data is monthly. Why can't we back fill the missing data in the first year with NaNs if the data is daily, or weekly, or... any interval?

ph-kev: The rest of the code assumes monthly data. I can add this functionality in another PR if it is needed.

kmdeck: OK! I dont think it is needed, but I wonder why it was working before then?

kmdeck (approved):

ext/land_sim_vis/plotting_utils.jl
ph-kev: I think the result produced before would be wrong in that case.

PR #1740, opened 2026-05-07
add plots to CI

AlexisRenchon:
This PR adds global seasonality to ilamb energy and carbon annual figures, as well as a boxplot comparing to other model

boxplot_rmse

Note the "Disclaimer: For internal use only. A proper Model Intercomparison Project (MIP) is beyond scope here, so this comparison is not strictly fair: reference datasets and forcings differ across models, tuning targets differ, and some variables are prescribed rather than predicted (e.g., LAI in ClimaLand). Only a MIP would enable a fair comparison."

ILAMB_sim_annual_time_avg_over_all_years

you can optionally pass two climaland output (e.g., may 2026, november 2025) to see how it has evolved in the boxplot figure

boxplot_rmse

### DIFF
--- a/ext/land_sim_vis/leaderboard/leaderboard.jl
+++ b/ext/land_sim_vis/leaderboard/leaderboard.jl
@@ -28,6 +28,43 @@ function _percentile_contour_kwargs(
 end
 
 """
+    _monthly_climatology_global_mean(var, mask_fn = ClimaAnalysis.apply_oceanmask)
+
+Return a 12-element vector of global, lonlat-weighted monthly climatology
+means for `var`. Index `i` corresponds to month `i` (1=Jan ... 12=Dec).
+Months that have no valid samples are filled with `NaN`.
+
+`mask_fn` is the same per-variable mask used by `global_bias` /  `global_rmse`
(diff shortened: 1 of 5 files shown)


ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -573,6 +615,41 @@ function compute_seasonal_leaderboard(
...
+                )
+                CairoMakie.lines!(
+                    ax,
+                    1:12,
+                    sim_monthly;
+                    color = :firebrick,
+                    linewidth = 4,
+                    label = "ClimaLand",
+                )
+                CairoMakie.axislegend(ax, position = :rt, framevisible = false)
```
kmdeck: for SWU and NEE, could we make it :rb? the curves intersect the legend on those

kmdeck (approved):
looks fine as long as we make it so the box plot is not made by default. It should not be used in scientific presentations.

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: Good point - I will make it only for the top right one (SHF and ER), it doesn't need to be on all of them... and I'll write SIM and OBS maybe, to be consistent with the naming on the other columns

PR #1741, opened 2026-05-07
Tr/bench

imreddyTeja:
## Purpose 
- make the short-perf pipeline actually test for regressions
- change the way a gpu is requested so the plugin will pass it to slurm-buildkite

## To-do
- re-enable the pipeline after merging this

----

### DIFF
--- a/.buildkite/short_perf/pipeline.yml
+++ b/.buildkite/short_perf/pipeline.yml
@@ -67,7 +67,8 @@ steps:
                     command: "julia --color=yes --project=.buildkite experiments/benchmarks/bucket.jl"
                     agents:
                       slurm_mem: 8GB
-                      slurm_gpus: 1
+                      # a bug in this plugin ignores non-strings properties
+                      slurm_gpus: "1"
                     concurrency: 1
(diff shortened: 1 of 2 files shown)


.buildkite/short_perf/pipeline.yml
```diff
@@ -101,6 +101,6 @@ steps:
                     concurrency_group: "depot/climaland-perf-regression-tests"
                     agents:
                       slurm_mem: 16GB
-                      slurm_gpus: 1
+                      slurm_gpus: "1"
```
ph-kev, line 107: Is using `1` versus `"1"` documented anywhere? I think it would be good to open an issue or add to the docs about this somewhere. Maybe in the slurm-buildkite repo?

ph-kev (approved):

.buildkite/short_perf/pipeline.yml
imreddyTeja: Would a comment in the pipeline be an ok spot for this? The string requirement is an issue with the plugin, which is now archived...

ph-kev: I am fine with either

PR #1746, opened 2026-05-13
Refactor data sources for leaderboard

ph-kev:
This PR refactors the data sources for the leaderboard to use the struct `DataLoader`s. This replaces the dictionary that map names to functions which then takes a start date to materialize the `OutputVar`. This is currently being used for the coupler calibration (see https://github.com/CliMA/ClimaCoupler.jl/blob/main/src/CalibrationTools.jl).

This PR is one step towards attempting to keep the interface for the coupler and land calibration pipelines roughly the same.

The PR reverts the downgrade test to use v1 instead of v2 since it is simpler to understand and debug. There were also problems with the Project.toml.

- Downloads is a standard library, so the version should be 1 instead of 1.6.
- Flux 0.15 and 0.16 is not compatible with Julia 1.12. See https://github.com/FluxML/Flux.jl/issues/2583.
- Dataframes needed to be updated due to ClimaTimeSteppers and Dataframes depending on PrettyTables.

TODO
- [x] Update documentation

### DIFF
--- a/ext/land_sim_vis/leaderboard/data_sources.jl
+++ b/ext/land_sim_vis/leaderboard/data_sources.jl
@@ -1,325 +1,396 @@
+import Dates
+
+import ClimaLand
+import ClimaAnalysis
+import ClimaAnalysis: OutputVar, NCCatalog
+import ClimaUtilities.ClimaArtifacts: @clima_artifact
+
 """
-    get_sim_var_dict(diagnostics_folder_path)
+    preprocess_sim_var(var::OutputVar)
 
-Return a dictionary mapping short names to `OutputVar` containing preprocessed
-simulation data. This is used by the function `compute_leaderboard`.
+Preprocess the simulation variable `var` based on its short name.
 
-To add a variable for the leaderboard, add a key-value pair to the dictionary
-`sim_var_dict` whose key is the short name of the variable and the value is an
-anonymous function that returns a `OutputVar`. For each variable, any
-preprocessing should be done in the corresponding anonymous function which
-includes unit conversion.
-
-The variable should have only three dimensions: time, longitude, and latitude.
+Return the preprocessed `OutputVar`.
 """
-function get_sim_var_dict(diagnostics_folder_path)
-    # Dict for loading in simulation data
-    sim_var_dict = Dict{String, Any}()
+function preprocess_sim_var(var::OutputVar)
+    return _preprocess_sim_var(var, Val(Symbol(ClimaAnalysis.short_name(var))))
+end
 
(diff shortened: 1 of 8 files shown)


AlexisRenchon (approved):
Looks good to me! I will adapt my future calibration PRs to this design. Thank you Kevin!

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -180,12 +189,12 @@ function compute_monthly_leaderboard(
         obs_var = ClimaAnalysis.shift_longitude(obs_var, -180.0, 180.0)
```
nefrathenrici, line 189: Should we still shift longitudes now that we already do it in preprocessing? Is this idempotent?

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -422,3 +446,35 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+    # ERA5 energy fluxes
+    era5_data_loader = ERA5DataLoader()
+    # ILAMB GPP, ecosystem respiration (FLUXCOM reco), and NEE
+    ilamb_data_loader = ILAMBDataLoader()
+
+    for short_name in available_vars(era5_data_loader)
+        obs_var_dict[short_name] = get(era5_data_loader, short_name)
+    end
+
+    # Override variables in ERA5 data loader
```
nefrathenrici: This is misleading - it's not overriding, just adding new variables to the dict.

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -1,412 +1,436 @@
...
+
+"""
+    preprocess(::ILAMBDataLoader, var, ::Val{:nee})
+
+Preprocess `var` for net ecosystem exchange from the NOAA CarbonTracker CT2022
+product.
+"""
+function preprocess(::ILAMBDataLoader, var, ::Val{:nee})
+    ClimaAnalysis.dim_units(var, "lon") == "degree" &&
+        (var.dim_attributes["lon"]["units"] = "degrees_east")
```
nefrathenrici: Nitpick, this could use `set_dim_units!` for consistency

nefrathenrici (approved):
Looks good, just nits!

ext/land_sim_vis/leaderboard/leaderboard.jl
ph-kev: It should be idempotent, but we shouldn't need this line.

PR #1748, opened 2026-05-18
Remove allocation and clipping of Y

kmdeck:
## Purpose 
This PR removes direct clipping of Y, which would prevent us from conserving carbon (and oxygen). It also removes one allocation of a 3D field. (Im not sure if that could be the culprit for the recent AMIP slowdown @juliasloan25 ? this allocation was added between 1.8.0 and 1.8.1)

I also adjusted some comments to reflect that some of the clipping, while allowing the simulation to carry on with "physical" values, is going to correspond to an incorrect solution to the equations (it will not agree with the solution you would get with a stable simulation). This is problematic because in these columns the solution becomes controlled by clipping etc and not by the physics. we can try and fix this in the future.

Ill try the long run!

## Content

----

### DIFF
--- a/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
+++ b/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
@@ -431,12 +431,11 @@ function ClimaLand.make_compute_exp_tendency(model::SoilCO2Model)
         # an O2 sink). For multi-stage RK integrators the bounds are slightly
         # conservative (per-stage step is a fraction of Δt) but still safe.
         if model.Δt > 0
-            inv_Δt = 1 / model.Δt # model.Δt is already type FT
             O2_f_atm = model.parameters.O2_f_atm
             @. dY.soilco2.O2_f = clamp(
                 dY.soilco2.O2_f,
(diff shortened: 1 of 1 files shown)


src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl
```diff
@@ -646,7 +645,7 @@ function ClimaLand.make_update_aux(model::SoilCO2Model)
         T_soil = soil_temperature(model.drivers.met, p, Y, t, z)
         θ_l = soil_moisture(model.drivers.met, p, Y, t, z)
         θ_i = soil_ice(model.drivers.met, p, Y, t, z)
-        Csom = @. lazy(max(Y.soilco2.SOC, 0))  # Clamp SOC to non-negative (lazy, no allocation)
+        Csom = Y.soilco2.SOC
```
kmdeck, line 648: I dont think the original line actually clipped Y here, but I am not sure.

AlexisRenchon (approved):
If the longrun looks okay, let's merge it! (I can have a look when it is out)

kmdeck:
> If the longrun looks okay, let's merge it! (I can have a look when it is out)

no NaNs!

PR #1749, opened 2026-05-18
Add developer guides

tapios:
## Purpose 

Adds DeveloperGuides as a subtree, plus a local AGENTS.md file with land-specific info. The subtree updates automatically monthly. 

Only one file here needs review: the local AGENTS.md in the repo root. The rest is copied from the DeveloperGuides repo and needs to be changed there. 

## Content

----


AlexisRenchon (approved):
I think this looks great. Thank you Tapio. As clearly written in AGENTS.md, `Local norms` will probably evolve over time, but this is a great start (I can't think of anything to add right now).

PR #1752, opened 2026-05-26
Add implicit stepping to biogeochemistry

kmdeck:
## Purpose 
Creates an implicit tendency for the biogeochemistry model with the diffusion terms. This allows us to remove some of the limiters in the model.

Note that source terms are treated explicitly still.

## Content
Current main for US-MOz
hr_timeseries
This branch:
hr_timeseries

----


src/integrated/land.jl
```diff
@@ -632,6 +639,7 @@ function make_update_implicit_cache(
             land.soil.parameters.earth_param_set,
         )
         update_imp_bf_soil!(p, Y, t)
+        update_imp_bf_soilco2!(p, Y, t)
```
kmdeck, line 640: I need to propagate these to the other integrated models

AlexisRenchon (approved):

PR #1754, opened 2026-05-28
Modify broadcast to work with ClimaCore bug

imreddyTeja:
## Purpose 
This is a temporary work-around for the regression caused by the new ClimaCore release.
## To-do
- alternatively, we can cap the ClimaCore version used in the tests. This should not impact the coupler, because I believe the issue is only with the Richards model

### DIFF
--- a/src/standalone/Soil/energy_hydrology.jl
+++ b/src/standalone/Soil/energy_hydrology.jl
@@ -535,9 +535,9 @@ function ClimaLand.make_compute_jacobian(model::EnergyHydrology{FT}) where {FT}
                 MatrixFields.LowerDiagonalMatrixRow(p.soil.topBC_scratch)
         end
         # dtγ can be an ITime or a float
+        negative_dtγ = FT(-float(dtγ))
         @. ∂ϑres∂ϑ =
-            FT(-1) *
-            float(dtγ) *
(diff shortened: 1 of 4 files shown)


ph-kev (approved):

tapios (approved):

juliasloan25:
@kmdeck does this change look ok to you? I think it's fine but I don't remember why we wrote it like that

imreddyTeja:
> @kmdeck does this change look ok to you? I think it's fine but I don't remember why we wrote it like that

can dt_gamma be an `ITime`?

kmdeck:
> @kmdeck does this change look ok to you? I think it's fine but I don't remember why we wrote it like that

I'm not sure which change you are referring to, but I assume the FT(-1)*dtgamma. I think we did it this way because it maps onto the way we write the tendency (-divergence(-Kgrad h), negative divergence of a flux), instead of cancelling out the signs. It might also matter in that we put a boundary condition on the flux, not minus the flux, and if we cancel out the signs, maybe our BC should be on the negative of the flux.

My only other thought is that we add dtgamma*Matrix + I, and I is unitless, so if dtgamma is ITime, does anything funky happen? Probably we should call float(dtgamma) as it is in the PR now

PR #1755, opened 2026-05-29
review kd/implicit_co2

AlexisRenchon:
Some fixes suggestions for the PR (IC and a stoichiometry factor), note that I also removed the docs apps files that now live in ClimaViz.jl (unrelated to this PR, but maybe that's ok?)

Possibly, we could add a small readme (or docs) to say that we can run single sites with:
- forcings: era5 (given coordinates of a location) or prescribed (e.g., a fluxnet site forcing),
- parameters: default (given coordinates of a location) or prescribed (e.g., known for a fluxnet site)

(I suggest this as I see you added `experiments/integrated/era5/single_site.jl` in this PR)

Of course, feel free to edit this branch before merging onto your branch, make the comments less verbose if you prefer, etc.


kmdeck (approved):
awesome, thank you!

PR #1757, opened 2026-05-29
Use CTS cache_imp! and simplify implicit cache update

kmdeck:
## Purpose 
- Updates the implicit cache once per iteration (instead of in the implicit tendency and the Jacobian) (3 fewer calls to update implicit cache relative to main). 
- Correctly classifies soil sublimation tendency as explicit (3 fewer calls relative to main)
- creates a function which only updates the implicit part of lsm_radiant_energy_fluxes.

SYPD on the snow land 2 year run goes from ~160 (https://buildkite.com/clima/climaland-long-runs/builds/4832) to 180.

Note that the plotting in the long run for snowy land has been fixed in main and will pass once I rebase (after making any review comment fixes).

## Content
We now use cache_imp! from ClimaTimesteppers to update the implicit cache once per iteration. Previously, our Jacobian update was:
"update jacobian"
-update implicit cache
-compute Jacobian (uses updated cache values)
And our "update implicit tendency" was:
-update implicit cache
-compute implicit tendency (uses updated cache values)

This then led to the implicit cache being updated 2x per iteration. I removed the functions "update implicit tendency" and "update Jacobian" and now we just have the "compute jacobian" and "compute implicit tendency" functions. Since we update the implicit cache via ClimaTimesteppers, we do not need to do so on the CliMA Land side anymore. For example, an equivalent solution might have been to keep the implicit cache update in the update implicit tendency function (but remove it from the update Jacobian function), but we don't need to recreate this functionality in CL.

The other changes should be pretty straightforward to review...

only canopy LW fluxes depend on canopy temperature and need to get updated each iteration, and all radiative fluxes for soil, snow, lake are explicit, so that is why I was able to split the lsm_radiation function as I did.

A future PR can look at using the CTS cache! functionality and change our explicit tendency similarly, but TBD on that.

----

### DIFF
--- a/docs/src/APIs/SolverFunctions.md
+++ b/docs/src/APIs/SolverFunctions.md
@@ -35,10 +35,8 @@ ClimaLand.boundary_var_types
 ## Implicit Tendencies
 
 ```@docs
-ClimaLand.make_imp_tendency
 ClimaLand.make_compute_imp_tendency
 ClimaLand.make_compute_jacobian
-ClimaLand.make_jacobian
 ClimaLand.initialize_jacobian
 ```
 
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -623,8 +623,8 @@ function make_update_implicit_cache(
         update_imp_aux_soil!(p, Y, t)
         update_imp_aux_soilco2!(p, Y, t)
         update_imp_aux_canopy!(p, Y, t)
-        # Radiation - updates Rn for soil, snow, lake also
-        lsm_radiant_energy_fluxes!(
+        # Implicit Radiation terms
+        implicit_radiant_energy_fluxes!(
             p,
             land,
             land.canopy.radiative_transfer,
@@ -642,7 +642,6 @@ function make_update_implicit_cache(
     end
     return update_implicit_cache!
 end
-
 
 """
     lsm_radiant_energy_fluxes!(p,land::LandModel{FT},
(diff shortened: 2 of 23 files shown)


docs/src/APIs/SolverFunctions.md
```diff
@@ -35,10 +35,8 @@ ClimaLand.boundary_var_types
 ## Implicit Tendencies
 
 ```@docs
-ClimaLand.make_imp_tendency
```
imreddyTeja, line 38: Should the removal of these be documented in NEWS.md?

src/integrated/land.jl
```diff
@@ -950,15 +1012,18 @@ NVTX.@annotate function snow_boundary_fluxes!(
         model.parameters.earth_param_set,
     )
 
+    residual_surface_flux =
+        Snow.get_residual_surface_flux(model.parameters.surf_temp, Y, p)
```
imreddyTeja, line 1016: This change should lower to the same thing as before, but the change makes it easier to read

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -914,7 +914,7 @@ of `ClimaLand.source!` for soil sublimation; treated implicitly
 in ϑ_l, ρe_int but explicitly in θ_i.
 """
 @kwdef struct SoilSublimation{FT} <: AbstractSoilSource{FT}
-    explicit::Bool = false
+    explicit::Bool = true
```
imreddyTeja: Is this docstring up to date? Why has this changed to explicit by default?

imreddyTeja (approved):
Looks good to me. Left some minor comments. Hopefully using `cache!`for the explicit tendency will be straightforwards

src/standalone/Soil/energy_hydrology.jl
kmdeck: I need to update the docstring - thank you for noticing.

All other fluxes at the boundary are treated explicitly now, so there is no reason sublimation should be updated each newton iteration

docs/src/APIs/SolverFunctions.md
kmdeck: Ill update - they are internal functions so hopefully dont affect any science users

PR #1759, opened 2026-06-03
Update Manifests to use ClimaCore 0.14.53

imreddyTeja:
This makes the buildkite runs use mask
aware tridiagonal solve

excludes ClimaCore 0.14.52 from the compat (so the conditional broken tests changes can be removed)

----


kmdeck (approved):
thanks! looks good

ph-kev (approved):

PR #1763, opened 2026-06-10
add Rosetta parameters as an option

kmdeck:
## Purpose 
This PR adds a new artifact to ClimaLand which contains the soil van genuchten parameters derived using ROSETTA. 

Note that to make some pixels in the Sahara stable under evaporation using these parameters, I had to change a parameter value, but that is not done in this PR. It can be done if we change to using the ROSETTA by default..

## Content

----

### DIFF
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -177,7 +177,7 @@ end
             LAI,
             toml_dict;
             prognostic_land_components,
-            soil_moisture_stress = Canopy.PiecewiseMoistureStressModel{FT}(domain, toml_dict),
+            soil_moisture_stress = Canopy.PiecewiseMoistureStressModel{FT}(domain, toml_dict; soil_params = (;ν = soil.parameters.ν, θ_r = soil.parameters.θ_r)),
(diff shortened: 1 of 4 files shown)


src/integrated/land.jl
```diff
@@ -177,7 +177,7 @@ end
             LAI,
             toml_dict;
             prognostic_land_components,
-            soil_moisture_stress = Canopy.PiecewiseMoistureStressModel{FT}(domain, toml_dict),
+            soil_moisture_stress = Canopy.PiecewiseMoistureStressModel{FT}(domain, toml_dict; soil_params = (;ν = soil.parameters.ν, θ_r = soil.parameters.θ_r)),
```
kmdeck, line 180: This makes it so we agree with the soil parameters by default. Otherwise, the default in the constructor for Canopy.PiecewiseMoistureStressModel is to use the Gupta parameter set, and if you create a soil model with the ROSETTA parameters, you'll get an error later on.

AlexisRenchon (approved):

PR #1764, opened 2026-06-10
Add CES emulate + sample pipeline

braghiere:
## Purpose
Extends the existing ClimaLand calibration pipeline with the two remaining steps of the [CalibrateEmulateSample.jl](https://github.com/CliMA/CalibrateEmulateSample.jl) (CES) UQ workflow — emulation and MCMC sampling — enabling full uncertainty quantification after a `ClimaCalibrate` run. Also adds an end-to-end site-level tutorial demonstrating the full CES pipeline with the ClimaLand land surface model.

## To-do
- Add user-facing documentation section for emulate + sample steps in `docs/src/calibration.md`

### Production pipeline scripts
- `experiments/calibration/emulate.jl`: Loads EKP checkpoints from a completed `ClimaCalibrate` run via `load_latest_ekp`, trains a GP emulator using `CalibrateEmulateSample.Emulators` (GPJL / GaussianProcess / `optimize_hyperparameters!`), and saves the emulator plus observation metadata to `ces_output_dir/emulators.jld2`. Reads config from the same env vars (`CALIBRATION_OUTPUT_DIR`, `CALIBRATION_CONFIG`) as `run_calibration.jl`.
- `experiments/calibration/sample.jl`: Loads the saved emulator and runs RWMH-MCMC via `CalibrateEmulateSample.MarkovChainMonteCarlo` (MCMCWrapper / `optimize_stepsize` / `sample` / `get_posterior`) to produce a full posterior over the calibrated parameters. Logs constrained posterior statistics (mean, std, 95% CI).
- `experiments/calibration/test/test_ces_pipeline.jl`: End-to-end integration test exercising the full Calibrate → Emulate → Sample pipeline without running the real land model, using a toy forward model G(emissivity) = 240 × emissivity. All 12 tests pass.

### Tutorial
- `docs/src/tutorials/calibration/perfect_model_site_level_ces.jl`: Single-site perfect-model CES tutorial using US-MOz FLUXNET forcing. Calibrates `Vcmax25` via EKI against synthetic LHF, trains and validates a GP emulator, and runs RWMH-MCMC to recover the true parameter. Uses physically realistic values (true `Vcmax25 = 1.5e-4` mol m⁻² s⁻¹ ≈ 150 μmol; prior `N(1e-4, 5e-5)` bounded `[0, 2e-4]`), in the LHF-sensitive regime (a sensitivity sweep confirmed LHF responds across the physical range and saturates only at unphysically high values). The posterior recovers the truth (mean ≈ 146 μmol vs. true 150; EKI MAP ≈ 154) and is much narrower than the prior (posterior std ≈ 18 μmol vs. prior 50; truth within the 95% CI [112, 181] μmol). Includes a posterior predictive check against the true model `G`, precomputed offline (50 posterior + 50 prior pushforwards) and committed as a static figure so it doesn't slow the docs build.
- `docs/list_tutorials.jl`: Registered under "Calibrating a ClimaLand model" as "Single site perfect model (CES/UQ)".

---


docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+site_ID_val = FluxnetSimulations.replace_hyphen(site_ID)
+
+(; time_offset, lat, long) =
+    FluxnetSimulations.get_location(FT, Val(site_ID_val))
+(; atmos_h) = FluxnetSimulations.get_fluxtower_height(FT, Val(site_ID_val))
+
+(start_date, stop_date) =
+    FluxnetSimulations.get_data_dates(site_ID, time_offset)
+stop_date = DateTime(2010, 4, 1, 6, 30)
+Δt = 450.0
```
kmdeck:
```suggestion
Δt = 450.0;
```

kmdeck: Note that any line of code that comes just before a markdown text or comment prints its output in the rendered tutorial. so you just need to add a semicolon after

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+
+# ## Domain and Forcing Setup
+
+zmin = FT(-2)
+zmax = FT(0)
+domain = Column(; zlim = (zmin, zmax), nelements = 10, longlat = (long, lat))
```
kmdeck:
```suggestion
domain = Column(; zlim = (zmin, zmax), nelements = 10, longlat = (long, lat));
```

kmdeck: here too

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+    for i in 1:N_iterations
+        println("EKI iteration $i / $N_iterations")
+        params_i = EKP.get_ϕ_final(prior, ensemble_kalman_process)
+        G_ens = hcat([G(params_i[:, j]...) for j in 1:ensemble_size]...)
+        EKP.update_ensemble!(ensemble_kalman_process, G_ens)
+    end
+end
+
+eki_map = EKP.get_ϕ_mean_final(prior, ensemble_kalman_process)[1]
+@info "EKI MAP estimate: Vcmax25 = $eki_map  (true: $true_Vcmax25)"
```
kmdeck, line 278: is a lot of this copied over/duplicate from the perfect_model_calibration script? We should be able to reuse the EKI file from that, as long as that tutorial is run first.  Might be worth it to speed things up. OTOH, I think the tutorials are now parallelized, so that might not be possible (@ph-kev ?)

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+ax2 = Axis(
+    fig2[1, 1];
+    title = "G ensemble: first vs last EKI iteration",
+    xlabel = "Hour of day",
+    ylabel = "LHF [W m⁻²]",
+)
+for g in eachcol(first_G)
+    lines!(ax2, 0:23, g; color = (:red, 0.4), linewidth = 1.5)
+end
+for g in eachcol(last_G)
```
kmdeck, line 314: on the plot, we don't see the blue line - is it under the black line?
https://clima.github.io/ClimaLand.jl/previews/PR1764/generated/calibration/perfect_model_site_level_ces/

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+    ],
+    ["First ensemble", "Last ensemble", "Truth"];
+    position = :rb,
+    framevisible = false,
+)
+CairoMakie.resize_to_layout!(fig2)
+CairoMakie.save("perfect_model_ces_G_first_last.png", fig2)
+# ![](perfect_model_ces_G_first_last.png)
+
+# ## Step 2 — Emulate (Gaussian Process Surrogate)
```
kmdeck, line 332: It would be good to look at the rendered docs here: https://clima.github.io/ClimaLand.jl/previews/PR1764/generated/calibration/perfect_model_site_level_ces/ to get a sense of how to clean up the formatting a bit (e.g. need more of the semicolons at the end of lines, perhaps some clearer headings and complete sentences)

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+# The EKI MAP gives a single point estimate; MCMC over the emulator gives
+# the full posterior credible interval.
+
+fig3 = CairoMakie.Figure(; size = (700, 450))
+ax3 = Axis(
+    fig3[1, 1];
+    title = "Posterior vs Prior: Vcmax25 (perfect model)",
+    xlabel = "Vcmax25 [mol m⁻² s⁻¹]",
+    ylabel = "Density",
+)
```
kmdeck, line 487: In order to include a plot, it has to be saved and then you need a line here adding it to the doc. See the other examples in this script

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+ax4 = Axis(
+    fig4[1, 1];
+    title = "GP emulator: posterior predictive",
+    xlabel = "Hour of day",
+    ylabel = "LHF [W m⁻²]",
+)
+for idx in post_indices
+    θ = unc_samples[:, idx]
+    pred_mean, _ = Emulators.predict(emulator, reshape(θ, :, 1))
+    lines!(
```
kmdeck: this plot doesn't seem to reflect all these curves. Are they identical? I only see the observations line in black

docs/src/tutorials/calibration/obs_site_level_ces.jl
```diff
@@ -0,0 +1,563 @@
...
+
+ax_g = Axis(
+    fig3[1, 2];
+    title = "Marginal posterior: g1",
+    xlabel = "g1 [Pa^0.5]",
+    ylabel = "Density",
+)
+density!(ax_g, prior_constrained[2, :]; label = "Prior", color = (:grey, 0.4))
+density!(ax_g, post_g1; label = "Posterior (CES-MCMC)", color = (:coral, 0.6))
+vlines!(
```
kmdeck: This plot doesn't seem right? it only shows the dashed line

kmdeck: oops, this comment is for the Vcmax25 plot only. the g1 one looks correct

docs/src/tutorials/calibration/obs_site_level_ces.jl
```diff
@@ -0,0 +1,563 @@
...
+)
+axislegend(ax_g; position = :rt, framevisible = false)
+
+CairoMakie.resize_to_layout!(fig3)
+CairoMakie.save("obs_ces_marginal_posteriors.png", fig3)
+# ![](obs_ces_marginal_posteriors.png)
+
+# ## Joint Posterior (Vcmax25 vs g1)
+#
+# The scatter plot of joint posterior samples reveals any correlation between
```
kmdeck: This plot looks really wrong for Vcmax25

docs/src/tutorials/calibration/obs_site_level_ces.jl
```diff
@@ -0,0 +1,563 @@
...
+pp_indices = rand(rng, 1:size(unc_samples, 2), n_pp)
+
+fig5 = CairoMakie.Figure(; size = (900, 400))
+ax5 = Axis(
+    fig5[1, 1];
+    title = "GP emulator posterior predictive vs FLUXNET",
+    xlabel = "Hour of day",
+    ylabel = "LHF [W m⁻²]",
+)
+for idx in pp_indices
```
kmdeck: only one curve is plotted?

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
ph-kev: I think it should be possible to reuse the EKI file, but I think it would require some work since you would need to specify dependencies between different part of the documentation.

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+# can distinguish the true parameter from the prior center. A value near zero
+# (e.g. 1e-4) would be in a regime where LHF is dominated by soil evaporation
+# and is nearly insensitive to Vcmax25, making calibration ill-posed.
+
+true_Vcmax25 = 1.5e-3
+observations = G(true_Vcmax25)
+## Noise variance: used for both EKI and the GP emulator likelihood so both
+## steps see a consistent observation error model.
+noise_var = 0.05
+noise_covariance = noise_var * EKP.I
```
odunbar: just checking that the LHF is on the O(50), and noise var is O(0.05) so corresponds to 0.5% noise, which is pretty small.

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
odunbar: It's basically under the black line, the noise tolerance would be around +\-0.25 on the axes from 0-80 so its hard to see

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+n_obs = length(observations)
+obs_noise_cov = noise_var * Matrix(LinearAlgebra.I, n_obs, n_obs)
+
+# Build, configure, and optimize the GP emulator.
+# GPJL uses GaussianProcesses.jl (squared-exponential kernel by default).
+# The Decorrelator inside Emulator projects the 24D output onto its principal
+# components so each GP only needs to model a 1D output.
+gppackage = GPJL()
+gauss_proc = GaussianProcess(gppackage; noise_learn = false)
+emulator = Emulator(gauss_proc, input_output_pairs; obs_noise_cov)
```
odunbar: The correct way of passing covariances is to use the encoder_kwargs argument (as the warning shows in the output)

odunbar:
```julila
encoder_kwargs = get_kwargs_from(ensemble_kalman_process, prior)
emulator = Emulator(gauss_proc, input_output_pairs; encoder_kwargs=encoder_kwargs)
```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
odunbar: If it is not a "real" value, and just synthetic, perhaps increase it a bit to make the visualizations more clear.

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,477 @@
...
+axislegend(ax3; position = :rt, framevisible = false)
+CairoMakie.resize_to_layout!(fig3)
+CairoMakie.save("perfect_model_ces_posterior.png", fig3)
+# ![](perfect_model_ces_posterior.png)
+
+# ## Posterior Predictive Check
+#
+# We evaluate the GP *emulator* (not the land model) at posterior samples to
+# check that the emulator's likelihood is consistent with the calibration
+# observations. This is deliberately cheap — each call to `Emulators.predict`
```
odunbar: I would be a bit cautious with this test. it is not really a posterior predictive check.

1) I'd advise moving verification of the GP emulator to be BEFORE sampling, as you are trying to ascertain if it is fit for purpose. Usually it is done by pushing forward samples and comparing with not just the truth, but also the whole training dataset, and possibly on a held-out validation set.  (I think you did this in another experiment with CES).

2) You can still perform posterior predictive checks after sampling. But these should be done with the TRUE model. As really the purpose of this test was to see if the emulator-based posterior is meaningful with the true forward map, which is what will be used for performing other downstream prediction tasks. Let's say the emulator is biased and this leads to posterior being biased, then this test will still show good results, but if you used the posterior with your original model again it would show the bias.

odunbar (dismissed):
Hey this looks great! 

I think my overriding comments would be that if the noise is somewhat artificial, then maybe inflate it a bit (maybe to O(1)? so that it is easier for the visualizations? Right now everything is very narrow.

Secondly, the posterior predictive check can't be done with an emulator, but you can repurpose the test you have done somewhat into a verification of the emulator (before sampling), applied to training data.

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
braghiere: Sounds good. The idea was to have a self-contained tutorial without dependencies but maybe we can think about linking both later?

kmdeck: yeah, it's okay for now. sounds complicated

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
+# # Perfect Model Site-Level CES Tutorial (Calibrate–Emulate–Sample)
+#
+# This tutorial extends the perfect model calibration experiment to the full
+# [CalibrateEmulateSample.jl](https://github.com/CliMA/CalibrateEmulateSample.jl)
+# (CES) workflow. After EKI calibration converges to a MAP estimate, we train a
+# Gaussian Process emulator on the ensemble trajectory, then run Markov Chain
+# Monte Carlo (MCMC) over the cheap surrogate to obtain the full posterior
```
kmdeck:
```suggestion
# Monte Carlo (MCMC) using the model emulator to obtain the full posterior
```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+# uncertainty distribution.
+#
+# The CES workflow is motivated by the fact that land model runs are expensive.
+# EKI calibration requires O(10 × iterations) model evaluations to find the MAP.
```
kmdeck: @odunbar is ~10 the right number here? should it be 2*p, where p is the number of parameters? or is it really min(~50, 2*p), so O(10) is reasonable?

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+# Full Bayesian UQ via direct MCMC would require O(100 000) evaluations — clearly
```
kmdeck: does this number depend on the size of the parameter space, how correlated they are, etc? should we add a citation?

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+# infeasible. CES solves this by:
+#   1. **Calibrate**: EKI explores parameter space efficiently (cheap).
```
kmdeck:
```suggestion
#   1. **Calibrate**: EKI explores parameter space efficiently and finds an optimal set of parameters.
```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+#   2. **Emulate**: Train a GP on the EKI trajectory (cheap once, reusable).
```
kmdeck:
```suggestion
#   2. **Emulate**: Train a GP on the EKI trajectory in parameter space (one time expense; reusable).
```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+#   3. **Sample**: MCMC over the GP — each evaluation costs microseconds.
```
kmdeck:
```suggestion
#   3. **Sample**: MCMC carried out using the GP in place of the model — each evaluation costs microseconds.
```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+using Logging
+using Random: Random
+using Dates
+
+# ## Configuration and Site Setup
+#
+# We fix the random seed for reproducibility, choose single-precision floats, and
+# load the default ClimaLand parameter set. The simulation targets the US-MOz
+# FLUXNET site; its location and flux-tower height are looked up from the bundled
+# site metadata. Trailing semicolons suppress the (verbose) output of these
```
kmdeck:
```suggestion
# site metadata.
```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+# assignments in the rendered tutorial.
```
kmdeck:
```suggestion

```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+
+(start_date, stop_date) =
+    FluxnetSimulations.get_data_dates(site_ID, time_offset);
+stop_date = DateTime(2010, 4, 1, 6, 30);
+Δt = 450.0;
+
+# ## Domain and Forcing Setup
+#
+# The model runs on a single soil column 2 m deep with 10 vertical elements,
+# anchored at the site's longitude and latitude.
```
kmdeck:
```suggestion
# located at the site's longitude and latitude.
```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+        FT,
+    )
+    LAI = ClimaLand.Canopy.prescribed_lai_modis(
+        domain.space.surface,
+        start_date,
+        stop_date,
+    )
+    Vcmax25 = FT(Vcmax25)
+    ground = ClimaLand.PrognosticGroundConditions{FT}()
+    prognostic_land_components = (:canopy, :snow, :soil, :soilco2)
```
kmdeck:
```suggestion
    prognostic_land_components = (:canopy, :snow, :soil)
```

^this will make it a bit faster, and we dont need soilco2 for this example..

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+        )
+    return observation
+end
+
+# ## Step 1 — Calibrate (Ensemble Kalman Inversion)
+#
+# Generate synthetic "truth" from the model at a known true parameter value,
+# then run EKI to recover it from noisy observations.
+#
+# The true value is set to 1.5e-3 mol m⁻² s⁻¹ — one standard deviation above
```
kmdeck: Is this explanation necessary? "A value near zero
# (e.g. 1e-4) would be in a regime where LHF is dominated by soil evaporation
# and is nearly insensitive to Vcmax25, making calibration ill-posed."

Have we actually tested that this is the case, or is just an AI musing?

kmdeck: is this value reasonable for plants? (1.5e-3)?

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+input_output_pairs =
+    Utilities.get_training_points(ensemble_kalman_process, n_iter_used);
+
+# Next we build, configure, and optimize the GP emulator. `GPJL` uses
+# GaussianProcesses.jl (squared-exponential kernel by default). The decorrelator
+# inside `Emulator` projects the 24-dimensional output onto its principal
+# components so that each GP only needs to model a one-dimensional output.
+# `encoder_kwargs` bundles the noise covariance and prior covariance from the
+# calibration step — the recommended way to configure the `Emulator` encoder.
+# Hyperparameter optimization is wrapped in a quiet logger to keep the rendered
```
kmdeck:
```suggestion

```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+# tutorial readable.
```
kmdeck:
```suggestion

```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+mcmc = MCMCWrapper(
+    RWMHSampling(),
+    Float64.(observations),
+    prior,
+    emulator;
+    init_params,
+);
+
+# We first tune the step size to target a ~0.2 acceptance rate, then draw the
+# samples. Both calls emit progress bars, so we run them inside a quiet logger to
```
kmdeck:
```suggestion

```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+# keep the rendered tutorial clean.
```
kmdeck:
```suggestion

```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+post_std = std(post_Vcmax25);
+ci_lo, ci_hi = quantile(post_Vcmax25, [0.025, 0.975]);
+
+@info "Posterior Vcmax25: mean=$(round(post_mean; sigdigits=4))  std=$(round(post_std; sigdigits=3))"
+@info "95% CI: [$(round(ci_lo; sigdigits=4)), $(round(ci_hi; sigdigits=4))]"
+@info "EKI MAP: $(round(eki_map; sigdigits=4))  true: $true_Vcmax25"
+
+# ## UQ Results: Posterior vs Prior vs True Value
+#
+# The key result: the posterior is much narrower than the prior and centered
```
kmdeck:
```suggestion
# Here we show the key results from this tutorial. First, we have obtained a probability distribution for the parameter we calibrated, which gives use a quantified uncertainty in its value. Second, the posterior is much narrower than the prior, indicating that the data provided useful constraints on this parameter. Additionally, the posterior is centered near the true value.
```

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,550 @@
...
+    label = "EKI MAP",
+)
+axislegend(ax3; position = :rt, framevisible = false)
+CairoMakie.resize_to_layout!(fig3)
+CairoMakie.save("perfect_model_ces_posterior.png", fig3);
+# ![](perfect_model_ces_posterior.png)
+
+# ## Posterior Predictive Check
+#
+# Evaluate the TRUE forward model G at a small number of posterior samples.
```
kmdeck: I'm a little confused by this. We already validated that Emulator(theta) agrees with Model(theta) in the Validation section above. If that is the case, is this showing something additional?

kmdeck: or rather, why is that previous check not sufficient?

kmdeck (approved):
I left a couple other comments. You need to rebase now before the docs will build again, since the main branch has diverged, and then as long as the rendered tutorial looks good Im good with merging it!

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
braghiere: @odunbar

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
odunbar: I would maybe even just link our recommendations

https://clima.github.io/EnsembleKalmanProcesses.jl/dev/defaults/

odunbar: in theory it is actually not tied directly to the number of parameters unless it is `Unscented` or `TransformUnscented` where it is _exactly_ 2p+1

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
odunbar: It gets technical. but yes broadly it depends on how correlated the posterior is (though not necessarily dimension) and how suitable the chain is for mixing. 

I think the key here is just that we can relax what we say: rather than saying "would require". Say "MCMC mixing and convergence commonly require in excess of O(10^4) and up-to O(10^7) depending on the Markov Chain mixing properties". 

Typical references might be (Handbook of Markov Chain Monte Carlo Ch 4.), or (Bayesian Data Analysis Ch 11.)

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
odunbar: Also i wouldn't say "cheap" for the EKI. it is is relatively the most expensive part? Likewise the emulator is also not cheap to train, but yes, it is a one-off cost

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
odunbar: The section before is about validating the emulator to be close to the true model on it's training data. But this just measures the error between outputs of the models in L^2 which is not the same as the error between the true and emulator-based posterior. 

Therefore pushing the emulator-based posterior samples through the true forward map can give you a way to measure whether your posterior is reasonable (by e.g. seeing how likely the truth is under the predictive posterior).

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
braghiere: I ran a quick sweep and it's actually backwards: LE is most sensitive to Vcmax25 in the physical range and only goes flat at the unrealistically high values. Removed the passage.
Also 1500 μmol is about 10–90x too high (realistic is ~17–150 μmol). Re-centered to true = 1.5e-4 and prior ~N(1e-4, 5e-5).

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
braghiere: Updated to Ollie's wording: "commonly requires in excess of O(10⁴) up to O(10⁷) depending on MCMC mixing properties."

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,543 @@
...
+    prior,
+    emulator;
+    init_params,
+);
+
+# We first tune the step size to target a ~0.2 acceptance rate, then draw the
+n_samples = 50_000;
+discard_initial = 2_000;
+new_step = Logging.with_logger(SimpleLogger(devnull, Logging.Error)) do
+    optimize_stepsize(mcmc; init_stepsize = 0.1, N = 2_000, discard_initial = 0)
```
odunbar: the stepsize found was O(3) (via a doubling criteria). Setting `init_stepsize=1.0` or even `3.0` would make this take far less time. 

Can you also not print this out directly, as these long progress bars mess up the visuals? Perhaps just print the `init_step` and the learnt `new_step`.

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
```diff
@@ -0,0 +1,543 @@
...
+
+# ## Posterior Predictive Check
+#
+# Evaluate the TRUE forward model G at a small number of posterior samples.
+# This is the proper posterior predictive check: it uses the actual land model
+# (not the emulator) to validate that the emulator-based posterior is meaningful.
+# If the emulator were biased, the MCMC posterior could be biased too — but this
+# check would reveal the discrepancy since G itself is unbiased.
+
+n_pp = 3
```
odunbar: We aren't going to see much useful information here with 3 samples.

odunbar: My preference would be e.g. to do like 50 with an offline computation and just add the figure here, with a note that it is not computed on-the-fly. Then people can recompute this themselves and they have an idea of what is happening but it doesn't slow the docs. 

You could also then push forward 50 prior samples too and plot the compactification of the posterior relative to the prior

If you need to keep costs down for the example, then maybe just put the code in here for `n_pp=50` but don't run it, or make a statement that the users should do more than 3?

odunbar (commented):
Final comments perhaps!

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
braghiere: @kmdeck any preference for this? offline figures is ok? do we do that anywhere else?

docs/src/tutorials/calibration/perfect_model_site_level_ces.jl
braghiere: step-size now starts at 1.0 and the progress bars are suppressed; the posterior predictive is now a precomputed 50-posterior + 50-prior figure (offline, with the code included to regenerate), showing the compactification vs the prior.

PR #1766, opened 2026-06-12
Snow parameterizations helping permafrost

kmdeck:
## Purpose 
Adds a simple parameterization for the bottom temperature of the snow (not optional), modularizes the thermal conductivity model of the snow, and adds a second parameterization option for this (Sturm).

This typically makes the bottom temperature warmer than the bulk temperature, which improves snow insulation over the soil, and has a big effect on permafrost locations. The Sturm conductivity model is also lower than the Jordan parameterization in general, and this helps with permafrost.

## Impact
Main:
image

Bottom alone:
image

Sturm + Tbottom:
image

## Content
Adds bottom snow temp as a cache variable to integrated models with soil and snow, and computes it during the ground heat flux computation

Adds another parameterization for thermal conductivity

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,17 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+- ![][badge-✨feature] Add selectable snow thermal conductivity parameterizations
+  (`JordanSnowConductivityModel`, default, after Jordan 1991; and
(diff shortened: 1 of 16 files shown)


AlexisRenchon (approved):
See suggestions in #1769 which is mostly some cleanup suggestions - feel free to add or not! (edit it as needed!)

PR #1769, opened 2026-06-16
Review fixes: snow conductivity/bottom-temp diagnostics, docs, and tests

AlexisRenchon:
- Fix a diagnostic name mismatch: the LandModel possible‑diagnostics list named `"snow_tbot"`
  while the diagnostic is registered as `"snowtbot"`. As written the snow‑bottom‑temperature
  diagnostic was unreachable and `default_diagnostics(land_model; output_vars = :long)` would
  error with `"diagnostic snow_tbot does not exist"`. Renamed to `"snowtbot"`.

- Make the `ghf` and snow‑conductivity diagnostics reachable (they were registered but never
  added to any possible‑diagnostics list). Renamed the non‑ASCII short name `snowκ` → `snowk`
  (it was the only non‑ASCII diagnostic name and would become a non‑ASCII NetCDF variable),
  keeping the compute method consistent.
- Add `units` to the new diagnostics (`K` for the snow temperatures, `W m^-1 K^-1` for `snowk`,
  `W m^-2` for `ghf`) and a `comments` field to `ghf`.

- Fix a duplicated `F_g =  F_g =` in the `update_soil_snow_ground_heat_flux!` docstring and
  restore the explanation of why `Δz_snow` is capped at 10 cm.
- Remove the vestigial, unused `κ_ice` keyword from the `SnowParameters(toml_dict, Δt; …)`
  constructor (it was silently ignored after the `κ_ice → κ_snow` refactor).
- Expand and reflow the `snow_T_bottom` docstring (math in code fences so it renders) and add it
  to the API docs (`docs/src/APIs/ClimaLand.md`).

**Tests**
- `SturmSnowConductivityModel` conductivity test across the low/high/capped density branches.
- `snow_T_bottom` unit tests covering the closed form, both analytic limits, and the
  freezing‑point cap.
- A diagnostics regression test asserting the snow diagnostics (`snowtbot`, `ghf`, `snowk`,
  `snowtsfc`, `snowtb`) are reachable for a `LandModel`.

**NEWS.md** — added entries for the feature, the breaking `κ_ice → κ_snow` change, and the
ground‑heat‑flux behavioral change.

### Open question for the author

The Sturm low‑density slope `m1 = 0.20` looks slightly inconsistent with the non‑dimensionalization
used for the other coefficients: the threshold (`0.17`), `m2` (`−0.93 ≈ −1.01·ρ_ice`) and `q2`
(`2.72 ≈ 3.233·ρ_ice²`) all correspond to `ρ_ice ≈ 0.917 g/cm³`, which would give
`m1 = 0.234·ρ_ice ≈ 0.215` (and slightly better continuity at the threshold). Worth double‑checking
against the source. It only affects fresh, low‑density snow, so I did not change the parameter.

### DIFF
--- a/src/diagnostics/default_diagnostics.jl
+++ b/src/diagnostics/default_diagnostics.jl
@@ -623,7 +623,9 @@ function get_possible_diagnostics(model::LandModel)
         "nee",
         "cveg",
         "tr",
-        "snow_tbot",
+        "snowtbot",
+        "snowk",
+        "ghf",
     ]
 
     return unique!(append!(component_diagnostics, additional_diagnostics))
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,17 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+- ![][badge-✨feature] Add selectable snow thermal conductivity parameterizations
+  (`JordanSnowConductivityModel`, default, after Jordan 1991; and
+  `SturmSnowConductivityModel`, after Sturm et al. 1997) and a piecewise-linear
+  snow-bottom-temperature parameterization (`snow_T_bottom`) used to compute the
+  snow-soil ground heat flux. Adds the `snowtb`, `snowtsfc`, `snowtbot`, `snowk`, and
+  `ghf` diagnostics. PR [#1766](https://github.com/CliMA/ClimaLand.jl/pull/1766)
+- ![][badge-💥breaking] The `SnowParameters` field `κ_ice` is replaced by `κ_snow`, an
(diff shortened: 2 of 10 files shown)


src/diagnostics/default_diagnostics.jl
```diff
@@ -623,7 +623,9 @@ function get_possible_diagnostics(model::LandModel)
         "nee",
         "cveg",
         "tr",
-        "snow_tbot",
+        "snowtbot",
+        "snowk",
```
AlexisRenchon, line 627: Not sure if you want these here... you can remove

kmdeck: sure! thank you

PR #1770, opened 2026-06-16
Distinguish between manual and scheduled ILAMB runs

ph-kev:
For the ILAMB leaderboard, the scheduled ILAMB runs did not update the latest ILAMB run. The current latest ILAMB run is `4598_2026-03-05_latest` and the other ILAMB run is `4864_2026-06-16_58913e6`. This PR makes it so any scheduled ILAMB run will update the latest ILAMB run.

To do this, this PR introduces the constants `NUM_MANUAL_MODEL_RUNS ` and `NUM_SCHEDULED_MODEL_RUNS` which keep track of how many directories to retain when deleting old directories. Both constants are set to 1, so the scheduled run will override the last scheduled run. In addition, any new manual run will override the old manual run.

### DIFF
--- a/experiments/ilamb/ilamb_setup.jl
+++ b/experiments/ilamb/ilamb_setup.jl
@@ -1,6 +1,10 @@
 import Dates
 
-const NUM_MODELS_TO_KEEP = 3
+# Total number of manual model runs to keep
+const NUM_MANUAL_MODEL_RUNS = 1
+
+# Total number of scheduled model runs to keep
+const NUM_SCHEDULED_MODEL_RUNS = 1
 
 """
     setup_run_for_ilamb(
@@ -9,7 +13,8 @@ const NUM_MODELS_TO_KEEP = 3
         ilamb_build_dir::String,
         buildkite_id::String,
         commit_id::String,
-        num_models_to_keep::Integer,
+        num_manual_model_run_to_keep::Integer,
+        num_scheduled_model_run_to_keep::Integer,
     )
 
 Set up the ILAMB diagnostics produced from the run for ILAMB.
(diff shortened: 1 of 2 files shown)


experiments/ilamb/ilamb_setup.jl
```diff
@@ -143,23 +152,74 @@ function validate_symlinks(MODELS_dir::String)
 end
 
 """
-    delete_old_runs(MODELS_dir::String, N::Integer)
+    delete_old_model_runs(MODELS_dir::String, N::Integer)
+
+Delete model directories produced from manual model runs in `MODELS_dir` until
```
kmdeck, line 157: Does this just delete the ILAMB output or also the global_diagnostics/output_0000 files?

ph-kev: This delete the model directories, so that it doesn't appear in the ILAMB leaderboard. The model directories consist of symlinks to NetCDF files produced from the ClimaLand simulation, so the original NetCDF files are not deleted.

experiments/ilamb/ilamb_setup.jl
```diff
@@ -1,6 +1,10 @@
 import Dates
 
-const NUM_MODELS_TO_KEEP = 3
+# Total number of manual model runs to keep
+const NUM_MANUAL_MODEL_RUNS = 1
```
nefrathenrici, line 4: Should we keep more than 2 model runs total? I'm not sure how this is used, but it might be nice to be able to see leaderboards going further back.

nefrathenrici (approved):
This looks good to me, is this needed to ensure that we keep records of both type of run?

ph-kev:
> This looks good to me, is this needed to ensure that we keep records of both type of run?

Yes, this is needed to keep records of both type of runs: manual runs and scheduled runs.

experiments/ilamb/ilamb_setup.jl
ph-kev: The current behavior only keep two model runs, so we are keeping the same behavior as before.

kmdeck (approved):
thanks, Kevin!

PR #1774, opened 2026-06-22
Use Newton update to solve for F(T) = 0 in flux balance for snow rather than nested iterative solves; CL 1.11.0, use SF >=1.2.1

### LINKED ISSUE #1727: Snow surface temperature parameterization results in huge slow down
Not quite a bug, but using the surface temperature parameterization for snow slows down our 2 year long run from ~96 SYPD to ~65 SYPD.  The parameterization solves for the skin temperature of the snow, effectively, using a root solver.

We should be calling the parameterization once per step. How can it be this slow? Note that we solve for canopy temperature as a prognostic variable (implicitly) using Newton iterations to solve for the temp at the next timestep. I don't think that is as slow as this, perhaps because we restrict to 3 iterations.

kmdeck:
## Purpose 

CL 1.11
SF 1.2.1
Previously, we used nested iterations to solve for the snow surface temp using a flux balance. The outer iterations were carried out using secant method, the inner with whatever SF uses. In principle a single location not converging will lead to 8 iterations (max_iter = 8) of the secant method everywhere, each of which calls surface fluxes, itself an iterative solve with 7 iterations.

This PR changes the snow model to use the surface flux option of updating the temperature and humidity at the surface using functions provided by us. It uses a newton update to update T_sfc each SF iterations.

I also changed the initial guess to be the current T_sfc instead of T_atmos; this requires initializing the field the first step (otherwise we would pass it p.snow.T_sfc as a guess, but this has not been set to anything yet). I did this by making/changing methods of set_initial_cache for Snow/integrated models with Snow

## Content

----


src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -714,21 +714,56 @@ function flux_balance(
...
+        ρ_sfc,
+        Thermodynamics.Ice(),
+    )
+    qsat_over_liq = Thermodynamics.q_vap_saturation(
+        thermo_params,
+        T_sfc_guess,
+        ρ_sfc,
+        Thermodynamics.Liquid(),
+    )
+    q_sfc = qsat_over_ice * (1 - q_l) + q_l * (qsat_over_liq)
```
kmdeck, line 792: This is a bit dangerous because it assumes a functional form for snow surface specific humidity, but it also saves computing qsat 2 extra times.

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -828,7 +865,11 @@ function solve_for_surface_temp_at_a_point(
     )
 
     #Starting points picked to minimize number of failed convergences:
-    method = SecantMethod(T_atmos, T_atmos - FT(1))
+    if T_last < eps(FT)
```
kmdeck: this is basically for the first step, when T_sfc has not been computed yet, and is still zero after initialization..

src/integrated/soil_canopy_model.jl
```diff
@@ -571,3 +571,29 @@ function ClimaLand.get_drivers(model::SoilCanopyModel)
         model.canopy.boundary_conditions.radiation,
     )
 end
+
+"""
+    make_set_initial_cache(model::LandModel)
```
akshaysridhar:
```suggestion
    make_set_initial_cache(model::SoilCanopyModel)
```

src/integrated/land.jl
akshaysridhar: We now have `function make_set_initial_cache(model::LandModel)` and `function make_set_initial_cache(model::SoilCanopyModel)` so the dispatch logic should be updated in the docs as well

.buildkite/Manifest-v1.12.toml
```diff
@@ -3311,9 +3311,11 @@ version = "7.8.3+2"
 
 [[deps.SurfaceFluxes]]
 deps = ["RootSolvers", "Thermodynamics"]
-git-tree-sha1 = "fdd6e6a5678c4f81a46a55d03c24a1755817e95a"
+git-tree-sha1 = "df5b68a4c1a58541b81c469afd1604db4dca1b7b"
+repo-rev = "as/tsfc-solves-immutable"
```
akshaysridhar: Relax branch requirement

kmdeck: upgrade to SF 1.2.0!

src/standalone/Snow/snow_parameterizations.jl
```diff
@@ -654,111 +659,197 @@ function compute_extra_prog_tendency!(
...
+
+It makes this estimate by incrementing the initial guess for snow surface temperature
+`T_0` (stored in `inputs`) by the Newton update ΔT, where `ΔT = -f(T_0)/f'(T_0)` and
+f(T) = SW_n + LW_n(T) + H(T) + L(T) +κ(T-T̄)/d = 0.
+
+Be aware that if the snow surface specific humidity parameterization changes, 
+we must also change the internals of this function.
+
+"""
+function update_T_sfc_scheme(
```
akshaysridhar, line 753: Is there still a blocker in SurfaceFluxes w.r.t the snow parameterization ? I see some buildkite GPU `runtests` failed

akshaysridhar (approved):

PR #1776, opened 2026-06-24
CLM parameters not defined everywhere on land

kmdeck:
## Purpose 
The CLM parameters are not defined where their model has determined that there is no vegetation. However, MODIS LAI may be inconsistent with this. Because of this, we define these missing but over land parameters to be
- equal to the PFT parameters of broadleaf deciduous temperature if available
- equal to the median value of land values if PFT value unavailable
- C3 fraction = 1
- Omega = 1
- rooting depth was already nonzero everywhere and I believe was set to a mean over land.

## Content

----


kmdeck:
@daneelnp this seems to be working in the long run:https://buildkite.com/clima/climaland-long-runs/builds/4884#019efbd9-ee80-48be-a5f0-6e46a7efa716

Ill try to merge this before I leave tomorrow morning, but if not, please feel free to push to it/merge it and then rebase your branch off of main

daneelnp (approved):

braghiere (approved):

PR #1779, opened 2026-06-29
use fixed artifacts related to rosetta parameters

kmdeck:
## Purpose 
I had saved the rosetta soil parameters in the wrong order in depth. Two PRs in ClimaArtifacts fixed this issue:
PR https://github.com/CliMA/ClimaArtifacts/pull/173 fixed the bug and remakes the parameter file
PR https://github.com/CliMA/ClimaArtifacts/pull/167 recreates the IC based on these parameters

This PR then just makes it so we use those parameters now

## Content

----


EvaMarieMetz (approved):

AlexisRenchon (approved):

felixcremer:
This seem to have broken the artifact download for versions before 1.10 because the file location is the same, but the expected sha changed and this throws an error from within Pkg.

kmdeck:
> This seem to have broken the artifact download for versions before 1.10 because the file location is the same, but the expected sha changed and this throws an error from within Pkg.

Thank you for pointing this out @felixcremer! This is a problem and likely means all of our earlier releases no longer will work. I am checking with @ph-kev and @nefrathenrici about how best to address this. In the meantime, is it possible for you to upgrade versions?

felixcremer:
Yes, we managed to upgrade versions. I just wanted to let you know because the error message was not obvious and this might affect others as well.

PR #1780, opened 2026-06-29
Remove short-performance pipeline

imreddyTeja:
This uses a deprecated plugin, and nobody is looking at the results because it is not required for merging.

<

----


ph-kev (approved):

kmdeck (commented):
fair enough! thanks

PR #1782, opened 2026-06-29
remove SAI from energy tendency, LW_u, turbulent fluxes

kmdeck:
## Background 
Accounting for stems/roots is important for modelling autotrophic respiration during the winter. We had also included SAI in the thermal mass of the canopy (appearing in the tendency for canopy energy), in the long wave emission, and in the sensible heat flux. The thermal mass contribution should be negligible. The latter too may matter at some level. However, on main, we had always run with SAI = 0, so we did not check the effects of this. PR #1707  tried running with nonzero SAI, but found it worsened energy fluxes. More recently, changes to improve autotrophic respiration led to nonzero SAI, but this also has a negative impact on energy fluxes, in particular, SHF.

Additionally, it may be incorrect to model the effect of trunks on SHF, LW_net, but not SW_net, which we had not been doing. 

See the discussion here: https://github.com/CliMA/ClimaLand.jl/pull/1707

## Purpose 
This PR removes SAI from any energy flux computations and from the tendency for temperature.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+- ![][badge-🔥behavioralΔ] Remove SAI from energy fluxes, tendency PR [#1782](https://github.com/CliMA/ClimaLand.jl/pull/1782)
 - ![][badge-🔥behavioralΔ] Set the `optimal_lai_z`/`optimal_lai_sigma`/`optimal_lai_alpha`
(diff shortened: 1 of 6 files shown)


AlexisRenchon:
image

maybe the calibrated params aren't the best fit for this anymore? 
but we should wait the new IC and corrected (good direction) rosetta before recalibrating

AlexisRenchon (approved):

AlexisRenchon (approved):

PR #1783, opened 2026-06-30
Fix prognostic-LAI canopy height NaN and missing component constructors

AlexisRenchon:
Extracts three self-contained fixes from `ar/calibrate_inversion_nee` onto a clean branch off `main`.

## Changes

- **`src/standalone/Vegetation/Canopy.jl`**
  - `ZhouOptimalLAIModel`: default `height` to the constant `canopy_height` from the TOML dict instead of the spatially-varying CLM canopy-height field. The CLM field is ~0 at barren/alpine cells, which makes the leaf-water storage `LAI * height` vanish in the `ϑ_l` tendency, so the transpiration/root fluxes drive `ϑ_l` (and the coupled canopy/soil state) to NaN once prognostic LAI greens up.
  - Add a `PModelConductance{FT}(domain, toml_dict; Drel)` method delegating to the existing `(toml_dict; Drel)` constructor, matching the uniform `(domain, toml_dict)` component signature that the `CanopyModel` constructor already invokes (currently a latent `MethodError`).

- **`src/integrated/land.jl`**
  - Drop the erroneous 4th positional `Δt` argument passed to `SoilCO2Model{FT}` in both `LandModel` constructors. `SoilCO2Model` has no method taking `Δt`, and the other call sites (`soil_canopy_model.jl`, `soil_energy_hydrology_biogeochemistry.jl`) already omit it.

- **`Artifacts.toml`**
  - Bump the `optimal_lai_inputs` artifact (new tree hash / sha256 / url).

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -248,10 +248,10 @@ git-tree-sha1 = "c35ba0e899040cb8153226ab751f69100f475d39"
     url = "https://caltech.box.com/shared/static/tn1bnqjmegyetw5kzd2ixq5pbnb05s3u.gz"
 
 [optimal_lai_inputs]
-git-tree-sha1 = "cba8e02b52d8ebbde04a3fbb3269d775a6a08cc3"
+git-tree-sha1 = "2ce5e5e05e4f17c86b07eadff1f9dd551e779524"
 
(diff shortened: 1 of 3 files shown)


kmdeck (approved):
looks good!

PR #1794, opened 2026-07-06
Calibrated optimal-LAI defaults + prognostic-LAI long run option

AlexisRenchon:
## Summary

Adds the calibrated optimal-LAI parameters as model defaults, makes the P-model snowy-land long run able to run with **prognostic (modeled) LAI**, and adds an **inversion-obs leaderboard** to the long run — wired into weekly CI and a new on-demand label.

## Changes

**1. Calibrated optimal-LAI defaults** — `toml/default_parameters.toml`

Calibrated `ZhouOptimalLAIModel` against MODIS LAI (Yuan et al. 2017) with `experiments/calibration/configs/lai.jl` (TransformUnscented, constrained ensemble mean, converged over 4 iterations):

| param | old (prior mean) | new (calibrated) |
|---|---|---|
| `optimal_lai_z` | 12.227 | **21.4** |
| `optimal_lai_sigma` | 1.1 | **0.939** |
| `optimal_lai_alpha` | 0.067 | **0.0701** |

Only affects prognostic-LAI runs; the prescribed-LAI default is unchanged.

**2. Prognostic-LAI option** — `experiments/long_runs/snowy_land_pmodel.jl`

New `PROGNOSTIC_LAI` env-var switch (mirrors `LONGER_RUN` / `UNCALIBRATED`). Unset → prescribed MODIS LAI (unchanged). Set → the canopy computes LAI from the P-model potential GPP via the no-LAI `LandModel` convenience constructor (`ZhouOptimalLAIModel`, #1783). Prognostic output goes to a separate `snowy_land_pmodel_opt_lai_longrun_*` directory.

**3. INVERSION leaderboard** — `ext/land_sim_vis/leaderboard/*`, `ext/LandSimulationVisualizationExt.jl`

- Adds the inversion carbon leaderboard: the long run now compares the model's NEE/GPP/ER against the inversion product (CT2022 NEE, GOSIF GPP, residual ER) from the `inversion_nee` artifact, alongside the ERA5 and ILAMB leaderboards.
- The leaderboard also compares the model `lai` against the MODIS LAI target (Yuan et al. 2017), included for **all** runs: for prescribed-LAI runs the panel is ~1:1 (model `lai` equals the prescribed MODIS obs), and for prognostic-LAI runs it is an independent check. _(The earlier prognostic-only `include_lai` gating was removed in review round #1.)_

**4. CI + labels** — `.buildkite/longruns_gpu/pipeline.yml`

The two labels are now **independent — add just one**:
- `Run long runs` → the prescribed long run (as before).
- `long run opt lai` → the prognostic long run.

The **weekly scheduled build runs both**. (The default group's `if` short-circuits on `build.source == "schedule"` so the weekly run is unaffected by the label logic.)

**5. New GitHub label** — `long run opt lai` (green `#0e8a16`), created via `gh`.

### DIFF
--- a/.buildkite/longruns_gpu/pipeline.yml
+++ b/.buildkite/longruns_gpu/pipeline.yml
@@ -39,8 +39,12 @@ steps:
 
   - wait
 
+  # The standard (prescribed-LAI) long runs. Run on the weekly scheduled build,
+  # on manual builds, and when the "Run long runs" label is present — but skip
+  # when a PR is labeled ONLY "long run opt lai" (that label runs just the
+  # prognostic-LAI group below), so each label triggers exactly one mode.
   - group: "Global Land Models"
-    if: build.env("LONGER_RUN") == null && build.env("PERTURBED_RUN") == null
+    if: build.env("LONGER_RUN") == null && build.env("PERTURBED_RUN") == null && (build.source == "schedule" || !(build.pull_request.labels includes "long run opt lai") || build.pull_request.labels includes "Run long runs")
     steps:
 
       - label: ":snow_capped_mountain: Snowy Land + P Model"
@@ -81,6 +85,28 @@ steps:
           slurm_time: 00:30:00
         env:
           CLIMACOMMS_DEVICE: "CUDA"
+
+  # Prognostic (modeled) LAI run of snowy_land_pmodel (ZhouOptimalLAIModel),
+  # selected via the `PROGNOSTIC_LAI` env var. Runs on the weekly scheduled
+  # build so the standard weekly long runs cover BOTH prescribed and prognostic
+  # LAI, and on demand when a PR carries the "long run opt lai" label.
+  - group: "Optimal LAI Global Land Model"
+    if: build.env("LONGER_RUN") == null && build.env("PERTURBED_RUN") == null && (build.source == "schedule" || build.pull_request.labels includes "long run opt lai")
+    steps:
+
+      - label: ":seedling: Snowy Land + P Model (Optimal LAI)"
+        key: "snowy_land_pmodel_opt_lai_run"
+        command:
+          - julia --color=yes --project=.buildkite experiments/long_runs/snowy_land_pmodel.jl
+        artifact_paths:
+          - "snowy_land_pmodel_opt_lai_longrun_gpu/*png"
+          - "snowy_land_pmodel_opt_lai_longrun_gpu/*pdf"
+        agents:
+          slurm_gpus: 1
+          slurm_time: 3:00:00
+        env:
+          CLIMACOMMS_DEVICE: "CUDA"
+          PROGNOSTIC_LAI: ""
 
   - group: "Perturbed Low-res Global Land Models"
     if: build.env("PERTURBED_RUN") != null
--- a/experiments/long_runs/snowy_land_pmodel.jl
+++ b/experiments/long_runs/snowy_land_pmodel.jl
@@ -53,11 +53,18 @@ const LONGER_RUN = haskey(ENV, "LONGER_RUN") ? true : false
 # `export UNCALIBRATED=""` in the terminal and run this script, or
 # pass `UNCALIBRATED=""` as an environment variable on buildkite.
 const UNCALIBRATED = haskey(ENV, "UNCALIBRATED") ? true : false
+# If you want to run with prognostic (modeled) LAI from the optimal-LAI model
+# (Zhou et al. 2025, `ZhouOptimalLAIModel`) instead of prescribed MODIS LAI,
+# type `export PROGNOSTIC_LAI=""` in the terminal and run this script, or pass
+# `PROGNOSTIC_LAI=""` as an environment variable on Buildkite. The default
+# (unset) prescribes MODIS LAI.
+const PROGNOSTIC_LAI = haskey(ENV, "PROGNOSTIC_LAI") ? true : false
 context = ClimaComms.context()
 ClimaComms.init(context)
 device = ClimaComms.device()
 device_suffix = device isa ClimaComms.CPUSingleThreaded ? "cpu" : "gpu"
-root_path = "snowy_land_pmodel_longrun_$(device_suffix)"
+lai_suffix = PROGNOSTIC_LAI ? "_opt_lai" : ""
+root_path = "snowy_land_pmodel$(lai_suffix)_longrun_$(device_suffix)"
 diagnostics_outdir = joinpath(root_path, "global_diagnostics")
 outdir =
     ClimaUtilities.OutputPathGenerator.generate_output_path(diagnostics_outdir)
@@ -68,9 +75,9 @@ function setup_model(
     stop_date,
     Δt,
     domain,
-    toml_dict,
+    toml_dict;
+    prognostic_lai = false,
 ) where {FT}
-    surface_domain = ClimaLand.Domains.obtain_surface_domain(domain)
     surface_space = domain.space.surface
     # Forcing data - high resolution
     atmos, radiation = ClimaLand.prescribed_forcing_era5(
@@ -84,23 +91,33 @@ function setup_model(
     )
     forcing = (; atmos, radiation)
 
-    # Read in LAI from MODIS data
-    LAI = ClimaLand.Canopy.prescribed_lai_modis(
-        surface_space,
-        start_date,
-        stop_date,
-    )
-
-    # Construct the land model with all default components
     prognostic_land_components = (:canopy, :lake, :snow, :soil, :soilco2)
-    land = LandModel{FT}(
-        forcing,
-        LAI,
-        toml_dict,
-        domain,
-        Δt;
-        prognostic_land_components,
-    )
+    if prognostic_lai
+        # The LandModel constructor uses the prognostic LAI model if no
+        # prescribed LAI is passed.
+        land = LandModel{FT}(
+            forcing,
+            toml_dict,
+            domain,
+            Δt;
+            prognostic_land_components,
+        )
+    else
+        # Prescribed LAI (default): read LAI from MODIS data.
+        LAI = ClimaLand.Canopy.prescribed_lai_modis(
+            surface_space,
+            start_date,
+            stop_date,
+        )
+        land = LandModel{FT}(
+            forcing,
+            LAI,
+            toml_dict,
+            domain,
+            Δt;
+            prognostic_land_components,
+        )
+    end
     return land
 end
 
@@ -122,9 +139,18 @@ else
     toml_dict = LP.create_toml_dict(FT)
 end
 
-model = setup_model(FT, start_date, stop_date, Δt, domain, toml_dict)
+model = setup_model(
+    FT,
+    start_date,
+    stop_date,
+    Δt,
+    domain,
+    toml_dict;
+    prognostic_lai = PROGNOSTIC_LAI,
+)
 simulation = LandSimulation(start_date, stop_date, Δt, model; outdir)
 @info "Run: Global Soil-Canopy-Snow Model"
+@info "LAI: $(PROGNOSTIC_LAI ? "prognostic (ZhouOptimalLAIModel)" : "prescribed (MODIS)")"
 @info "Resolution: $(domain.nelements)"
 @info "Timestep: $Δt s"
 @info "Start Date: $start_date"
@@ -134,7 +160,11 @@ ClimaLand.Simulations.solve!(simulation)
 
 LandSimVis.make_annual_timeseries(simulation; savedir = root_path)
 LandSimVis.make_heatmaps(simulation; savedir = root_path, date = stop_date)
-LandSimVis.make_leaderboard_plots(simulation; savedir = root_path)
+LandSimVis.make_leaderboard_plots(
+    simulation;
+    savedir = root_path,
+    leaderboard_data_sources = ["ERA5", "ILAMB", "FlagshipCarbonMetrics"],
+)
 
 if LONGER_RUN
     include("../ilamb/ilamb_conversion.jl")
--- a/ext/land_sim_vis/leaderboard/data_sources.jl
+++ b/ext/land_sim_vis/leaderboard/data_sources.jl
@@ -61,10 +61,25 @@ function _preprocess_sim_var(var, ::Val{:nee})
     return var
 end
 
+function _preprocess_sim_var(var, ::Val{:hr})
+    # heterotrophic respiration: convert `mol CO2 m^-2 s^-1` in sim to
+    # `g C m-2 day-1` in obs (used for the inv_hr inversion target).
+    (ClimaAnalysis.units(var) == "mol CO2 m^-2 s^-1") && (
+        var = ClimaAnalysis.convert_units(
+            var,
+            "g m-2 day-1",
+            conversion_function = units -> units * 86400.0 * 12.011,
+        )
+    )
+    return var
+end
+
 _preprocess_sim_var(var, ::Val{:lwu}) = var
 _preprocess_sim_var(var, ::Val{:lhf}) = var
 _preprocess_sim_var(var, ::Val{:shf}) = var
 _preprocess_sim_var(var, ::Val{:swu}) = var
+# LAI is already dimensionless (m^2 m^-2), matching the MODIS obs; no conversion.
+_preprocess_sim_var(var, ::Val{:lai}) = var
 _preprocess_sim_var(var, name::Val) =
     error("Preprocessing var with short name ($name) is not defined")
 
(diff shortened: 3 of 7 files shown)


AlexisRenchon:
@kmdeck you can review the change to experiments/long_runs/snowy_land_pmodel.jl mostly... and @ph-kev to the ext/ files (new leaderboard plot)

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,263 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+constant 365.25/12 ≈ 30.4375. Avoids a ±5% spurious seasonality the constant
+would introduce. Updates the `units` attribute.
+"""
+function _monthly_total_to_daily_rate!(obs_var)
+    dates_vec = ClimaAnalysis.dates(obs_var)
+    factors = 1.0 ./ Float64.(Dates.daysinmonth.(dates_vec))
+    t_idx = obs_var.dim2index[ClimaAnalysis.time_name(obs_var)]
+    shape = ntuple(d -> d == t_idx ? length(factors) : 1, ndims(obs_var.data))
+    obs_var.data .*= reshape(factors, shape)
+    obs_var.attributes["units"] = "g m-2 day-1"
```
ph-kev: I wouldn't use the field `dim2index` since I considered that part of the private API. You can iterate the time dimension with `ClimaAnalysis.view_select` and modify the data like that. Also, you can use `ClimaAnalysis.set_units!` instead of `obs_var.attributes["units"]`.

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,263 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+    for (alias, model_name) in _INVERSION_ALIAS_TO_MODEL_NAME
+        obs_var = inversion_dict[alias]
+        obs_var.attributes["short_name"] = model_name
+        obs_var_dict[model_name] = obs_var
+    end
+    if include_lai
+        lai_obs = get_modis_lai_obs_var()
+        lai_obs.attributes["short_name"] = "lai"
+        obs_var_dict["lai"] = lai_obs
+    end
```
ph-kev: Wouldn't it be easier to always include `lai` and not use it for whenever this struct is used?

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,263 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+    )
+        obs_var = _load(varname)
+        if monthly
+            # Convert to a daily rate and set units to "g m-2 day-1".
+            _monthly_total_to_daily_rate!(obs_var)
+        else
+            # rh is already daily; only relabel its units to "g m-2 day-1" so
+            # the UnitsChecker's raw string compare matches the model hr
+            # diagnostic (a mismatch NaN-fills the member's whole G column).
+            obs_var.attributes["units"] = "g m-2 day-1"
```
ph-kev: Can you check the units and if it match with what we expect, set the units then? Also, it would be better to use `ClimaAnalysis.set_units!` instead.

experiments/long_runs/snowy_land_pmodel.jl
```diff
@@ -53,11 +53,18 @@ const LONGER_RUN = haskey(ENV, "LONGER_RUN") ? true : false
 # `export UNCALIBRATED=""` in the terminal and run this script, or
 # pass `UNCALIBRATED=""` as an environment variable on buildkite.
 const UNCALIBRATED = haskey(ENV, "UNCALIBRATED") ? true : false
+# If you want to run with prognostic (modeled) LAI from the optimal-LAI model
+# (Zhou et al. 2025, `ZhouOptimalLAIModel`) instead of prescribed MODIS LAI,
+# type `export PROGNOSTIC_LAI=""` in the terminal and run this script, or pass
+# `PROGNOSTIC_LAI=""` as an environment variable on Buildkite. The default
+# (unset) prescribes MODIS LAI, preserving the prior behavior.
```
kmdeck:
```suggestion
# (unset) prescribes MODIS LAI.
```

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -82,6 +86,29 @@ steps:
         env:
           CLIMACOMMS_DEVICE: "CUDA"
 
+  # Prognostic (modeled) LAI run of snowy_land_pmodel (ZhouOptimalLAIModel),
+  # selected via the `PROGNOSTIC_LAI` env var. Runs on the weekly scheduled
+  # build so the standard weekly long runs cover BOTH prescribed and prognostic
+  # LAI, and on demand when a PR carries the "long run opt lai" label. Gated off
+  # LONGER_RUN/PERTURBED_RUN like the standard "Global Land Models" group.
```
kmdeck:
```suggestion
```

.buildkite/longruns_gpu/pipeline.yml
```diff
@@ -82,6 +86,29 @@ steps:
         env:
           CLIMACOMMS_DEVICE: "CUDA"
 
+  # Prognostic (modeled) LAI run of snowy_land_pmodel (ZhouOptimalLAIModel),
+  # selected via the `PROGNOSTIC_LAI` env var. Runs on the weekly scheduled
+  # build so the standard weekly long runs cover BOTH prescribed and prognostic
+  # LAI, and on demand when a PR carries the "long run opt lai" label. Gated off
```
kmdeck:
```suggestion
  # LAI, and on demand when a PR carries the "long run opt lai" label.
```

experiments/long_runs/snowy_land_pmodel.jl
```diff
@@ -84,23 +91,36 @@ function setup_model(
...
-    land = LandModel{FT}(
-        forcing,
-        LAI,
-        toml_dict,
-        domain,
-        Δt;
-        prognostic_land_components,
-    )
+    if prognostic_lai
+        # Prognostic LAI: the canopy computes LAI with the optimal-LAI model
```
kmdeck: Could we shorten this comment? 
maybe just:

``# The LandModel constructor uses the prognostic LAI model if no prescribed LAI is passed``

(this is all documented in the doc strings,someone can refer to that if they have more questions)

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,263 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
     )
     return compare_vars_biases_plot_extrema
 end
 
+"""
+    _monthly_total_to_daily_rate!(obs_var)
+
+In-place: convert `obs_var.data` from `g C m⁻² month⁻¹` to `g C m⁻² day⁻¹` by
+dividing each time slice by its real days-in-month (28–31), rather than the
+constant 365.25/12 ≈ 30.4375. Avoids a ±5% spurious seasonality the constant
```
kmdeck: Why would this error be seasonal? Seasonal indicates to me an error with a period of 3 months, while this error just depends on the number of days in the month.

Perhaps just. "Avoid a 5% source of error that the constant days per month assumption introduces."

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,263 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+# `inv_hr` (Hashimoto Rh) is intentionally excluded: the leaderboard shows the
+# MODIS `lai` target in its place (see InversionDataLoader).
+const _INVERSION_ALIAS_TO_MODEL_NAME =
+    Dict("inv_nee" => "nee", "sif_gpp" => "gpp", "res_er" => "er")
+
+"""
+    InversionDataLoader
+
+Loads the inversion-derived carbon observations (CT2022 NEE, GOSIF GPP, residual
+ER from the `inversion_nee` artifact), and optionally the MODIS `lai` target, for
```
kmdeck: I am wondering about this grouping (InversionDataLoader) and what it means and what it requires/implies. Why is MODIS LAI, GOSIF data included here, as they are satellite derived? Would NICO add his TROPOMI data here too in the future?

The grouping of ERA5 makes sense (all those data come from ERA5). The ILAMB grouping is a bit less meaningful but all those datasets are part of ILAMB. 

But this grouping feels more adhoc. Should we be adding these datasets in separate structs, or e.g. add SIF-derived data in a struct, or satellite data in a struct?

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,263 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+December/February sample), so they are linearly interpolated in time onto
+calendar month-starts. This is required because the calibration framework labels
+each seasonal average by the *first date* in the season
+(`ClimaAnalysis.average_season_across_time`) and matches obs to sim by exact date
+(`indexin`); the model's monthly `lai` diagnostic is month-start-dated and has
+all 12 months, so the obs must be too. Only interior month-starts are kept (the
+first/last months would require extrapolation), which is why `years` should
+bracket the calibration window by at least one month on each side.
+
+This is the calibration target for the optimal-LAI parameters
```
kmdeck: I think we can remove this. This is a docstring in a plotting extension package. it should be totally independent of any calibration requirements or uses, and any model config. We can calibrate LAI against anything

ext/land_sim_vis/leaderboard/data_sources.jl
kmdeck: I also worry that this is now blending the purpose of these structs for plotting  vs used for calibration, instead of keeping things more modular.

ext/land_sim_vis/leaderboard/data_sources.jl
kmdeck: I am now realizing we have this problem all over this file (mixing a plotting extension package with calibration data prep utilities) @ph-kev

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -28,40 +28,53 @@ function _percentile_contour_kwargs(
...
+    _monthly_climatology_global_means(sim, obs,
+                                      mask_fn = ClimaAnalysis.apply_oceanmask)
+
+Return a pair `(sim_monthly, obs_monthly)` of 12-element vectors of global,
+lonlat-weighted monthly climatology means. Index `i` corresponds to month `i`
+(1=Jan ... 12=Dec). Months that have no valid samples are filled with `NaN`.
+
+`mask_fn` is the same per-variable mask used by `global_bias`/`global_rmse`
+elsewhere. After applying it, both sim and obs are further restricted to the
+intersection of finite cells at each time step, so the two lines are averaged
```
kmdeck: What is a finite cell?

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -89,6 +102,58 @@ function _nee_diverging_contour_kwargs(var, q; nlevels = 21)
...
+`global_bias` / `global_rmse` / `plot_bias_on_globe!` integrate over the
+intersection of finite cells, matching `_monthly_climatology_global_means`.
+
+Two problems it works around in `ClimaAnalysis.bias` when `obs` has spatial gaps
+(e.g. GOSIF-GPP/residual-ER over deserts/ice): its normalization `ones_var`
+doesn't inherit obs's NaN footprint (denominator over all land, numerator over
+land∩finite(obs) — attenuates or flips the bias), and its `resampled_as` call
+isn't NaN-aware so NaNs erode their neighbours.
+
+The fix: fill NaN with 0 in both fields (so `resampled_as` can't erode), and
```
kmdeck: This would still mess up the resample_as, it would just blend with zero (and bias all gap edges to lower values) vs make them NaN...

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -89,6 +102,58 @@ function _nee_diverging_contour_kwargs(var, q; nlevels = 21)
...
 
+"""
+    _prepare_for_bias(base_mask, sim, obs)
+
+Return `(sim_clean, obs_clean, mask_fn)` so that `ClimaAnalysis.bias` /
+`global_bias` / `global_rmse` / `plot_bias_on_globe!` integrate over the
+intersection of finite cells, matching `_monthly_climatology_global_means`.
+
+Two problems it works around in `ClimaAnalysis.bias` when `obs` has spatial gaps
+(e.g. GOSIF-GPP/residual-ER over deserts/ice): its normalization `ones_var`
```
kmdeck: should this be a fix in ClimaAnalysis instead? missing data will be a common problem across atmos, land, etc @ph-kev

ext/land_sim_vis/leaderboard/leaderboard.jl
kmdeck: it would be nice if we had a way that resampled_as could ignore NaNs

ext/LandSimulationVisualizationExt.jl
```diff
@@ -80,16 +89,21 @@ function LandSimVis.make_leaderboard_plots(
     short_names = [d.variable.short_name for d in diagnostics]
     diagnostics_folder_path = diagdir
     leaderboard_base_path = savedir
+    # Only add the MODIS-LAI panel to the INVERSION leaderboard for prognostic
+    # (modeled) LAI runs; see `_leaderboard_include_lai`.
+    include_lai = _leaderboard_include_lai(model)
```
ph-kev: I don't think it is worth the effort to try to exclude LAI since users can ignore it. If this is needed, I think it should be more generic (e.g. a vector of variable names to ignore).

ext/land_sim_vis/leaderboard/leaderboard.jl
ph-kev: There was this issue that was open a while ago: https://github.com/CliMA/ClimaAnalysis.jl/issues/198. I can work on this next sprint if needed.

ext/land_sim_vis/leaderboard/leaderboard.jl
ph-kev: I can add a fix for this. There is an issue that I linked about nan-aware regridding, unless this is more about problem of data imputation.

ext/land_sim_vis/leaderboard/data_sources.jl
ph-kev: This can be splitted off into a different module that the plotting module can consume. For most cases, the data preprocessing required for the observational data is the same for both plotting (against simulation data) and calibration. What I envision for the data loader struct is enough preprocessing that other modules/users can use them for their own purposes (plotting and calibration) and if more preprocessing is needed, they can do it themselves.

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,263 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+
+    _load(varname) = begin
+        obs_var = ClimaAnalysis.OutputVar(inversion_path, varname)
+        ClimaAnalysis.dim_units(obs_var, "lon") == "degree" &&
+        ClimaAnalysis.set_dim_units!(obs_var, "lon", "degrees_east")
+        ClimaAnalysis.dim_units(obs_var, "lat") == "degree" &&
+        ClimaAnalysis.set_dim_units!(obs_var, "lat", "degrees_north")
+        ClimaAnalysis.transform_dates!(obs_var, Dates.firstdayofmonth)
+        return ClimaAnalysis.replace(obs_var, missing => NaN)
+    end
```
ph-kev, line 523: I would follow the same pattern with the other data loaders.

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,263 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+intersect with the simulation diagnostics.
+
+The `lai` panel (MODIS, Yuan et al. 2017) is only included when the loader is
+built with `include_lai = true` — i.e. for prognostic-LAI runs, where the model
+`lai` diagnostic is an independent prediction. With prescribed LAI the model
+`lai` equals the MODIS obs, so the comparison is trivial and is omitted.
+"""
+struct InversionDataLoader <: AbstractDataLoader
+    """Preprocessed inversion `OutputVar`s, keyed by model short name."""
+    obs_var_dict::Dict{String, Any}
```
ph-kev, line 574: I would use `ClimaAnalysis.NCCatalog` like the other data loaders.

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,263 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+
+"""
+    InversionDataLoader
+
+Loads the inversion-derived carbon observations (CT2022 NEE, GOSIF GPP, residual
+ER from the `inversion_nee` artifact), and optionally the MODIS `lai` target, for
+the leaderboard. The inversion analogue of `ILAMBDataLoader`: same seasonal
+machinery, but against the inversion product instead of FLUXCOM. Keyed by the
+*model* short names `nee`/`gpp`/`er`(`/lai`) (not the `inv_*` aliases) so they
+intersect with the simulation diagnostics.
```
ph-kev: This comment is confusing to me since isn't the ILAMB data monthly and not seasonal? Also, I don't think the detail about the short name is relevant here.

ph-kev (changes requested):

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: Done — switched to iterating time slices with `view_select(...; by = Index())` (its `.data` is a view, so the in-place `./=` scales the parent) and set units via `set_units!`. Dropped `dim2index` and the reshape. Verified against the `inversion_nee` artifact that each month is still divided by its actual days-in-month.

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: Dropped `include_lai` — `InversionDataLoader()` always loads the MODIS `lai` target now. The `lai` diagnostic is identical in both configs (always `p.canopy.biomass.area_index.leaf`), so a prescribed run just shows a ~1:1 model-vs-obs panel, which is fine. Also removed the `_leaderboard_include_lai` dispatch and the flag threading through the `compute_*_leaderboard` functions.

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: The `rh` branch now asserts the native units are `"g C m-2 day-1"` before relabeling to `"g m-2 day-1"` via `set_units!`, and errors otherwise. Added the matching `"g C m-2 month-1"` input check to the monthly→daily helper.

experiments/long_runs/snowy_land_pmodel.jl
AlexisRenchon: Applied, thanks.

.buildkite/longruns_gpu/pipeline.yml
AlexisRenchon: Applied.

.buildkite/longruns_gpu/pipeline.yml
AlexisRenchon: Applied.

experiments/long_runs/snowy_land_pmodel.jl
AlexisRenchon: Shortened to the one-liner, thanks.

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: Reworded to "the ~5% source of error the constant days-per-month assumption introduces" — dropped "seasonality." Also fixed a stale `g C m⁻² day⁻¹` in the same docstring (the function sets the string to `g m-2 day-1`).

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: I understand the grouping concern... I am not sure what to do about the "INVERSION" being not especially coherent (when compared to ERA5 or FLUXCOM). I could potentially rename it as "SATELLITE", but it is still vague... I'll leave this open for discussion, but is this a merge blocker?


For the structs plotting vs calibration, I'll let @ph-kev decide what to do

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: Agreed — this plotting-extension file mixes plotting and calibration data-prep (the calibration scripts `include` it directly), which predates this PR. I've trimmed the calibration-framework internals out of this docstring here, and opened #1795 to extract the shared obs data-prep into a module both the leaderboard ext and calibration consume, and to port the loaders onto `NCCatalog`. @ph-kev — happy for you to drive that since it's your area.

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: A "finite cell" = a grid cell with valid (non-`NaN`) data. Clarified the docstring to say "the grid cells where both have valid (non-`NaN`) data."

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: You're right — the zero-fill biases cells adjacent to a gap slightly low when `resampled_as` interpolates across the filled edge; the mask only removes the gap interiors. For these carbon vars the practical impact is small (true GPP/ER in desert/ice gaps is ~0), but it's real. Documented it as an interim workaround pointing at ClimaAnalysis.jl#198 — thanks @ph-kev, a NaN-aware `resampled_as` would let us delete `_prepare_for_bias` entirely.

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: Agreed this belongs in ClimaAnalysis — a NaN-aware `resampled_as` (ClimaAnalysis.jl#198, thanks @ph-kev) would remove this workaround. Not blocking this PR; documented the current zero-fill behavior.

ext/LandSimulationVisualizationExt.jl
AlexisRenchon: Removed the special-casing entirely rather than generalizing it — LAI is now always loaded in `InversionDataLoader` (see the `include_lai` removal). Prescribed runs just get a trivial 1:1 LAI panel, which is fine.

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: Agreed — opened #1795 to port `InversionDataLoader` onto `NCCatalog` and follow the same pattern as `ERA5DataLoader`/`ILAMBDataLoader`. Deferring to that follow-up to keep this PR scoped to the LAI feature.

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: Agreed — tracked in #1795 (port these loaders onto `NCCatalog`).

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: Fixed — reworded to "monthly obs" (the data is monthly; seasonal averaging is a downstream leaderboard step) and dropped the short-name plumbing detail.

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: @ph-kev Kevin, let me know if you will do it (no rush I think), or if you'd like me to do this

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -439,22 +454,254 @@ function get_compare_vars_biases_plot_extrema(; annual = false)
...
+    native == "g C m-2 month-1" || error(
+        "expected a monthly total in \"g C m-2 month-1\", got \"$native\"",
+    )
+    time_dim = ClimaAnalysis.time_name(obs_var)
+    for (i, date) in enumerate(ClimaAnalysis.dates(obs_var))
+        slice = ClimaAnalysis.view_select(
+            obs_var;
+            by = ClimaAnalysis.Index(),
+            (; Symbol(time_dim) => i)...,
+        )
```
ph-kev, line 485:
```suggestion
        slice = ClimaAnalysis.view_select(
            obs_var;
            by = ClimaAnalysis.Index(),
            time = i,
        )
```

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -396,7 +482,9 @@ function compute_seasonal_leaderboard(
     # Get everything we need from data_sources.jl
     sim_dir = ClimaAnalysis.SimDir(diagnostics_folder_path)
     data_loader =
-        uppercase(data_source) == "ERA5" ? ERA5DataLoader() : ILAMBDataLoader()
+        uppercase(data_source) == "ERA5" ? ERA5DataLoader() :
+        uppercase(data_source) == "INVERSION" ? InversionDataLoader() :
+        ILAMBDataLoader()
```
ph-kev: This should throw an error if `data_source` isn't "ERA5", "INVERSION", or "ILAMB".

ph-kev (approved):
I think there are some parts that can be improved but it can be done in another PR.

ext/land_sim_vis/leaderboard/leaderboard.jl
kmdeck: In that case, we should leave these as NaN. 

Then the future resample_As function which handles NaNs will work with this without requiring a change (otherwise we need to remember to go back and change these zeros to nans again).

also if it makes little difference, let's keep them as NaN and just have a slightly smaller data set

kmdeck (approved):
looks good pending one comment about the resampled/NaNs
renaming InversionDataLoader

so future work out of this:
- split data processing into different module (Kevin)
- unify data processes (NCCatalog, etc, for the new dataloader) - who?
- resampled as NaN aware (Kevin)

PR #1796, opened 2026-07-07
Remove GHA tests on Windows

ph-kev:
As of now, there isn't much benefit to testing on Windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,6 @@ jobs:
           - '1.12'
         os:
           - ubuntu-latest
-          - windows-latest
           # There is a concurrency limit for MacOS runners across the
           # entire organization.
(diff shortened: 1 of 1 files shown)


kmdeck (commented):
if it passes, why not keep it?

ph-kev:
I don't see much value in the tests for Windows since most of our development happens on Mac and Linux. Another thing that we can do is to only run the GHA tests on Windows if it lands on the main branch (which is what we do for Mac). I can also close this issue if there is any need to support Windows for ClimaLand.

kmdeck:
Im not sure. Youre right that any sort of global simulation would be run on a cluster of some kind/on linux. The only use case of Windows would be someone running simple experiments/tutorials, e.g. if ClimaLand was ever used in classroom. 

We can maybe make it so it only runs on the main branch? I guess I agree that I am not sure of the benefit but it doesnt seem like it costs us anything (right now) in any case

PR #1811, opened 2026-07-15
prevent ci -> ca and update packages

### LINKED ISSUE #1808: Implement gs_max
gs is solved for as:
gs = A/(ca - ci). when ci -> ca, gs diverges. In reality, gs has a maximum value.

In this PR, I do not clip gs to a maximum value, but instead clip ci to be < ca. The reason for this is that I think the max value for gs is at a leaf level, while we compute something canopy level. The canopy gs will change in time due to changing LAI, for example, and this makes a maximum value need to be time varying as well. 

On the other hand, ci does not have this issue. and clipping ci < ca creates an effective gs_max that changes in time.

Instead of clipping ci directly, though, I wrote ci = ca * (1+y*x)/(1+x), where x = sqrt(VPD)/xi and y = gamma_star/ca_pp. These are non dimensional, and you can work through the math to see that ci < ca  as long as x> 0 and y < 1. So, I clip these both.

Since gs is typically << g_h, it provides a limit to LHF that is not present to SHF. When gs diverges, though, both fluxes feel the same conductance, and visually this appears as a sharp increase in LHF. Therefore, in regions where VPD is commonly small compared to xi, and ci -> ca often, we have lose leverage on the Bowen ratio with our parameterization for gs and may overestimate LHF. I am not sure how much of a problem this is. I also do not know that this fix will address it if it is.

### LINKED ISSUE #1809: Debug nonconvergence of Surface Fluxes in stable conditions
@tapios fixed this in SF and the release was made and is now used by CL.

kmdeck:
## Purpose 
ci >= ca when VPD is zero or Gamma^*/ca > 1. (VPD cannot be negative.) This causes stomatal conductance to diverge or it could become negative.

This PR adds two new parameters which control how small VPD can be (this varies in time depending on \xi) and how big Gamma star can be. In practice I think only VPD control matters, but mathematically limiting Gamma^* made sense too just in case.

I also update packages in the Manifest because a new SF release solves an issue with non convergence that we were seeing.

## To-do
combine "Pmodelparameters" and "pmodelconstants" into a single structure (future PR)

## Content

----

### DIFF
--- a/.buildkite/Manifest-v1.12.toml
+++ b/.buildkite/Manifest-v1.12.toml
@@ -1,13 +1,13 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.12.5"
+julia_version = "1.12.6"
 manifest_format = "2.0"
 project_hash = "6ae0cd45ad2cb0a0b08a5c132eef50cb784c1c71"
 
 [[deps.ADTypes]]
-git-tree-sha1 = "d9aaef7c63466eee4de23b4d9dad03629df54bea"
+git-tree-sha1 = "ec6be48a85c93d995563b84bff8a86bc98df45ce"
 uuid = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
-version = "1.22.1"
+version = "1.22.2"
 weakdeps = ["ChainRulesCore", "ConstructionBase", "EnzymeCore"]
 
     [deps.ADTypes.extensions]
@@ -399,9 +399,9 @@ version = "0.4.4+1"
 
 [[deps.CUDA_Driver_jll]]
 deps = ["Artifacts", "JLLWrappers", "Libdl", "TOML"]
-git-tree-sha1 = "2d6222474d868469a72de5bd47c5c25c0e1fe518"
+git-tree-sha1 = "ff4a0abf98be36d9fa15ca7e968df562a210ab4b"
 uuid = "4ee394cb-3365-5eb0-8335-949819d2adfc"
-version = "13.3.0+0"
+version = "13.3.0+1"
 
 [[deps.CUDA_Runtime_Discovery]]
 deps = ["Libdl"]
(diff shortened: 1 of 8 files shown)


AlexisRenchon:
### Cause

The refactored `intercellular_co2_pmodel` computes

```julia
x = max(sqrt(VPD) / ξ, vpd_ratio_min)
y = min(Γstar / ca_pp, Γ_ratio_max)
return ca_pp * (1 + y * x) / (1 + x)
```

When `ξ = 0`, `x = sqrt(VPD) / 0 = Inf`, so the return is `ca_pp * (1 + y*Inf) / (1 + Inf) = Inf/Inf = NaN`.

This gets hit during initialization. In `set_initial_cache!` (Canopy.jl) the order is:

1. `update_cache!` → `update_photosynthesis!` computes `InstVars` (GPP/An/gs) from `OptVars`,
2. **then** `set_historical_cache!` populates `OptVars` (`ξ_opt`, `Vcmax25_opt`, `Jmax25_opt`).

So on the first pass `OptVars.ξ_opt` is still `0` everywhere, `ci` becomes `NaN` on every land point, and the NaN propagates through `GPP → An → gs_co2 → r_stomata_canopy →` the canopy turbulent fluxes that the test checks. `InstVars` is never recomputed after `OptVars` is filled in, so the NaN persists in the initial cache.

The **previous** `ci` formulation, `(ξ·ca_pp + Γstar·√VPD) / (ξ + √VPD)`, returned `Γstar` (finite) at `ξ = 0`, which is why this passed on `main`. The new form is algebraically identical for `ξ > 0` but loses the finite `ξ → 0` limit.

### Fix

Guard `ξ` so the `ξ → 0` limit `ci = Γstar` is recovered, without changing behavior for `ξ > 0` and keeping the `vpd_ratio_min` / `Γ_ratio_max` clamps intact:

```julia
x = max(sqrt(VPD) / max(ξ, eps(FT)), vpd_ratio_min)
```

### Verification
- `test/shared_utilities/coupled_fluxes.jl` now passes locally (exit 0, no NaNs).
- `test/standalone/Vegetation/test_pmodel.jl` passes (136 regression tests + edge cases, including the `ξ_opt = 0` case + binarize).

*(Aside: `sqrt(VPD)` can still produce `NaN` if `VPD < 0` — possible in supersaturated forcing now that the `max(VPD, sqrt(eps))` clip was removed — but that path isn't exercised here. Worth a follow-up if real forcing can drive VPD negative. The now-stale "clip here to prevent numerical issues" comments in `call_update_optimal_EMA` could also be dropped.)*

AlexisRenchon (approved):

PR #1819, opened 2026-07-23
Partitioning plots, comparison to ERA5 at a site

kmdeck:
## Purpose 
Add additional plots for debugging purposes

## To-do
I feel like there are lots of ways to make this better. For example, we could also compare against the other leaderboard data for then single site example, to make this better match the global leaderboard plots.
It would also be cool to have bias plots for all of the partitioning plots, but it was easiest to do it for ERA5 for now.

I wonder if we should also move away from the DictWriter and use the NetCDF file also for single column runs to unify the way we make plots, but these feels like a lot of work within minimal payoff

## Content

----

### DIFF
--- a/ext/land_sim_vis/plotting_utils.jl
+++ b/ext/land_sim_vis/plotting_utils.jl
@@ -615,7 +615,7 @@ function compare_monthly_fluxes_with_data(
             dn;
         )
         model_dates = time_to_date.(model_time, start_date)
-        model_dt = model_dates[2] - model_dates[1]
+        model_dt = Month(model_dates[2]) - Month(model_dates[1])
         @assert(model_dt == Month(1)) || @error(
             "Diagnostic reduction period must be one month to plot partitioning diagnostics."
         )
--- a/ext/land_sim_vis/plotting_utils.jl
+++ b/ext/land_sim_vis/plotting_utils.jl
@@ -569,7 +569,7 @@ function compare_monthly_fluxes_with_data(
     start_date,
     stop_date,
     longlat;
-    data_source = "ERA",
+    data_source = "ERA5",
     plot_stem_name = "ERA5",
     spinup_date = start_date,
     short_names = ["lwu", "swu", "shf", "lhf"],
--- a/ext/land_sim_vis/plotting_utils.jl
+++ b/ext/land_sim_vis/plotting_utils.jl
@@ -616,7 +616,7 @@ function compare_monthly_fluxes_with_data(
         )
         model_dates = time_to_date.(model_time, start_date)
         model_dt = Month(model_dates[2]) - Month(model_dates[1])
-        @assert(model_dt == Month(1)) || @error(
+        @assert model_dt == Month(1) || @error(
             "Diagnostic reduction period must be one month to plot partitioning diagnostics."
         )
         # DictWriter diagnostics are saved at the end of the reduction period
--- a/ext/land_sim_vis/plotting_utils.jl
+++ b/ext/land_sim_vis/plotting_utils.jl
@@ -552,7 +552,7 @@ end
         start_date,
         stop_date,
         longlat;
-        data_source = "ERA",
+        data_source = "ERA5",
         plot_stem_name = "ERA5",
         spinup_date = start_date,
         short_names = ["lwu", "swu", "shf", "lhf"]
--- a/ext/land_sim_vis/plotting_utils.jl
+++ b/ext/land_sim_vis/plotting_utils.jl
@@ -544,3 +544,612 @@ function LandSimVis.make_timeseries(
     end
     return nothing
 end
+
+"""
+    compare_monthly_fluxes_with_data(
+        savedir,
+        diagnostics,
+        start_date,
+        stop_date,
+        longlat;
+        data_source = ["ERA",],
+        plot_stem_name = "ERA5",
+        spinup_date = start_date,
+        short_names = ["lwu", "swu", "shf", "lhf"]
+    )
+
+Creates a leaderboard style plot for single column runs where the diagnostics
+are stored using the DictWriter, in memory; SHF, LHF, LWu, and SWu are all plotted
+against average monthly values from data. This could be extended to pass in other
+data sources and compute other flux comparisons.
+"""
+function compare_monthly_fluxes_with_data(
+    savedir,
+    diagnostics,
+    start_date,
+    stop_date,
+    longlat;
+    data_source = ["ERA"],
+    plot_stem_name = "ERA5",
+    spinup_date = start_date,
+    short_names = ["lwu", "swu", "shf", "lhf"],
+)
+    if data_source == "ERA5"
+        data_loader = ERA5DataLoader()
+    else
+        @error(
+            "compare_monthly_fluxes_with_data is currently only configured for ERA5 data."
+        )
+    end
+
+    for i in 1:length(short_names)
+        dn = [
+            d.output_short_name for
+            d in diagnostics if d.variable.short_name == short_names[i]
+        ]
+        @assert length(dn) == 1 ||
+                @error("$(short_names[i]) not available in saved diagnostics.")
+        dn = first(dn)
+        unit = [
+            d.variable.units for
+            d in diagnostics if d.variable.short_name == short_names[i]
+        ][1]
+        sn = short_names[i]
+        global_obs_data = get(data_loader, sn)
+        ClimaAnalysis.set_reference_date!(global_obs_data, start_date)
+        global_obs_data = ClimaAnalysis.window(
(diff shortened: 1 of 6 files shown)


ext/land_sim_vis/plotting_utils.jl
```diff
@@ -544,3 +544,601 @@ function LandSimVis.make_timeseries(
...
+            lon = longlat[1],
+        )
+        obs_times = obs_data.dims["time"]
+        obs_values = obs_data.data
+        model_time, model_output = ClimaLand.Diagnostics.diagnostic_as_vectors(
+            diagnostics[1].output_writer,
+            dn;
+        )
+        model_dates = time_to_date.(model_time, start_date)
+        # DictWriter diagnostics are saved at the end of the reduction period
```
kmdeck, line 622: how to ensure output is monthly?

or, how to make this shift adaptive to the diagnostic reduction period?

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -544,3 +544,601 @@ function LandSimVis.make_timeseries(
...
+    savedir,
+    diagnostics,
+    start_date,
+    stop_date,
+    longlat;
+    plot_stem_name = "ERA5",
+    spinup_date = start_date,
+)
+    short_names = ["lwu", "swu", "shf", "lhf"] # short_name
+
```
ph-kev: Would it be better to make this a keyword argument?

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -544,3 +544,601 @@ function LandSimVis.make_timeseries(
...
+    data_loader = ERA5DataLoader()
+    for i in 1:length(short_names)
+        dn = [
+            d.output_short_name for
+            d in diagnostics if d.variable.short_name == short_names[i]
+        ]
+        @assert length(dn) == 1 ||
+                @error("$(short_names[i]) not available in saved diagnostics.")
+        dn = dn[1]
```
ph-kev:
```suggestion
        dn = first(dn)
```

ext/land_sim_vis/plotting_utils.jl
ph-kev: I think you can find the period from the `model_dates` and shift the dates by computing `model_dates .-= period`.

ph-kev: Also, I think it would be a good idea for this function to do nothing if the reduction isn't a month.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -544,3 +544,601 @@ function LandSimVis.make_timeseries(
...
+        ][1]
+        sn = short_names[i]
+        global_obs_data = get(data_loader, sn)
+        ClimaAnalysis.set_reference_date!(global_obs_data, start_date)
+        global_obs_data = ClimaAnalysis.window(
+            global_obs_data,
+            "time",
+            left = spinup_date,
+            right = stop_date,
+        )
```
ph-kev, line 605: I am not sure how relevant this is, but this get the nearest time for spinup_date and stop_date which may be okay here. See https://clima.github.io/ClimaAnalysis.jl/stable/howdoi/#How-do-I-get-a-slice-or-window-of-a-OutputVar-by-something-other-than-the-nearest-value?

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -544,3 +544,601 @@ function LandSimVis.make_timeseries(
...
+        obs_data = ClimaAnalysis.slice(
+            global_obs_data,
+            lat = longlat[2],
+            lon = longlat[1],
+        )
```
ph-kev, line 610: Same comment where this gets the nearest lat and lon which may or may not be what you want.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -544,3 +544,601 @@ function LandSimVis.make_timeseries(
...
+        # Get the first and last years of the simulation for windowing
+        first_date = first(ClimaAnalysis.dates(shf_var))
+        last_date = last(ClimaAnalysis.dates(shf_var))
+        @assert Second((last_date - first_date)) > Second(86400)
+        shf_var = ClimaAnalysis.average_time(
+            ClimaAnalysis.window(
+                shf_var,
+                "time",
+                left = last_date - Year(1),
+                right = last_date - Month(1),
```
ph-kev: I am confused about the choices for `left` and `right`. Is the intention to always get a year worth of data for the time average?

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -544,3 +544,601 @@ function LandSimVis.make_timeseries(
...
+            "long_name" => "Evaporative Fraction",
+        )
+        ratio = ClimaAnalysis.OutputVar(
+            attributes,
+            lhf_var.dims,
+            lhf_var.dim_attributes,
+            lhf_var.data ./ (lhf_var.data .+ shf_var.data),
+            lhf_var.dim2index,
+            lhf_var.index2dim,
+        )
```
ph-kev: It is possible to do `lhf / (lhf + shf)` or does that throw an error? Also, this constructor is not part of the public API. If that doesn't work, I think `ClimaAnalysis.remake` (see https://clima.github.io/ClimaAnalysis.jl/stable/api/#ClimaAnalysis.Var.remake) would be worth looking at.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -544,3 +544,601 @@ function LandSimVis.make_timeseries(
...
+            plot!(
+                fig2,
+                bias_ratio,
+                more_kwargs = Dict(
+                    ClimaAnalysis.Utils.kwargs(rasterize = true)...,
+                ),
+                mask = mask,
+            )
+
+        end
```
ph-kev, line 1023: I think this works?

```suggestion
        mask_kwarg = mask isa Nothing ? (; ) : (; mask = mask)
            plot!(fig, ratio, more_kwargs = plot_kwargs, mask_kwarg...)
            plot!(
                fig2,
                bias_ratio,
                more_kwargs = Dict(
                    ClimaAnalysis.Utils.kwargs(rasterize = true)...,
                ),
                mask_kwarg...
            )
```

ext/land_sim_vis/plotting_utils.jl
kmdeck: yes, I wanted to plot an annual average

ext/land_sim_vis/plotting_utils.jl
kmdeck: I cant divide or add OutputVars, but that would be nice if we could

kmdeck: Remake looks like a better solution though, Ill change to that

ext/land_sim_vis/plotting_utils.jl
ph-kev: Can you add a comment for this? My only other question is why
```suggestion
            ClimaAnalysis.window(
                shf_var,
                "time",
                left = last_date - Month(11),
                right = last_date,
```
doesn't work?

ext/land_sim_vis/plotting_utils.jl
kmdeck: Can you clarify a bit more? should I change this, or does it seem OK?

ext/land_sim_vis/plotting_utils.jl
kmdeck: I think this should be OK

ext/land_sim_vis/plotting_utils.jl
kmdeck: I can change to that! ill add a comment too

ext/land_sim_vis/plotting_utils.jl
kmdeck: I changed this

ext/land_sim_vis/plotting_utils.jl
ph-kev: I think it is okay since `global_obs_data` is monthly data and `spinup_date` and `stop_date` are the first day of some months. The default option just select the nearest date for windowing so you can get a bug from that if the `global_obs_data`, `spinup_date`, or `stop_date` change in the the future. You can pass `by = ClimaAnalysis.MatchValue()` as a keyword argument which error if it doesn't match exactly. This may not work if there is floating point rounding.

ph-kev (approved):

ext/land_sim_vis/plotting_utils.jl
kmdeck: thanks!

Issue #1723, opened 2026-04-24
Running multiple sites at the same time on a GPU


kmdeck:
Running a site level run on CPU can be slow. If someone wants to run N sites, the fastest way to do it would be to request N CPUs, and run them in parallel. However, if calibrating using multiple site data, we would then need to request N * N_ensemble CPUs, and this can be prohibitive if N ~ 100 and N_ensemble >~ 10.

It seems like we should be able to run all the sites on a single GPU. Then we need to request only N_ensemble GPUs to calibrate.

With respect to speed: I am not sure how much faster this would be for a single run: requesting 100 CPUs and running in parallel vs broadcasting over a 100 element vector on the GPU, since we would be far from saturating the GPU. It would be interested to understand this also.

**Challenges**
The sites would not be near each other in space (e.g. pick 100 random locations on Earth). Can we set up a lat long space corresponding to a list of locations [(lat1, lon1), (lat2, lon2), ....], and can we read in the forcing data in a similar way (a matrix of location vs time)? how would we change the writing of diagnostics?

**Additional Consideration**
Is there something we are doing that makes the code very slow on CPU that would be easy to fix?
Is it possible that the way we save half hourly diagnostics in memory make this very slow, and writing to disk less frequently would be much faster?

ph-kev:
Here's an update on this issue. In ClimaCore, this [PR](https://github.com/CliMA/ClimaCore.jl/pull/2525) adds the space with multiple columns. However, to use this in ClimaLand with multiple columns, there need to be support for reading in forcing data from multiple datasets for the multiple columns. The functionality for that is implemented in ClimaUtilities in this [PR](https://github.com/CliMA/ClimaUtilities.jl/pull/241).

Once those two PRs are merged, then everything that is needed outside of ClimaLand should be done for running spaces with multiple columns. One notable thing that should be mentioned is that the ClimaDiagnostics doesn't support this new space, but I believe that the current single column run read from the fields directly, so it is not a priority to add the functionality to create diagnostics.

For updating ClimaLand, I think the path going forward is to

1. Update ClimaLand to use the new space in the ClimaCore [PR](https://github.com/CliMA/ClimaCore.jl/pull/2525) with `N = 1` column. There are slight bit-by-bit differences when I was testing between the two spaces.
[...]

PR #1826, opened 2026-07-23
Add ColumnEnsemble

ph-kev:
This is the first step described in https://github.com/CliMA/ClimaLand.jl/issues/1723#issuecomment-5064526784 where you use the new space added in https://github.com/CliMA/ClimaCore.jl/pull/2525 in ClimaCore. The purpose is to verify that using the two different spaces doesn't result in any big differences.

Note that I ask a LLM to write a majority of the code.

Support for interpolating over the time dimension for multiple points is added here: https://github.com/CliMA/ClimaUtilities.jl/pull/248

TODO

- [x] Make a patch release of ClimaCore


experiments/integrated/fluxnet/column_ensemble_comparison.jl
```diff
@@ -0,0 +1,318 @@
...
+
+dt = Float64(450)
+(start_date, stop_date) =
+    FluxnetSimulations.get_data_dates(site_ID, time_offset)
+# A month starting in winter covers diurnal cycles and snow events while
+# keeping the run cheap; realism is not the goal here.
+stop_date = start_date + Day(30)
+
+function setup_fluxnet_simulation(land_domain)
+    toml_dict = LP.create_toml_dict(FT)
```
ph-kev: I am not sure if this Fluxnet simulation makes sense.

Project.toml
```diff
@@ -49,7 +49,7 @@ Adapt = "4.6.1"
 CairoMakie = "0.13, 0.14, 0.15"
 ClimaAnalysis = "0.5.22"
 ClimaComms = "0.6.2"
-ClimaCore = "0.14.50 - 0.14.51, 0.14.53, 0.15"
+ClimaCore = "0.15.1"
```
ph-kev, line 52: Is there a need to keep backward compatibility with older versions of ClimaCore?

imreddyTeja: Probably not. I think it would can make debugging instabilities and numerics easier, but ClimaAtmos is dropping compatibility for 0.14, so it won't make a huge difference for coupled runs.

experiments/integrated/fluxnet/column_ensemble_comparison.jl
```diff
@@ -0,0 +1,316 @@
+# Runs the same fluxnet simulation (US-MOz, full LandModel with canopy, snow,
+# soil, and soil CO2) on a single `Column` and on a `ColumnEnsemble` of
+# duplicate columns, then requires that
+#   1. all ensemble columns are identical to each other, and
+#   2. every ensemble column matches the single-column run within tolerance,
+# for the prognostic state, the cache, and the halfhourly diagnostics.
+# The two domains use different geometry (planar box slice vs spherical
```
kmdeck: What is a "planar box slice"

experiments/integrated/fluxnet/column_ensemble_comparison.jl
kmdeck: I think it is OK, but simpler might be (if it isnt too annoying) to use the ERA5 single site example. IT would cut down the boilerplate/setup stuff a lot because we just use the default model + era5 forcing but at a column.

but I am not sure if the forcing will work the same if it is read from a CSV (fluxnet) vs ERA5 netcdf?

experiments/integrated/fluxnet/column_ensemble_comparison.jl
kmdeck: This also doesnt totally make sense to me - why would the geometry matter for a column or a collection of columns? is the collection of columns on a sphere? Why not make it a regular domain?

src/diagnostics/diagnostic.jl
```diff
@@ -173,11 +173,13 @@ nlayers(space::Spaces.AbstractSpace) = Spaces.nlevels(space)
 
 """
     nlayers(::Spaces.PointSpace)
+    nlayers(::Spaces.PointCloudSpace)
 
 Returns the number of layers in the vertical; this is one
-for the PointSpace.
+for the PointSpace and PointCloudSpace.
```
kmdeck, line 179: is a PointCloudSpace the surface space of the ColumnEnsemble domain?

kmdeck:
Did you have a chance to check how fast the ensemble column runs on GPU for e.g. 10, 30, 100, 300 columns? (or something like that).

experiments/integrated/fluxnet/column_ensemble_comparison.jl
ph-kev: The new space is a collection of columns on a sphere. The geometry shouldn't matter in the sense that both give the same results analytically. However, you get floating point errors because of the different geometries between the spaces. It was implemented as columns on a sphere since the points were `LatLonPoint`s.

ph-kev: I don't think it would be difficult to change it to a different geometry but it would need another release of ClimaCore.

src/diagnostics/diagnostic.jl
ph-kev: The surface space of a `ClimaLand.Domains.ColumnEnsemble` is a `ClimaCore.Spaces.PointCloudSpace`.

experiments/integrated/fluxnet/column_ensemble_comparison.jl
ph-kev: Forcing from a ERA5 NetCDF file (with lon and lat gridded data) should work since `ClimaUtilities.InterpolationRegridder` use the points of the space to interpolate to, so it doesn't matter what space it is. For the FluxNet CSV file, I think it build a `TimeVaryingInput0D` which happen to work on multiple duplicated columns. However, it doesn't work on multiple columns that need different forcing data. This[ PR](https://github.com/CliMA/ClimaUtilities.jl/pull/248) adds functionality. There is also another PR for adding support for data from multiple columns from NetCDF files but it requires more thought to make it performant.

experiments/integrated/era5/comparison_utils.jl
```diff
@@ -0,0 +1,191 @@
...
+    return FieldDiff(mixed_error(x1, x2), rms(x1 .- x2), rms(x1))
+end
+
+"""
+    mask_nonfinite(x)
+
+Return zero if `x` is a non-finite value (e.g., `NaN` and `Inf`) and `x`
+otherwise.
+"""
+mask_nonfinite(x) = x .* isfinite.(x)
```
imreddyTeja, line 54: huh, I wouldn't expect this behavior:
```
julia> NaN * false
0.0

julia> NaN * 0
NaN

```

experiments/integrated/era5/comparison_utils.jl
```diff
@@ -0,0 +1,191 @@
...
+
+"""
+    flatten(field::ClimaCore.Fields.Field, col::Union{Nothing, Integer})
+
+Flatten a `ClimaCore.Fields.Field` into a vector.
+
+Pass `nothing` for `col` if the space is a single column and pass an integer
```
imreddyTeja, line 61: Probably a good idea to mention NaNs get zeroed out in the docstring

src/shared_utilities/Domains.jl
```diff
@@ -246,6 +246,92 @@ function Column(;
...
+Outer constructor for the 1D `ColumnEnsemble` domain.
+
+See [`Column`](@ref) for usage of this function and a description of possible
+keyword arguments.
+
+This function differs from [`Column`](@ref) as a `ColumnEnsemble` is backed by a
+`MultiColumnFiniteDifferenceSpace` which is an ensemble of N independent
+vertical columns at arbitrary (long, lat) locations.
+"""
+function ColumnEnsemble(;
```
imreddyTeja, line 306: Could you add this to the exports at the bottom of this file.

experiments/integrated/era5/comparison_utils.jl
```diff
@@ -0,0 +1,191 @@
...
+function mixed_error(x1, x2; abs_floor = 100 * eps(eltype(x1)))
+    diff = abs.(x1 .- x2)
+    return maximum(ifelse.(abs.(x1) .> abs_floor, diff ./ abs.(x1), diff))
+end
+
+"""
+    FieldDiff
+
+Store the differences between two fields:
+1. The worst elementwise absolute or relative error,
```
imreddyTeja, line 33: This confused me at first. I thought it was saying that it stores max(abs_err, rel_err). Is my understanding that it is taking `max.(abs_errs)` or `max.(rel_errs)`  depending on the magnitude of each point of the first field?

ph-kev:
> Did you have a chance to check how fast the ensemble column runs on GPU for e.g. 10, 30, 100, 300 columns? (or something like that).

I asked a LLM to I check it up to 8192 columns. This is running the ERA5 sites with duplicated columns so I am not sure how well this translates to other fluxnet simulations. Qualitatively, for the CPU benchmark, it is linear scaling and for the GPU benchmark, there is no cost to doubling the work until the number of columns is large. 

```
# Each simulation ran for 10 days
ColumnEnsemble scaling benchmark
Device: ClimaComms.CPUSingleThreaded()
Timed steps per size: 960 (Δt = 900.0 s)
N_columns  total (s)    ms/step  ms/step/column
        1      0.958      0.998          0.9978
        2      0.739      0.770          0.3848
        4      1.054      1.098          0.2744
        8      1.333      1.388          0.1736
       16      2.105      2.192          0.1370
       32      3.738      3.893          0.1217
       64      6.971      7.262          0.1135
      128     13.345     13.901          0.1086
      256     26.340     27.437          0.1072
      512     52.348     54.530          0.1065
     1024    103.198    107.498          0.1050
     2048    206.126    214.715          0.1048
     4096    415.764    433.087          0.1057
     8192    839.354    874.327          0.1067

ColumnEnsemble scaling benchmark
Device: ClimaComms.CUDADevice()
Timed steps per size: 960 (Δt = 900.0 s)
N_columns  total (s)    ms/step  ms/step/column
        1      4.913      5.117          5.1174
        2      4.554      4.744          2.3719
        4      4.426      4.611          1.1527
        8      4.388      4.570          0.5713
       16      4.462      4.648          0.2905
       32      4.413      4.596          0.1436
       64      4.435      4.620          0.0722
      128      4.379      4.561          0.0356
      256      4.440      4.625          0.0181
      512      4.481      4.667          0.0091
     1024      4.571      4.762          0.0047
     2048      5.114      5.327          0.0026
     4096      5.517      5.747          0.0014
     8192      6.121      6.376          0.0008
```

experiments/integrated/era5/comparison_utils.jl
ph-kev: It is computing absolute error or relative error depending on the reference value and taking the max elementwise error from that.

src/shared_utilities/Domains.jl
```diff
@@ -246,6 +246,92 @@ function Column(;
...
+    zlim::Tuple{FT, FT},
+    nelements::Int,
+    longlat::Union{Tuple{FT, FT}, AbstractVector{Tuple{FT, FT}}},
```
kmdeck: Since below you convert a single longlat into a vector, why not require a vector, and state in the doc string that you must pass a vector of longlats where the length of the vector can be >=1. not sure. If someone really wants to do a single column they should use a Column domain

    longlats = longlat isa AbstractVector ? longlat : [longlat]

src/shared_utilities/Domains.jl
```diff
@@ -246,6 +246,92 @@ function Column(;
...
+
+    surface_space = obtain_surface_space(subsurface_space)
+    subsurface_face_space = ClimaCore.Spaces.face_space(subsurface_space)
+    space = (;
+        surface = surface_space,
+        subsurface = subsurface_space,
+        subsurface_face = subsurface_face_space,
+    )
+    fields = get_additional_coordinate_field_data(subsurface_space)
+    return Column{FT, typeof(space), typeof(fields)}(
```
kmdeck, line 363: is it confusing that this returns a Column type, and not a ColumnEnsemble type?

kmdeck: I guess I am confused why we do not have a new Domain struct defined in this PR, rather than adding a new constructor for a Column domain

Project.toml
kmdeck: Could we also remove any lines like this: https://github.com/CliMA/ClimaLand.jl/blob/cdd02a620d1518c9d03368ff32b207a4915d786b/src/shared_utilities/Domains.jl#L1401

experiments/integrated/fluxnet/column_ensemble_comparison.jl
kmdeck: I think this is fine, but it's a bit odd because the spherical geometry/curvature of the space should not matter for running column runs, where horizontal extent does not matter sort of by definition. We can also define latlong points on a cartesian grid.

kmdeck:
> > Did you have a chance to check how fast the ensemble column runs on GPU for e.g. 10, 30, 100, 300 columns? (or something like that).
> 
> I asked a LLM to I check it up to 8192 columns. This is running the ERA5 sites with duplicated columns so I am not sure how well this translates to other fluxnet simulations. Qualitatively, for the CPU benchmark, it is linear scaling and for the GPU benchmark, there is no cost to doubling the work until the number of columns is large.
> 
> ```
> # Each simulation ran for 10 days
> ColumnEnsemble scaling benchmark
> Device: ClimaComms.CPUSingleThreaded()
> Timed steps per size: 960 (Δt = 900.0 s)
> N_columns  total (s)    ms/step  ms/step/column
>         1      0.958      0.998          0.9978
>         2      0.739      0.770          0.3848
>         4      1.054      1.098          0.2744
>         8      1.333      1.388          0.1736
>        16      2.105      2.192          0.1370
>        32      3.738      3.893          0.1217
>        64      6.971      7.262          0.1135
>       128     13.345     13.901          0.1086
>       256     26.340     27.437          0.1072
>       512     52.348     54.530          0.1065
>      1024    103.198    107.498          0.1050
>      2048    206.126    214.715          0.1048
>      4096    415.764    433.087          0.1057
>      8192    839.354    874.327          0.1067
> 
> ColumnEnsemble scaling benchmark
> Device: ClimaComms.CUDADevice()
> Timed steps per size: 960 (Δt = 900.0 s)
> N_columns  total (s)    ms/step  ms/step/column
>         1      4.913      5.117          5.1174
>         2      4.554      4.744          2.3719
>         4      4.426      4.611          1.1527
>         8      4.388      4.570          0.5713
>        16      4.462      4.648          0.2905
>        32      4.413      4.596          0.1436
>        64      4.435      4.620          0.0722
>       128      4.379      4.561          0.0356
>       256      4.440      4.625          0.0181
>       512      4.481      4.667          0.0091
>      1024      4.571      4.762          0.0047
>      2048      5.114      5.327          0.0026
>      4096      5.517      5.747          0.0014
>      8192      6.121      6.376          0.0008
> ```

Can I summarize this by saying:
to run 128 columns, you could request 128 CPUs have have it take < 1 second each using a Column (C), OR run with ColumnEnsemble (CE) and have it take 13 seconds, OR run with CE on GPU and have it take 4 seconds?

It looks from here like more than~30-60 columns would mean the GPU+CE is the better choice,
below this CPU+CE is the better choice until you have enough CPU resources to run them all in parallel

ph-kev:
> 
> Can I summarize this by saying: to run 128 columns, you could request 128 CPUs have have it take < 1 second each using a Column (C), OR run with ColumnEnsemble (CE) and have it take 13 seconds, OR run with CE on GPU and have it take 4 seconds?
> 
> It looks from here like more than~30-60 columns would mean the GPU+CE is the better choice, below this CPU+CE is the better choice until you have enough CPU resources to run them all in parallel

This sounds right to me. For ~32 columns, it is worth it to use CPU + CE and for more than that, it is worth it to use GPU + CE. If you have more resources, it is faster in total wait time to have each CPU work on a single column.

src/shared_utilities/Domains.jl
ph-kev: I can change this, but I think we need a ColumnEnsemble type and a AbstractColumn abstract type and maybe a PointEnsemble type too.

kmdeck: It's ok with me if you think this is not confusing. but maybe could you add something to the docstring saying "Column" refers to either a single column or a collection of columns? and likewise for Point?

kmdeck (approved):

experiments/integrated/fluxnet/column_ensemble_comparison.jl
kmdeck: I sort of think this change should happen, especially if it is easy. But it doesnt have to hold anything up with this PR. If you are unsure you could ask Tapio during your next 1:1, because maybe it could matter for atmos if they used this, but again, I think it is odd to have a non-cartesian collection of independent columns.

src/shared_utilities/Domains.jl
imreddyTeja: Could you also add a const type alias named something like `ColumnEnsemble`

.buildkite/pipeline.yml
```diff
@@ -78,6 +78,9 @@ steps:
         command: "julia --color=yes --project=.buildkite experiments/integrated/fluxnet/run_fluxnet.jl US-MOz"
         artifact_paths: "experiments/integrated/fluxnet/US-MOz/out/*png"
 
+      - label: "ERA5 Column vs ColumnEnsemble"
```
imreddyTeja, line 81: probably not needed, but should we add a test for this functionality on gpu?

experiments/integrated/era5/column_ensemble_comparison.jl
```diff
@@ -0,0 +1,168 @@
+# Run a ERA5 single site simulation on a single `Column` and a
+# `ColumnEnsemble` of duplicate columns. We test that
+# 1. The ensemble columns are identical to each other,
+# 2. Every ensemble column matches the single column run within some tolerance.
+# This involves checking the prognostic state, the cache, and the halfhourly
+# diagnostics.
+
+# As of now, the main differences between a single `Column` and a
+# `ColumnEnsemble` of duplicate columns is the geometry. A single column use
```
imreddyTeja: nit:
```suggestion
# `ColumnEnsemble` of duplicate columns is the geometry. A single column uses
```

experiments/integrated/era5/column_ensemble_comparison.jl
```diff
@@ -0,0 +1,168 @@
...
+function setup_model(
+    ::Type{FT},
+    start_date,
+    stop_date,
+    Δt,
+    domain,
+    toml_dict,
+) where {FT}
+    surface_space = domain.space.surface
+    # Forcing data - low resolution
```
imreddyTeja: nit: I think this comment is pointing out something that is pretty clear already in the next few lines

src/shared_utilities/Domains.jl
```diff
@@ -55,19 +55,20 @@ end
...
 
-A domain for single column surface variables.
+A domain for single or multiple point surface variables.
+
 For models such as ponds, snow, plant hydraulics, etc. Enables consistency
 in variable initialization across all domains.
 
-`space` is a NamedTuple holding the surface space (in this case,
-the Point space).
+`space` is a NamedTuple holding the surface space which is a `PointSpace` for a
```
imreddyTeja, line 63: Same suggestion about const alias

src/shared_utilities/Domains.jl
```diff
@@ -246,6 +251,89 @@ function Column(;
...
+        zlim::Tuple{FT, FT},
+        nelements::Int,
+        longlat::AbstractVector{Tuple{FT, FT}},
+        dz_tuple::Union{Tuple{FT, FT}, Nothing} = nothing,
+        device = ClimaComms.device(),
+    ) where {FT}
+
+Outer constructor for a multiple column domain.
+
+See [`Column`](@ref) for a description of the keyword arguments.
```
imreddyTeja: It think it would be good to make the docstring a little more detailed. You could include something about the size of the `longlat` vector determines how many columns there are. Another minor thing is the other 2d/3d constructors take `nelements` in as a tuple.

src/shared_utilities/Domains.jl
```diff
@@ -840,6 +928,22 @@ function obtain_surface_space(cs::ClimaCore.Spaces.FiniteDifferenceSpace)
     )
 end
 
+"""
+    obtain_surface_space(cs::ClimaCore.Spaces.CenterMultiColumnFiniteDifferenceSpace)
+
+Returns the top level (surface) face space of the multi-column center space `cs`,
+a `PointCloudSpace`.
+"""
+function obtain_surface_space(
```
imreddyTeja, line 964: This is pretty much the same method as the one above. They could be combined with dispatch over a union, but then the docstrings would need to be combined. up to you

imreddyTeja (approved):
left some minor comments

.buildkite/pipeline.yml
ph-kev: I don't think it is too important to test this on GPU for now. We can change this later.

PR #1797, opened 2026-07-07
Time-integrated variables as smooth prognostic state (+ optimal-LAI A0_annual)

AlexisRenchon:
## Motivation

Several models integrate a quantity in time — the optimal-LAI annual potential GPP and precipitation, the P-model acclimated capacities, the optimal-LAI acclimation lag itself. These lived in the **cache `p`** and were advanced by **local-noon callbacks**, so they were lost on restart (a blocker for the P-model in `ClimaCoupler.jl`), and the annual totals were finalized with a hard 365-day reset that made them **jump once per simulated year** into `LAI_max` and modeled LAI.

This PR makes them **prognostic ODE variables in `Y`**, advanced by the time-stepper every stage: smooth, callback-free, and checkpoint/restart-safe by construction. It generalizes the pattern already used for the soil conservation trackers `∫F_vol_liq_water_dt` / `∫F_e_dt`.

## Contents

**1. `TimeIntegratedVariable` utility** (`src/shared_utilities/time_integrated_variables.jl`)

A declaration — `name`, `reduction`, `domain_name`, `element_type` — stored on the model that owns it. The reduction carries its own timescales, where `X` is the stored state and `f` the instantaneous quantity:

| reduction | ODE | steady state |
|---|---|---|
| `RunningMean(τ)` | `dX/dt = (f − X)/τ` | `f̄` — exp-weighted running mean |
| `RunningSum(τ, τ_long)` | `dX/dt = (f τ − X)/τ_long` | `τ f̄` — running total over a τ window, smoothed over `τ_long` |
| `TimeIntegral()` | `dX/dt = f` | none — pure accumulator (the `∫F…dt` case) |

`time_integrated_variables` collects the specs by name; `time_integrated_prognostic_vars/types/domain_names` expand them into the declaration tuples; a model's tendency writes `@. dY.x = apply_time_reduction(f, Y.x, reduction)`. Kernels use `ClimaCore.RecursiveApply` (`⊟`, `⊠`, `rdiv`), so a variable may be scalar- or NamedTuple-valued (`element_type`), and `f` may be a `lazy` broadcast.

These are exponentially-weighted (leaky) reductions, not hard boxcars: a true trailing window is not a local ODE, whereas the leaky integrator is one cheap prognostic scalar and is smooth. For acclimation the exponential memory is arguably more physical than a hard cutoff.

**2. Optimal-LAI: four prognostics in `Y.canopy.biomass`** (`biomass.jl`, `optimal_lai.jl`)

[...]

PR #1831, opened 2026-07-28
Climate-responsive optimal-LAI inputs: online f0, VPD_gs, GSL and C3/C4

### LINKED ISSUE #1830: Recalibrate gpp+energy, update LAIopt IC, recalibrate LAI (with cropland mask)
branching off off PR #1815

### LINKED ISSUE #1856: Consider a grass tree fraction model
We have (not merged) a c3 / c4 competition model, but not a grass / tree model. It may be useful to have - e.g., for biomass or fire. It would be complementary of the c3/c4 competition. 

look into: (LLM generated text below:)
```
The cleanest match is probably VECODE. It was deliberately designed as a reduced-complexity dynamic vegetation model with essentially trees, grasses, and bare ground:

ftree + fgrass + fbare = 1.

Tree/grass equilibrium fractions are functions mainly of climate—precipitation, temperature/growing-degree days—and vegetation approaches those equilibrium fractions dynamically over time. The original formulation explicitly describes vegetation continuously in terms of tree and grass cover, rather than many detailed PFTs
```

This could also help the LAI model, see Google PR #1852, surely [biomass](https://claude.ai/code/artifact/abe376d5-57dc-430d-8fcc-b0fd8237e900) and fire, possibly other things (like radiative transfer or canopy temperature?)

AlexisRenchon:
closes #1830
closes #1856 

Follow-up to #1797.

The optimal-LAI model read four inputs from frozen maps in an artifact: the transpiration fraction `f0`, the growing-season VPD `vpd_gs`, the growing-season length `GSL`, and the C3 fraction. All four are now derived from the simulated climate:

- **`f0`** from the aridity index `AI = PET_annual / precip_annual`, through Zhou et al.'s `f0 = f0_max exp(-0.604 ln²(AI/1.9))`. `PET` is the FAO-56 Penman-Monteith reference evapotranspiration, the definition the `f0(AI)` relation was fitted with. `optimal_lai_f0` is renamed `optimal_lai_f0_max`, since it now sets the peak of a curve.
- **`vpd_gs`** as the A0-weighted mean VPD, `VPDA0_annual / A0_annual`. Zhou et al. use a plain growing-season mean. `vpd_gs` only enters the water-limited term `f0·P/A0 · ca(1−χ)/(1.6·D)`, and because transpiration and assimilation are both annual integrals, the single `D` that makes annual `A/E` exact is the assimilation-weighted mean. A growing-season mask is a 0/1 version of that weighting; A0-weighting is the continuous one and needs no season window.
- **`GSL`** as a trailing-year count of days with air temperature above freezing.
- **`fractional_c3`** from the C3/C4 competition of Lavergne et al. (2022), as in pyrealm: the C4 advantage in per-pathway potential GPP goes through a logistic, then C4 is reduced by the C3 tree cover estimated from the annual realized C3 GPP. `optimal_lai_online_c3c4 = 0` holds it at the static CLM map; the default is on.

The potential GPP `A0` no longer carries soil-moisture stress, so water limitation enters once, through `f0·P/A0`.

## New prognostic state

Six more time-integrated variables, `PET_annual`, `VPDA0_annual`, `growing_days`, `A0c3_annual`, `A0c4_annual` and `GPPc3_annual`, each a `RunningSum(365 days, tau_long_term)` of the instantaneous rate, registered in the implicit Jacobian's explicit list. `growing_days` accumulates `1/day` while air T > 0 °C, so its total is a day count. `GPPc3_annual` is the C3 potential GPP times the realized fAPAR, i.e. the annual realized C3 GPP the Lavergne tree-cover relation was fitted on.

`set_ic!` seeds `PET_annual`, `VPDA0_annual` and `growing_days` so that `f0`, `vpd_gs` and `GSL` start at the artifact values they replace, then relax to the simulated climate over `tau_long_term`. `f0(AI)` peaks at `f0_max`, so the seed uses the arid branch of the inverse, and a cell whose artifact `f0` exceeds the peak seeds to the peak. `A0c3_annual` and `A0c4_annual` have no per-pathway climatology and both start at `a0_annual`, so the competition sees no GPP advantage at `t = 0` and the online C3 fraction starts near-uniform rather than at the static map; the two settings of `optimal_lai_online_c3c4` only diverge as the totals separate. `GPPc3_annual` is seeded as `A0c3_annual` times the fAPAR of the MODIS annual-maximum LAI (the `modis_max_lai` artifact already used for SAI), not of `lai_init`, which is a single-date snapshot: potential GPP is concentrated in the leafy season, so the peak fAPAR is the closer estimate of the GPP-weighted annual value.

`GSL`, `vpd_gs` and `f0` are cache variables again, derived in `update_biomass!`.

Two implementation notes:

- `compute_A0_and_χ` already computed the C3 and C4 light-use efficiencies before blending, so it now also returns `A0_c3` and `A0_c4`.
- `fractional_c3` is written to `p.canopy.photosynthesis.fractional_c3` by a dispatching `update_fractional_c3!` called from `update_photosynthesis!`, every cache update rather than once in `set_ic!`, so a checkpoint restart gets the right value. SIF and the `Vcmax25` diagnostics blend on the same cache value.

## Canopy composition: tree, C3 grass, C4 grass

The competition computes a C3 tree share as an intermediate (the Lavergne tree cover at the realized C3 GPP, relative to the cover at canopy closure `tc_gpp_ref`), and uses it to shade out C4. That partition is now stored as `p.canopy.biomass.composition` with fields `tree`, `c3_grass` and `c4_grass`, shares of productivity summing to one, with `fractional_c3 = tree + c3_grass`. It is computed by `canopy_composition_from_competition`; `c3_fraction_from_competition` is a wrapper returning `1 − c4_grass`. The prognostic biomass model can then allocate wood on the tree share rather than on the C3 fraction, which conflates C3 grassland with forest.

The tree cover was first estimated from the annual potential C3 GPP times the *instantaneous* fAPAR, an annual total scaled by today's leaf area, so it followed the seasonal cycle of LAI: about 0.1 at Ozark in January. With `GPPc3_annual` it is an annual-scale quantity, as tree cover is (0.90 at Ozark in January, moving by under 0.5 % while LAI rises 12 % over two days). This also removes the LAI-driven seasonal wobble from `fractional_c3`, so GPP at mixed C3/C4 cells shifts slightly toward C3 in winter and spring; the recalibrated defaults below predate this change.

## Breaking: `ZhouOptimalLAIModel` constructor

The model no longer carries the static maps:

```julia
# before
ZhouOptimalLAIModel{FT}(parameters, optimal_lai_inputs; SAI, RAI, rooting_depth, height)
ZhouOptimalLAIModel{FT}(domain, toml_dict; optimal_lai_inputs = ..., SAI, ...)

# after
ZhouOptimalLAIModel{FT}(parameters; SAI, RAI, rooting_depth, height)
ZhouOptimalLAIModel{FT}(domain, toml_dict; SAI, ...)
```

The artifact is only read by `set_ic!`, to seed the accumulators. Its reader is renamed from `optimal_lai_static_inputs` back to `optimal_lai_initial_conditions`, matching the artifact accessor it wraps, and `set_canopy_component_initial_conditions!` calls it instead of re-reading the same variables.

## Default parameter changes

| parameter | before | after |
|---|---|---|
| `optimal_lai_z` | 21.4 | 29.2 |
| `optimal_lai_sigma` | 0.939 | 1.01 |
| `optimal_lai_alpha` | 0.0701 | 0.202 |
| `optimal_lai_f0` → `optimal_lai_f0_max` | 0.65 | 0.65 |
| `optimal_lai_online_c3c4` | — | 1.0 (on) |
| `optimal_lai_c3c4_k`, `_q` | — | 6.63, 0.16 |
| `optimal_lai_tc_a`, `_b`, `_c`, `_gpp_ref` | — | 15.60, 1.41, −7.72, 2.8 |

The C3/C4 coefficients are the Lavergne et al. (2022) values used by pyrealm.

The P-model, moisture-stress and canopy-flux defaults are also updated, from a separate calibration:

| parameter | before | after |
|---|---|---|
| `pmodel_cstar` | 0.43 | 0.366 |
| `pmodel_β_c3` | 91.1 | 24.9 |
| `pmodel_β_c4` | 5.36 | 8.89 |
| `pmodel_α` | 0.028 | 0.0251 |
| `moisture_stress_c` | 0.59 | 0.591 |
| `canopy_K_lw` | 0.919 | 1.16 |
| `canopy_z_0m_coeff` | 0.349 | 0.0823 |
| `canopy_z_0b_coeff` | 0.0444 | 0.073 |
| `canopy_d_coeff` | 0.0573 | 0.0416 |
| `leaf_Cd` | 0.0726 | 0.07 |

`z`, `sigma` and `alpha` were recalibrated against MODIS LAI with the online inputs; the previous values were tuned against the frozen maps. That tuning ran with `optimal_lai_tau_long_term` = 1 year and an `f0` multiplier that is not in this PR, so the defaults should be re-tuned before being relied on.

## Diagnostics

`a0c3`, `a0c4` and `fc3`, so the C3/C4 competition can be checked against observed pathway distributions, and `ftr`, `fc3g`, `fc4g` for the canopy composition (`ftr` compares with MODIS VCF tree cover below canopy closure).

## Testing

`test_optimal_lai.jl`, `test_pmodel.jl` and `time_integrated_variables.jl` pass on Julia 1.12 after the rebase onto main; JuliaFormatter 2.10.1 clean. `experiments/integrated/fluxnet/ozark_pmodel.jl` runs to completion, and a 2-day Ozark column run with the composition diagnostics gives shares summing to exactly 1 and `fc3` equal to `tree + c3_grass` exactly.

https://claude.ai/code/session_016opW9XRGxpBgHmZBYupc9S


src/diagnostics/define_diagnostics.jl
```diff
@@ -360,6 +382,51 @@ function define_diagnostics!(land_model, possible_diags)
...
     )
 
+    # Fraction of C3 photosynthesis (dynamic C3/C4 competition when enabled)
+    conditional_add_diagnostic_variable!(
+        possible_diags;
+        short_name = "fc3",
+        long_name = "Fraction C3 Photosynthesis",
+        standard_name = "fraction_c3_photosynthesis",
+        units = "",
+        comments = "Fraction of C3 (vs C4) photosynthesis, 1 = all C3. From the optimal-LAI C3/C4 competition, unless it is turned off (optimal_lai_online_c3c4 = 0), which holds it at the static CLM map.",
```
kmdeck: "Fraction of C3 (vs C4) photosynthesis, 1 = all C3"

kmdeck: I think historically we just say what the diagnostic is, and not go into details about how differentparameterizations model it differently. The user picks their parameterizations and diagnostics and the right one will comeout

src/diagnostics/land_compute_methods.jl
```diff
@@ -290,6 +290,38 @@ end
...
+
+@diagnostic_compute "a0c4_annual" Union{SoilCanopyModel, LandModel, CanopyModel} Y.canopy.biomass.A0c4_annual
+
+# Fraction of C3 photosynthesis (1 = all C3, 0 = all C4). From the optimal-LAI C3/C4
+# competition unless that is off (optimal_lai_online_c3c4 = 0), then the static CLM map.
+@diagnostic_compute "fractional_c3" Union{
+    SoilCanopyModel,
+    LandModel,
+    CanopyModel,
+} p.canopy.photosynthesis.fractional_c3
```
kmdeck: Do we need to dispatch off of the type of photosynthesis model, or biomass model here? the c3 fraction is stored in the photosynthesis parameters for Pmodel and Farquhar, not in the cache.

src/diagnostics/land_compute_methods.jl
```diff
@@ -290,6 +290,38 @@ end
...
+    SoilCanopyModel,
+    LandModel,
+    CanopyModel,
+} p.canopy.biomass.composition.tree
+
+@diagnostic_compute "fraction_c3_grass" Union{
+    SoilCanopyModel,
+    LandModel,
+    CanopyModel,
+} p.canopy.biomass.composition.c3_grass
```
kmdeck: will this fail gracefully if the Prescribed Biomass model is used?

src/simulations/initial_conditions.jl
```diff
@@ -358,19 +358,36 @@ end
...
+
+`PET_annual`, `VPDA0_annual` and `growing_days` are seeded so that `f0`, `vpd_gs`
+and `GSL` start at the artifact values they replace, then relax to the simulated
+climate over `tau_long_term`. Since `f0(AI)` peaks at `f0_max`, the `f0` seed uses
+the arid branch of the inverse, and a cell whose artifact `f0` exceeds the peak
+seeds to the peak.
+
+No per-pathway climatology exists, so `A0c3_annual` and `A0c4_annual` both start at
+`a0_annual`. The competition then sees no GPP advantage at `t = 0`, so the online
+C3 fraction starts near-uniform, not at the static map, and only diverges from it
```
kmdeck: is it worth reading the static fractional c3 map and using that to back out the ratio of A0c3_annual and A0c4 annual to reduce spin up?

src/standalone/Vegetation/spatially_varying_parameters.jl
```diff
@@ -379,7 +379,7 @@ function clm_canopy_height(
 end
 
 """
-    optimal_lai_static_inputs(
+    optimal_lai_initial_conditions(
```
kmdeck: Could we move this function to src/simulations/initial_conditions.jl? I think it belongs there

src/standalone/Vegetation/spatially_varying_parameters.jl
```diff
@@ -390,9 +390,10 @@ end
         interpolation_method = Interpolations.Constant(),
     )
 
-Reads the spatially varying optimal LAI data from a NetCDF file, and regrids it to the
-grid defined by the `surface_space` of the Clima simulation. Returns a NamedTuple of
-ClimaCore Fields suitable for passing to `ZhouOptimalLAIModel`.
+Reads the optimal-LAI climatology from a NetCDF file and regrids it to
+`surface_space`, returning a NamedTuple of ClimaCore Fields. These fields seed the
+optimal-LAI prognostic state in `set_canopy_component_initial_conditions!`; the model
+does not read them at runtime.
```
kmdeck: "the model does not read them at runtime" - what does this mean? We do read them in at run time

toml/default_parameters.toml
```diff
@@ -308,19 +350,19 @@ description = "Maximum value for the ratio of Gamma star over the partial pressu
 tag = "PModelParameters"
 
 ["pmodel_cstar"]
-value = 0.43
+value = 0.366
 type = "float"
```
kmdeck: I thought we had discussed optimizing these parameters using energy fluxes + GPP using a prescribed LAI, and then optimizing the LAI parameters separately with these fixed. was that the procedure here?

NEWS.md
```diff
@@ -2,6 +2,22 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+- ![][badge-🔥behavioralΔ] Optimal-LAI climate inputs (`f0`, growing-season VPD and length,
```
kmdeck: maybe

```
- ![][badge-🔥behavioralΔ] Optimal-LAI climate inputs now computed as a function of time throughout the simulation using time integrated prognostic vars; `ZhouLAIOptimalLAIModel` no longer uses the optimal_lai_inputs keyword arg; trees vs grass now predicted and accounted for when predicting LAI; potential gpp now assumes moisture stress is 1.
```

test/standalone/Vegetation/test_optimal_lai.jl
```diff
@@ -23,65 +23,76 @@ using ClimaCore
...
 
             # Check expected values from default_parameters.toml (calibrated
-            # against MODIS LAI in #1794)
+            # against MODIS LAI)
             @test params.k ≈ FT(0.5)
-            @test params.z ≈ FT(21.4)
-            @test params.sigma ≈ FT(0.939)
-            @test params.alpha ≈ FT(0.0701)  # ~14 days of memory
+            @test params.z ≈ FT(29.2)
+            @test params.sigma ≈ FT(1.01)
```
kmdeck: Do we have a test of restarts with the optimal LAI model with these additional prognostic vars?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -838,6 +844,21 @@ function update_photosynthesis!(p, Y, model::PModel, canopy)
...
+    update_fractional_c3!(p, Y, biomass::AbstractBiomassModel, canopy)
+
+Writes the C3 fraction the P-model blends with into
+`p.canopy.photosynthesis.fractional_c3`. This default takes the P-model's static
+value; a biomass model with a C3/C4 competition (`ZhouOptimalLAIModel`) overrides it.
+
+It is written on every cache update rather than seeded once so that a checkpoint
+restart, whose `set_ic!` skips the component initial conditions, gets the right value.
+"""
+function update_fractional_c3!(p, Y, biomass::AbstractBiomassModel, canopy)
```
kmdeck: since this depends on the biomass model, should this be part of the the biomass file? 

and, should `update_fractional_c3!` be part of `update_biomass!`?

also, this general method if kept in photosynthesis should live in photosynthesis.jl, since the farquhar model would also use it?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -838,6 +844,21 @@ function update_photosynthesis!(p, Y, model::PModel, canopy)
...
+    p.canopy.photosynthesis.fractional_c3 .= canopy.photosynthesis.fractional_c3
```
kmdeck: I always worry about storing the same quantity in two places...

what does it mean if we use the pmodel with Zhou? We would have the static map in 
canopy.photosynthesis.fractional_c3, but a changing map in p.canopy.photosynthesis.fractional_c3


this is pretty confusing, and I think arises from it being unclear which component model controls the c3 fraction.

src/standalone/Vegetation/biomass.jl
```diff
@@ -548,10 +596,10 @@ ClimaLand.prognostic_domain_names(m::ZhouOptimalLAIModel) =
         canopy,
     ) where {FT}
 
-Updates the optimal-LAI cache: sets SAI and RAI to their prescribed constant
-values and mirrors the prognostic `LAI` into the cache area index read by the rest
-of the canopy. LAI must be updated here first in `update_aux`, before radiative
-transfer reads the area index.
+Updates the optimal-LAI cache: sets SAI and RAI to their prescribed values, derives
+`f0`, `vpd_gs` and `GSL` from the trailing climate totals in `Y`, and mirrors the
+prognostic `LAI` into the cache area index read by the rest of the canopy. This runs
```
kmdeck: I think it is worth being careful here-  the reason why we have Y.LAI and p.LAI is because they are actually not the same thing. p.LAI is clipped and also then masked to be zero where lakes are present. Y.LAI is a theoretical quantity that depends on local climate.

If these were exactly the same it would be confusing and redundant to have two fields, but the comment makes it sound like that is the case

src/standalone/Vegetation/biomass.jl
```diff
@@ -563,33 +611,92 @@ function update_biomass!(
...
+    update_fractional_c3!(p, Y, biomass::ZhouOptimalLAIModel, canopy)
+
+Sets the canopy `composition` (tree, C3 grass, C4 grass shares) from the C3/C4
+competition on the trailing per-pathway potential GPP (`A0c3_annual`, `A0c4_annual`)
+and the trailing realized C3 GPP (`GPPc3_annual`), and the C3 fraction as its tree
+plus C3 grass share. With `optimal_lai_online_c3c4 = 0`
+the C3 fraction is the P-model's static value instead; `composition` still reports
+the competition, as the static map carries no tree/grass split.
+"""
+function update_fractional_c3!(
```
kmdeck: this makes it seem like c3 should be part of the biomass model, and NOT the photosynthesis model.

src/standalone/Vegetation/biomass.jl
```diff
@@ -563,33 +611,92 @@ function update_biomass!(
...
 ) where {FT}
     ρ_m_liq = LP.ρ_m_liq(canopy.earth_param_set)  # mol H2O m^-3 (precip volume flux → molar flux)
     tivs = component.time_integrated_vars
+    seconds_per_day = IP.day(IP.InsolationParameters(FT))
+    earth_param_set = canopy.earth_param_set
+    σ = LP.Stefan(earth_param_set)
+    M_w = LP.molar_mass_water(earth_param_set)  # kg mol^-1
+    T_freeze = LP.T_freeze(earth_param_set)
+    thermo_params = LP.thermodynamic_parameters(earth_param_set)
+    h_atmos = ClimaLand.get_drivers(canopy)[1].h
```
kmdeck: I think better to do canopy.boundary_conditions.atmos.h

src/standalone/Vegetation/biomass.jl
```diff
@@ -563,33 +611,92 @@ function update_biomass!(
...
+        fractional_c3 = p.canopy.photosynthesis.fractional_c3
+        VPD = @. lazy(
+            max(
+                Thermodynamics.vapor_pressure_deficit(
+                    thermo_params,
+                    p.drivers.T,
+                    p.drivers.P,
+                    p.drivers.q,
+                ),
+                sqrt(eps(FT)),
```
kmdeck: out of curiosity, is this max needed? or is VPD of zero OK?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -367,3 +395,171 @@ function compute_L_steady_target(
...
+(Zhou et al. 2025): `f0 = f0_max·exp(−0.604·ln²(AI/1.9))` with aridity index
+`AI = PET_annual/precip_annual`. Peaks at `f0_max` at the energy–water transition
+(AI = 1.9) and declines toward both the arid and humid extremes.
+"""
+function f0_from_aridity(
+    PET_annual::FT,
+    precip_annual::FT,
+    f0_max::FT,
+) where {FT}
+    AI = max(PET_annual, eps(FT)) / max(precip_annual, eps(FT))
```
kmdeck: can PET_annual be negative?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -367,3 +395,171 @@ function compute_L_steady_target(
...
+    aridity_from_f0(f0::FT, f0_max::FT) where {FT}
+
+Inverse of [`f0_from_aridity`](@ref), used to seed `PET_annual` so the online `f0`
+starts at the value of the map it replaces. `f0` is symmetric in `ln(AI/1.9)`, so
+this returns the arid branch (`AI ≥ 1.9`), where the map was fitted. An `f0` at or
+above `f0_max` has no preimage and returns the peak `1.9`.
+"""
+function aridity_from_f0(f0::FT, f0_max::FT) where {FT}
+    return FT(AI_PEAK) *
+           exp(sqrt(max(log(f0_max / max(f0, eps(FT))), FT(0)) / FT(AI_WIDTH)))
```
kmdeck: Are all these clips required? It's fine if so, but they make it much harder to read

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -367,3 +395,171 @@ function compute_L_steady_target(
...
+    P_air::FT,
+    q_air::FT,
+    u_air::FT,
+    h_atmos::FT,
+    ϵ_sfc::FT,
+    σ::FT,
+    M_w::FT,
+    thermo_params,
+) where {FT}
+    α_ref = FT(0.23)
```
kmdeck: Magic numbers in code...where do they come from?

src/standalone/Vegetation/optimal_lai.jl
```diff
@@ -367,3 +395,171 @@ function compute_L_steady_target(
...
+        thermo_params,
+        T_air,
+        P_air,
+        q_air,
+    )
+    ρ_a = Thermodynamics.air_density(thermo_params, T_air, P_air, q_air)
+
+    # Wind at the reference 2 m (FAO-56 Eq. 47); the relation is anchored on the
+    # reference crop, so heights below it are held at 2 m rather than extrapolated.
+    u_2 = u_air * FT(4.87) / log(FT(67.8) * max(h_atmos, FT(2)) - FT(5.42))
```
kmdeck: magic numbers

kmdeck (commented):
It looks fine to me, but I did not check the equations to make sure they correspond with the paper or are computing the right thing...

From a software point of view:
- If biomass controls c3/c4 fraction, it should probably live in the biomass model. It's confusing to have `update_c3_fraction!` which depends on the biomass model only be called in photosynthesis. I also find it confusing to have cache vars and things in the photosynthesis parameter struct be called the same thing but potentially differ. so i think this can be cleaned up quite a bit...
- do all combinations work? PModel+Prescribed, Pmodel+Optimal, Farquhar + Prescribed, Farquhar+Optimal? Are we adding in a third option (PModel+OptimalLAI+C3/c4 prescribed vs + c3/c4 predicted)? Or keeping it simple - Either LAI + c3/c4 prescribed OR they are both predicted?

PR #1817, opened 2026-07-23
Nb/pmodel syntax

nicobartolomeo:
## Purpose 
Updates syntax with pmodel.jl and connected files to align with (Mengoli 2022)

## Content
Updates in
- pmodel.jl
- defualt_parameters.toml
- uncalibrated_parameters.toml
- gpp_er_nee_lhf.jl
- gpp.jl
- gpp_lhf.jl
- gpp.jl
- pmodel.md

Ran validation test comparing ozark_pmodel.jl for old and new changes. Showed agreement.

Confirmed changes in priors for config files (gpp*.jl)
image
Orange -> Old prior
Blue -> New prior

Used in gpp_er_nee_lhf.jl

image

Orange -> Old prior
Blue -> New prior

Used in gpp_lhf.jl & gpp.jl

PR #1832, opened 2026-07-28
Unsupervised calibration pipeline: GPP/energy, rebuilt LAI initial conditions, prognostic LAI

AlexisRenchon:
## Pipeline status — 2026-08-03

| stage | status | outcome |
|---|---|---|
| 1 — GPP + energy | ✅ **done** | 10 parameters committed at 3 s.f. |
| 2 — rebuild LAI IC | ✅ **done** | IC rebuilt, override verified |
| 3 — prognostic LAI | ✅ **done** | recalibrated: 3 parameters, no mask |
| follow-up experiments | ✅ **4 of 4 done** | nothing config-level improved LAI RMSE |
| model simplification | ✅ **done** | C3/C4 split removed, mask dropped, recalibrated |

📊 **[Full report with the leaderboard figures →](https://claude.ai/code/artifact/03f98cf9-3976-4162-b00d-36120226e467)**

## ✅ Final calibration — 3 parameters, no mask

The C3/C4 parameter split is removed from the model and the natural-vegetation mask is removed from the objective. Three parameters were recalibrated against MODIS LAI over **all vegetated land**. EKP's `DataMisfitController` terminated on its own convergence criterion, and the parameters **settled rather than oscillating** — the first run in this work to do so.

| parameter | before | calibrated |
|---|---|---|
| `optimal_lai_z` | 15.0 | **29.2** |
| `optimal_lai_sigma` | 0.939 | **1.01** |
| `optimal_lai_alpha` | 0.0701 | **0.202** |
| `optimal_lai_z_c4`, `optimal_lai_sigma_c4` | 15.0, 0.939 | *removed from the model* |

Leaderboard metric (mean of per-month area-weighted `global_rmse`), scored over all vegetated land:

| configuration | free params | LAI RMSE | bias |
|---|---|---|---|
| uncalibrated default | — | 1.046 | +0.225 |
| calibrated **with** the mask, 5 params | 5 | 0.926 | +0.081 |
| **calibrated without the mask, 3 params** | **3** | **0.895** | **−0.103** |

**14.4% better than the default**, 3.3% better than the masked five-parameter fit, with two fewer parameters. On the leaderboard's annual-mean field the LAI row reads **RMSE 0.788, bias −0.102**.

The narrow lesson: the earlier configuration was optimised against a masked subset and then reported on a different domain. Calibrating on the quantity actually being measured recovers most of the gap.

⚠️ **Convergence caveat.** The scheduler terminated after 5 incorporated EKI updates; iterations 6–7 re-ran identical parameters (verified — same values to 16 digits), so the last ~2.5 h of compute was redundant. The parameters are genuinely converged.

⚠️ **Two changes at once.** The C4 pair and the mask were both removed before this run, so the newly-stable convergence can't be attributed to either alone. Experiment B showed that dropping to 3 parameters *by itself* did not stop the oscillation, which points at the mask as the likelier cause — but that isn't isolated here.

### ❌ Correction: "calibrate against more years" was wrong

The report originally ranked this the highest-value change. **Experiment A falsified it.** Quadrupling the target reproduced the single-year trajectory to *three significant figures* at every iteration (`z` 27.55→27.44, `sigma` 1.062→1.065) and finished at RMSE 0.887 vs 0.885. Null on parameters, loss, and fit. The ridge exists at every instant, so averaging over years doesn't break it.

### Why nothing helps: a degeneracy plus an objective that wants the bias

- **The ridge.** `z`↔`sigma` correlate at **+0.86** within the ensemble; both enter `compute_m` (Eq. 20) in the same direction, so the data constrains their product, not each. Parameters oscillate rather than converge, in every run including a single-batch one.
- **The noise scalar is a no-op.** Halving it scaled the misfit by *exactly* the variance ratio (2.00/2.01/2.01/1.99) while parameters moved <2%. `DataMisfitController` divides it back out. `lai.jl`'s docstring on this is misleading for single-target calibrations.
- **The bias is what the loss asks for.** Every run drives masked bias negative and `z` up (which lowers peak LAI). Unpenalised squared error rewards trading one large over-prediction for widespread mild under-prediction.

### ⭐ Where the error actually lives — and why no scalar can reach it

Partitioning the masked MSE exactly, per cell over the 12-month cycle, into `(μs−μo)² + (σs−σo)² + 2σsσo(1−r)`:

| term | share of masked MSE |
|---|---|
| **mean offset** | **73.7%** |
| amplitude | 14.6% |
| phase / shape | 11.8% |

and splitting the mean offset once more:

| | share of masked MSE |
|---|---|
| globally **uniform** offset (−0.114) | **1.3%** |
| **spatial pattern** of the offset | **72.4%** |

**This closes the question of why the plateau exists.** `z`, `sigma` and `alpha` are global scalars — they translate or reshape the whole LAI field at once. 72% of the error is the *spatial pattern* of the annual mean being wrong cell by cell, which no setting of three or five global numbers can remove. Hence five configurations within 0.7% of each other.

The pattern is strongly organised, not noise: **cor(offset, observed mean LAI) = −0.52**. The model compresses the spatial dynamic range — mildly too green in sparse vegetation, badly too sparse in dense forest.

| | mean offset | share of the offset MSE |
|---|---|---|
| tropics (MODIS LAI 3.21) | **−0.442** | **73.6%** |
| N temperate | +0.156 | 11.8% |
| S extratropics | +0.257 | 7.9% |
| N boreal | −0.142 | 6.7% |

By observed-LAI quintile, the top quintile (LAI 1.9–6.0) alone carries **57%** of it at offset −0.577.

**❌ Withdrawing my own recommendation: "add a bias term to the objective."** The globally uniform offset is **1.3%** of the MSE — penalising it perfectly moves pooled RMSE 0.998 → 0.991. It was a fair guess from the −0.098 masked bias, but that bias is a thin global film over a large spatial-pattern error.

**The reparameterisation onto `log(z·sigma)`/`log(z/sigma)` is still worth doing — for convergence, not skill.** It should end the period-2 limit cycle by putting the ridge on a coordinate axis. But the ridge direction *is* a global rescaling, i.e. the 1.3% term. Expect a stable inversion at the same RMSE.

**How much a spatially varying parameter could actually buy.** Every row below is an **oracle** — each group gets its own constant fitted directly to the offsets, strictly better than any calibration could find:

| grouping | groups | removes | pooled RMSE |
|---|---|---|---|
| one global `z` (status quo) | 1 | 0.0% | 0.998 |
| C3 / C4 | 2 | 1.5% | 0.992 |
| 4 latitude bands | 4 | 8.9% | 0.965 |
| 5 observed-LAI classes | 5 | 11.9% | 0.954 |
| latitude × LAI class | 20 | **24.3%** | **0.905** |

Three consequences:

1. **A per-PFT `z` is worth ~9% at best, not the 48% that "remove the whole pattern" implies.** Removing the pattern entirely (pooled RMSE → 0.52) needs a free constant *per grid cell*; a PFT parameter has ~10 groups. Zhou et al. do treat `z` as a PFT property and it's defensible physics, but it is not the answer to the RMSE question.
2. **A fourth independent verdict on the C3/C4 split** — fitting the two constants *optimally to the residuals themselves* removes 1.5% of the pattern. There is almost no between-pathway signal to find.
3. **The error is fine-scale.** Three quarters of the pattern survives a 20-group stratification, so it lives at grid scale, not biome scale. That points at the **input fields** — `A0_annual` from the P-model, precipitation, the aridity index behind `f0` — and removes the parameter-stratification escape route. **Improving LAI past ~0.885 is a stage-1 and input-field problem, not a stage-3 one.**

<sub>Level note: the partition needs the *pooled* cell-month RMSE (0.998); the leaderboard reports the *mean over months* of each month's RMSE (0.885). Mean-of-roots < root-of-mean, so they differ by construction. Read the shares here; 0.885 stays the headline.</sub>

### 🐛 A bug worth fixing regardless

Members crash with a GPU `DomainError` at parameter sets **inside** the priors — **7 of 275 member-runs (2.5%)** across the five stage-3 configurations.

❌ **Correction: I earlier reported the rate "tracks `optimal_lai_alpha`" and capped the prior at 0.18 on that basis. It does not, and the cap is reverted.** D crashed a member at α = 0.1788 — *below* the cap and the lowest α in its ensemble — which prompted a proper per-member census:

| α | crashes / runs | rate |
|---|---|---|
| 0.00–0.10 | 0 / 57 | 0.0% |
| 0.10–0.15 | 2 / 82 | 2.4% |
| 0.15–0.18 | 1 / 48 | 2.1% |
| 0.18–0.21 | 1 / 69 | 1.4% |
| 0.21–0.30 | 3 / 19 | 15.8% |

Crashes span α 0.122–0.242, successes 0.027–0.234 — near-total overlap. The one elevated bin is confounded: all three are `lai_noc4split` (3/11 there) while two other configs crashed 0/8 above 0.21. **My error was binning by each iteration's ensemble-*mean* α and attributing that iteration's crashes to it**; members within an iteration span a wide α range.

A 0.18 cap would prevent 4 of 7 crashes while excluding **84 of 268 successful member-runs**, and since every calibration settles at α 0.18–0.21 it *binds on the posterior* — an accidental modelling choice, not a safety rail. Reverted to 0.3.

**The real fix** is still to guard the faulting expression. **Do not debug from the stack trace** — it points at the ClimaDiagnostics writer, which is where the deferred kernel exception surfaces, not where it's raised. Use `CUDA_LAUNCH_BLOCKING=1`.

### Committed parameters

**Stage 1** (terminated by EKP's own scheduler at iteration 8 = one epoch): `pmodel_cstar` 0.43→**0.366** · `pmodel_β_c3` 91.1→**24.9** · `pmodel_β_c4` 5.36→**8.89** · `pmodel_α` 0.028→**0.0251** · `moisture_stress_c` 0.59→**0.591** · `leaf_Cd` 0.0726→**0.07** · `canopy_z_0m_coeff` 0.349→**0.0823** · `canopy_z_0b_coeff` 0.0444→**0.073** · `canopy_d_coeff` 0.0573→**0.0416** · `canopy_K_lw` 0.919→**1.16**

**Stage 3**: `optimal_lai_z` 15.0→**24.3** · `z_c4` 15.0→**14.6** · `sigma` 0.939→**1.05** · `sigma_c4` 0.939→**1.4** · `alpha` 0.0701→**0.13**

⚠️ Stage 3's values are one point on the ridge, not a unique optimum. ⚠️ Stage 1's `z_0b/z_0m` ≈ **0.89** against ≈0.1 conventionally, and `canopy_d_coeff` has 30% ensemble CV.

Live detail in [`STATE.md`](experiments/calibration/unsupervised_loop/STATE.md); narrative in [`LOG.md`](experiments/calibration/unsupervised_loop/LOG.md).

Sets up an unattended Claude Code loop on NCAR Derecho that runs a three-stage calibration ending in a calibrated prognostic-LAI configuration.

Stacked on #1831 — **base is `ar/climate_responsive_lai_inputs`**, since the pipeline calibrates the parameters that PR introduces.

## The pipeline

| stage | what | config | params | ensemble |
|---|---|---|---|---|
| 1 | GPP + energy fluxes, prescribed MODIS LAI | `configs/sifgpp_lhf_shf_lwu_rosetta.jl` | 10 | 21 |
| 2 | rebuild the optimal-LAI IC artifact from a 10-year run | `long_runs/snowy_land_pmodel.jl` | — | — |
| 3 | prognostic LAI vs MODIS | `configs/lai.jl` | 5 | 11 |

Strictly sequential. Stage 1 fits GPP against an *observed* canopy so prognostic-LAI error does not contaminate the P-model parameters; those parameters then set the potential GPP that drives LAI in stage 3. Each calibration commits its result into `toml/default_parameters.toml` (3 s.f., matching how existing calibrated values are stored), so later stages inherit it with no override file to keep in sync.

Stage 2 is not optional bookkeeping. With `optimal_lai_tau_long_term` at 2 years, the trailing totals need about that long to forget a bad initial condition. Starting stage 3 from the stock artifact would mean either a 3-year spinup on every ensemble member of every iteration, or an inversion that fits spin-up transients rather than the seasonal cycle.

## What this PR adds to make that possible

**`use_rosetta` / `prognostic_lai` on `CalibrateConfig`**, threaded through `setup_model`. Ported from `ar/calibrate_inversion_nee` (#1745), which is a draft, conflicting, and based on v1.10.1 — so this adapts the fields rather than waiting on that branch to land.

**Three new diagnostics — `olf0`, `olvpd`, `olgsl`.** Rebuilding the IC needs six fields: `lai_init`, `a0_annual`, `precip_annual`, `vpd_gs`, `gsl`, `f0`. Only the first three were diagnosable. The other three are exactly the quantities #1831 turned from frozen maps into climate-derived ones, so they had no diagnostics at all — and without them a spun-up run cannot be turned back into an initial condition. The round-trip is clean: `set_canopy_biomass_initial_conditions!` inverts `f0 → PET_annual`, `vpd_gs × a0_annual → VPDA0_annual`, `gsl → growing_days`, so writing the model's own equilibrated values reseeds all five accumulators consistently.

**Both calibration configs, rewritten rather than copied.** Two traps in the originals:

- `pmodel_α` changed meaning in #1817, from `1 - 1/τ` to `1/τ`. The old config's prior reads mean 0.9723 with bounds 0.85–0.999. Used unchanged it would explore τ ≈ 1 day instead of ≈ 36 days and quietly mis-calibrate GPP. The prior here is in the current convention (0.028, bounds 0.001–0.15), and the config says so in a comment so it does not get "fixed" back.
- `lai.jl` justified excluding `optimal_lai_f0` by pointing at the `optimal_lai_inputs` NamedTuple, which #1831 deleted. `f0` is still inert, but for a different reason — it is computed from the aridity index — and the config now says that.

Prior means are the current TOML defaults throughout. The LAI inversion goes from 3 parameters to 5 with the C3/C4 split on `z` and `sigma`; `alpha` is *not* split, because it enters as the `RunningMean` timescale of a `TimeIntegratedVariable`, which is a scalar for the whole field and cannot vary per column.

## Loop harness

`.claude/unsupervised-loop.settings.json` (staged, not auto-loaded) plus the guard hook, adapted from `ar/derecho_loop`. Under `experiments/calibration/unsupervised_loop/`:

- `loop_prompt.md` — the task, the stage gates, the failure signatures, the rules.
- `STATE.md` — where the pipeline is; read first and committed every iteration.
- `LOG.md` — how it got there.

The loop never blocks on a PBS job: it submits, records the job id, exits, and checks `qstat` next iteration. The prompt carries the failure signatures worth knowing — in particular that frozen parameters plus NaN loss usually means every member *crashed* rather than that the model is insensitive, and that a swallowed GPU compile error looks identical to a numerical NaN from the EKI side.

## Testing

ClimaLand loads with the new diagnostics; `test_optimal_lai.jl` passes (250 assertions); JuliaFormatter clean. The pipeline itself is untested here by construction — it runs on Derecho.

Commits use `[skip ci]`.


AlexisRenchon:
## ✅ Stage 1 complete — calibrated parameters committed; stage 2 running

**The EKI stopped itself at iteration 8.** Chunk 8's log ends with:

```
Warning: Termination condition of scheduler `DataMisfitController` has been exceeded,
returning `true` from `update_ensemble!` and preventing futher updates
```

`run_calibration.jl:140` sets `DataMisfitController(terminate_at = 100)`, and `LearningRateSchedulers.jl:284` terminates once `sum_Δt >= T` — the inversion has consumed its allotted algorithmic time. **This is a principled stopping criterion, not a failure.** Iteration 8's members ran and were processed, but the update was refused, which is why `iteration_009/eki_file.jld2` carries no new block. I cancelled chunk 9 (`6973790`) rather than let it burn 21 GPU jobs producing nothing.

Worth noting: it stopped at **exactly one epoch** (8 iterations at minibatch 2 over 16 samples) — independent corroboration of the epoch reasoning I used to raise `n_iterations` from 5 to 9.

### Calibrated parameters (committed at 3 s.f., `ec4774837`)

| parameter | old default | **calibrated** | ensemble CV |
|---|---|---|---|
| pmodel_cstar | 0.43 | **0.366** | 0.2% |
| pmodel_β_c3 | 91.1 | **24.9** | 0.8% |
| pmodel_β_c4 | 5.36 | **8.89** | 3.0% |
| pmodel_α | 0.028 | **0.0251** | 0.4% |
| moisture_stress_c | 0.59 | **0.591** | 0.3% |
| leaf_Cd | 0.0726 | **0.07** | 0.8% |
| canopy_z_0m_coeff | 0.349 | **0.0823** | 3.4% |
| canopy_z_0b_coeff | 0.0444 | **0.073** | 6.3% |
| canopy_d_coeff | 0.0573 | **0.0416** | **30%** |
| canopy_K_lw | 0.919 | **1.16** | 0.4% |

### Sanity checks — all passed

- **G ensemble finite over all 8 iterations**: `(647824, 21)`, 0 NaN columns, finite fraction 1.0. GPP is nowhere zero or NaN.
- **No parameter pinned at a bound.** Closest is `canopy_z_0b_coeff` at 73% of its 0–0.1 range.
- **Ensemble genuinely collapsed** — most CVs under 1%.
- The prompt's *"loss decreased across iterations"* check **does not apply**: `minibatch_size = 2` means each iteration scores a different pair of years, so the misfit series is not comparable across iterations. EKP's own termination is the sound signal.

**Two caveats worth carrying forward:**

⚠️ **`canopy_d_coeff` has 30% ensemble CV** — the flux targets barely constrain it. Treat 0.0416 as weakly determined, not a precise result.

⚠️ **The roughness ratio.** Final `z_0b/z_0m` = 0.073/0.0823 ≈ **0.89**, against ≈0.13 in the prior and a conventional ≈0.1 — about 7× higher, i.e. scalar and momentum roughness end up nearly equal where surface-layer theory expects an order of magnitude between them. `z_0m` fell by a factor of 4 while `z_0b` rose. Both stayed inside their bounds and the ensemble is tight, so this is what the flux data genuinely prefer under this parameterisation. Either the pair is compensating for a missing process in the canopy turbulence scheme, or the scheme is being pushed outside its valid regime — **not resolvable from EKI output, so I am reporting it rather than tuning it away.** This is the main thing I would want a second opinion on before these become defaults.

*(Correcting myself: I earlier described this ratio as "monotonically climbing" and flagged an inversion above 1.0. Both were overstated — the sequence oscillates and the >1 excursion lasted one iteration. The elevated ratio itself is real.)*

### Stage 2 started

Long run `6973822` queued: `snowy_land_pmodel.jl`, `PROGNOSTIC_LAI`, `RUN_YEARS=10`, GPU. Added `experiments/long_runs/long_run_gpu.pbs` since no launcher existed.

It asks for **12 h in `main`** rather than the faster-starting `develop` queue, because **a long run has no checkpoint** — a walltime kill loses everything, and ~4.5 h estimated leaves too little margin against `develop`'s hard 6 h cap. The calibration could safely use `develop` only because ClimaCalibrate resumes at `last_completed_iteration + 1`; that argument does not transfer.

Stage 1 ran 9 chunks over ~14 h with **one crashed member out of ~168** (and that one's data turned out valid).

AlexisRenchon:
## ✅ Stage 2 complete — IC rebuilt and override verified; stage 3 running

The 10-year prognostic-LAI run finished and produced all six IC diagnostics at 120 monthly steps. The IC is rebuilt, the artifact override is verified live, both stage-3 masks are built, and **stage 3's first iteration is running with all 11 members.**

### The rebuilt IC vs the existing one

| field | rebuilt | existing | note |
|---|---|---|---|
| lai_init | mean 0.578 | mean 0.502 | comparable |
| a0_annual | mean 276 | mean 126 | higher — calibrated GPP |
| precip_annual | mean 3.302e4 | mean 3.281e4 | **within 0.6%** |
| vpd_gs | mean 1455 | mean 630 | higher |
| gsl | mean 182, min 0.21 | mean 193, min 30 | see below |
| f0 | mean 0.457 | mean 0.428 | comparable |

**The `precip_annual` means agree to 0.6%** — an end-to-end check on the m yr⁻¹ → mol m⁻² yr⁻¹ conversion. A raw write would have been ~4.7 orders of magnitude out, giving a silently hyper-arid IC.

### Two judgement calls, both departing from the stock file

**1. NaN fill over zero-vegetation cells.** `a0a` and `olvpd` were NaN on 6921 land cells — **6843 of them Antarctica**, and every one with LAI < 0.01. Potential GPP is undefined without vegetation and `vpd_gs = VPDA0/A0` is 0/0. The runtime's `Ao_annual_safe` guards protect the *computation*, not the IC, so a NaN in the file propagates into model state. Filled `a0_annual = 0` (physically right; the stock file's minimum is also 0) and `vpd_gs`/`f0` with medians — with A0 = 0 those cells are inert, so the values cannot affect the solution and need only be finite and safe in a denominator. **All six fields are now finite over all 22420 land cells.**

**2. I did *not* floor `gsl` at 30 days**, though the stock file advertises that minimum and ours reaches 0.21. `optimal_lai.jl:224-231` puts `GSL` in the **numerator** of Eq. 20 and guards the denominator with `Ao_annual_safe`/`fAPAR_max_safe`, commenting that "m ~ 0 naturally" as A0 → 0. So the floor is a generation-time choice in the stock file, not a model requirement — and applying it would overwrite the model's own equilibrium, which is the whole point of rebuilding the IC. Flagging in case you disagree.

The previous local IC is untouched and the pre-edit `Overrides.toml` is backed up, so this is revertible by changing one line.

### Verifications worth noting

- **Override resolves**: `ClimaLand.Artifacts.optimal_lai_initial_conditions_path()` returns the rebuilt file. Members inherit it — `run_calibration.jl` sets no `JULIA_DEPOT_PATH`, so they use the same depot.
- **The validity mask is a valid reference despite differing domains.** The long run uses `global_box_domain` defaults; the calibration sets `nelements = (180, 360, 15)`. Compared the long run's `lai` NaN footprint against a stage-1 member's `gpp`: **22420 finite cells in both, identical, zero disagreement either way.**

### One bug caught in my own setup

The `N_ITERATIONS` chunking override had only been added to the **stage-1** config. My first stage-3 submission silently ignored it and would have attempted all 5 iterations in one 5h30m job — overrunning the walltime and orphaning members that then race the resubmitted orchestrator. Cancelled within a minute, added the override to `lai.jl` (default still 5), resubmitted.

### Note for stage 3's convergence check

Unlike stage 1, **stage 3's misfit series *is* a valid convergence signal.** `lai.jl` uses `minibatch_size = 1` over a single sample range, so every iteration scores the same data — the minibatching argument that invalidated the "loss decreased" check for stage 1 does not apply here.

Stage 3 scores over **natural vegetation only** (10113 of 22420 land cells). Caveats as documented: the CLM surface dataset is `simyr2000`, so post-2000 cropland expansion is not captured, and GFED4.1s ends in 2016, so the fire criterion is climatological rather than specific to the calibrated year.

AlexisRenchon:
## Follow-up experiments: three levers tested, and one recommendation withdrawn

Stage 3's parameters are committed. Since the run lowered its loss without identifying its parameters, I ran three follow-up calibrations — each changing exactly one thing against the same baseline — to find what actually moves LAI RMSE.

### Results

| configuration | RMSE masked | bias masked | verdict |
|---|---|---|---|
| default parameters | 0.941 | +0.081 | baseline |
| calibrated, 1 yr, 5 params | **0.885** | −0.098 | reference |
| **A** — 4 years instead of 1 | 0.887 | −0.088 | no effect |
| **C** — noise scalar 0.5 → 0.25 | 0.883 | −0.089 | no effect |
| **B** — drop the C3/C4 split | running | running | damps oscillation 20–35% |

**Every calibrated variant lands between 0.883 and 0.887.** Calibration is worth ~6% against the default; nothing after that moves the needle by more than 0.5%.

### ❌ I have to withdraw my top recommendation

The report ranked "calibrate against more years" highest, arguing five parameters fitted to one annual cycle were under-determined. **That was wrong.** Quadrupling the target reproduced the single-year trajectory *to three significant figures at every iteration* — `z` 27.55→27.44, `sigma` 1.062→1.065 — with a nearly identical loss series, finishing at 0.887 vs 0.885. Null on parameters, loss, and fit.

The reasoning failed because the `z`–`sigma` ridge exists at every instant in time. Averaging over more years constrains directions the loss already constrained and says nothing about the one it doesn't.

### The noise scalar is a no-op here

Halving `NOISE_SCALARS["lai"]` scaled the misfit by **exactly the variance ratio at every iteration** (2.00, 2.01, 2.01, 1.99) while parameters tracked within 1–2%. `DataMisfitController` sets its step from the misfit magnitude and divides a uniform rescaling straight back out.

This makes `lai.jl`'s docstring misleading — it advises tuning the scalar to trade seasonal amplitude against timing. True for *relative* weighting across targets (as stage 1 uses it), not for a single-target calibration under an adaptive-timestep scheduler. Worth scoping rather than deleting.

### 🐛 A bug worth fixing regardless of any of this

Members crash with a GPU `DomainError` at parameter sets **inside the priors**, and the crash rate tracks `optimal_lai_alpha` across experiments: **1/88** at α≤0.21, **3/28** at α 0.233–0.242, **0/22** at α=0.188. Every calibration drives alpha up from 0.0701, so the optimiser reliably walks into the unsafe region.

Two fixes: cap the alpha prior near **0.18** (no run settles above 0.19; would have prevented all four crashes), and guard the faulting expression. **Do not start from the stack trace** — it points at `interpolate!` in the ClimaDiagnostics writer, but the CUDA frames show a deferred kernel exception surfacing at the next synchronisation. Reproduce with `CUDA_LAUNCH_BLOCKING=1`.

### What I'd do next

The remaining idea with a mechanism behind it is **reparameterising onto `log(z·σ)` and `log(z/σ)`** — the first absorbs the ridge and should converge, the second is the weak direction and its posterior spread would then report that weakness honestly instead of hiding it in an oscillating mean. That attacks the ridge rather than working around it.

Separately, the persistent negative bias is **a property of the objective, not a tuning failure**: unpenalised squared error rewards trading one large regional over-prediction for widespread mild under-prediction, which is exactly the signature across all runs. Removing it needs a bias term, LAI-weighted cells, or amplitude and phase scored separately.

📊 [Full report, updated with all of this →](https://claude.ai/code/artifact/03f98cf9-3976-4162-b00d-36120226e467)

AlexisRenchon:
## ⭐ Why LAI RMSE plateaus at ~0.885 — the error is spatial pattern, not tuning

Experiment D (reversing the C3/C4 split) is at 2 of 8 iterations. While it runs I partitioned the masked LAI MSE exactly, and it answers the improvement question more decisively than any of the four experiments did.

Per cell, over the 12-month cycle, `MSE = (μs−μo)² + (σs−σo)² + 2σsσo(1−r)`:

| term | share of masked MSE |
|---|---|
| **mean offset** | **73.7%** |
| amplitude | 14.6% |
| phase / shape | 11.8% |
| — of which, globally **uniform** offset (−0.114) | **1.3%** |
| — of which, **spatial pattern** of the offset | **72.4%** |

`z`, `sigma` and `alpha` are global scalars — they translate or reshape the whole field at once. **72% of the error is the annual mean being wrong cell by cell, which no setting of three or five global numbers can remove.** That is the complete explanation for five configurations landing within 0.7% of each other, and it predicts D will too.

The pattern is organised, not noise: **cor(offset, observed mean LAI) = −0.52.** The model compresses LAI's spatial dynamic range — mildly too green where vegetation is sparse, badly too sparse where it is dense. The tropics carry **73.6%** of the offset error at −0.442; the densest quintile of cells alone carries **57%** at −0.577.

Two structural checks confirm it:

- **`z` shifts the ceiling by a constant everywhere.** Where energy-limited, `LAI_max = (1/k)·[ln(k·A0) − ln(z)]` — additive in `ln z`, i.e. −2.0 m² m⁻² per e-fold at *every* point. The ceiling's spatial pattern is inherited entirely from `A0_annual`.
- **The ceiling is mostly slack.** Against the long run's own `a0a`, it sits at 4.0–4.7 versus observed LAI 0.6–0.8 outside the tropics and binds in <1% of those cells — but in **34%** of tropical cells.

### What this changes

❌ **I'm withdrawing my own recommendation to add a bias term to the objective.** The globally uniform offset is **1.3%** of the MSE; penalising it perfectly moves pooled RMSE 0.998 → 0.991. It was a fair read of the −0.098 masked bias, but that bias is a thin global film over a large pattern error.

⚠️ **The `log(z·sigma)`/`log(z/sigma)` reparameterisation is still worth doing — for convergence, not skill.** It should end the period-2 limit cycle by putting the +0.86 ridge on a coordinate axis, but the ridge direction *is* a global rescaling, i.e. the 1.3% term.

✅ **Lifting the peak-LAI ceiling is the one targeted lever left.** It acts only where the cap binds — the tropical third holding 74% of the offset error. Not free: the cap enters Eq. 20 even when slack.

⭐ **The real prize is elsewhere.** Removing the spatial-pattern offset would take pooled RMSE 0.998 → **0.52**. That needs the parameters setting mean LAI to vary in space (Zhou et al. treat `z` as a plant-functional-type property; we carry one C3 and one C4 value globally), or the fields that already vary — `A0_annual` from the P-model, the aridity index behind `f0` — to be more accurate. **Improving LAI past ~0.885 is a stage-1 and input-field problem, not a stage-3 one.** That's a model change, so I'm reporting it rather than making it.

<sub>Level note: the partition needs the pooled cell-month RMSE (0.998); the leaderboard reports the mean over months of each month's RMSE (0.885). Mean-of-roots < root-of-mean, so they differ by construction — read the shares, not the level. Full report: https://claude.ai/code/artifact/03f98cf9-3976-4162-b00d-36120226e467 · commits `c9949b5f2`, `d80c60498`, `7582b1f1a`</sub>


---

**Correction (same day).** I wrote above that removing the spatial-pattern offset takes pooled RMSE 0.998 → 0.52 and pointed at a per-PFT `z`. The arithmetic holds but the attribution didn't: 0.52 needs a free constant **per grid cell**, and a PFT parameter has ~10 groups. Bounding each grouping as an *oracle* (each group's constant fitted directly to the offsets — better than any calibration could find):

| grouping | groups | removes | pooled RMSE |
|---|---|---|---|
| one global `z` (status quo) | 1 | 0.0% | 0.998 |
| C3 / C4 | 2 | 1.5% | 0.992 |
| 4 latitude bands | 4 | 8.9% | 0.965 |
| 5 observed-LAI classes | 5 | 11.9% | 0.954 |
| latitude × LAI class | 20 | 24.3% | **0.905** |

So a per-PFT `z` is worth **~9% at best**, not 48%. Defensible physics, not the answer to the RMSE question.

The residual is the real message: **three quarters of the pattern survives a 20-group stratification**, so the error lives at grid scale, not biome scale. That closes off parameter stratification as an escape route and points at the input fields — `A0_annual`, precipitation, the aridity index behind `f0`. It also gives a fourth independent verdict on the C3/C4 split: fitting its two constants optimally to the residuals removes 1.5% of the pattern.

<sub>`lai_stratification.jl` · commit `bb548d74b`</sub>

AlexisRenchon:
## 🐛 Two fixes: the GPU crash is non-deterministic, and one bad member was killing whole iterations

Experiment D's chunk for iteration 5 died outright — `update_ensemble!(ekp, nothing)`, a `MethodError`, after `observation_map.jl` logged "G ensemble matrix is not completed". No iteration produced. Diagnosing it turned up two things worth acting on.

### 1. One dead member was costing the entire iteration — `325aab037`

All seven members reported `checkpoint.txt = completed`, but two had written **31 of 36** monthly time steps: they hit the GPU `DomainError` late and left a *valid but truncated* NetCDF. `process_member_data!` doesn't throw on short-but-valid data, so the existing `try/catch` never NaN-filled their columns, the builder stayed incomplete, and `observation_map` fell out of its `if` returning `nothing`.

That defeats `SampleSuccGauss`, which exists precisely to absorb dead members. The fix NaN-fills any column with missing short names before the completeness check, and raises a clear error instead of returning `nothing` if the matrix is still incomplete.

**Verified end-to-end:** resubmitting took 2 minutes (member output was reused), iteration 5 now shows **2 NaN columns**, absorbed cleanly, and D advanced rather than losing the chunk.

### 2. ⭐ The crash is not parameter-determined — `73795873e`

| member | z | sigma | alpha | outcome |
|---|---|---|---|---|
| 001 | 23.164587167583516 | 1.0357 | 0.1367 | **crashed** |
| 002 | 23.425649461555658 | 1.0424 | 0.1395 | **crashed** |
| 003 | 23.164587167583516 | 1.0406 | 0.1376 | completed |

members 001 and 003 carry the **identical `z`** — the same Float64 to all 17 digits — with sigma and alpha differing in the third decimal. One crashed; the other finished. No parameter-space explanation survives that.

Together with the earlier census (7 crashes in 275 member-runs; crashes from α 0.122 to 0.242, successes to 0.234), the fault reads as **non-deterministic** — a race, an uninitialised read, or a value that only occasionally goes negative — not a region the optimiser walks into.

Three consequences:
- It independently confirms **reverting the α cap was right**: there was never a safe band to stay inside.
- It **raises the priority of the guard**, since no prior choice can avoid a ~2.5% per-member failure rate.
- **Resubmitting a failed iteration is legitimate recovery**, not a gamble on different numbers.

The error class is already narrowed: a negative-real `sqrt`/`log`/`^` inside a kernel (the device message is CUDA.jl's override of `Base.Math.throw_complex_domainerror`). The optimal-LAI candidates are ruled out — `lambertw0`'s branch-point `sqrt` is exactly 0 at its `MINARG` guard in both precisions, and `compute_LAI_max`'s `log` is clamped — so it's probably not in `optimal_lai.jl` at all. Reproduce with `CUDA_LAUNCH_BLOCKING=1` **and** `-g2`; only the latter yields a device stacktrace.

### Housekeeping — `8d19119d4`

The orchestrator writes `clima_calibration.o<jobid>` into `$PBS_O_WORKDIR`, which is the repo root, and 41 of them had been committed by my earlier `git add -A` calls. Now gitignored and untracked.

<sub>D is at 5 of 8 iterations; misfit 0.384/0.402/0.361/0.334/0.346, z settling near 23.1, oscillation damping.</sub>

AlexisRenchon:
## ✅ Follow-up complete — all four experiments finished

**Experiment D** ran all 8 iterations and landed at masked **RMSE 0.8888, bias −0.0886** (unmasked 0.9457 / +0.0889), measured at `iteration_008/member_001`.

| configuration | free params | RMSE masked | bias masked |
|---|---|---|---|
| default | — | 0.941 | +0.081 |
| single-year | 5 | 0.885 | −0.098 |
| A — 4 years | 5 | 0.887 | −0.088 |
| B — 3 params, C4 frozen | 3 | 0.889 | −0.083 |
| C — noise 0.25 | 5 | 0.883 | −0.089 |
| **D — 3 params, C4 tied to calibrated C3** | **3** | **0.889** | **−0.089** |

### D confirms the C3/C4 split should be dropped

D lands on B **to three decimal places** (0.8888 vs 0.8886). The distinction D existed to test — C4 *tied to the calibrated C3* rather than *frozen at inherited values* — makes no measurable difference. That's exactly what the pathway budget predicted: C4-dominated cells carry 4.3% of the squared error, so whatever the C4 parameters do is invisible in the total. The prediction recorded *before* D ran ("D will not improve RMSE; it should confirm the split can be dropped entirely") holds.

Four independent lines now agree the split earns nothing:
1. **B** — freezing `z_c4`/`sigma_c4` costs nothing
2. **D** — tying them to calibrated C3 costs nothing
3. **Pathway budget** — C4 cells are 20.8% of the domain but 4.3% of the error
4. **Oracle stratification** — a C3/C4 split fitted directly to the residuals removes 1.5% of the spatial pattern

### The overall answer on LAI skill

All five calibrated variants sit in **0.883–0.889**, a 0.7% spread. Calibration is worth ~6% against the 0.941 default; **nothing config-level moves it further**, and the reason is now established rather than guessed: **72% of the masked MSE is the spatial pattern of the annual-mean offset**, and `z`/`sigma`/`alpha` are global scalars that cannot reach it. Only 1.3% is a globally-uniform offset, 14.6% amplitude, 11.8% phase.

**Improving LAI past ~0.885 is a stage-1 and input-field problem** — `A0_annual` from the P-model, precipitation, the aridity index behind `f0` — not a stage-3 one. Even an oracle 20-group stratification reaches only 0.905, and three quarters of the pattern survives it, so the error lives at grid scale.

### What changed in the repo along the way

- `325aab037` — `observation_map.jl` NaN-fills incomplete member columns, so one truncated member no longer costs an entire iteration (this bug ate a chunk; verified fixed in production)
- `a1a2e8a9d` — reverted the α prior cap after a 275-run census showed α doesn't separate crashes
- `73795873e` — recorded that the GPU `DomainError` is **non-deterministic**: two members with an identical `z` to 17 digits, one crashed, one didn't
- `8d19119d4` — gitignored 41 PBS logs that had been committed into the repo

Remaining known bug: the `DomainError` itself, narrowed to a negative-real `sqrt`/`log`/`^` in a kernel. Reproduce with `CUDA_LAUNCH_BLOCKING=1` **and** `-g2` — the host trace points at the ClimaDiagnostics writer, which is where the deferred exception surfaces, not where it's raised.

<sub>Full report: https://claude.ai/code/artifact/03f98cf9-3976-4162-b00d-36120226e467 · commit `cbb78e790`</sub>

PR #1834, opened 2026-07-31
Prognostic carbon pools (WIP — design + unsupervised loop scaffolding)

### LINKED ISSUE #1567: Inputs to soil carbon
Implement something like:

```
C_input_leaves = LAI × LMA × τ_leaf × f_C
C_input_roots  = RAI × RMA × τ_root × f_C
C_input_total  = C_input_leaves + C_input_roots

Where:
LAI = leaf area index (m² leaf / m² ground)
RAI = root area index (m² root / m² ground)
LMA = leaf mass per area (kg dry mass / m² leaf)
RMA = root mass per area (kg dry mass / m² root)
τ_leaf, τ_root = turnover rates (1/year)
f_C ≈ 0.5 = carbon fraction of dry mass

with:
Typical values (ballpark):
LMA: 0.02–0.08 kg/m² (thin vs. thick leaves)
τ_leaf: 0.5–2 /year (evergreen vs. deciduous)
τ_root: 0.5–1 /year (fine roots turn over faster)
```

AlexisRenchon:
closes #1567 

**Draft / WIP.** Design and loop scaffolding only — no model code yet.

Based on `ar/climate_responsive_lai_inputs` (not `main`) so the diff shows only the carbon work.

## What this adds

A simple prognostic live/dead carbon model for ClimaLand, in
`src/standalone/Vegetation/biomass.jl`:

- **Live C**: `C_sugar` (in = GPP; out = maintenance respiration + structure building),
  and three structure pools `C_leaf` / `C_stem` / `C_root`
  (in = `a·f_i·S` with a ≈ 0.7; out = turnover).
- **Dead C**: `Y.soilco2.SOC`, currently pinned at its SoilGrids initial condition,
  becomes prognostic — in = leaf/stem/root litter, out = the existing microbial
  decomposition. Today `MicrobeProduction` adds `Sm` to CO₂ but never debits SOC,
  so the soil carbon balance is open; this closes it.
- **Autotrophic respiration** is reworked to respire the actual pools instead of
  JULES' `Rd_ref·RAI` / `Rd_ref·μs·SAI` on constant prescribed area indices, with
  `Rg = (1−a)·S` an actual cost of actual growth rather than `Rel·max(An−Rpm, 0)`.

## Constraints

1. The carbon model does not change GPP or LAI — both are inputs. Autotrophic
   respiration is the one deliberate exception.
2. No PFTs. Allocation fractions and turnover times are global constants,
   optionally split C3/C4; climate proxies (mean annual temperature, mean annual
   precipitation) only if constants demonstrably fail.
3. Phase 1 is one-way coupled: SAI, RAI and canopy height stay prescribed. This
   makes constraint 1 verifiable, and makes centuries-long spinup cheap — the
   pools can be integrated offline against a saved driver record.

## Files

- `experiments/integrated/prognostic_carbon/MODEL.md` — the specification.
- `experiments/integrated/prognostic_carbon/unsupervised_loop/` — loop prompt,
  state, log, README.
- `.claude/unsupervised-loop.settings.json`, `.claude/hooks/guard.sh` — staged
  sandbox config for the unattended loop (not auto-loaded).

## Plan

| stage | what | where |
|---|---|---|
| 0 | port the parametrized single-pixel ERA5 driver + 20-site battery from `ar/derecho_loop` | CPU |
| 1 | the four carbon pools | CPU battery |
| 2 | pool-based autotrophic respiration | CPU battery |
| 3 | prognostic SOC | CPU battery |
| 4 | offline spinup of the pools to steady state | offline |
| 5 | global run, comparison to ILAMB biomass, tuning | GPU |

Reference data is already on Derecho at
`/glade/campaign/cesm/community/lmwg/diag/ILAMB/DATA/{biomass,soilc}` — the
ClimaLand `ilamb_data` artifact has no biomass in it.


AlexisRenchon:
## 📌 STATUS — kept current by the unsupervised loop

> ### 📄 [**Model description & results →**](https://claude.ai/code/artifact/abe376d5-57dc-430d-8fcc-b0fd8237e900)
> Equations, rationale, global maps, and where the formulation reaches its limit.

**Last updated:** 2026-08-03 · iteration 180 · `optimal_lai_z = 24.3`

### Verification — all against the current code

| check | result |
|---|---|
| rule 1 (GPP/LAI unchanged) | **`rel_diff = 0.0`**, 20/20 |
| offline integrator vs coupled | 20/20 within 2.4% |
| global map vs single-column battery | 17/20 within 2× |
| tests | green, Float32 and Float64 |

### Drivers now match their targets

| | value |
|---|---|
| GPP vs **GOSIF** (calibration target) | **1.06×** |
| LAI vs MODIS / AVHRR / AVH15C1 | 0.85 / 0.91 / 0.98× |

### Biomass

Binned model ÷ observed over ~20 000 cells:

| observed | XuSaatchi | ESACCI | Saatchi2011 |
|---|---|---|---|
| **> 10** — real forest | **0.8×** | **0.8×** | 1.3× |
| 5–10 | 1.1× | 1.0× | 1.6× |
| 2–5 | 1.4× | 1.7× | 2.8× |
| 0.5–2 | 2.8× | 3.7× | 3.1× |
| **< 0.5** — no wood | **7.6×** | **15.7×** | 7.6× |

Global live pool **635 Pg C** (540 wood / 56 root / 29 leaf). Site score 10/20;
global bias improved in all six products, correlation improved or held in all six.

⚠️ **Known limitations, documented not tuned away:**
1. **Fire is absent** and the residual scales with it — 2.8× from least- to
   most-burnt bin. Diagnosed, deliberately *not* prescribed.
2. **Root carbon is 8.8%** of the live pool against a typical 20–25%.
3. Site score and global metrics disagree; the map is the better evidence but
   the disagreement is real.

### Negative results (each cost real compute)

1. No constant `f_stem` satisfies forests and grasslands simultaneously.
2. Precipitation at **site** level cannot separate them — but **globally** a
   precipitation limit on woody allocation helps a lot (this was a correction).
3. Soil properties cannot separate them.
4. **Aridity (P/PET) cannot** — Alaskan tundra scores highest, Congo lowest half.
5. **`f_stem(MAT, MAP)` cannot** — fixes boreal, but spatial *r* falls
   monotonically in all three products; it boosts wood across the whole cold band.
6. The stage-1 GPP recalibration alone did **not** fix the biomass bias; the
   leaf cost did.

### Corrections made to this PR's own record

- **Four of six biomass products are dry matter, not carbon** (`long_name`, not
  `units`). Corrected: the model is biased high against *all six*; the previously
  reported ESACCI ≈ 0 and negative GEOCARBON biases were artefacts.
- **Observational floor is 2.88**, not 4.38 — the products agree better than
  reported, so there was more room to improve than claimed.
- **A GFED input was recommended and is now withdrawn** — a prescribed
  burned-area map is as unextrapolatable as a PFT map.

### Where the excess actually is (iteration 180)

The overprediction is **not spread across grassland** — it is concentrated in a
small wet minority. XuSaatchi `< 0.5` bin, 5827 cells, observed mean 0.09:

| | dry cells (94%) | wet cells (6%) | share of bin carbon in that 6% |
|---|---|---|---|
| current `n = 2` | 0.54 | 6.79 | **45%** |
| sharpened `n = 8` | 0.28 | 8.50 | **66%** |

Sharpening the rainfall ramp works as intended on cells it can reach and is then
swamped by cells it cannot. **Five climate predictors have now been rejected**,
all failing on the same cells — Cerrado, Sahel fringe, Pampas, miombo — whose
rainfall *is* forest rainfall.

Global budget: GPP 135.4, NPP 65.5, litterfall 65.5 Pg C/yr, cVeg 635 Pg C,
τ_veg 9.7 yr. `litterfall == NPP` is an equilibrium closure check.

### ✅ All known gates green (iteration 180)

| gate | result |
|---|---|
| rule 1 — GPP/LAI unchanged | **PASS 30/30**, relative difference exactly `0.0` |
| package test suite | **PASS**, 0 failures (Optimal LAI 242/242) |
| formatter | clean |
| docs | `SoilCarbonLitterInput` documented — it would have failed `checkdocs = :exports`; new model page added; no duplicate `@docs` |
| unit tests for the new mechanism | `seasonality_limit` + `monthly_pet`, 5/5 each in Float32 and Float64 |

Diff against base: 23 files, ~1990 insertions.

### ⚠️ Prescribed LAI was broken and untested (iteration 180)

Asked which supported configurations hadn't been exercised since the seasonality
limit landed. **Prescribed LAI + carbon** was one, and it was broken two ways:

1. **The harness couldn't run it at all** — `site_driver.jl` requested `fc3` and `pra`
   unconditionally, both optimal-LAI diagnostics that don't exist under prescribed LAI,
   so `default_diagnostics` asserted before the run started. All 30 sites failed.
   Those diagnostics were added at stage 4, *after* the last batteries that exercised
   this path, so **it had been unrunnable for most of this PR** with no visible signal.
2. **The deficit seeding over-suppressed** — with no `precip_annual`, `P_annual` starts
   at zero, so `D_annual` was seeded to `12·pet`, the maximum possible deficit, switching
   the seasonality limit almost fully on everywhere including the aseasonal tropics.

Both fixed; **30/30 pass**, and the limit now behaves correctly there:

| site | D_annual | limit |
|---|---|---|
| amazon_central | 0.064 | **1.000** |
| congo_basin | 0.077 | **1.000** |
| cerrado_brazil | 0.363 | 0.841 |
| sahara | 0.711 | **0.264** |

**Two claims withdrawn.** The seeding bug was described as latent in a supported path —
that path was already broken upstream of it, so it was never reachable. And *"the model
works under prescribed LAI too"* was repeated from design intent, including into the new
docs page, without ever being verified on this branch. It is now.

The existing initial-condition testset already wrapped a `PrescribedBiomassModel` — it
was running the broken path and only asserting on `T_annual`. Extended to assert
`P_annual`, `D_annual` and the limit, so a millisecond unit test now covers what took a
30-site battery to find.

### 📋 Acceptance criteria — audited (iteration 185)

Every `DONE WHEN` in the loop prompt checked against its own wording. **Eight of nine met.**

| stage | criterion | status |
|---|---|---|
| 0 | battery green; baseline recorded | ✅ (30 sites) |
| 1 | tests; battery; carbon conservation; GPP & LAI unchanged | ✅ rule 1 30/30, diff exactly `0.0` |
| **2a** | **NPP/GPP in 0.3–0.6 at forest & grassland sites** | ❌ **9/13** — 3 boreal + central Europe at 0.62–0.65 |
| 2b | Ra neither zero nor > GPP at any site | ✅ no violations |
| 3a | soil C conservation; SOC drifts slowly | ✅ worst 4.1%/2 yr |
| 3b | Rh within ~2× of stage-0 baseline everywhere | ✅ 29/29 |
| 4a | offline reproduces coupled pools to a few percent | ✅ 19/19 where seed and run share a climate |
| 4b | global equilibrium pools finite & non-negative over land | ✅ 22 420 cells, no NaN, no negatives |
| 5 | global biomass defensible, or say which term is wrong | ✅ bias 0.97, RMSE 3.48, r 0.603 |

**Three of these had never been tested before this audit** — 2b, 3b and 4b. All three
pass, but *passing* and *never checked* are different states, and they were being
reported as the same one.

The one failure is stated rather than softened: NPP/GPP is too high in boreal forest.
Its fix adjusts autotrophic respiration — the one sanctioned exception to rule 1 — so
it is yours to decide.

### ⚠️ An unmet acceptance criterion (iteration 180)

The stage-2 `DONE WHEN` is *"the battery's annual NPP/GPP lands in 0.3–0.6 **at the
forest and grassland sites**, and Ra neither collapses to zero nor exceeds GPP"*.

I had checked only the first clause and only **globally**, where it passes at 0.479.
Per site, restricted to the 13 sites the criterion actually names:

| site | biome | NPP/GPP |
|---|---|---|
| cerrado_brazil · congo · borneo · amazon · n_australia | tropical | 0.44 – 0.47 |
| pampas · us_great_plains | temperate grassland | 0.51 – 0.54 |
| ozark_us · ne_china | temperate deciduous | 0.564 |
| **canada_boreal** | boreal forest | **0.619** |
| **central_europe** | temperate deciduous | **0.637** |
| **fennoscandia** | boreal forest | **0.647** |
| **central_siberia** | boreal forest | **0.654** |

**9/13 in range.** The failures are the three boreal forest sites plus central Europe,
all just above 0.6. The second clause — Ra bounds — **passes at every site**, and had
never been tested.

**Cause:** with `Q10 = 2`, `T_ref = 298.15 K`, a cold column respires **9–14%** of its
reference maintenance cost, so nearly all GPP survives as NPP, against an observed
NPP/GPP near 0.5 including in boreal forest.

**Why the global number hid it:** the tropics dominate the global flux and sit at the low
end, so a flux-weighted mean cannot see a bias confined to low-flux cells.

**Not fixed, deliberately** — lowering `Q10`, raising `T_ref` or flooring `f_T` all adjust
**autotrophic respiration**, the one sanctioned exception to rule 1 and so the one
parameter whose adjustment moves a flux this project otherwise holds fixed.

**What would fix it, with numbers.** `Ra = Rm + Rg`, and only `Rm` carries Q10, so the
required change solves directly. To put each failing site exactly at NPP/GPP = 0.60:

| site | T_annual | Rm/Ra | Q10 needed | *or* T_ref needed |
|---|---|---|---|---|
| canada_boreal | 275.0 K | 0.31 | 1.88 | 296.1 K |
| central_europe | 284.6 K | 0.23 | 1.53 | 292.9 K |
| central_siberia | 269.3 K | 0.12 | 1.50 | 286.1 K |
| fennoscandia | 277.3 K | 0.13 | 1.42 | 287.8 K |

**Q10 ≈ 1.4 brings all four inside**, against the current 2.0 and a usual published range
of 1.5–2.5 — close enough to argue, far enough to say out loud.

The more informative number is **`Rm/Ra` = 0.12–0.31**: maintenance respiration is a
minority of Ra at these sites, so Q10 has weak leverage (fennoscandia needs `Rm` roughly
doubled to gain 0.047). The alternative lever, construction efficiency `a`, raises Ra at
*every* site including the tropics, which are currently correct at 0.44–0.47. Since only
cold sites fail, the temperature-dependent lever is the structurally right one — an
argument from the shape of the error, not from either parameter's literature range.

*Two corrections to my own reporting: this criterion was previously called met on the
strength of a global number that does not test it; and my first per-site figure (11/27)
applied it to tundra, steppe and desert sites the criterion does not name.*

### 📊 Skill by climate zone — and what the global mean was hiding (iteration 180)

Delivers a stage-5 requirement that had been missed: RMSE and bias by broad climate
zone. Zones come from the model's own drivers (semi-arid = MAP < 0.6 m yr⁻¹ at any
latitude, taken first; then tropics/temperate/boreal by latitude), so no extra
boundary condition enters. XuSaatchi, kg C m⁻²:

| zone | cells | bias (base → limit) | RMSE | r |
|---|---|---|---|---|
| tropics | 2916 | +2.98 → **+1.76** | 6.35 → 5.74 | 0.438 → **0.503** |
| temperate | 2188 | +3.32 → **+2.83** | 4.98 → 4.74 | 0.247 → 0.257 |
| boreal | 2325 | +0.06 → −0.02 | 2.37 → 2.38 | 0.256 → 0.256 |
| semi-arid | 8047 | −0.28 → **−0.43** | 1.19 → 1.25 | 0.617 → 0.616 |

**This revises three things stated earlier in this PR:**

1. **The largest bias is temperate (+2.83), not dry.** "The model builds forests where
   there are none" is true of treeless *cells* — which the binned-by-observation analysis
   isolates correctly — but it is not the same as where the error is largest, and I let
   those blur together.
2. **Semi-arid is biased low, and the limit lowers it further.** Both hold at once because
   they describe different populations: binning by observed biomass selects the treeless
   cells in that zone, where the model is far too high; the zone mean is carried by the
   vegetated ones, where it is now too low.
3. **The limit's benefit is almost entirely tropical** — bias −41% and r +0.065 there,
   against ~nothing in the other three zones. The headline mean Δr of +0.011 is a tropical
   result diluted across four zones, not a uniform improvement.

**`deficit_half` deliberately not retuned.** It was tuned on a global mean, which is
arguably the wrong objective for a regionally structured error, and the semi-arid
over-suppression is a cost that objective cannot see. But changing the objective is a
modelling decision rather than a calibration detail, and the current value has held
across three independent driver changes on the objective it was chosen for. **Flagged for
you, not acted on.**

### ✅ Merged and green (iteration 180)

Base branch merged (twice — it moved mid-verification). **PR is MERGEABLE.**

| gate | result |
|---|---|
| rule 1 — GPP/LAI unchanged | **PASS 30/30**, relative difference exactly `0.0` |
| package test suite | **PASS**, 0 failures |
| drift (where forcing agrees) | 19/19 within 10% |
| formatter, docs | clean |

Both rule-1 arms regenerated on the merged tree; the earlier PASS was against a
different GPP and would not have transferred.

`carbon_deficit_half_woody = 0.55` m remains the interior optimum — it has now held
through the GPP recalibration, the LAI recalibration, and a merge that replaced the
aridity index.

| | mean bias | mean RMSE | mean r |
|---|---|---|---|
| base | 1.39 | 3.71 | 0.592 |
| **+ seasonality limit** | **0.97** | **3.48** | **0.603** |

Budget: cVeg **570 Pg C** · cSoil 2136 (*SoilGrids IC, not equilibrated*) · GPP 135.9 ·
NPP = litterfall 65.1 · Rh 34.0 Pg C yr⁻¹.

### 🔬 Rainfall seasonality — the first predictor that works (iteration 180)

Five climate predictors failed on the same cells: treeless land with forest-sized
*annual* rainfall. The discriminator is the rain's **distribution through the year**,
not its total. Validated first as a contingency table (medians differ 3–4×, IQRs
barely touch, five products), then tested offline:

| product | base bias/r | **floor 0.6, dh 550** |
|---|---|---|
| XuSaatchi | 0.96/0.614 | 0.43/**0.645** |
| Thurner | −0.35/0.537 | −0.82/**0.552** |
| ESACCI | 1.37/0.614 | 0.86/**0.653** |
| GEOCARBON | 1.15/0.572 | 0.44/**0.631** |
| Saatchi2011 | 3.23/0.633 | 2.38/**0.677** |
| USForest | 2.40/0.575 | 2.05/**0.576** |

**Spatial correlation improves in all six products, bias in five** — the first change
here to move both across the board. Optimum is interior in both parameters. Untuned:
cVeg 635 → 560 Pg C, root share 8.8% → 10.7%, NPP unchanged.

⚠️ **Offline against GPCC; no model code yet.** Model precipitation vs GPCC is
r = 0.879 but **+15.7% wet**, so `deficit_half` must be recalibrated for ERA5, not copied.

### ✅ Seasonality limit is in the model — rule 1 clean (iteration 180)

`check_rule1.jl`, 20/20 sites: GPP and LAI relative difference **exactly 0.0**,
including every dry and seasonal site the limit acts on. Three time-integrated
variables (`P_month`, `T_month`, `D_annual`) drive `seasonality_limit(D_annual)`
on woody allocation; root takes what stem gives up.

Optimum is **interior at the already-committed `deficit_half = 0.55` m**, so the
GPCC calibration transfers unchanged.

| | r improved | \|bias\| improved | mean Δr |
|---|---|---|---|
| observed precipitation (GPCC) | 5/6 | 5/6 | **+0.0302** |
| model precipitation (ERA5) | 4/6 | 5/6 | **+0.0103** |

The model-side gain is a third of the offline one, and the cause is **not** ERA5
damping seasonality — the two deficit fields agree at r = 0.941. It is that the
treeless/forest separation collapses to a fifth over the contiguous US
(USForest 37 mm vs GPCC's 176 mm), which is exactly the product whose r falls at
every half-point. A forcing-data limitation, documented rather than tuned around.

### ✅ Rebuilt on the recalibrated LAI parameters (iteration 180)

`z = 29.2, sigma = 1.01, alpha = 0.202`. **RULE 1 PASS at 30/30 sites**, relative
difference exactly `0.0`, both arms regenerated under the new parameters and now
including ten cold non-forest sites.

The seasonality limit survives: interior optimum still at the committed
`deficit_half = 0.55` m, mean Δr **+0.0117** (was +0.0103). Global skill:

| | mean bias | mean RMSE | mean r |
|---|---|---|---|
| base | 1.37 | 3.71 | 0.592 |
| **+ seasonality limit** | **0.95** | **3.48** | **0.603** |

**First live + dead budget:** cVeg **568 Pg C** (equilibrium) · cSoil 2136 Pg C
(*SoilGrids IC + 2 yr, not equilibrated*) · GPP 135.3 · NPP = litterfall 64.8 ·
Rh **33.5** Pg C yr⁻¹. Rh below litterfall means the soil is still gaining carbon;
the two sides are on different footings and are deliberately not summed.

### Open for the maintainer

All three earlier items are now done: `soc_int` + `hr` are in the driver output and
the budget is above; the leaf cost was settled by the maintainer's recalibration; and
the site set carries ten cold non-forest additions.

Open now:
1. **LAI runs slightly low** under the new calibration — 0.81× MODIS, 0.87× AVHRR,
   0.94× AVH15C1, with GPP unchanged at 135 Pg C yr⁻¹. Worth confirming that is intended,
   since a global annual mean is not what a calibration necessarily targets.
2. **The soil equilibrates to at least ~2.7× observed soil carbon.** `microbe_source`
   is Michaelis–Menten in substrate, and at realistic SOC and moisture the model sits in
   the *intermediate* regime (`MM_sx` 0.07 dry/lean → **0.80** wet/rich). Respiration is
   therefore strictly increasing in SOC — equilibrium exists, an offline soil integrator
   is well-posed — but **concave**, so a first-order extrapolation is a rigorous lower
   bound: `SOC_eq > cSoil · litterfall/Rh = ` **4132 Pg C**, against a SoilGrids initial
   condition of 2136 and observed 1 m soil carbon of ~1500–1600.

   The live pools sit close to observations while the dead pool would run away from them.
   That sublinearity is not only a defect, though — together with the `MM_o2` oxygen
   limitation it is exactly the mechanism by which a wetland accumulates peat.

   *Caveats held: `Rh` is from a 2-year run, `cSoil` integrates to 1 m while the pool is
   deeper, and `litterfall` is the live equilibrium. The direction survives all three
   (litterfall exceeds Rh by ~2×); the number is a bound, not an estimate.* This is the
   soil model, which this PR does not touch — flagged, not changed.
3. **Disturbance remains absent** — the residual sits in wet treeless cells whose
   rainfall is forest rainfall.

AlexisRenchon:
### Iterations 59–61 — the precipitation limit is **in the model**, and it costs something

#### Ported, and rule 1 survives it

Battery `6979917`, 20/20, **`RULE1 PASS` with `rel_diff = 0.0` exactly**. Adding a new prognostic state (`P_annual`) and rewriting the stem/root allocation left GPP and LAI bit-identical — the one-way coupling holds.

`f_stem` is now scaled by `w(MAP) = xⁿ/(1+xⁿ)`, `x = MAP/map_half`, with roots taking whatever stem gives up. The carbon model declares its **own** `P_annual` rather than reading the optimal-LAI model's `precip_annual`, for the same reason `T_annual` exists: the wrapper must work under prescribed LAI too, where that state is absent. It is seeded from `precip_annual` where available — a running sum starting at zero would suppress all wood for its full 2-year memory.

#### It trades one failure for another

**10/20 → 12/20** inside the observational range. Fixed: `us_great_plains`, `iberia`, `california_vaira`, `n_australia_savanna`, `ozark_us`.

**But boreal forest is now under-predicted:**

| site | pre | post |
|---|---|---|
| `central_siberia` | 6.87 — inside | **1.74 — 2.7× below** |
| `canada_boreal` | 9.69 — 1.1× over | **3.17 — 1.3× below** |
| `fennoscandia` | 8.40 — inside | **3.84 — 1.1× below** |

Boreal forest is dry in millimetres yet carries wood, so a raw-precipitation ramp suppresses it. Global statistics still improved because that binning is by observed biomass and boreal cells sit in bins that stayed healthy — a real site-versus-global disagreement, not a contradiction.

The wet-end failures are untouched, as the disturbance diagnosis predicts: `pampas` 4.6×, `cerrado` 2.3×.

#### The obvious fix does not work — tested before implementing

The natural repair is the aridity index P/PET: cold columns are dry but not water-*limited*. Sorting the sites by a temperature-driven aridity proxy:

| site | P/PET | observed woody |
|---|---|---|
| `alaska_north_slope` | **2.89** — highest | **0.00–2.09** (tundra) |
| `borneo` | 2.72 | 13.9–43.6 |
| `central_siberia` | 2.36 | 4.70–13.9 |
| **`congo_basin`** | **1.12** — low | **16.3–42.9** — highest biomass |
| `sahel` | 0.12 | 0.00–0.13 |

**It orders wrongly at both ends.** The highest-aridity site is tundra with almost no wood; one of the highest-biomass sites sits in the lower half. Such a ramp would give tundra full woody allocation and penalise rainforest. **Rejected — not implemented.**

*(The PET proxy is crude, but both failures come from temperature dominating evaporative demand, which is equally true of real PET.)*

#### What this establishes

Three predictors have now been tested and rejected for separating boreal forest from warm grassland: **precipitation at site level, soil properties, and aridity.** The trade the MAP ramp makes is therefore **structural, not a bad parameter choice** — and lowering `map_half` would only slide along the same axis rather than escape it.

**Current state:** ramp on, 12/20 inside, global bias several-fold better, spatial correlation up, boreal under-predicted and documented. I have not tuned `map_half` to hide the regression.

Still yours: the **GFED fire/land-use input** (a genuinely new input, and the only remaining lever for the wet savanna cases), and the **ClimaLand 1.11 bump** for the downstream coupler check.

AlexisRenchon:
### Iterations 66–67 — `f_stem(MAT, MAP)` rejected; global RMSE and the observational floor

#### A temperature-dependent precipitation half-point does not work

Suggested as the fix for the boreal under-prediction: let `map_half` depend on mean annual temperature, so cold columns hold wood at lower rainfall — the same trick `τ_stem(MAT)` uses. Implemented **offline only**, disabled by default.

**At the sites it is a trade, not a gain** — the count stays flat at 10/18 because equal numbers move in and out:

| site | q=1.0 | q=2.0 | observed |
|---|---|---|---|
| `central_siberia` | 1.74 | **6.14 ✓** | 4.7–13.9 |
| `canada_boreal` | 3.17 | **8.36 ✓** | 4.0–8.64 |
| `fennoscandia` | 3.84 | **6.89 ✓** | 4.08–8.77 |
| `ozark_us` | 6.84 ✓ | 8.59 ✗ | 2.08–7.03 |
| `us_great_plains` | 4.05 ✓ | 5.78 ✗ | 0.05–5.06 |
| `borneo` | 14.05 ✓ | 11.93 ✗ | 13.95–43.61 |
| `congo_basin` | 13.04 | **7.65** | 16.31–42.93 |

**Globally it is a clear loss** — spatial correlation falls *monotonically* in all three products, and bias worsens in two of three:

| q_map | XuSaatchi *r* | ESACCI *r* | Saatchi *r* |
|---|---|---|---|
| **1.0 (shipped)** | **0.631** | **0.623** | **0.657** |
| 1.5 | 0.566 | 0.543 | 0.592 |
| 2.0 | 0.485 | 0.461 | 0.524 |

The reason is worth recording: a temperature-shifted half-point raises woody allocation across the **whole cold band**, not only in real boreal forest — tundra, steppe and cropland gain wood they should not have. The 20-site set holds three boreal *forests* and almost no cold non-forest, so the sites flattered it and the 20 000-cell map caught it. **Not ported.**

That is the fourth predictor tested and rejected for the boreal/grassland conflict, after site-level precipitation, soil properties and aridity.

#### Global RMSE — computed, not previously reported

| product | n | bias | **RMSE** | cRMSE | obs σ |
|---|---|---|---|---|---|
| XuSaatchi | 15476 | +1.73 | **4.31** | 3.95 | 3.38 |
| Thurner | 6408 | +0.79 | **2.81** | 2.70 | 2.24 |
| ESACCI | 16164 | +0.16 | **5.35** | 5.35 | 6.75 |
| GEOCARBON | 9798 | −1.71 | **6.87** | 6.66 | 8.14 |
| Saatchi2011 | 5787 | +1.53 | **5.31** | 5.09 | 5.08 |
| USForest | 1035 | +1.91 | **4.28** | 3.82 | 4.51 |

**bias −1.7 … +1.9 (median ≈ +1.2), RMSE 2.8 … 6.9 (median ≈ 4.8) kg C m⁻².**

Three caveats that matter more than the numbers:

1. **The error is pattern, not offset.** Centred RMSE ≈ RMSE everywhere. ESACCI has bias +0.16 and RMSE 5.35 — removing the bias buys nothing. *Reducing bias is not the lever.*
2. **RMSE is comparable to the observed spatial σ.** Worse than a constant field for XuSaatchi and Thurner; better for ESACCI, GEOCARBON, USForest. Consistent with *r* ≈ 0.5–0.66.
3. **There is an observational floor.** Product-vs-product RMSE on overlapping cells: **median 4.38**, range 2.2–8.3. The model's ~4.8 is only slightly above the observations' own disagreement — so pushing below ~4 would be fitting to a chosen product rather than to reality.

Data: [`harness/global_skill.tsv`](../blob/ar/prognostic_carbon/experiments/integrated/prognostic_carbon/harness/global_skill.tsv). Model, observed and difference maps are now in the [results page](https://claude.ai/code/artifact/abe376d5-57dc-430d-8fcc-b0fd8237e900).

AlexisRenchon:
### Iterations 68–76 — a unit error corrected, the base recalibration adopted, and it does **not** fix the biomass bias

#### 1. Four of six biomass products are dry matter, not carbon

Caught by comparing against an ILAMB leaderboard table: its GlobalCarbon benchmark is 364 Pg where this code read 774 Pg for GEOCARBON, and 774 × 0.5 = 387. The distinction appears only in `long_name`; the unit check tested only whether `units` contained `kg`.

| product | long_name | |
|---|---|---|
| XuSaatchi | "annual **carbon** density…" | ✅ |
| Thurner | "**Carbon** Mass in Vegetation" | ✅ |
| ESACCI / GEOCARBON / Saatchi2011 / USForest | "…**biomass**" | ❌ ×0.5 |

**Corrected, the model is biased high against all six** (+0.79 … +4.00), where it had appeared *unbiased* against ESACCI (+0.16) and *too low* against GEOCARBON (−1.71). Both were artefacts. The **observational floor** was likewise wrong: the true median obs-vs-obs RMSE is **2.88**, not 4.38 — the products agree *better* than reported, so the model has more room to improve than I claimed. Fixed with an explicit `CARBON_FRACTION` table.

#### 2. GPP is ~30% too high — the upstream cause

| vs | ratio | *r* |
|---|---|---|
| FLUXCOM | 1.40× | 0.831 |
| GBAF / FLUXNET-MTE | 1.29× | 0.825 |
| WECANN | 1.28× | 0.848 |

Pattern right, magnitude wrong — and equilibrium woody carbon scales roughly with GPP.

#### 3. The recalibrated base was adopted — and it does not fix it

Branch brought up to date with the four new base commits (C3/C4 split dropped, A0-mapped leaf cost removed, transpirable fraction parameterised, stage-1 GPP/energy parameters). **`RULE1 PASS`, `rel_diff = 0.0` at all 20 sites.**

But:

| | before | after |
|---|---|---|
| GPP vs FLUXCOM | 1.40× | **1.39×** |
| GPP vs GBAF | 1.29× | **1.28×** |
| LAI vs MODIS | 0.98× | **1.10×** |
| LAI vs AVHRR | 1.06× | **1.18×** |

**GPP essentially unchanged; LAI moved the wrong way.** The increase is concentrated in seasonally dry vegetation — `n_australia` LAI +117%, `pampas` +84%, `cerrado` +65% — which raised *their* GPP and biomass. Site score **12/20 → 11/20**; global bias improved slightly (XuSaatchi 1.73 → 1.51) while spatial correlation fell in five of six products.

Binned model/observed on the recalibrated base:

| obs bin | XuSaatchi | ESACCI | Saatchi2011 |
|---|---|---|---|
| **> 10** — real forest | **0.9×** | **0.9×** | 1.4× |
| 5–10 | 1.2× | 1.2× | 1.8× |
| 2–5 | 1.6× | 1.9× | 3.1× |
| 0.5–2 | 3.2× | 4.3× | 3.4× |
| **< 0.5** — no wood | **9.1×** | **18.8×** | 8.3× |

Model, observed and difference maps are in the [results page](https://claude.ai/code/artifact/abe376d5-57dc-430d-8fcc-b0fd8237e900).

#### This reopens `optimal_lai_z`

`z` is the leaf construction cost, so **higher z ⇒ lower LAI**, and LAI is now 10–27% too high. `z = 24.3` on `ar/calibrate_lai_pipeline` is exactly that direction; this branch took the base's 15.0 when LAI still looked unbiased — true of the *pre*-recalibration run, no longer true now.

#### Two process failures recorded

- **Near-miss:** the artifact's binned array was first written by adjusting previous values *by eye* rather than computing them, caught before publishing. The guesses were close enough to look right. Every published number must come from a command in the transcript.
- **`pass=16 fail=0` was not success.** The first carbon-off arm died mid-run; the missing sites had *no status file* rather than a failure. Absence of a FAIL is not evidence of a pass.

AlexisRenchon:
### Iterations 77–85 — the leaf cost fixes the drivers; the residual is fire, and stays diagnosed

#### `optimal_lai_z` 15 → 24.3

Confirmed by @AlexisRenchon that GPP is unbiased under **prescribed** MODIS LAI on these parameters — which isolates the error in the optimal-LAI solution rather than in photosynthesis. `z` is the leaf construction cost, so raising it lowers LAI.

| metric | z = 15 | **z = 24.3** |
|---|---|---|
| GPP vs **GOSIF** (the calibration target) | 1.18× | **1.06×** |
| LAI vs MODIS / AVHRR / AVH15C1 | 1.10 / 1.18 / 1.27× | **0.85 / 0.91 / 0.98×** |
| biomass bias, XuSaatchi | +1.51 | **+0.96** |
| biomass bias, GEOCARBON | +1.92 | **+1.15** |
| biomass *r* | — | improved or held in **all six** |
| **site score** | 11/20 | **10/20** |

**`RULE1 PASS`, `rel_diff = 0.0` at all 20 sites** (batteries `7002249` / `7002250`).

Two honesties. Earlier GPP ratios here were referenced to the **FLUXCOM family**; the target is **GOSIF-GPP v2**, which ships in the `inversion_nee` artifact. Against it the change looks better, not worse. And the **site score fell** while every global metric improved — I take the map (22 419 columns) over twenty biome-balanced sites, and the same disagreement appeared in the `q_map` test where the map proved right, but it is a real disagreement.

#### Fire: diagnosed, and an earlier recommendation withdrawn

I previously recommended admitting a GFED input. **That was wrong on this project's terms** — a prescribed burned-area map is as unextrapolatable as a PFT map, and generalising under climate change is exactly why PFTs are excluded.

Used diagnostically instead ([`harness/fire_diagnostic.tsv`](../blob/ar/prognostic_carbon/experiments/integrated/prognostic_carbon/harness/fire_diagnostic.tsv)):

| burnt % | n | model | obs | residual | MAT |
|---|---|---|---|---|---|
| < 0.05 | 8626 | 3.82 | 3.00 | **+0.82** | 6.0 °C |
| 0.05–0.2 | 2111 | 4.31 | 3.17 | +1.14 | 12.3 °C |
| 0.2–1 | 1569 | 4.00 | 2.30 | +1.70 | 16.8 °C |
| 1–3 | 616 | 4.37 | 2.19 | +2.18 | 24.3 °C |
| > 3 | 296 | 4.57 | 2.28 | **+2.29** | 25.1 °C |

Modelled biomass is **flat** across every bin while observations fall — the model is blind to fire and the residual grows **2.8×** with it.

On predictability: burned area rises steeply with MAT while MAP barely moves. That is **not** sufficient for a `fire(MAT)` term — `τ_stem` already uses MAT, and the `q_map(MAT)` experiment failed precisely by re-scaling a temperature band. Real fire needs fuel load and moisture, i.e. prognostic biomass and litter. **This PR is the prerequisite, not the fix.**

#### Global live carbon pool

| pool | Pg C |
|---|---|
| C_stem | 540 |
| C_root | 56 |
| C_leaf | 29 |
| **cVeg** | **635** |

Plausible against published ~450–650 Pg C for a *potential-vegetation* equilibrium with no land use, fire or harvest. **Caveat: root carbon is 8.8% of the live pool** against a typical 20–25% belowground share — the allocation fractions are global constants no observational product constrained.

Maps, binned table and skill numbers rebuilt at z = 24.3: [results page](https://claude.ai/code/artifact/abe376d5-57dc-430d-8fcc-b0fd8237e900).

AlexisRenchon:
### Iterations 86–90 — the excess is **6% of the cells**, and it is not a rainfall problem

Two numbers that looked like they answered the same question turned out to answer
opposite ones. Both are now produced by committed scripts rather than ad hoc.

#### The global budget says the model is fine

`global_equilibrium.jl` was already averaging the equilibrium fluxes over the last
driver cycle and discarding them. It now writes them, so the budget comes out of
the same run as the pools (which are bit-identical, and the six-product scoring
table reproduces exactly — the change is additive):

| | model | independent constraint |
|---|---|---|
| GPP | 135.4 Pg C/yr | 120–130 (GOSIF) |
| NPP | 65.5 Pg C/yr | 55–65 |
| litterfall | 65.5 Pg C/yr | = NPP at equilibrium |
| cVeg | 635 Pg C | ~450–650 |
| **τ_veg** | **9.7 yr** | ~8–12 |

`litterfall == NPP` is a closure check on the offline integrator, not a
coincidence — it is what equilibrium means, and it would break if the accumulator
or the allocation fractions were inconsistent.

Turnover is normal, production is ~5% high. **Read alone this says the model is
basically right, and that reading is wrong.**

#### Binned by *observed* biomass it says the opposite

(Binning by the model would sort every wrongly-forested cell into the forest bin
and hide exactly the error being looked for.) XuSaatchi, 15 475 cells:

| observed | cells | obs | model | ratio | τ_veg | NPP |
|---|---|---|---|---|---|---|
| > 10 | 695 | 14.24 | 13.21 | 0.9 | 11.6 | 1.13 |
| 2–5 | 4435 | 3.34 | 5.46 | 1.6 | 9.3 | 0.57 |
| **< 0.5** | 5827 | 0.09 | 0.92 | **9.9** | **5.8** | **0.13** |

**NPP in the treeless bin is 0.13 — an 8× drop from forest.** Production there is
already low and correctly so. The model is not photosynthesising too much in
grassland; it is holding what it fixes for 5.8 years instead of ~1.

#### So I sharpened the woody ramp. It failed — informatively

τ_stem is 30 yr against τ_leaf 1.5 and τ_root 2.0, so allocation to wood is
weighted ~20× and even 5% of it supplies half the standing pool. Sharpening the
exponent `n` leaves `w(map_half) = 0.5` untouched, so it cannot move forests by
construction. Every product got worse, monotonically:

| n | XuSaatchi bias | r | ESACCI bias | r |
|---|---|---|---|---|
| **2 (kept)** | **0.96** | **0.614** | **1.37** | **0.614** |
| 4 | 1.13 | 0.598 | 1.53 | 0.598 |
| 8 | 1.18 | 0.574 | 1.58 | 0.574 |

**Fifth climate predictor rejected.** But the split by whether the ramp can reach
a cell at all shows the ramp is not what failed — XuSaatchi `< 0.5` bin:

| n | bin mean | dry cells (94%) | wet cells (6%) | share of bin carbon in that 6% |
|---|---|---|---|---|
| 2 | 0.92 | 0.54 | 6.79 | 45% |
| 8 | 0.78 | **0.28** | **8.50** | **66%** |

Observed: 0.09.

Sharpening works **exactly as intended** on the cells it can reach — dry-cell
biomass halves, 0.54 → 0.28, toward 0.09. It is then swamped: the same steepening
pushes the wet minority 6.79 → 8.50, and since those carry forest-sized biomass,
**6% of the cells end up holding two-thirds of the bin's carbon.**

#### What this closes

"The model builds forests where there are none" is now localised instead of
described: **the excess sits in a small wet minority of the treeless cells, not
spread across them.** Cerrado, the Sahel fringe, the Pampas, miombo — no function
of rainfall can remove wood from them because their rainfall *is* forest rainfall.
That is why five climate predictors have now failed on exactly those cells, and it
bounds the disturbance residual: ~half the treeless bin's excess at the current
`n`, two-thirds at `n = 8`.

Also checked and ruled out: the non-structural `C_sugar` pool, which inflates cVeg
and τ_veg without ever becoming litter, is 2% of cVeg in every bin.

Nothing in the model changed this iteration — `n` stays at 2. New:
`global_budget.jl`, `bin_by_obs.jl`, `run_global_equilibrium.pbs`, and the flux
fields in the equilibrium output.

AlexisRenchon:
### Iterations 91–98 — rainfall **seasonality** works, and it is the first change to improve every product

Five climate predictors had failed on the same cells: treeless land whose *annual*
rainfall is forest-sized. MAP is an annual total, and the classical discriminator
of the forest/savanna boundary is not the total but how the rain is distributed
through the year. That had not been tried.

#### First, does the separation exist? (before touching anything)

`check_seasonality.jl` — contingency table restricted to the cells the MAP ramp
**cannot** reach (MAP ≥ 0.8 m/yr), classified by observed biomass, precipitation
from GPCC so the test is of the real world rather than of the forcing:

| product | class | cells | dry months (q25/50/75) | annual deficit mm |
|---|---|---|---|---|
| XuSaatchi | treeless | 349 | 6/**8**/11 | 316/**476**/638 |
| | forest | 1691 | 2/**4**/7 | 33/**168**/331 |
| ESACCI | treeless | 810 | 7/**8**/11 | 365/**531**/655 |
| | forest | 1425 | 2/**4**/6 | 16/**156**/321 |
| GEOCARBON | treeless | 303 | 6/**8**/9 | 444/**552**/649 |
| | forest | 1856 | 2/**4**/7 | 16/**169**/337 |

It separates in all five products that contain both classes — medians differ 3–4×
and the interquartile ranges barely touch. **Annual deficit scores as well as
maximum cumulative water deficit**, which decides the form: MCWD needs a running
maximum with hemisphere-aware phasing, annual deficit is a running sum of a
pointwise function of precipitation — the machinery `P_annual` already uses.

#### Then, does it help? (offline, still no model code)

I wrote the failure mode down *before* running: a fixed 100 mm/month reference
assumes tropical evaporative demand everywhere, so it scores a **frozen** cell as
water-stressed. It duly appeared — Thurner (the boreal product) went −0.35 → −2.02.
Temperature-ramping the reference fixed it, but by not suppressing cold cells at
all, which gave up most of the correction. The two references bracket the answer,
so `pet_floor` interpolates: 0 = pure temperature ramp, 1 = fixed reference.

Bias / spatial *r*:

| product | base | f0.6 dh550 | f0.8 dh550 | fixed (f1) |
|---|---|---|---|---|
| XuSaatchi | 0.96/0.614 | 0.43/**0.645** | 0.16/0.644 | −0.26/0.640 |
| Thurner | −0.35/0.537 | −0.82/**0.552** | −1.33/0.541 | −2.02/**0.484** |
| ESACCI | 1.37/0.614 | 0.86/**0.653** | 0.61/0.656 | 0.20/0.654 |
| GEOCARBON | 1.15/0.572 | 0.44/**0.631** | 0.07/0.644 | −0.52/0.659 |
| Saatchi2011 | 3.23/0.633 | 2.38/**0.677** | 2.24/0.690 | 1.89/0.706 |
| USForest | 2.40/0.575 | 2.05/**0.576** | 1.69/0.583 | 1.18/0.583 |

**At f0.6 dh550, spatial correlation improves in all six products and bias in five.**
Nothing in this PR has moved both across the board before. Mean Δr rises 0.3 → 0.6
→ 0.8 and falls at floor 1, so the optimum is **interior in both parameters**.

Thurner's mean bias worsens while its *r* becomes the best anywhere in the sweep
(0.552, vs 0.484 for the fixed reference). Its bins show these aren't in conflict:
every low bin improves (`<0.5` 3.4× → 2.4×) and only its dominant forest bin gets
slightly more under-predicted, which is what moves the mean.

XuSaatchi bins: 0.9 / 1.2 / **1.4** / **2.7** / **7.7** against base 0.9/1.2/1.6/3.3/9.9
— forest held, everything else improved, and the treeless bin's dry *and* wet
halves both better (0.54|6.79 → 0.39|5.67). Untuned side effects: cVeg 635 → **560 Pg C**,
root share 8.8% → 10.7%, NPP unchanged (this is allocation, not production).

#### Recommendation and what's still missing

**floor 0.6, dh 550** — not the top scorer (f0.8 dh550 has mean Δr +0.0355 vs
+0.0315) but the gap is small, floor 0.8 costs Thurner much more, and a reference
that is 80% temperature-*independent* is hard to defend: evaporative demand over
frozen ground is not 80 mm/month. Both are recorded; it's a reasonable call to take
differently.

⚠️ **This is all offline against GPCC — no model code yet, deliberately.** Transfer
risk is measured, not assumed: model precipitation vs GPCC is **r = 0.879 but +15.7%
wet**, so the pattern should survive the move to ERA5 while `deficit_half` must be
**recalibrated, not copied** — transplanting 550 would under-suppress and look like
the predictor failing.

Also: one floor sweep was wasted by a harness bug (`PET_FLAG` built from
`${PET_FLOOR:-0}` above the loop that sets it, so four runs used one value). Caught
only because duplicate runs give bit-identical numbers; had they merely been close,
"the floor barely matters" would have been a plausible and wrong conclusion. The
monitor now reports the distinct values it actually observed.

AlexisRenchon:
### Iterations 99–110 — the seasonality limit is **in the model**, rule 1 passes, and the gain is a third of the offline one

#### ✅ RULE 1 PASS — 20/20 sites, relative difference exactly `0.0`

`check_rule1.jl` over battery 7006957 against the CARBON=0 baseline: GPP and LAI
identical to the last digit at every site, including every dry and seasonal site
the limit acts on (cerrado_brazil, sahel, n_australia_savanna, sahara,
pampas_argentina, us_great_plains, iberia, mojave_sw_us). `Ra` differs — the one
sanctioned exception.

#### What went into the model

Three time-integrated variables (`P_month`, `T_month`, `D_annual`) and
`seasonality_limit(D_annual)` multiplying woody allocation, root taking whatever
stem gives up. `RunningSum` semantics were **verified, not assumed**:
`(f·τ − X)/τ_long` ⇒ steady state `f·τ`, so `P_month` is a 30-day total in m and
`D_annual` is 12.17 × mean monthly deficit — exactly what the offline calibration used.

#### Three defects, two of which produced plausible wrong answers

1. **Missing Jacobian registration** — new prognostic variables must be listed in
   `implicit_timestepping.jl`. Hard failure at all 20 sites; the harmless kind.
2. **`D_annual` initialised to zero** makes the limit exactly 1 for its whole
   2-year memory. A 2-year run would have completed cleanly and reported a model
   *without* the limit.
3. **Overcorrecting (2)** by seeding `P_month` from the *instantaneous* rate —
   which is zero at almost every instant, so every column started at maximum
   deficit. Measured: **5× suppression in Amazon, Congo and Borneo**, aseasonal
   rainforest the limit must not touch. Caught by comparing against a run with
   bit-identical GPP and LAI, which left my change as the only possible cause.

`P_month` now starts from the annual mean — assume aseasonal until the run
observes otherwise, so short runs *under*-suppress, failing visibly rather than silently.

Site runs now emit `D_annual` and `seasonality_limit` directly: `C_stem` in a
2-year run cannot distinguish "inert" from "not filled yet", and that ambiguity
is what hid both bugs.

#### The model-side gain is real but smaller — and the reason is not what I assumed

Driver run now carries `precip`, so the offline integrator computes the deficit
the way the coupled model does. Optimum is **interior at `dh` = 550 mm**, which is
the already-committed `0.55` m — the GPCC calibration transfers unchanged, contrary
to my expectation that a wetter model would need a lower half-point.

| | r improved | \|bias\| improved | mean Δr |
|---|---|---|---|
| observed precipitation (GPCC) | **6/6** | 5/6 | **+0.0315** |
| model precipitation (ERA5) | 4/6 | 5/6 | **+0.0103** |

**The obvious explanation — ERA5 damping rainfall seasonality — is wrong.** The two
deficit fields agree at **r = 0.941** with the model slightly *more* variable.

What differs is the separation the predictor actually depends on. Treeless minus
forest median deficit, wet cells only:

| product | model | GPCC | ratio |
|---|---|---|---|
| XuSaatchi | 160 | 185 | 0.86 |
| ESACCI | 209 | 251 | 0.83 |
| GEOCARBON | 402 | 388 | 1.04 |
| Saatchi2011 | 146 | 119 | 1.23 |
| **USForest** | **37** | **176** | **0.21** |

Four hold; **USForest collapses to a fifth** — and USForest is exactly the product
whose *r* falls at every half-point in the model sweep while staying flat under GPCC.
So ERA5's rainfall seasonality is serviceable almost everywhere and poor over the
contiguous US, where it barely separates the dry plains from the eastern forests.

A global correlation of 0.941 looks like agreement right up until the separation is
measured *where the predictor acts*.

#### Also fixed: the driver harness reported `GLOBAL_OK` for a crashed run

`solve!` catches an integration failure, logs it and returns, so Julia exits 0 and
the PBS script took the PASS branch for a run that produced nothing. `GLOBAL_OK` was
never evidence of success on any driver run. Now requires exit status, absence of the
crash marker, and the completion marker.

**Status:** mechanism verified in the model, rule 1 clean, parameters validated at the
committed values. The remaining shortfall is a forcing-data limitation, documented rather
than tuned around.


---

### ⚠️ Correction (iteration 112): the GPCC row is **5/6**, not 6/6

A two-month phase error — the monthly driver files carry no reference date, so
record 1 is the run's first month (March), not January — paired GPCC precipitation
with the model's temperature misaligned by two months in every `--pet` GPCC sweep.

Recomputed with the climatology rotated onto calendar months:

| | as reported above | corrected |
|---|---|---|
| r improved | ~~6/6~~ | **5/6** |
| mean Δr | ~~+0.0315~~ | **+0.0302** |

USForest flips from +0.001 to −0.003 — a margin smaller than the error that
produced it, so "improves every product" was never robust and is withdrawn.

**The finding is unchanged:** observed precipitation still gives ~3× the gain of the
model's own (+0.0302 vs +0.0103), still because the treeless/forest separation
collapses over the contiguous US. Equilibrium pools and the model-precipitation
numbers are unaffected — both pair model months with each other and sum over a whole
cycle. The original contingency table is unaffected too; it uses GPCC alone.

`global_driver.jl` now writes a `start_date.txt` sidecar and `read_monthly` rotates
by it, warning rather than silently assuming January.

AlexisRenchon:
### Iterations 120–135 — rebuilt on the recalibrated LAI parameters; seasonality result holds

Maintainer supplied a fresh optimal-LAI calibration: `z` 24.3 → **29.2**,
`sigma` 0.939 → **1.01**, `alpha` 0.0701 → **0.202** (LAI memory ~14 d → ~5 d).
Everything downstream was rebuilt against it.

#### ✅ RULE 1 PASS — 30/30 sites, relative difference exactly `0.0`

Both arms regenerated under the new parameters, so this is not inherited from the
earlier 20-site check. Now includes **ten cold non-forest sites** — tundra, boreal
peatland, cold and alpine steppe — added because with a single tundra site out of
twenty the battery could not distinguish a predictor that works from one that
merely leaves the sites alone. That is exactly how `q_map(MAT)` once looked neutral
at sites while degrading spatial correlation in all three products globally.

#### The seasonality limit survives the recalibration

| dh (mm) | r improved | \|bias\| improved | mean Δr |
|---|---|---|---|
| 450 | 4/6 | 5/6 | +0.0095 |
| **550** | **4/6** | 5/6 | **+0.0117** |
| 650 | 4/6 | 5/6 | +0.0100 |

Still an **interior** optimum, still at the already-committed
`carbon_deficit_half_woody = 0.55` m, and slightly better than before (+0.0117 vs
+0.0103). No parameter change needed.

Global skill, kg C m⁻², averaged over the six ILAMB products:

| | mean bias | mean RMSE | mean r |
|---|---|---|---|
| base | 1.37 | 3.71 | 0.592 |
| **+ seasonality limit** | **0.95** | **3.48** | **0.603** |

#### First global live + dead carbon budget

| quantity | value | footing |
|---|---|---|
| cVeg | **568 Pg C** | offline equilibrium of the pools |
| cSoil (1 m) | 2136 Pg C | *SoilGrids initial condition + 2 yr — **not** equilibrated* |
| GPP | 135.3 Pg C yr⁻¹ | GOSIF ≈ 120–130 |
| NPP = litterfall | 64.8 Pg C yr⁻¹ | equal by construction at equilibrium |
| Rh | **33.5 Pg C yr⁻¹** | below litterfall — the soil is still **gaining** carbon |

**Deliberately not summed into a headline.** The two sides are on different
footings: live carbon is an equilibrium the pools reach, soil carbon is an
observational field integrated for two years against a turnover of centuries. Rh at
half of litterfall is the visible consequence — the litter the canopy now supplies
has not yet been balanced by decomposition. Quoting a single total would imply an
equilibrium half of it has not reached.

#### Things that went wrong, and what changed as a result

- **Rh first came out at 2792 Pg C yr⁻¹**, ~50× any physical value: `hr` is reported
  in mol CO₂ m⁻² s⁻¹ like `gpp` and was read as kg C. Third unit error of this kind
  in this harness, so `global_budget.jl` now derives the conversion from the file's
  own `units` attribute rather than a hard-coded constant — removing the class, not
  the instance.
- **`canadian_arctic` (-100, 68) is not land** — 0/12 valid months in the model's mask.
  Moved to (-105, 66). Its stale baseline row was *removed*, not kept: comparing a row
  keyed to the old coordinate against a run at the new one would surface as a GPP/LAI
  difference, i.e. a false rule-1 failure from a cause unrelated to the carbon model.
- **The CSV edit was held until the battery finished** — `run_battery.pbs` reads
  `test_sites.csv` per site at runtime, so a mid-run edit would give the carbon arm a
  different coordinate than its baseline.

#### Worth flagging on the LAI recalibration itself

Under the new parameters global GPP is essentially unchanged (135.1 vs 135.4 Pg C yr⁻¹)
and LAI is slightly **lower**: 0.81× MODIS (was 0.85), 0.87× AVHRR (0.91), 0.94× AVH15C1
(0.98). Reported rather than assumed wrong — the calibration may target seasonal cycle,
spatial pattern or site-level fit, none of which a global area-weighted annual mean measures.

📄 [**Artifact updated**](https://claude.ai/code/artifact/abe376d5-57dc-430d-8fcc-b0fd8237e900) — new maps, skill table and the budget.

AlexisRenchon:
### Iterations 150–165 — merged the moved base twice; every result rebuilt and holding

The base branch (`ar/climate_responsive_lai_inputs`) had advanced and the PR was
**CONFLICTING**. It then moved again mid-verification. Both merges resolved by hand;
**PR is now MERGEABLE**.

#### ✅ All gates green on the merged tree

| gate | result |
|---|---|
| rule 1 — GPP/LAI unchanged | **PASS 30/30**, relative difference exactly `0.0` |
| package test suite | **PASS**, 0 failures |
| drift (where forcing agrees) | 19/19 within 10% |
| formatter, docs | clean |

Both rule-1 arms were regenerated on the merged tree — the earlier PASS was against
a different GPP and would not have transferred.

#### The merge changes GPP and LAI locally, and nothing globally

Two behavioural changes came in, both isolated by diffing baselines:

- **`pmodel.jl` VPD clamp** (my branch had it, upstream doesn't): GPP moves up to
  **17%** at cold humid sites — california_vaira −11.4%, alaska_interior_bog +6.1% —
  tropics bit-identical. Took upstream's, since the base defines what GPP should be
  and a carbon-pools PR must not carry a GPP-changing divergence from it.
- **FAO-56 Penman-Monteith aridity** (upstream's new commits): LAI moves up to
  **25%** at Mojave, 11 of 30 sites, all arid — the expected signature of
  `AI = PET/P → f0 → LAI_max`.

Yet the **global** carbon results barely move:

| product | pre-merge base | merged base |
|---|---|---|
| XuSaatchi | 0.87 / 0.610 | 0.90 / 0.612 |
| Saatchi2011 | 3.18 / 0.630 | 3.18 / 0.630 |

Global GPP 135.1 → 135.9 Pg C yr⁻¹. Both effects land in cells carrying almost no
biomass, so a large *relative* change doesn't move an area-weighted statistic. The
site diagnostics said **where**; only the global rerun said **whether it mattered**.

#### The seasonality calibration has now held three times

| dh (mm) | r improved | mean Δr |
|---|---|---|
| 450 | 4/6 | +0.0092 |
| **550** | **4/6** | **+0.0113** |
| 650 | 4/6 | +0.0102 |

Interior optimum, still the committed `carbon_deficit_half_woody = 0.55` m. That value
has now survived the **GPP recalibration**, the **LAI recalibration**, and a **merge that
replaced the aridity index** — three independent perturbations of the drivers it was
calibrated against.

| | mean bias | mean RMSE | mean r |
|---|---|---|---|
| base | 1.39 | 3.71 | 0.592 |
| **+ seasonality limit** | **0.97** | **3.48** | **0.603** |

Budget: cVeg **570 Pg C** · cSoil 2136 (*SoilGrids IC, not equilibrated*) · GPP 135.9 ·
NPP = litterfall 65.1 · Rh **34.0** Pg C yr⁻¹.

#### Two silent hazards the merge produced

- **Duplicate TOML keys.** `optimal_lai_tau_long_term` and `optimal_lai_online_c3c4`
  existed on this branch from the #1797 lineage and upstream's hunk re-added them
  *outside* the conflict region, so both survived and git reported the file clean. Caught
  only because a verification grep printed one value twice.
- **A stale comment in the base**: it describes `alpha = 0.202` as "~15 days of memory",
  but `tau_LAI = seconds_per_day/alpha` = **4.95 days**. The "~15" matches the previous
  `alpha = 0.0701`. Corrected here; worth fixing at source.

#### For the maintainer

The VPD clamp came from this PR's own #1797 lineage, so it may have been intended
upstream and lost. Whether VPD should be clipped away from zero is a P-model question,
and **11% at a Mediterranean site is not negligible** — even though it doesn't move the
global carbon numbers.

📄 [**Artifact updated**](https://claude.ai/code/artifact/abe376d5-57dc-430d-8fcc-b0fd8237e900)

PR #1836, opened 2026-08-03
Partitioning leaderboard, zonal-mean and interannual-spread panels

### LINKED ISSUE #1813: Inter-annual variability + decadal trend on 20 year CI
When assessing model performance, we currently focus on:
- global RMSE and bias 
- seasonality 
- spatial heterogeneity

It would be good to start looking at inter-annual variability and decadal trends. We could add plots and metrics to the weekly 20 year longrun. (timeseries, linear 1-1 regression, for example).

### LINKED ISSUE #1833: Add a new leaderboard plot: EF, ETF, TF, RP. Add latitude average side axis.
Add a new leaderplot (follow-up to #1819): 
- evaporative fraction (LHF/(LHF+SHF))
- runoff partition (SR/(SSR+SR))
- transpiration fraction (T/ET)
- ET fraction (ET/Precip)

also, add latitude mean (model and obs) for all leaderboard plots, as a side axis.

AlexisRenchon:
closes #1833 
closes #1813 

Follows up on #1819: turns the partitioning plots into a leaderboard, adds panels to the annual leaderboards, and makes site and global runs share one plotting path.

## Partitioning leaderboard

A third leaderboard alongside ERA5 and FlagshipCarbonMetrics, with one row per fraction:

| Row | Observations |
|---|---|
| `LHF/(LHF+SHF)` | ERA5 |
| `SR/(SR+SSR)` | ERA5 |
| `ET/P` | ERA5 evaporation over the simulation's own (prescribed) precipitation |
| `T/ET` | none gridded — shown against Wei et al. (2017), 0.57 ± 0.07 |

The ERA5 artifact already contains `msror`, `mssror` and `mer`, so the first three rows get real bias maps; they are registered through a separate name list so the standard ERA5 leaderboard gains no rows. Every panel aggregates numerator and denominator before dividing: the mean of a ratio is not the ratio of the means, and a per-cell `ET/P` diverges wherever precipitation approaches zero. Replaces `create_partitioning_plots` for global runs, whose hardcoded `[:, 1:180]` obs slice only lined up at the default resolution.

## New panels on the leaderboards

All three of these are on the annual leaderboards and on the partitioning one.

- `LAT`: zonal mean of the annual mean, with a band of ±1 standard deviation along each latitude band. For a fraction the mean line still aggregates before dividing, but the band has to come from the per-cell ratios — there is no other sense in which a ratio varies along a band — which is safe because the small-denominator mask has already dropped the cells where a per-cell ratio would blow up.
- `MON`: global seasonal cycle, with a band of the spread across years. Note this is the spread of a *global average*, so a thin band means the global mean is reproducible between years, not that the field is uniform.
- `IAV`: area-weighted global annual mean, SIM vs OBS, one point per year, with the 1:1 line, the least squares fit, slope, r² and σ_sim/σ_obs. Points run blue to red with the year and the first and last are labelled in place, so a drift reads off the plot without spending a strip of the panel on a colorbar. Only years covering all twelve months count — a partial year's mean is aliased by whichever part of the seasonal cycle it sampled — and the column is drawn only for runs with more than three of them. A fraction aggregates its numerator and denominator over the year before dividing, as its other panels do.

r² and the amplitude ratio answer different questions: whether the model varies in the right years, and whether it varies by the right amount. On this run `gpp` gets r² 0.61 with σ_sim/σ_obs 0.41, so it tracks the interannual signal but damps it to 40% of observed; `er` gets 0.14 and 0.17, i.e. almost no interannual variability at all.

`SR/(SR+SSR)` is worse than either: slope −0.47, r² 0.18, amplitude ratio 1.10. The model varies its surface runoff fraction by about the right amount from year to year, but *anticorrelated* with ERA5 — which compounds the mean bias noted at the bottom of this description.

## Single-column output

Following the review suggestion in #1819: site plotting now goes through `site_timeseries`, which reads either a `DictWriter` or a `NetCDFWriter` and reports the start of each reduction period in both cases. `experiments/integrated/era5/single_site.jl` now writes NetCDF.

I did not change the `default_output_writer` default for `Column`/`Point`. In-memory output is deliberate for the site and calibration workflows, and making the plotting writer-agnostic achieves the unification without it. Happy to flip the default separately.

## Bugs fixed along the way

- `trans` was declared `m s^-1`, but `compute_canopy_transpiration!` returns a mass flux. The ratios were fine — both sides were mass fluxes — but anything dispatching on the units string would be misled.
- The `DictWriter` timestamp correction used `Month(dates[2]) - Month(dates[1])`, which returns `Month(-11)` when a run's first two outputs straddle a year boundary. It worked only because `single_site.jl` happens to start in March.
- `create_partitioning_plots` saved `fig`/`fig2` outside the block that defined them, so any run without `shf`/`lhf` hit an `UndefVarError`.
- Site `ET/P` came out negative: `precip` is stored downward-positive and the site path bypassed the `abs` in `_preprocess_sim_var`.

## Performance

`ClimaAnalysis.apply_oceanmask`, and the masks `make_lonlat_mask` builds, rebuild their mask and resample it onto the target grid on *every* call — and a leaderboard masks every monthly slice of every variable several times over (`_prepare_for_bias` alone runs three times per slice, for the sim, RMSE and bias series). Since every one of these masks NaNs cells multiplicatively, which cells are masked is the same for any field on the grid and only needs resolving once. `_resolved_mask` does that against an all-zero template on the sim grid — all-zero rather than the field itself, so a variable's own missing values are not folded into the base mask — and falls back to the original function for a field on another grid.

On 19 years of monthly output:

| Leaderboard | Before | After |
|---|---|---|
| FlagshipCarbonMetrics | 443 s | 126 s |
| ILAMB | 632 s | 121 s |
| ERA5 | 1169 s | 130 s |
| Partitioning | 108 s | 108 s (already hoisted) |
| **Total** | **~40 min** | **~8 min** |

All ten output PNGs are byte-for-byte identical before and after.

## Validation

Run against 19 years of monthly output from a snowy-land long run, across all three seasonal leaderboards plus the partitioning one. That run predates `et` being added to the LandModel short diagnostics (in #1819), so `ET/P` and `T/ET` were exercised with a synthetic `et = lhf/λ`; production long runs will have a real `et`. The single-column path was validated with a 5-month column run.

One result worth a look independently of this PR: `SR/(SR+SSR)` has a global bias of +0.244 (RMSE 0.442), peaking at 0.68 in May against ERA5's 0.41, with the zonal profile at ~0.8 where ERA5 is 0.05–0.2 across 40–70°N. The model routes far too much water to surface rather than subsurface runoff in the northern extratropics, worst at snowmelt.

Partitioning_leaderboard

FlagshipCarbonMetrics_sim_annual_time_avg_over_all_years

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+
+v1.12.0
+----
+- ![][badge-✨feature] Partitioning leaderboard, zonal-mean, and interannual-spread panels PR[#1836](https://github.com/CliMA/ClimaLand.jl/pull/1836)
 - ![][badge-✨feature] Change initial condition function name from set_subseasonal to set_from_era5land PR[#1858](https://github.com/CliMA/ClimaLand.jl/pull/1858)
   
 v1.11.2
@@ -25,6 +29,18 @@ v1.11.1
   custom initial-condition function must call `set_canopy_biomass_initial_conditions!`.
   Adds the `pra` diagnostic and the `optimal_lai_tau_long_term` parameter.
   PR [#1797](https://github.com/CliMA/ClimaLand.jl/pull/1797)
+- ![][badge-✨feature] Add a leaderboard of the model's energy and water partitioning against
+  ERA5 (`Partitioning_leaderboard.png`), and zonal-mean, seasonal-cycle and
+  interannual-variability panels to the annual leaderboards.
+  PR [#1836](https://github.com/CliMA/ClimaLand.jl/pull/1836)
+- ![][badge-🚀performance] Resolve each leaderboard's land-sea mask once instead of on every
+  masked slice, taking a 19-year run of all leaderboards from ~40 min to ~8 min.
+  PR [#1836](https://github.com/CliMA/ClimaLand.jl/pull/1836)
+- ![][badge-🐛bugfix] Correct the units of the `trans` diagnostic from `m s^-1` to
+  `kg m^-2 s^-1`. PR [#1836](https://github.com/CliMA/ClimaLand.jl/pull/1836)
+- ![][badge-🔥behavioralΔ] Read single-site plotting data through `site_timeseries`, which works
+  with either output writer, and fix the `DictWriter` timestamps across a year boundary.
+  PR [#1836](https://github.com/CliMA/ClimaLand.jl/pull/1836)
 
 v1.11.0
 -----
--- a/docs/src/leaderboard/leaderboard.md
+++ b/docs/src/leaderboard/leaderboard.md
@@ -123,3 +123,71 @@ compare_vars_biases_plot_extrema = Dict(
     ...
 )
 ```
+
+Note that `get_mask_dict(::ERA5DataLoader)` gives every ERA5 variable the ocean
+mask, so step 3 is only needed for loaders whose observations have gaps over
+land.
+
+## Annual leaderboard columns
+
+`compute_seasonal_leaderboard` draws one row per variable and the columns below.
+The names are the short headings drawn on the figure:
+
+- `SIM` (simulation), the simulated annual mean,
+- `ANN` (annual bias), its bias against observations,
+- `LAT` (zonal mean), the mean along each latitude band of both, banded by the
+  standard deviation along that band,
+- `MON` (seasonal cycle), their global monthly means, banded by the spread
+  across years,
+- `IAV` (interannual variability), the area-weighted global annual mean of one
+  against the other, one point per year, with the 1:1 line, the least squares
+  fit, its slope, r² and `σ_sim / σ_obs`.
+
+Only years covering all twelve months enter `IAV`, since the annual mean of a
+partial year is aliased by whichever part of the seasonal cycle it sampled, and
+the column is drawn only when some variable has more than `_MIN_IAV_YEARS` of
+them.
+
+## Partitioning leaderboard
+
+`partitioning.jl` produces a separate leaderboard of the dimensionless fractions
+describing how the model splits energy and water: the evaporative fraction
+`lhf / (lhf + shf)`, the surface runoff fraction `sr / (sr + ssr)`, the
+evaporative index `et / precip`, and the transpiration fraction `trans / et`.
+Its layout matches the annual leaderboard, with the same `SIM`, `ANN`, `LAT`,
+`MON` and `IAV` columns.
+
+Each row is a `PartitionFraction`, so adding one means appending an entry to
+`PARTITION_FRACTIONS` rather than touching the plotting code:
+
+```julia
+PartitionFraction(
+    "ef",                # short name used in file names
+    "Evaporative Fraction",
+    "LHF/(LHF+SHF)",     # label shown on rows and axes
+    ["lhf"],             # diagnostics summed for the numerator
+    ["lhf", "shf"],      # diagnostics summed for the denominator
+    String[],            # components taken from the simulation on the obs side
+    (-0.2, 0.2),         # bias panel color scale
+    nothing,             # published constraint, when there are no gridded obs
+)
+```
+
+Two things to keep in mind:
+
+**Aggregate, then divide.** Every panel sums the numerator and the denominator
+over its domain or period and divides only at the end. The mean of a ratio is
+not the ratio of the means, and a per-cell `et / precip` diverges wherever
+precipitation approaches zero, so the order matters both physically and
+numerically. Cells whose aggregated denominator falls below
+`_MIN_DENOMINATOR_FRACTION` of its global mean are masked out for the same
+reason.
+
+**Observations are optional.** A row is compared against ERA5 only when every
+component that is not listed as prescribed is available from the loader, which
+is built from the `era5_to_clima_names` keyword argument
+(`ERA5_PARTITION_TO_CLIMA_NAMES` by default). Components that *are* listed as prescribed
+(such as precipitation, which is forcing) are taken from the simulation on both
+sides. A row with no gridded observations, like the transpiration fraction,
+shows the simulation alone next to the published global constraint given in its
+`reference` field.
--- a/experiments/integrated/era5/single_site.jl
+++ b/experiments/integrated/era5/single_site.jl
@@ -82,7 +82,11 @@ model = setup_model(FT, start_date, stop_date, Δt, domain, toml_dict);
 diagnostics = ClimaLand.default_diagnostics(
     model,
     start_date;
-    output_writer = ClimaDiagnostics.Writers.DictWriter(),
+    output_writer = ClimaDiagnostics.Writers.NetCDFWriter(
+        domain.space.subsurface,
+        outdir;
+        start_date,
+    ),
     reduction_period = :monthly,
     reduction_type = :average,
     output_vars = [
--- a/ext/land_sim_vis/leaderboard/data_sources.jl
+++ b/ext/land_sim_vis/leaderboard/data_sources.jl
@@ -74,6 +74,51 @@ function _preprocess_sim_var(var, ::Val{:hr})
     return var
 end
 
+"""
+    _water_mass_flux_to_mm_per_day(var)
+
+Convert a simulation water mass flux (`kg m^-2 s^-1`) to `mm / day`.
+"""
+_water_mass_flux_to_mm_per_day(var) =
+    ClimaAnalysis.units(var) != "kg m^-2 s^-1" ? var :
+    ClimaAnalysis.convert_units(
+        var,
+        "mm / day",
+        conversion_function = units -> units * 86400.0,
+    )
+
+"""
+    _water_volume_flux_to_mm_per_day(var)
+
+Convert a simulation water volume flux (`m^3` of water per `m^2` of ground per
+second, i.e. `m s^-1`) to `mm / day` using the density of liquid water.
+"""
+_water_volume_flux_to_mm_per_day(var) =
+    ClimaAnalysis.units(var) != "m s^-1" ? var :
+    ClimaAnalysis.convert_units(
+        var,
+        "mm / day",
+        conversion_function = units -> units * 1000.0 * 86400.0,
+    )
+
+_preprocess_sim_var(var, ::Val{:sr}) = _water_volume_flux_to_mm_per_day(var)
+_preprocess_sim_var(var, ::Val{:ssr}) = _water_volume_flux_to_mm_per_day(var)
+
+# `compute_canopy_transpiration!` returns a mass flux, so `m s^-1` in older
+# output denotes the same quantity as `kg m^-2 s^-1`.
+function _preprocess_sim_var(var, ::Val{:trans})
+    ClimaAnalysis.units(var) == "m s^-1" &&
+        (var = ClimaAnalysis.set_units(var, "kg m^-2 s^-1"))
+    return _water_mass_flux_to_mm_per_day(var)
+end
+
+# Precipitation is reported downward-positive; the partitioning ratios use its
+# magnitude.
+function _preprocess_sim_var(var, ::Val{:precip})
+    var = _water_mass_flux_to_mm_per_day(var)
+    return ClimaAnalysis.remake(var; data = abs.(var.data))
+end
+
 _preprocess_sim_var(var, ::Val{:lwu}) = var
 _preprocess_sim_var(var, ::Val{:lhf}) = var
 _preprocess_sim_var(var, ::Val{:shf}) = var
@@ -149,6 +194,23 @@ const ERA5_TO_CLIMA_NAMES = [
     "msuwlwrf" => "lwu",
     "msuwswrf" => "swu",
 ]
+
+"""
+`ERA5_TO_CLIMA_NAMES` plus the runoff and evaporation the partitioning
+leaderboard needs; the default `era5_to_clima_names` of
+`compute_partitioning_leaderboard` and `create_partitioning_plots`.
+
+These are kept out of `ERA5_TO_CLIMA_NAMES` because the standard ERA5
+leaderboard draws a row per loaded variable the simulation also has, so adding
+them there would add rows to it.
+"""
+const ERA5_PARTITION_TO_CLIMA_NAMES = [
+    ERA5_TO_CLIMA_NAMES
+    "msror" => "sr"
+    "mssror" => "ssr"
+    "mer" => "et"
+]
+
 const STANDARD_UNITS = Dict("W m**-2" => "W m^-2", "W m-2" => "W m^-2")
 
 """
@@ -210,6 +272,28 @@ preprocess(::ERA5DataLoader, var, ::Val{:lhf}) =
     _preprocess_var(var; flip_sign = true)
 preprocess(::ERA5DataLoader, var, ::Val{:lwu}) = _preprocess_var(var)
 preprocess(::ERA5DataLoader, var, ::Val{:swu}) = _preprocess_var(var)
+preprocess(::ERA5DataLoader, var, ::Val{:sr}) =
+    _preprocess_var(_era5_water_flux_to_mm_per_day(var))
+preprocess(::ERA5DataLoader, var, ::Val{:ssr}) =
+    _preprocess_var(_era5_water_flux_to_mm_per_day(var))
+# ERA5 evaporation is negative downward-positive; CliMA reports it positive.
+preprocess(::ERA5DataLoader, var, ::Val{:et}) =
+    _preprocess_var(_era5_water_flux_to_mm_per_day(var); flip_sign = true)
+
+"""
+    _era5_water_flux_to_mm_per_day(var)
+
+Convert an ERA5 water mass flux (`kg m**-2 s**-1`) to `mm / day`, matching the
+units the simulation water fluxes are preprocessed into.
+"""
+function _era5_water_flux_to_mm_per_day(var)
+    ClimaAnalysis.units(var) in ("kg m**-2 s**-1", "kg m^-2 s^-1") || return var
+    return ClimaAnalysis.convert_units(
+        var,
+        "mm / day",
+        conversion_function = units -> units * 86400.0,
+    )
+end
 
 function _preprocess_var(var; flip_sign = false)
     short_name = ClimaAnalysis.short_name(var)
@@ -380,18 +464,16 @@ containing observational data, and returns a masking function. The masking
 function is used to correctly normalize the global bias and global RMSE.
 """
 function get_mask_dict(data_loader::ERA5DataLoader)
-    # Dict for loading in masks
-    mask_dict = Dict{String, Any}()
-
+    # ERA5 is gap-free over land, so every variable uses the same ocean mask.
     make_mask_fn =
         (sim_var, obs_var) -> begin
             return ClimaAnalysis.apply_oceanmask
         end
 
-    mask_dict["shf"] = make_mask_fn
-    mask_dict["lhf"] = make_mask_fn
-    mask_dict["swu"] = make_mask_fn
-    mask_dict["lwu"] = make_mask_fn
+    mask_dict = Dict{String, Any}(
+        short_name => make_mask_fn for
+        short_name in available_vars(data_loader)
+    )
 
     @assert keys(mask_dict) == available_vars(data_loader)
     return mask_dict
--- a/ext/land_sim_vis/leaderboard/leaderboard.jl
+++ b/ext/land_sim_vis/leaderboard/leaderboard.jl
@@ -28,53 +28,314 @@ function _percentile_contour_kwargs(
 end
 
 """
-    _monthly_climatology_global_means(sim, obs,
-                                      mask_fn = ClimaAnalysis.apply_oceanmask)
+    _global_mean_series(sim, obs, mask_fn = ClimaAnalysis.apply_oceanmask)
 
-Return a pair `(sim_monthly, obs_monthly)` of 12-element vectors of global,
-lonlat-weighted monthly climatology means. Index `i` corresponds to month `i`
-(1=Jan ... 12=Dec). Months that have no valid samples are filled with `NaN`.
+Return `(dates, sim_global, obs_global)`: the date of every output time and the
+lonlat-weighted global mean of `sim` and `obs` there.
 
-`mask_fn` is the same per-variable mask used by `global_bias`/`global_rmse`
-elsewhere. After applying it, at each time step both sim and obs are restricted
-to the grid cells where *both* have valid (non-`NaN`) data, so the two lines are
-averaged over the same domain even where obs has data gaps that sim does not —
-otherwise the SIM/OBS gap would not match the global bias in the ANN column.
+Both fields are restricted to the cells where *both* are finite, so the SIM/OBS
+gap here matches the global bias in the ANN column even where obs has gaps that
+sim does not.
 """
-function _monthly_climatology_global_means(
-    sim,
-    obs,
-    mask_fn = ClimaAnalysis.apply_oceanmask,
-)
+function _global_mean_series(sim, obs, mask_fn = ClimaAnalysis.apply_oceanmask)
     times = ClimaAnalysis.times(sim)
-    isempty(times) && return (fill(NaN, 12), fill(NaN, 12))
-    start_date = Dates.DateTime(sim.attributes["start_date"])
-    months =
-        [Dates.month(start_date + Dates.Second(round(Int, t))) for t in times]
+    dates = ClimaAnalysis.dates(sim)
     sim_global = Float64[]
     obs_global = Float64[]
     for t in times
         sim_t = mask_fn(ClimaAnalysis.slice(sim, time = t))
         obs_t = mask_fn(ClimaAnalysis.slice(obs, time = t))
-        # Keep only cells where both fields have valid data, so the two
-        # weighted global means are taken over the same domain.
         nan_either = isnan.(sim_t.data) .| isnan.(obs_t.data)
         sim_t.data[nan_either] .= NaN
         obs_t.data[nan_either] .= NaN
         push!(sim_global, ClimaAnalysis.weighted_average_lonlat(sim_t).data[])
         push!(obs_global, ClimaAnalysis.weighted_average_lonlat(obs_t).data[])
     end
-    out_sim = fill(NaN, 12)
-    out_obs = fill(NaN, 12)
+    return (dates, sim_global, obs_global)
+end
+
+"""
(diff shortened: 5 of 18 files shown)


NEWS.md
```diff
@@ -2,6 +2,24 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+- ![][badge-✨feature] Add a partitioning leaderboard (`Partitioning_leaderboard.png`) covering
```
kmdeck: Would it be possible to shorten this NEWS update? Many of these details can be learned from the PR description. What is the single high level change? ("Add zonal means of energy fluxes and energy/water partitioning metrics to leaderboard plots")

AlexisRenchon: yes definitely this PR isn't ready for review!

experiments/integrated/era5/single_site.jl
```diff
@@ -82,7 +82,13 @@ model = setup_model(FT, start_date, stop_date, Δt, domain, toml_dict);
 diagnostics = ClimaLand.default_diagnostics(
     model,
     start_date;
-    output_writer = ClimaDiagnostics.Writers.DictWriter(),
+    # Written to NetCDF, like the global runs, so the site plots go through the
+    # same ClimaAnalysis path and the output outlives the session.
```
ph-kev: I don't think you need this comment.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -345,6 +345,82 @@ function time_to_date(t::ITime, start_date)
...
 end
 
+"""
+    _reduction_starts(dates)
+
+Convert timestamps recorded when a reduction period closes into the start of
+that period, the convention `ClimaAnalysis` reads out of NetCDF output. Each
+period starts where the previous one ended, and the first is extrapolated from
+the length of the second, so this stays exact for reduction periods of uneven
+length such as calendar months.
```
ph-kev: I don't get what this is saying. All ClimaAnalysis does it read the time dimension from the NetCDf files. It doesn't have any convention that it follows.

docs/src/leaderboard/leaderboard.md
```diff
@@ -123,3 +123,68 @@ compare_vars_biases_plot_extrema = Dict(
...
+
+`compute_seasonal_leaderboard` draws one row per variable and the columns
+
+- `SIM`, the simulated annual mean,
+- `ANN`, its bias against observations,
+- `LAT`, zonal means of both, banded by the standard deviation along each
+  latitude band,
+- `MON`, their global seasonal cycle, banded by the spread across years,
+- `IAV`, the area-weighted global annual mean of one against the other, one point
+  per year, with the 1:1 line, the least squares fit, its slope, r² and
```
ph-kev: Are these abbreviations? It is not clear to me what the full names are.

ext/land_sim_vis/leaderboard/data_sources.jl
```diff
@@ -149,6 +195,21 @@ const ERA5_TO_CLIMA_NAMES = [
...
+rows to the standard ERA5 leaderboard, whose variable list is the intersection
+of the simulation and loader variables.
+"""
+const ERA5_PARTITION_TO_CLIMA_NAMES = [
+    "mslhf" => "lhf",
+    "msshf" => "shf",
+    "msror" => "sr",
+    "mssror" => "ssr",
+    "mer" => "et",
+]
```
ph-kev, line 209: It would be better to just reuse `ERA5_TO_CLIMA_NAMES` and specify a default for the partition plots using a keyword argument.

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -102,24 +378,61 @@ function _nee_diverging_contour_kwargs(var, q; nlevels = 21)
...
-erodes cells adjacent to a gap into NaN, leaving a slightly smaller dataset
-there, which is preferable to biasing those edges low with zero-fill.
+cells, matching `_global_mean_series`.
+
+`ClimaAnalysis.bias` normalizes by a `ones_var` that does not inherit obs's NaN
+footprint, so wherever obs has spatial gaps (GOSIF-GPP and residual-ER over
+deserts and ice) the denominator covers all land while the numerator covers only
+the cells obs sees, attenuating or flipping the bias. `mask_fn` combines
+`base_mask` with the union of the sim/obs NaN footprints so both share a domain.
+
```
ph-kev: I don't understand what this paragraph is saying. I opened an issue for it in ClimaAnalysis which I think which this paragraph is trying to say: https://github.com/CliMA/ClimaAnalysis.jl/issues/424.

ext/land_sim_vis/leaderboard/partitioning.jl
```diff
@@ -0,0 +1,779 @@
...
+summed from the `diagnostics` of one side of the comparison.
+"""
+function _ratio_global_series(diagnostics, fraction, mask_fn)
+    num = _sum_vars([diagnostics[c] for c in fraction.numerator])
+    den = _sum_vars([diagnostics[c] for c in fraction.denominator])
+    start_date = Dates.DateTime(num.attributes["start_date"])
+    dates = [
+        start_date + Dates.Second(round(Int, t)) for
+        t in ClimaAnalysis.times(num)
+    ]
```
ph-kev: Does `ClimaAnalysis.dates(var)` not work?

ph-kev (commented):
I still need to review it more, but I gave a preliminary review.

experiments/integrated/era5/single_site.jl
AlexisRenchon: Removed.

docs/src/leaderboard/leaderboard.md
AlexisRenchon: They are the short column headings drawn on the figure. I have spelled each one out in the docs (`SIM` simulation, `ANN` annual bias, `LAT` zonal mean, `MON` seasonal cycle, `IAV` interannual variability), and added the `IAV` column to the partitioning section, which was missing it.

ext/land_sim_vis/plotting_utils.jl
AlexisRenchon: The convention is not ClimaAnalysis'. It is ClimaDiagnostics': `append_time_values!` stamps a reduction at the *start* of its period (`init_time` for the first write, the previous period'"'"'s end after that), and ClimaAnalysis just reads that back. The `DictWriter` instead records the moment the period closes, so this function exists only to line the two up.

While rewording it I also made the function take `start_date` rather than extrapolating the first period from the length of the second, which is what `init_time` gives you in the NetCDF path and is exact. The old form was off by a day or two on the first monthly point.

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: Done, `ERA5_PARTITION_TO_CLIMA_NAMES` is now `[ERA5_TO_CLIMA_NAMES; runoff; evaporation]`. The keyword argument was already there; it is the constant that was duplicating entries.

ext/land_sim_vis/leaderboard/partitioning.jl
AlexisRenchon: It does, thanks. Used it here and in the two other places that were building dates by hand (`_global_mean_series` and the SIM panel title).

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: Rewritten, and it is indeed what your issue describes. `ClimaAnalysis.bias` area-weights `sim - obs` against an all-ones field, so where `sim - obs` is NaN the numerator skips the cell (`nansum`) but the denominator still counts it, and the global bias comes out attenuated by the area ratio. That bites on obs with land gaps, GOSIF-GPP and residual-ER over deserts and ice.

I saw ClimaAnalysis.jl#425 went in this morning, thanks for the quick fix. Once we depend on a release that carries it the `extra_nan` half of `_prepare_for_bias` can go; the registry is still at 0.5.22 so I have left it in for now.

NEWS.md
AlexisRenchon: Shortened; it is now four entries, one per badge, with the two feature lines merged into one. The detail lives in the PR description.

docs/src/leaderboard/leaderboard.md
AlexisRenchon: Maybe we could remove those columns name to save space? what do you think? (and what about the title)

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -28,53 +28,316 @@ function _percentile_contour_kwargs(
...
+"""
+    _zonal_std(var, mask_fn)
+
+Return the standard deviation over longitude of the time-averaged `var` within
+each latitude band. Computed as `sqrt(mean(x²) - mean(x)²)` so that both terms
+skip masked cells the same way `_zonal_means` does.
+"""
+function _zonal_std(var, mask_fn)
+    masked = mask_fn(var)
+    squared = ClimaAnalysis.Var.remake(masked; data = masked.data .^ 2)
```
ph-kev, line 191:
```suggestion
    squared = ClimaAnalysis.remake(masked; data = masked.data .^ 2)
```

ext/land_sim_vis/leaderboard/data_sources.jl
ph-kev: This should just be a keyword argument to `compute_partitioning_leaderboard`.

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -28,53 +28,316 @@ function _percentile_contour_kwargs(
...
+
+"""
+    _zonal_std(var, mask_fn)
+
+Return the standard deviation over longitude of the time-averaged `var` within
+each latitude band. Computed as `sqrt(mean(x²) - mean(x)²)` so that both terms
+skip masked cells the same way `_zonal_means` does.
+"""
+function _zonal_std(var, mask_fn)
+    masked = mask_fn(var)
```
ph-kev, line 190: Is this different from what `ClimaAnalysis.variance_lon` compute?

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -102,24 +365,55 @@ function _nee_diverging_contour_kwargs(var, q; nlevels = 21)
     )
 end
 
+"""
+    _mask_template(var)
+
+An all-zero lon-lat field on `var`'s grid, used to resolve a mask's footprint
+without folding `var`'s own missing values into it.
```
ph-kev, line 372: I don't get what this is trying to say

ext/land_sim_vis/leaderboard/leaderboard.jl
```diff
@@ -102,24 +365,55 @@ function _nee_diverging_contour_kwargs(var, q; nlevels = 21)
...
+    return function (var)
+        size(var.data) == size(blank) || return mask_fn(var)
+        new_data = copy(var.data)
+        new_data[blank] .= NaN
+        return ClimaAnalysis.OutputVar(
+            var.attributes,
+            var.dims,
+            var.dim_attributes,
+            new_data,
+        )
```
ph-kev, line 401: You should use `ClimaAnalysis.remake`.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -345,6 +345,80 @@ function time_to_date(t::ITime, start_date)
     return isnothing(t.epoch) ? start_date + t.counter * t.period : date(t)
 end
 
+"""
+    _reduction_starts(dates, start_date)
+
+Convert timestamps recorded when a reduction period closes into the start of
+that period: each period starts where the previous one ended, and the first at
+`start_date`. This is how `ClimaDiagnostics` stamps a reduction in NetCDF.
```
ph-kev, line 353: This is confusing to read without any context of the DictWriter. It should also note that the difference between times saved for the DictWriter and NetCDFWriter and explain why we need it and when this should be used.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -345,6 +345,80 @@ function time_to_date(t::ITime, start_date)
...
+function _site_timeseries(
+    writer::ClimaDiagnostics.Writers.NetCDFWriter,
+    diagnostic,
+    start_date,
+    layer,
+)
+    var = get(
+        ClimaAnalysis.SimDir(writer.output_dir),
+        diagnostic.variable.short_name,
+    )
```
ph-kev, line 413: I don't get this design. Why are we constantly recreating a `SimDir`? Does the `SimDir` not have all the information you need? Why do we need to access the `NetCDFWriter` and `diagnostic` still?

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -659,497 +685,139 @@ end
...
-        lhf_dn = [
-            d.output_short_name for
-            d in diagnostics if d.variable.short_name == "lhf"
-        ][1]
-        shf_dn = [
-            d.output_short_name for
-            d in diagnostics if d.variable.short_name == "shf"
-        ][1]
+    data_loader =
+        ERA5DataLoader(; era5_to_clima_names = ERA5_PARTITION_TO_CLIMA_NAMES)
```
ph-kev, line 716: Why not load all the variables and just choose the relevant ones to get? `ERA5DataLoader` are cheap to make.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -659,497 +685,139 @@ end
...
-        )
-        _, ssr = ClimaLand.Diagnostics.diagnostic_as_vectors(
-            diagnostics[1].output_writer,
-            ssr_dn;
-        )
-        start_date = model_time[1].epoch
-        model_dates = time_to_date.(model_time, start_date)
-        model_dates = model_dates .- Month(1)
+        obs_components = setdiff(components, fraction.prescribed)
+        if issubset(obs_components, available_vars(data_loader))
```
ph-kev, line 748: Isn't it fine to iterate over the intersection?

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -659,497 +685,139 @@ end
...
+function _era5_at_site(data_loader, short_name, longlat, dates)
+    var = get(data_loader, short_name)
+    ClimaAnalysis.set_reference_date!(var, first(dates))
+    var = ClimaAnalysis.window(
+        var,
+        "time",
+        left = first(dates),
+        right = last(dates),
+    )
+    return ClimaAnalysis.slice(var, lat = longlat[2], lon = longlat[1]).data
```
ph-kev, line 822: This gets the nearest lonlat coordinate which I am not sure is the behavior you want.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -659,497 +685,139 @@ end
...
+    scaled = if units == "kg m^-2 s^-1"
+        values .* 86400.0
+    elseif units == "m s^-1"
+        values .* 1000.0 .* 86400.0
+    else
+        values
     end
+    # Precipitation is reported downward-positive; the ratios use its magnitude.
+    return short_name == "precip" ? abs.(scaled) : scaled
+end
```
ph-kev, line 806: This function seems error prone. How do you know if this function is already called before?

ext/land_sim_vis/leaderboard/partitioning.jl
```diff
@@ -0,0 +1,767 @@
...
+                            "A",
+                            "M",
+                            "J",
+                            "J",
+                            "A",
+                            "S",
+                            "O",
+                            "N",
+                            "D",
+                        ],
```
ph-kev, line 694:
```suggestion
   Dates.monthname.(1:12) |> (s -> first.(s, 1))
```

ext/land_sim_vis/leaderboard/partitioning.jl
```diff
@@ -0,0 +1,767 @@
...
+        return nothing
+    end
+
+    groups = ["SIM", "ANN", "LAT", "MON"]
+    any(row -> length(row.years) > _MIN_IAV_YEARS, rows) && push!(groups, "IAV")
+    fig = CairoMakie.Figure(; size = (600 * length(groups), 400 * length(rows)))
+
+    # Note: Changing the order things are added to this figure
+    # may result in an error depending on the GeoMakie version.
+    # See https://github.com/MakieOrg/GeoMakie.jl/issues/330
```
ph-kev, line 557: Is this comment actually true?

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: No difference — it is `variance_lon` now:

```julia
variance = ClimaAnalysis.variance_lon(mask_fn(var); corrected = false)
return sqrt.(vec(variance.data))
```

`corrected = false` because a latitude band is the whole population of cells at that latitude rather than a sample of it, which is what the old `mean(x²) − mean(x)²` computed. I checked the two against each other on 19-year annual means of `gpp`, `lhf` and `sr`: same NaN pattern, max |Δ| ≤ 7e-14 on values of order 3 to 43. `variance_lon` is also the better-conditioned of the two, since it does not subtract two large numbers.

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: That line is gone with it — no squared field left to remake. The remaining `ClimaAnalysis.Var.remake` calls are `ClimaAnalysis.remake` now.

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: Done, and in the three other places that were building an `OutputVar` by hand: `_prepare_for_bias` here, `_intersect_mask` and `_small_denominator_mask` in `partitioning.jl`.

ext/land_sim_vis/leaderboard/leaderboard.jl
AlexisRenchon: Rewritten:

```
Return a lon-lat field on `var`'s grid whose values are all zero.

`_resolved_mask` needs some field on the grid to find out which cells a mask
removes. Zeros are used rather than `var`'s own values so that what comes back
is the mask's footprint alone, with none of `var`'s missing cells in it.
```

ext/land_sim_vis/plotting_utils.jl
AlexisRenchon: Rewritten with the difference spelled out:

```
Move the timestamps of a reduction (a monthly average, say) from the end of each
averaging period to its start, so that output kept in memory is dated the way
output written to disk is.

The two writers disagree on when a reduction happened. A `DictWriter` stores it
under the time the period closed, while `ClimaDiagnostics` writes the start of
the period to NetCDF: `init_time` for the first output, the end of the previous
period after that. Reading the two paths through `site_timeseries` therefore
needs this on the `DictWriter` side, and only for reductions — an instantaneous
diagnostic carries no period and is already stamped at its own time.
```

ext/land_sim_vis/plotting_utils.jl
AlexisRenchon: There is now a `site_source(diagnostics)` returning what the output is read from: a `SimDir` for a `NetCDFWriter`, the `DictWriter` itself otherwise. Each plotting function builds it once before its loop and passes it to `site_timeseries` as `source`, and `_site_timeseries` dispatches on that rather than on the writer.

The NetCDF method now takes nothing but the `SimDir` and the short name, as you expected. The `DictWriter` one still needs the diagnostic: in-memory output is keyed by `output_short_name`, its units live on `diagnostic.variable`, and `reduction_time_func` is what says whether the timestamps need shifting — none of that exists anywhere else.

ext/land_sim_vis/plotting_utils.jl
AlexisRenchon: It is a keyword argument now, `era5_to_clima_names`, defaulting to `ERA5_PARTITION_TO_CLIMA_NAMES` — same as on `compute_partitioning_leaderboard`.

I stopped short of "load everything and pick", because the loader's `available_vars` is also what the standard ERA5 leaderboard iterates over to decide its rows (`intersect(available_vars(sim_dir), available_vars(data_loader))`), and each loaded name needs a `preprocess` method and a mask entry. So widening the default set is not free: it adds rows to that leaderboard.

ext/land_sim_vis/plotting_utils.jl
AlexisRenchon: A fraction needs *all* of its components — half of `T/ET` is not a ratio — so this is deliberately all-or-nothing rather than an intersection. When ERA5 is missing one, only the overlay is dropped and the model line is still drawn; that is how `T/ET` ends up model-only, since ERA5 has no transpiration. Added a comment saying so, here and on the `available` check above.

ext/land_sim_vis/plotting_utils.jl
AlexisRenchon: Fair — nothing recorded that the conversion had happened, so calling it twice would multiply by 86400 twice. Two changes:

it is now applied in exactly one place, where the components are read

```julia
model = Dict(
    c => _site_component(
        site_timeseries(diagnostics, c, start_date; source)...,
        c,
    ) for c in components
)
```

and it returns the units it produced, so a second pass over its own output does nothing:

```julia
units == "kg m^-2 s^-1" &&
    return (; dates, values = values .* 86400.0, units = "mm / day")
```

ext/land_sim_vis/plotting_utils.jl
AlexisRenchon: Nearest cell is what I want here: the site sits inside one ERA5 cell, and that cell is what the simulated column is compared against. It is also how `compare_monthly_fluxes_with_data` just above reads its observations. That was nowhere in writing, so the docstring now says it:

```
Return the ERA5 timeseries of `short_name` over the months spanned by `dates`,
taken from the grid cell nearest `longlat`: a site sits within one ERA5 cell,
and that cell is what the simulated column is being compared against.
`compare_monthly_fluxes_with_data` reads its observations the same way.
```

Happy to move both to interpolation if you would rather, but that changes the existing site comparison too, so I left it out of this PR.

ext/land_sim_vis/leaderboard/partitioning.jl
AlexisRenchon: Done.

ext/land_sim_vis/leaderboard/partitioning.jl
AlexisRenchon: Yes — it is your bug: MakieOrg/GeoMakie.jl#330 is still open, and the reproducer in it is exactly this, a `Label` added to a figure that already holds a `GeoAxis` sending GridLayoutBase into infinite recursion. What the comment did not say is *which* order, so it now does:

```julia
# The title and the column labels go in before any globe panel: adding a
# `Label` to a figure that already holds a `GeoAxis` can send GridLayoutBase
# into infinite recursion. See MakieOrg/GeoMakie.jl#330.
```

ext/land_sim_vis/leaderboard/data_sources.jl
AlexisRenchon: Done: `compute_partitioning_leaderboard(...; era5_to_clima_names = ERA5_PARTITION_TO_CLIMA_NAMES)`, and the same argument on `create_partitioning_plots`. I kept the constant as the default value rather than inlining the pairs, since both entry points want the same list.

docs/src/leaderboard/leaderboard.md
```diff
@@ -123,3 +123,71 @@ compare_vars_biases_plot_extrema = Dict(
...
+them.
+
+## Partitioning leaderboard
+
+`partitioning.jl` produces a separate leaderboard of the dimensionless fractions
+describing how the model splits energy and water: the evaporative fraction
+`lhf / (lhf + shf)`, the surface runoff fraction `sr / (sr + ssr)`, the
+evaporative index `et / precip`, and the transpiration fraction `trans / et`.
+Its layout matches the annual leaderboard, with the same `SIM`, `ANN`, `LAT`,
+`MON` and `IAV` columns.
```
ph-kev, line 158: It seems weird to me to have a description of code in `ext` unless this is how you expect the user to interact with the code.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -659,497 +718,155 @@ end
...
-        )
-        if mask isa Nothing
-            plot!(fig, ratio, more_kwargs = plot_kwargs)
-        else
-            plot!(fig, ratio, more_kwargs = plot_kwargs, mask = mask)
-        end
-        CairoMakie.save(joinpath(savedir, "water_partition.png"), fig)
-    end
+This mirrors the `_preprocess_sim_var` methods, which normalize gridded
+`OutputVar`s rather than the plain vectors a single site produces.
```
ph-kev, line 839: I don't understand what this comment is trying to say.

ph-kev (approved):

docs/src/leaderboard/leaderboard.md
kmdeck: I dont think this is unique to this PR (the leaderboard code is all in ext/, and this documentation already references it.

ext/land_sim_vis/plotting_utils.jl
```diff
@@ -345,6 +345,110 @@ function time_to_date(t::ITime, start_date)
...
 end
 
+"""
+    _reduction_starts(dates, start_date)
+
+Move the timestamps of a reduction (a monthly average, say) from the end of each
+averaging period to its start, so that output kept in memory is dated the way
+output written to disk is.
+
+The two writers disagree on when a reduction happened. A `DictWriter` stores it
```
kmdeck, line 355: @ph-kev is this true? or is it the opposite? I thought the NetCDF was now saved correctly (no shift needed) while for the DictWriter it still saved it with the timestamp at the end of the file

ext/land_sim_vis/plotting_utils.jl
kmdeck: I think it is saying that we convert the units differently for sites (DictWriter diagnostics - saved as a vector) vs something read from netcdf as an OutputVar. But I dont understand why this would be the case

ext/land_sim_vis/plotting_utils.jl
kmdeck: @ph-kev let me know. I'd like to fix this comment and then merge this PR

ph-kev: The times for the NetCDF files are at the start of the reduction period and the times for the DictWriter are at the end of the reduction period.

PR #1840, opened 2026-08-05
Add a restart test for the global land model with the P-model

### LINKED ISSUE #1829: Add a test checking restarts for land default (pmodel), prescribed and prognostic LAI
AMIP still uses PFTs with Farquhar, but we are mainly developping pmodel. The reason we currently don't use pmodel in AMIP is due to restart issues. As we are fixing this in PR #1797, we should ensure restart work with a test, for prescribed and prognostic LAI.

AlexisRenchon:
closes #1829 

We have a restart test for the bucket model, but none for the full land model. This adds one for the configuration of `experiments/long_runs/snowy_land_pmodel.jl` — soil, snow, soil CO2, inland water, and a canopy using the P-model for photosynthesis — on a global domain.

`test/integrated/restart.jl` runs the model for 18 steps at `Δt = 900 s`, checkpointing at 3 hours (a driver-update boundary, so the restarted simulation sees the same forcing), then restarts from that checkpoint and integrates to the same final time. It checks that

- the checkpoint round-trips every prognostic variable exactly, including the P-model's acclimated capacities;
- `find_restart`/`initial_time_from_checkpoint` recover the checkpoint date;
- the restarted simulation reaches the same final state as the uninterrupted one.

### Restarting a dated simulation used to error

`save_checkpoint` computed the output file name with `t / (60 * 60 * 24)`, which is undefined for an `ITime` — so any simulation set up from a `start_date` and a `stop_date` (i.e. every long run) failed at the first checkpoint. The time is now written as a number of seconds together with the start date it carries, and `read_checkpoint` rebuilds the `ITime` from the two. Checkpoints without a start date are read back as before.

### The restart is not bit-for-bit, and why

Only `Y` is checkpointed; the cache is rebuilt from it on restart. That is exact for nearly every cache variable, but not for `p.snow.T_sfc`: it is the initial guess of the snow surface temperature root find, which takes a single Newton step per timestep, so in an uninterrupted run it carries the previous step's value, while a restart reseeds it from the air temperature. Three hours into the run above, the restarted and uninterrupted states differ by a relative `1e-4` at worst (`soil.∫F_e_dt`, `canopy.energy.T`).

`p.snow.T_sfc` is the *only* such quantity: restoring it and nothing else makes the restart bit-for-bit identical. The test asserts both — the `1e-3` tolerance on a plain restart, and exact agreement once `p.snow.T_sfc` is restored — so it will catch any new lagged cache variable. `docs/src/restarts.md` documents this alongside the existing driver and diagnostic caveats.

Making restarts exact would mean checkpointing `p.snow.T_sfc` (or promoting it to prognostic state); happy to do that here or in a follow-up.

### Note on `global_box_domain`

The test uses `global_domain` (cubed sphere) rather than the `global_box_domain` of the long run: ClimaCore's HDF5 reader only understands `XPoint`/`YPoint`/`ZPoint` interval domains, so reading a checkpoint written on a lat/long box fails with `Invalid coord type LatPoint`. Restarts of the long run as currently configured would hit this — it needs a fix in ClimaCore's `InputOutput._scan_coord_type`.

### Testing

`test/integrated/restart.jl` passes (51 tests, ~2 min including compilation); `test/standalone/Bucket/restart.jl` still passes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+- ![][badge-✨feature] Add a restart test for the global land model with the P-model
+  (`test/integrated/restart.jl`): restarting from a checkpoint reproduces an uninterrupted
+  run bit-for-bit. PR [#1840](https://github.com/CliMA/ClimaLand.jl/pull/1840)
 
 v1.11.1
 -------
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,15 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+- ![][badge-🐛bugfix] Let `save_checkpoint`/`read_checkpoint` round-trip an `ITime`: the time
+  is written as a number of seconds together with the start date it carries, so a simulation
+  set up from a `start_date` and a `stop_date` can now be checkpointed and restarted (it
+  previously errored when saving). PR [#1836](https://github.com/CliMA/ClimaLand.jl/pull/1836)
+- ![][badge-✨feature] Add a restart test for the global land model with the P-model
+  (`test/integrated/restart.jl`), the configuration of `experiments/long_runs/snowy_land_pmodel.jl`.
+  It checks that the checkpoint round-trips the state exactly, and that restarting from it
+  reproduces an uninterrupted run — bit-for-bit once `p.snow.T_sfc`, the only lagged quantity
+  in the cache, is restored. PR [#1836](https://github.com/CliMA/ClimaLand.jl/pull/1836)
 - ![][badge-✨feature] Add a `TimeIntegratedVariable` utility for prognostic time-integrated
   variables — a running mean/sum or plain time-integral stored in `Y` and advanced by the
   time-stepper (no callback, checkpoint/restart-safe). PR [#1797](https://github.com/CliMA/ClimaLand.jl/pull/1797)
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,12 +5,12 @@ main
 - ![][badge-🐛bugfix] Let `save_checkpoint`/`read_checkpoint` round-trip an `ITime`: the time
   is written as a number of seconds together with the start date it carries, so a simulation
   set up from a `start_date` and a `stop_date` can now be checkpointed and restarted (it
-  previously errored when saving). PR [#1836](https://github.com/CliMA/ClimaLand.jl/pull/1836)
+  previously errored when saving). PR [#1839](https://github.com/CliMA/ClimaLand.jl/pull/1839)
 - ![][badge-✨feature] Add a restart test for the global land model with the P-model
   (`test/integrated/restart.jl`), the configuration of `experiments/long_runs/snowy_land_pmodel.jl`.
   It checks that the checkpoint round-trips the state exactly, and that restarting from it
   reproduces an uninterrupted run — bit-for-bit once `p.snow.T_sfc`, the only lagged quantity
-  in the cache, is restored. PR [#1836](https://github.com/CliMA/ClimaLand.jl/pull/1836)
+  in the cache, is restored. PR [#1839](https://github.com/CliMA/ClimaLand.jl/pull/1839)
 - ![][badge-✨feature] Add a `TimeIntegratedVariable` utility for prognostic time-integrated
   variables — a running mean/sum or plain time-integral stored in `Y` and advanced by the
   time-stepper (no callback, checkpoint/restart-safe). PR [#1797](https://github.com/CliMA/ClimaLand.jl/pull/1797)
--- a/docs/src/restarts.md
+++ b/docs/src/restarts.md
@@ -70,6 +70,24 @@ restart_file = ClimaLand.find_restart(output_dir)
 Y, t = ClimaLand.read_checkpoint(restart_file; model)
 ```
 
+The time is returned as an `ITime` if the checkpoint was saved with one carrying
+a start date, and as a number of seconds otherwise. Simulations built from a
+`start_date` and a `stop_date` use `ITime`s, so restarting one amounts to using
+the date of the checkpoint as the new start date:
+
+```julia
+restart_file = ClimaLand.find_restart(output_dir)
+t = ClimaLand.initial_time_from_checkpoint(restart_file; model)
+simulation = LandSimulation(
+    date(t),
+    stop_date,
+    Δt,
+    model;
+    set_ic! = (Y, p, t, model) ->
+        ClimaLand.set_initial_conditions_from_checkpoint!(Y, restart_file; model),
+)
+```
+
 ## Output Structure
 
 `ClimaLand` utilizes the `OutputPathGenerator` from `ClimaUtilities` to manage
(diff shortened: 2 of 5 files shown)


docs/src/restarts.md
```diff
@@ -130,3 +148,16 @@ checkpointing and updating the drivers are compatible.
...
+
+### Checkpoints and the cache
+
+Only the state `Y` is checkpointed; the cache is rebuilt from it when the
+simulation restarts. Nearly every cache variable is a function of the state and
+the time, so rebuilding it is exact, but `p.snow.T_sfc` is not: it is the initial
+guess of the snow surface temperature root find, which takes a single Newton step
+per timestep, and so carries the previous step's value in an uninterrupted run.
+
+A restarted land simulation therefore does not reproduce an uninterrupted one
```
kmdeck, line 172: Thanks, Alexis! This is a good catch. Hopefully it is OK because this is just the initial guess and then the solved root should be within a tolerance we accept of the value we would get if we did not restart.

src/shared_utilities/checkpoints.jl
```diff
@@ -164,6 +186,10 @@ function read_checkpoint(
         end
     end
     t = attributes["time"]
+    epoch = get(attributes, "epoch", "nothing")
+    if epoch != "nothing"
+        t = ITime(t; epoch = Dates.DateTime(epoch))
```
kmdeck, line 191: I think we need to double check if we want this behavior. @akshaysridhar @imreddyTeja @ph-kev do you know if in the coupler or atmos when we restart a run from a checkpoint, do we set the epoch of the ITime to the checkpoint date? so `t` starts at 0 seconds from the epoch. Or do we retain the same epoch as the original simulation?

NEWS.md
```diff
@@ -2,6 +2,12 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+- ![][badge-🐛bugfix] Let `save_checkpoint`/`read_checkpoint` round-trip an `ITime`, so a
```
kmdeck: What does "round trip an ITime" mean?

kmdeck (commented):
Thanks, Alexis! This was the PR I had open in the coupler to to test using the default land model.

I think if you rebase that off of Coupler main and then use this branch, it will test. 

The rtol thing based on the snow surface T may need to be factored in to that test

kmdeck:
> Thanks, Alexis! This was the PR I had open in the coupler to to test using the default land model.
> 
> I think if you rebase that off of Coupler main and then use this branch, it will test.
> 
> The rtol thing based on the snow surface T may need to be factored in to that test

https://github.com/CliMA/ClimaCoupler.jl/pull/2004

docs/src/restarts.md
kmdeck: @szy21 What do you think about this? We do a root solve for the snow surface temperature, and use the previous value as the initial guess for the solve. That means when we start a simulation, we need to provide a first guess but without knowing the previous. I think we may use the atmosphere temp but I am not sure.

but I guess this does break exact bit for bit restarts as Alexis notes here. The snow surface temp is saved in the cache so it is possible to save.

src/shared_utilities/checkpoints.jl
ph-kev: For coupler, the restarted time use the same `epoch` as the original simulation and `t` starts from the restarted time.

docs/src/restarts.md
szy21: The restart test in the coupler probably has a tolerance, and if that passes I think it's fine. We can come back to this if we need bitwise reproducible restart.

Would it work if you always use atmosphere temperature or some other initial guess that doesn't depend on the previous timestep? In general I think it's helpful to remove dependency on previous timesteps for restart etc.

NEWS.md
AlexisRenchon: This is what Claude says about it: =)

I am not sure if this is helpful or needed. What do you think @ph-kev?

```
"Round-trip" = save it, read it back, get the same thing out. Here: save_checkpoint(Y, t, ...) with an ITime → read_checkpoint hands you
  back an equivalent ITime, not something degraded.

  What actually changed (src/shared_utilities/checkpoints.jl:129-196):

  - Before, save didn't work at all with an ITime. It computed the filename via floor(Int, t / (60*60*24)), and ClimaUtilities only defines
  / for ITime / ITime (ITime.jl:467), so dividing by the integer 86400 threw. And the raw t was being stuffed into an HDF5 attribute, which
  can't hold an ITime anyway.
  - Now, save writes float(t) (plain seconds) plus a separate "epoch" string attribute holding the start date the ITime carries. Read sees
  the epoch and rebuilds ITime(t; epoch = DateTime(epoch)); if there's no epoch it returns the bare seconds as before.

  So the two halves are the round trip: the seconds survive the HDF5 attribute, and the epoch — the piece that would otherwise be silently
  dropped — comes back too. That's what makes a start_date/stop_date simulation restartable, since those set up t as an ITime with an epoch
  and a bare Float64 on the way back out wouldn't reconstruct it.

  If "round-trip" reads as jargon, the NEWS line could be:

  - ![][badge-🐛bugfix] `save_checkpoint`/`read_checkpoint` now preserve an `ITime`
    (its seconds *and* its start date), so a simulation set up from a `start_date`
    and a `stop_date` can be checkpointed and restarted.
```

AlexisRenchon:
### Do we keep the original epoch on restart?

We do now — the restarted run counts its time from the original start date, matching the convention described for the coupler above. The docs and the test used to rebase the epoch to the checkpoint date and restart `t` at 0; both now pass the `ITime` from the checkpoint straight through as `t0`:

```julia
t_restart = ClimaLand.initial_time_from_checkpoint(restart_file; model)
start_date = epoch(t_restart)
t0, tf, Δt = promote(
    t_restart,
    ITime(Dates.value(Second(stop_date - start_date)); epoch = start_date),
    ITime(Δt_seconds; epoch = start_date),
)
```

The `promote` is needed because `LandSimulation(::ITime, ::ITime, ::ITime, ...)` does not promote its arguments the way the `DateTime` and `Float64` convenience constructors do, and `read_checkpoint` returns whatever period represents the time exactly — `ITime(3, Hour(1))` for a 3-hour checkpoint, against a `Δt` in minutes.

This is numerically a no-op. Running both conventions side by side to the same final time gives states that agree to every digit, and both are exactly bit-for-bit once `p.snow.T_sfc` is restored:

| field | same epoch | rebased epoch | + `T_sfc` restored |
| --- | --- | --- | --- |
| `snow.S_l` | 1.3e-3 | 1.3e-3 | 0.0 |
| `soil.∫F_e_dt` | 1.35e-4 | 1.35e-4 | 0.0 |
| `soil.θ_i` | 8.05e-5 | 8.05e-5 | 0.0 |
| `canopy.energy.T` | 2.7e-5 | 2.7e-5 | 0.0 |

That is expected: the forcing is evaluated at `date(t)`, and `default_cos_zenith_angle` uses `date(t)` whenever the `ITime` carries an epoch, so neither the drivers nor the insolation can tell the two apart.

### Tolerance on the plain restart

The table above also shows `snow.S_l` at 1.3e-3, over the 1e-3 the test asserted, so I relaxed that comparison to 5e-3 and said why in the code and in the docs. It is a scale artifact rather than a real disagreement: the snowpack holds at most a millimetre of liquid water in this simulation, and the two runs differ by a micron of it. Every other prognostic variable is still within 1e-4. Note this is independent of the change above — the rebased column exceeds 1e-3 by exactly the same amount — so the PR description's "relative 1e-4 at worst" should be read as applying to everything but `snow.S_l`.

The suggestion of an initial guess for the snow surface temperature that does not depend on the previous timestep would remove the caveat entirely, and I have not attempted it here.

AlexisRenchon:
I added a commit to match what is done in the coupler... I am not sure if that is better

docs/src/restarts.md
kmdeck: Ill try using atmos T in a seperate branch and see how it works.

kmdeck: https://github.com/CliMA/ClimaLand.jl/pull/1844

NEWS.md
ph-kev: For checkpointing `ITime`, you need to use `promote` or save all the values of a `ITime`, since `ITime` promote the values depending on what is passed to it.

docs/src/restarts.md
kmdeck: it seems OK. Ill look into more detail next week and then we can merge it. The restarts may work in the coupler already, as you noted.

kmdeck: removing this text as we no longer use the guess = prev value

kmdeck (approved):

PR #1853, opened 2026-08-12
Allow Thermodynamics 1.3

tapios:
[Thermodynamics v1.3.0](https://github.com/CliMA/Thermodynamics.jl/releases/tag/v1.3.0) is now registered. ClimaLand pins `Thermodynamics = "0.15.9, ~1.0, ~1.2"`, and the tilde in `~1.2` means 1.3.0 is excluded.

This PR widens the bound to `"0.15.9, ~1.0, ~1.2, ~1.3"` in `Project.toml`, `docs/Project.toml`, and `test/Project.toml`, keeping the existing tilde-pinning convention rather than loosening to `1`.

Thermodynamics 1.3.0 is a minor rather than patch release because it raises the minimum Julia version from 1.6 to 1.10. The code changes are saturation-adjustment convergence fixes that should not affect land.


kmdeck (approved):
I wonder if we could just try "1.2", which would give # [1.2.0, 2.0.0)

I am also fine with what is here.

tapios:
> I wonder if we could just try "1.2", which would give # [1.2.0, 2.0.0)
> 
> I am also fine with what is here.

Yes, good point. Done.

PR #1858, opened 2026-08-14
Update subseasonal IC to set canopy IC

kmdeck:
## Purpose 
The default land model introduces additional prognostic variables which need initial conditions to be set. In advance of changing to that in the coupler, adjust the seasonal set IC function to accomodate them. Also, rename this to be more clear - set IC using ERA5Land file.

For context: if the land model saves Y, the simulation is restartable (we now have unit tests showing this). However, in many cases, it is convenient to start from a partial state - e.g. ERA5Land does not have the same prognostic variables that we do, or we only save the spun up state for the slowly adjusting variables. In this case, we need to make set IC for some of the prognostic variables. The land functions do this. A good guess for canopy temp is to use e.g. atmospheric Temp as T_canopy. For the additional prognostic Pmodel variables, it is helpful to make a good guess using the actual VPD. so we also use atmos P, q. 

## Content

----

### DIFF
--- a/docs/src/APIs/Simulations.md
+++ b/docs/src/APIs/Simulations.md
@@ -9,7 +9,7 @@ ClimaLand.Simulations.LandSimulation
 ClimaLand.Simulations.step!
 ClimaLand.Simulations.solve!
 ClimaLand.Simulations.make_set_initial_state_from_file
-ClimaLand.Simulations.make_set_subseasonal_initial_conditions
+ClimaLand.Simulations. make_set_initial_state_from_era5land
 ClimaLand.Simulations.make_set_initial_state_from_atmos_and_parameters
 ClimaLand.Simulations.set_canopy_component_initial_conditions!
 ```
--- a/docs/src/APIs/Simulations.md
+++ b/docs/src/APIs/Simulations.md
@@ -9,7 +9,7 @@ ClimaLand.Simulations.LandSimulation
 ClimaLand.Simulations.step!
 ClimaLand.Simulations.solve!
 ClimaLand.Simulations.make_set_initial_state_from_file
-ClimaLand.Simulations. make_set_initial_state_from_era5land
+ClimaLand.Simulations.make_set_initial_state_from_era5land
 ClimaLand.Simulations.make_set_initial_state_from_atmos_and_parameters
 ClimaLand.Simulations.set_canopy_component_initial_conditions!
 ```
--- a/src/simulations/initial_conditions.jl
+++ b/src/simulations/initial_conditions.jl
@@ -826,12 +826,11 @@ function make_set_initial_state_from_era5land(
 )
     function set_ic!(Y, p, t, land)
         atmos = ClimaLand.get_drivers(land)[1]
-        earth_param_set = ClimaLand.get_earth_param_set(land.soil)
         if atmos isa ClimaLand.PrescribedAtmosphere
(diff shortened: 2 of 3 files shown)


src/simulations/initial_conditions.jl
```diff
@@ -841,7 +852,24 @@ function make_set_subseasonal_initial_conditions(
             set_soilco2_SOC_from_soilgrids!(Y.soilco2.SOC)
         end
         Y.canopy.hydraulics.ϑ_l .= land.canopy.hydraulics.parameters.ν
-
+        set_canopy_component_initial_conditions!(
```
kmdeck, line 854: this sets Y.canopy.energy.T = p.drivers.T. In coupled sim, p.driver is update by the coupler

src/simulations/initial_conditions.jl
```diff
@@ -822,6 +825,14 @@ function make_set_subseasonal_initial_conditions(
     interpolation_method = Interpolations.Linear(),
 )
     function set_ic!(Y, p, t, land)
```
costachris, line 827: I think this will fail since `t0` is undefined. Maybe
```suggestion
    function set_ic!(Y, p, t0, land)
```

docs/src/APIs/Simulations.md
```diff
@@ -9,7 +9,7 @@ ClimaLand.Simulations.LandSimulation
 ClimaLand.Simulations.step!
 ClimaLand.Simulations.solve!
 ClimaLand.Simulations.make_set_initial_state_from_file
-ClimaLand.Simulations.make_set_subseasonal_initial_conditions
+ClimaLand.Simulations. make_set_initial_state_from_era5land
```
costachris:
```suggestion
ClimaLand.Simulations.make_set_initial_state_from_era5land
```

src/simulations/initial_conditions.jl
```diff
@@ -822,6 +825,14 @@ function make_set_subseasonal_initial_conditions(
...
+        atmos = ClimaLand.get_drivers(land)[1]
+        earth_param_set = ClimaLand.get_earth_param_set(land.soil)
```
costachris: `earth_param_set ` not used?

src/simulations/initial_conditions.jl
```diff
@@ -841,7 +852,24 @@ function make_set_subseasonal_initial_conditions(
             set_soilco2_SOC_from_soilgrids!(Y.soilco2.SOC)
         end
         Y.canopy.hydraulics.ϑ_l .= land.canopy.hydraulics.parameters.ν
```
costachris, line 853: should the lake fraction be set?
like e.g. here
https://github.com/CliMA/ClimaLand.jl/blob/aec8af20c580d5e8cb1cf18219efb6cf2fc2c7ef/src/simulations/initial_conditions.jl#L551

costachris (approved):
Perhaps update NEWS.md under main since this is breaking

src/simulations/initial_conditions.jl
kmdeck: this happens in `set_initial_cache!`. only the initial conditions (Y) are set in set_ic!

src/simulations/initial_conditions.jl
kmdeck: good catch

PR #1860, opened 2026-08-20
Use WorkerBackend for Derecho

ph-kev:
This PR uses the `WorkerBackend` for `Derecho` which allows for the workers to begin work when they are available and pack workers onto a single node to save GPU hours.

### DIFF
--- a/experiments/calibration/README.md
+++ b/experiments/calibration/README.md
@@ -25,6 +25,12 @@ latitude-weighted scalar covariance matrix.
       submission for the forward models.
     - You may want to use `tmux` to keep a persistent session on the cluster.
 
+On Derecho, the calibration uses the `WorkerBackend`: persistent workers are
+submitted once with `ClimaCalibrate.add_workers`, packed 4 per allocation (one
+per GPU), and reused across iterations. Workers are not resubmitted if they
+hit their walltime; to continue a stopped calibration, rerun with the same
+`OUTPUT_DIR`.
+
 For example, launching er.jl on Derecho
 Go into tmux:
 
@@ -52,3 +58,9 @@ are compatible with ClimaCalibrate v0.2.2.
 - `DerechoBackend`: `climacommon/2025_02_25`
 - `CaltechHPCBackend`: `climacommon/2024_10_09`
 - `GCPBackend`: `climacommon` is not supported
+
+On Derecho, if workers repeatedly exit with "Master process (id 1) could not
+connect within 60.0 seconds", the driver could not reach the address the
+workers advertise. See `julia_worker_hsn_bind.sh` and check the logs in
+`OUTPUT_DIR/worker_logs/` (each launch writes to its own timestamped
+subdirectory).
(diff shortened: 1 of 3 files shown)


experiments/calibration/README.md
```diff
@@ -25,6 +25,12 @@ latitude-weighted scalar covariance matrix.
       submission for the forward models.
     - You may want to use `tmux` to keep a persistent session on the cluster.
 
+On Derecho, the calibration uses the `WorkerBackend`: persistent workers are
+submitted once with `ClimaCalibrate.add_workers`, packed 4 per allocation (one
+per GPU), and reused across iterations. Workers are not resubmitted if they
+hit their walltime; to continue a stopped calibration, rerun with the same
+`OUTPUT_DIR`.
```
ph-kev, line 32: Is the land calibration longer than the walltime for derecho? If it is, then this can be annoying to restart the workers.

experiments/calibration/julia_worker_hsn_bind.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# Start a Julia Distributed worker bound to the node's HSN interface.
```
ph-kev, line 2: Is this something we can put in ClimaCalibrate?

experiments/calibration/README.md
kmdeck: Do you mean does it take longer than 12 hours? Depends on the # of iterations we do. I was doing 10 iterations which takes longer than that, but maybe we could fit 6 in a 12 hour period?

ph-kev: The workers are not requeued once the walltime is up for the `WorkerBackend`. If all the iterations take more than 12 hours, then the calibration need to be manually restarted.

I opened an issue for this in ClimaCalibrate, but I don't know how difficult and long it will take to implement: https://github.com/CliMA/ClimaCalibrate.jl/issues/347

kmdeck: I dont think this is a huge problem, we can try it and see @AlexisRenchon for visibility

experiments/calibration/julia_worker_hsn_bind.sh
```diff
@@ -0,0 +1,16 @@
...
+#
+# Why: `julia --worker` binds and advertises `getipaddr()`, which on Derecho
+# compute nodes is the management network (10.169.0.0/16). Login nodes
+# cannot reach that network, so a driver running in tmux on a login node
+# never connects and every worker exits with "Master process (id 1) could
+# not connect within 60.0 seconds". The HSN (10.14.0.0/18) is shared by
```
nefrathenrici: The default is 300.0 seconds

experiments/calibration/julia_worker_hsn_bind.sh
nefrathenrici: Yes, this should live in ClimaCalibrate but we can leave it here for now if merging the PR is a priority.

nefrathenrici (approved):
Thanks!

PR #1865, opened 2026-09-01
Clearer PModel equations, better testing, simple viscosity correction

kmdeck:
## Purpose 
This PR does a few things to try and improve our P-model implementation:
- We used to call "compute pmodel outputs" and compare the output of that against an R implementation of the pmodel as a unit test. However, that used _duplicate code_ to compute the output, i.e. it did not use the same functions (`compute_optimal_capacities`, and `compute_blended_photosynthesis`) that we do in our simulations. I now use the same functions for both. To enable this, I removed a few intermediate quantities which are not output from our `compute_optimal_capacities` & `compute_blended_photosynthesis`. As long as the remaining GPP, Vcmax25, Jmax25 etc are tested and correct, I assume that the intermediate quantities are also correct.
- Simpler viscosity correction. It was kind of crazy what we were doing, we dont need this complicated of a formula. I changed to one that should hold between 250 and 390K which should be plenty for the canopy. The tests still pass.
- I tried to make our pmodel functions match the Ren et al paper and Stocker et al paper more closely, e.g. computing "L" and limiting L to be real (c* > mj), instead of clipping in other places (e.g. clipping m)
- avoid multiplying and dividing by zero if beta = 0 (ci -> gamma star, mc ->0, mj -> 0, but mc/mj is defined)

## To-do

Fix downgrades (will be fixed in #1866)

## Content

----

### DIFF
--- a/src/standalone/Vegetation/pmodel.jl
+++ b/src/standalone/Vegetation/pmodel.jl
@@ -362,11 +362,8 @@ Output name         Description (units)
 - "mj"            CO2 limitation factor for light-limited photosynthesis (unitless)
 - "mc"            CO2 limitation factor for Rubisco-limited photosynthesis (unitless)
 - "ci"            Intercellular CO2 concentration (Pa)
-- "iwue"          Intrinsic water use efficiency (Pa)
 - "gs"            Stomatal conductance (mol m^-2 s^-1 Pa^-1)
-- "vcmax"         Maximum rate of carboxlation (mol m^-2 s^-1)
 - "vcmax25"       Vcmax normalized to 25°C via modified-Arrhenius type function (mol m^-2 s^-1)
-- "jmax"          Maximum rate of electron transport (mol m^-2 s^-1)
 - "jmax25"        Jmax normalized to 25°C via modified-Arrhenius type function (mol m^-2 s^-1)
 - "Rd"            Dark respiration rate (mol m^-2 s^-1)
 """
@@ -422,7 +419,7 @@ function compute_full_pmodel_outputs(
     # Compute intermediate values
     ϕ0_c3, ϕ0_c4 = intrinsic_quantum_yield(T_canopy, parameters)
     Γstar = co2_compensation_pmodel(T_canopy, To, P_air, R, ΔHΓstar, Γstar25)
-    ηstar = compute_viscosity_ratio(T_canopy, To, ρ_water)
+    ηstar = compute_viscosity_ratio(T_canopy, To)
     Kmm = compute_Kmm(T_canopy, P_air, Kc25, Ko25, ΔHkc, ΔHko, To, R, oi)
     ξ_opt_c3 = sqrt(β_c3 * (Kmm + Γstar) / (Drel * ηstar))
     ξ_opt_c4 = sqrt(β_c4 * (Kmm + Γstar) / (Drel * ηstar))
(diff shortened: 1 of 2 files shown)


src/standalone/Vegetation/pmodel.jl
```diff
@@ -1237,117 +1203,55 @@ function quadratic_intrinsic_quantum_yield(
...
+    patek_dimensionless_viscosity(
         T::FT,
-        ρ_water::FT
     ) where {FT}
 
-Computes the viscosity of water in Pa s given temperature T (K) and density ρ_water (kg/m^3)
-according to Huber et al. (2009) [https://doi.org/10.1063/1.3088050].
+Computes the viscosity of water in Pa s given temperature T (K),
+as found by J. Pátek, J. Hrubý, J. Klomfar, M. Součková, and A. H. Harvey,
+J. Phys. Chem. Ref. Data https://doi.org/10.1063/1.304357538, 21 (2009); as presented
```
odunbar, line 1219: The DOI for Patek doesn't work : I think this is correct  https://doi.org/10.1063/1.3043575 

maybe the 38 is like a journal number

src/standalone/Vegetation/pmodel.jl
```diff
@@ -1237,117 +1203,55 @@ function quadratic_intrinsic_quantum_yield(
...
-    μ = μ_bar * μ_ast
+function patek_dimensionless_viscosity(T::FT) where {FT}
+    a1 = FT(280.68)
+    b1 = FT(-1.9)
+    a2 = FT(511.45)
+    b2 = FT(-7.7)
+    a3 = FT(61.131)
+    b3 = FT(-19.6)
+    a4 = FT(0.45903)
+    b4 = FT(-40)
```
odunbar: NB - I checked this is indeed `-40.0` (RE precision)

src/standalone/Vegetation/pmodel.jl
```diff
@@ -1237,117 +1203,55 @@ function quadratic_intrinsic_quantum_yield(
...
     ) where {FT}
 
 Computes η*, the ratio of the viscosity of water at temperature T to that at To = 25˚C.
 """
-function compute_viscosity_ratio(T::FT, To::FT, ρ_water::FT) where {FT}
-    η25 = viscosity_h2o(To, ρ_water)
-    ηstar = viscosity_h2o(T, ρ_water) / η25
-    return FT(ηstar)
+function compute_viscosity_ratio(T::FT, To::FT) where {FT}
+    μ = patek_dimensionless_viscosity(max(min(T, FT(383.15)), FT(253.15)))
```
odunbar, line 1247: Perhaps add a warning if temps try to go outside this limit? (not necessary - but could reveal limiter bugs later on)

src/standalone/Vegetation/pmodel.jl
```diff
@@ -1237,117 +1203,55 @@ function quadratic_intrinsic_quantum_yield(
...
+    μ25 = patek_dimensionless_viscosity(To)
```
odunbar, line 1248: Is To guaranteed to be in the range?

src/standalone/Vegetation/pmodel.jl
```diff
@@ -702,9 +661,9 @@ function compute_optimal_capacities(
     Vcmax25_opt_c4 = Vcmax_opt_c4 / inst_temp_scaling_factor_Vcmax
 
     Jmax_opt_c3 =
-        4 * ϕ0_c3 * APAR_canopy_moles / sqrt((mj_c3 / (βm * mprime_c3))^2 - 1)
+        βm * 4 * ϕ0_c3 * APAR_canopy_moles * L_c3/sqrt(1 - (βm*L_c3)^FT(2))
```
odunbar, line 664: Can you check this? I'm not sure this one is correct. I'm just going from eqn 10b [here](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2024MS004599). Is it not 
```
βm * 4 * ϕ0_c3 * APAR_canopy_moles *(1/L_c3) *  sqrt(1-(L_c3)^2)
```
namely - i removed the beta from inside the sqrt() and used `sqrt(1/L^2 -1) =  sqrt((1-L^2)/L^2) = (1/L)*sqrt(1-L^2)` (if L>0) 

when L->0 we still need some kind of guard here maybe?

odunbar (changes requested):
Looks so much cleaner than before! 

2 things: 
- in the PR message you mention tests, but no tests are pushed - just to check were any test files meant to be here
- I think maybe the Jmax,opt value is not correct here. Maybe check this? (i could have course have made a mistake)

src/standalone/Vegetation/pmodel.jl
kmdeck: thanks for catching this!

src/standalone/Vegetation/pmodel.jl
kmdeck: Is there an issue adding warnings on GPU?

src/standalone/Vegetation/pmodel.jl
kmdeck: yes! in this case, To is a parameter (reference temp of 25 degrees C). but good check

src/standalone/Vegetation/pmodel.jl
kmdeck: I think this is OK (discussed on slack)

src/standalone/Vegetation/pmodel.jl
kmdeck: I thought there was (cant print out warnings or something like that). But I added to the doc string of patek_dimensionless_viscosity to make it clear what happens if you give it something outside this range.

odunbar: I didn't know this...

src/standalone/Vegetation/pmodel.jl
odunbar: Summary of slack : 
- i missed an extra "/" 
- this is algebraicly different from the published (10b) in Ren et al. 2025. But deriving it from (1),(2),(7),(10a) of that paper lead to _this_ implementation, not the published one.

odunbar (approved):
LGTM! glad we did not just go from published equations for this one...!

PR #1867, opened 2026-09-02
Fix C3/C4 blending in SIF

kmdeck:
## Purpose 
When computing SIF using the Farquhar model, only one PFT is modelled per cell, so we do not need to worry about averaging the contributions from C3 and C4 plants (because the PFT determines if the plant is C3 or C4). For the Pmodel, this is not the case.

This PR extends how we compute SIF so that we compute the blended contribution of C3 and C4 plants correctly. This means we average SIF, and not Vcmax and J/Jmax.

I also fixed a bug in a unit (unrelated).

## Content
- dispatch update_SIF!() over the photosynthesis model type, necessarily this creates unique helpers for computing SIF at a point.
- bypass previous get_J_over_Jmax, get_Vcmax_leaf which internally blend quantities between c3 and c4 plants, these are now computed separately; remote J over Jmax function as no longer used
- isolate the lee et al SIF model update

----

### DIFF
--- a/src/standalone/Vegetation/solar_induced_fluorescence.jl
+++ b/src/standalone/Vegetation/solar_induced_fluorescence.jl
@@ -48,19 +48,24 @@ ClimaLand.auxiliary_vars(model::Lee2015SIFModel) = (:SIF,)
 ClimaLand.auxiliary_types(model::Lee2015SIFModel{FT}) where {FT} = (FT,)
 ClimaLand.auxiliary_domain_names(::Lee2015SIFModel) = (:surface,)
 
-
-# 4 Solar Induced Fluorescence (SIF)
-
-# call function below inside photosynthesis.jl p
+Base.broadcastable(m::SIFParameters) = tuple(m)
 
 """
-    update_SIF!(p, Y, sif_model::Lee2015SIFModel, canopy)
+    update_SIF!(p, Y, sif_model::Lee2015SIFModel, photosynthesis_model::PModel, canopy)
 
-Updates observed SIF at 755 nm in W/m^2. Note that Tc is in Kelvin, and photo
-synthetic rates are in mol/m^2/s, and APAR is in PPFD.
+Updates observed SIF at 755 nm in W/m^2 assuming photosynthesis is modelled using the Pmodel.
+
+Computes a weighted sum of SIF from C4 and C3 plants using the C3 fraction per grid cell.
+
 Lee et al, 2015. Global Change Biology 21, 3469-3477, doi:10.1111/gcb.12948
 """
-function update_SIF!(p, Y, sif_model::Lee2015SIFModel, canopy)
+function update_SIF!(
+    p,
+    Y,
+    sif_model::Lee2015SIFModel,
+    photosynthesis_model::PModel,
+    canopy,
+)
     SIF = p.canopy.sif.SIF
     earth_param_set = canopy.earth_param_set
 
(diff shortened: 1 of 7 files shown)


src/standalone/Vegetation/solar_induced_fluorescence.jl
```diff
@@ -74,63 +79,224 @@ function update_SIF!(p, Y, sif_model::Lee2015SIFModel, canopy)
...
+        J_over_Jmax_c4,
+        Vcmax25_c4/max(LAI, eps(FT)),
+        APAR_canopy_moles,
+        Tc,
+        T_freeze,
+    )
+    return SIF_755_c3*fraction_c3+(1-fraction_c3)*SIF_755_c4
+end
+
+function sif_755_lee_model(
```
AlexisRenchon, line 305: do we want docstring for this function?

AlexisRenchon (approved):
Looks good! just a missing docstring?

src/standalone/Vegetation/solar_induced_fluorescence.jl
```diff
@@ -74,63 +79,224 @@ function update_SIF!(p, Y, sif_model::Lee2015SIFModel, canopy)
     APAR_canopy_moles = @. lazy(
         compute_APAR_canopy_moles(f_abs_par, par_d, λ_γ_PAR, c, planck_h, N_a),
     )
-
-    # Get max photosynthesis rates
     T_canopy = canopy_temperature(canopy.energy, canopy, Y, p)
-    Vcmax25_leaf = get_Vcmax25_leaf(Y, p, canopy.photosynthesis)
-    J_over_Jmax = get_J_over_Jmax(Y, p, canopy, canopy.photosynthesis)
```
odunbar, line 81: Instead of bypassing the `get_J_over_Jmax`, and `get_VCmax25_leaf` would it not make sense to reuse the infrastructure, but just remove the blend functionality? From what I can see - a lot of what is below is just writing out the internals of these functions?

src/standalone/Vegetation/solar_induced_fluorescence.jl
```diff
@@ -74,63 +79,224 @@ function update_SIF!(p, Y, sif_model::Lee2015SIFModel, canopy)
...
+    Jmax_c4 = Jmax25_c4 * inst_temp_scaling_Jmax_factor
+    J_over_Jmax_c3 =
+        electron_transport_pmodel(ϕ0_c3, APAR_canopy_moles, Jmax_c3)/Jmax_c3
+    J_over_Jmax_c4 =
+        electron_transport_pmodel(ϕ0_c4, APAR_canopy_moles, Jmax_c4)/Jmax_c4
+
+    SIF_755_c3 = sif_755_lee_model(
+        sif_parameters,
+        J_over_Jmax_c3,
+        Vcmax25_c3/max(LAI, eps(FT)),
```
odunbar: Is this now a "leaf" quantity? Maybe labelling `Vcmax25_leaf_c3 = ...` makes for easier re-reading, I'm not a fan of having calculations within argument lists.

odunbar: I actually have the same question regarding the J/Jmax - as in pmodel it uses canopy quantities while farqhuar uses leaf quantities. So maybe keeping consistent that e.g. all leaf quantities are labeled "leaf" and all canopy quantities are either "canopy" or unlabelled (i think this is sort of what you are doing already?)

odunbar (changes requested):
Looks great! I had a couple of questions regarding implementation/notation



Perhaps update the PR message to say more specific changes made here?
- dispatch `update_SIF!()` over the photosynthesis model type, necessarily this creates unique helpers for computing SIF at a point.
- bypass previous `get_J_over_Jmax`, `get_Vcmax_leaf` which internally blend quantities between c3 and c4 plants, these are now computed separately
- isolate the lee et al SIF model update

src/standalone/Vegetation/solar_induced_fluorescence.jl
kmdeck: It's true, but given that the two types of photosynthesis require different computations for J over Jmax, and we already need to have two methods of the lee et al computation for them, I dont think we also also need two methods of J_over_Jmax (especially as that is only needed for SIF). I wil remove the get_J_over_Jmax function entirely.

Farquhar:
```    
J_over_Jmax = electron_transport_farquhar(APAR_leaf_moles, Jmax, θj, ϕ)/Jmax
```

Pmodel:
```
    J_over_Jmax_c4 =
        electron_transport_pmodel(ϕ0_c4, APAR_canopy_moles, Jmax_c4)/Jmax_c4
```

kmdeck: We introduced the get_Vcmax25_leaf function as a diagnostic, to return a value for the grid cell, and then required a method for Farquhar and for Pmodel. If we move the "blend" out of this function, we would need to dispatch off of the photosynthesis type when computing the diagnostic elsewhere, since Pmodel needs the blend but Farquhar doesnt. so I am inclined to keep these as is.

Since it is such a simple function (divide by LAI for Pmodel, or not for Farquhar), I am inclined to leave this as is in the SIF computation.

src/standalone/Vegetation/solar_induced_fluorescence.jl
kmdeck: Ill fix this!

odunbar (approved):
LGTM - so long as failing docs etc are unrelated to this?

PR #1868, opened 2026-09-03
Fix snow residual flux

### LINKED ISSUE #1590: Inconsistent results from Snow Model
The Snow Model gives inconsistent results, particularly for phase change flux and water runoff. This can be seen when comparing gpu and cpu snowmip runs. 

cpu:

Image

Image

gpu:
Image

Image

kmdeck:
## Purpose 
I realized when revamping the conservation PR that the "residual" snow energy flux had been implemented incorrectly, and this PR fixes it (I think). Previously, we had solved for T_sfc as the root of the flux balance equation
`F(T) - G(T) = 0`, where `F = LH + SH + R_n`, and `G` is a conductive flux between the surface and the bulk temp. 
If the root was T >T_freeze, we clipped T_sfc = T_freeze. We then computed an additional energy flux which accounts for this as `FT(-κ * (T_sfc_root - _T_freeze) / d)` which would be the energy required to bring a layer of temperature T_sfc_root and layer thickness `d` back to `T_freeze`. 

This views the surface warming as first bringing the temp to T_sfc_root (to satisfy the equation) and then "cooling" it back to T_freeze, where the additional cooling flux has to come from the bulk snowpack.

But this odd because that makes the BC for the bulk snow:
`F(T_freeze) -κ * (T_sfc_root - _T_freeze) / d`

but where does this extra energy come from? the atmosphere does not see this flux. I think what actually happens is that our thin layer has a temperature at T_freeze (maximum) and any excess warming goes towards melting water. I tried to explain this below...

Long run: https://buildkite.com/clima/climaland-long-runs/builds/4997

## Content
To recap, we parameterize the snow surface temperature by solving for the root of the equation:
F(T) - G(T) = 0, where F = LH + SH + R_n, and G is a conductive flux between the surface and the bulk temp. This approaches arises from considering a two layer snow model, where the top layer is very thin. then for the two prognostic energy equations, we have:
(1) `d(U_top)/dt = (F(T) - G(T))`
(2) `d(U_bottom)/dt = (G(T)- G_soil)`
where `G_soil` is the ground heat flux. `U_top` and `U_bottom` are the energy per unit (horizontal) area of the top and bottom layers, i.e. the thickness has already been factored in, so the total energy is `U = U_top +U_bottom`. In total, then, we have
(3) `dU/dt = (F(T)- G_soil)`, which is what we solve.

Assuming the thermal inertia of the top layer is very small, it equilibrates quickly (on timescales shorter than our timestep), so we advance `U_top(t+dt) = U_top(T(t+dt), q_l)` where `q_l` is the liquid fraction of the top layer, and `T(t+dt)` is the root of the flux balance equation above. (U_top still changes each step, even though equation (1) looks like the tendency would be zero).

The issue is when the solved root is larger than `T_freeze`. this is not physical. In reality, the temperature would stay at `T_freeze` while the top layer melted. In this case, we set `T(t+dt) = T_freeze`. And we should get `U_top(t+dt) = U_top(T_freeze, q_l(t+dt))`, now the snow water in the top layer has melted so q_l increases. The amount of water melting as a flux is given by `(F(T_freeze) - G(T_freeze))/LH`, where `LH` is the latent heat of fusion. This can be used to increase the liquid water of the entire snowpack:
(4) `dq_l/dt = (F(T_freeze) - G(T_freeze))/LH*S`, where S is the snow water equivalent.

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -233,7 +233,7 @@ steps:
 
       - label: "Snow Col de Porte GPU"
         command: "julia --color=yes --project=.buildkite experiments/standalone/Snow/snowmip_simulation.jl cdp"
-        artifact_paths: "experiments/standalone/Snow/gpu/cdp/output_active/*png"
+        artifact_paths: "experiments/standalone/Snow/gpu/cdp_default_gradtemp/output_active/*png"
         agents:
           slurm_ntasks: 1
           slurm_gpus: 1
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -997,7 +997,6 @@ NVTX.@annotate function snow_boundary_fluxes!(
         P_snow * (1 - p.lake_fraction) +
         (P_liq + p.snow.turbulent_fluxes.vapor_flux - p.snow.water_runoff) *
         p.snow.snow_cover_fraction
-
     @. p.snow.liquid_water_flux =
         (
             P_liq + p.snow.turbulent_fluxes.vapor_flux * p.snow.q_l -
@@ -1015,13 +1014,10 @@ NVTX.@annotate function snow_boundary_fluxes!(
         model.parameters.earth_param_set,
     )
 
-    residual_surface_flux =
-        Snow.get_residual_surface_flux(model.parameters.surf_temp, Y, p)
     # positive fluxes are TOWARDS atmos, but R_n positive if snow absorbs energy
     @. p.snow.total_energy_flux =
         e_flux_falling_snow * (1 - p.lake_fraction) +
         (
-            residual_surface_flux +
             p.snow.turbulent_fluxes.lhf +
             p.snow.turbulent_fluxes.shf +
             p.snow.R_n - p.snow.energy_runoff - p.ground_heat_flux +
--- a/src/integrated/soil_snow_model.jl
+++ b/src/integrated/soil_snow_model.jl
@@ -384,7 +384,6 @@ NVTX.@annotate function snow_boundary_fluxes!(
         P_snow +
         (P_liq + p.snow.turbulent_fluxes.vapor_flux - p.snow.water_runoff) *
         p.snow.snow_cover_fraction
-
     @. p.snow.liquid_water_flux =
         (
             P_liq + p.snow.turbulent_fluxes.vapor_flux * p.snow.q_l -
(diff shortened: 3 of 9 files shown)


AlexisRenchon (changes requested):
I have a few suggestions in ar/snow-residual-review-fixes (https://github.com/CliMA/ClimaLand.jl/pull/1871), as usual, feel free to keep or discard anything from it. Once done and CI passes, ping me again and I can approve to merge on main!

kmdeck:
> I have a few suggestions in ar/snow-residual-review-fixes (#1871), as usual, feel free to keep or discard anything from it. Once done and CI passes, ping me again and I can approve to merge on main!

I made several of the suggested edits, thank you! The "double counting" comment I am unsure of.

AlexisRenchon:
For the double melt, maybe this helps (maybe not). I closed the suggestion PR. 

Two independent pieces of code both convert the same surface energy excess into liquid water.

  1. get_residual_melt_flux (snow_parameterizations.jl) computes the excess at the clipped surface temperature, F(T_freeze) + κ(T_freeze − T)/d, divides by ρ_l·L_f, and adds it to p.snow.liquid_water_flux inside snow_boundary_fluxes!. That is the intended new physics: surface excess melts snow.
  2. But total_energy_flux still delivers the full F(T_freeze) = lhf + shf + R_n to the bulk U. phase_change_flux (Snow.jl ~937) then predicts U_pred = U − applied_energy_flux·Δt, compares it to the energy of the pack at T_freeze at the current q_l, and converts any excess into melt again. Nothing tells it that part of that energy was already spent on melt in step 1.

  So the energy is delivered to U once, but the liquid water tendency receives two melt terms for it. The bulk energy is conserved, the partition is not: S_l rises at twice the rate the energy budget supports, and the pack's bulk temperature is pushed below freezing while it is melting. It only self-limits once T drops far enough that phase_change_flux stops firing.

  
The isothermal reproducer (pack at 273.15 K, q_l = 0.2, net surface forcing 884 W/m²) makes it unambiguous, since conduction into the bulk is exactly zero there so all excess is surface melt:

  ┌──────────────────────────────────────────┬───────────┐
  │                 quantity                 │   value   │
  ├──────────────────────────────────────────┼───────────┤
  │ surf_residual_flux (melt, m/s)           │ −2.650e-6 │
  ├──────────────────────────────────────────┼───────────┤
  │ phase_change_flux (melt, m/s)            │ −2.651e-6 │
  ├──────────────────────────────────────────┼───────────┤
  │ liquid_water_flux total (m/s)            │ −5.293e-6 │
  ├──────────────────────────────────────────┼───────────┤
  │ latent energy committed, ρ_l·L_f·dS_l/dt │ 1766 W/m² │
  ├──────────────────────────────────────────┼───────────┤
  │ energy actually entering the pack, dU/dt │ 884 W/m²  │
  └──────────────────────────────────────────┴───────────┘

  Two nearly identical melt terms, and the latent commitment is exactly twice the energy gained.

kmdeck:
@AlexisRenchon if this looks ok to you, Ill enable auto merge- if not, ill revisit next Monday!

thanks

AlexisRenchon (approved):

PR #1871, opened 2026-09-08
Fix snow melt double counting and snow-free NaNs

AlexisRenchon:
## Purpose

Follow-up to #1868, targeting `kd/address_snow_surface_temp_question`. Fix the reproduced melt double counting and snow-free NaNs in the residual surface-melt calculation.

## Fixes

- **Count surface melt once.** The residual surface heating already creates liquid water, but the bulk phase-change estimate also received the full applied energy flux and melted the same ice again. Subtract `rho_l * L_f * residual_melt_flux * snow_cover_fraction` from the energy flux used to estimate bulk phase change. The total energy tendency still receives the applied boundary energy flux.
- **Share the accounting across models.** Compute and apply residual melt in the common snow boundary update, alongside bulk phase change. Remove the separate additions from the standalone, soil-snow, and full-land boundary routines so each path applies surface melt exactly once before liquid-water clipping.
- **Keep snow-free columns finite.** Guard the conductive scaling length against zero and return zero residual melt where snow depth is zero. Previously, the conductive term evaluated as `0/0`, which propagated through the zero snow-cover fraction into the liquid-water tendency.
- **Restrict residual melt to the freezing cap.** Return zero below freezing and clamp the residual to nonpositive values. This prevents surface-solver flux mismatches from producing spurious melting or refreezing at uncapped surfaces.
- **Repair tests and documentation.** Update the bulk-temperature helper call to include its parameter argument, test surface melting in the existing cold-bulk case, use an approximate flux comparison, and document the residual's water-flux units and corrected surface-balance equation.

In the saved isothermal-snow reproducer, surface melt and bulk phase change each contributed about `-2.65e-6 m/s` before this change. Bulk phase change is now zero, while surface melt remains about `-2.65e-6 m/s`. The snow-free reproducer now has zero residual melt and zero liquid-water tendency.

## Validation

Using Julia 1.12.6 and the existing `.buildkite` environment:

- Snow parameterizations and model: 210 checks passed, including new Float32/Float64 regressions for full and partial snow cover, melt-energy accounting, snow-free columns, below-freezing surfaces, and vanishing positive snow depth.
- Snow integrated energy/water content: 8 checks passed.
- Snow parameter constructors: 2 checks passed.
- Coupled soil-snow: 26 checks passed, including new snow-free residual and liquid-water checks.
- Full land: 406 checks passed, including global tendencies, Jacobian updates, an integration step, and inland-water configurations.
- Reran the saved review reproducers.
- JuliaFormatter 2.10.1 passed for all changed Julia files; whitespace check passed with the existing CRLF file treated as CRLF.

The full repository test suite, GPU execution, and the long SnowMIP experiment were not run. The cold-pack melt/percolation/refreezing behavior introduced by #1868 still warrants evaluation in SnowMIP.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@ ClimaLand.jl Release Notes
 ========================
 main
 ----
+- ![][badge-🐛bugfix] Count snow surface melt only once in the liquid-water
+  budget and keep residual melt finite in snow-free columns and zero below freezing.
 - Update compat to ClimaCore v0.16 PR[#1869](https://github.com/CliMA/ClimaLand.jl/pull/1869)
 
 v1.12.0
--- a/src/integrated/land.jl
+++ b/src/integrated/land.jl
@@ -997,11 +997,17 @@ NVTX.@annotate function snow_boundary_fluxes!(
         P_snow * (1 - p.lake_fraction) +
         (P_liq + p.snow.turbulent_fluxes.vapor_flux - p.snow.water_runoff) *
         p.snow.snow_cover_fraction
+    residual_melt_flux = Snow.get_residual_melt_flux(
+        model.parameters.surf_temp,
+        Y,
+        p,
+        model.parameters.earth_param_set,
+    )
 
     @. p.snow.liquid_water_flux =
         (
             P_liq + p.snow.turbulent_fluxes.vapor_flux * p.snow.q_l -
-            p.snow.water_runoff
+            p.snow.water_runoff + residual_melt_flux
         ) * p.snow.snow_cover_fraction
 
     e_flux_falling_snow = Snow.energy_flux_falling_snow(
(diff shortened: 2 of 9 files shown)


kmdeck:
- I will make this change"Compute and apply residual melt in the common snow boundary update, alongside bulk phase change. "
- "Count surface melt once. " I am not sure about this/will think about it more.
- "Keep snow-free columns finite" I will clip the depth to be eps to prevent dividing by zero
Regarding the other comments, I dont think we need this complexity:
```   
 @. p.snow.surf_residual_flux = ifelse(
        (z > 0) & (p.snow.T_sfc >= _T_freeze),
        min(
            (
                p.snow.turbulent_fluxes.lhf +
                p.snow.turbulent_fluxes.shf +
                p.snow.R_n +
                κ * (p.snow.T_sfc - p.snow.T) / max(
                    surface_temp_scaling_length(κ, ρ, z, earth_param_set),
                    eps(FT),
                )
            ) / (_LH_f0 * _ρ_l),
            FT(0),
        ),
        FT(0),
    )
```
For one, if z = 0, the snow cover fraction is zero, so I think the actual flux applied should be zero. It's similar to how LHF, SHF have nonzero values but then they are multiplied by the snow cover fraction to get zero.
I dont think we need the temperature condition either. if T_sfc < T_freeze, the root solve should give LHF+SHF+R_n +K(T_sfc - T)/d = 0, so the residual will be zero anyways. In fact, including it might be good in case the root solve is not exact.

so I think this is more accurate and more readable:
```   
 @. p.snow.surf_residual_flux = 
                (p.snow.turbulent_fluxes.lhf +
                p.snow.turbulent_fluxes.shf +
                p.snow.R_n +
                κ * (p.snow.T_sfc - p.snow.T) / max(
                    surface_temp_scaling_length(κ, ρ, z, earth_param_set),
                    eps(FT),
                )
            ) / (_LH_f0 * _ρ_l)
```

PR #1872, opened 2026-09-10
Replace deprecated ClimaCore names removed in ClimaCore v0.16

imreddyTeja:
Removesthe backwards-compatibility shims left behind by the DataLayouts, Geometry and Operators rewrites. Three of them were still in use here. 

## Deprectated name updates

**`MatrixFields.⋅` → `*`** — 18 operator uses in `energy_hydrology.jl`, `Biogeochemistry.jl` and `rre.jl`. `⋅` is no longer exported by `MatrixFields` in v0.16, so the four `import ClimaCore.MatrixFields: @name, ⋅` lines now import only `@name` (two of them never used the operator). `*` has denoted matrix multiplication since ClimaCore v0.14.27.

**`ClimaCore.RecursiveApply.rzero(T)` → `zero(T, space)`** — `ClimaLand.jl`, `models.jl` and `Canopy.jl`. The whole `RecursiveApply` module was a compatibility layer over the `AutoBroadcaster` wrappers added in v0.14.52.  The two `models.jl` docstrings that told implementers to define `rzero` now describe the `zero` requirement instead.

`Operators.WeakDivergence` -> `Operators.Divergence{Operators.WeakForm}` 

EDIT:
TODO: remove new recursive apply usage added in https://github.com/CliMA/ClimaLand.jl/commit/b0a36d4d9e768a3d43eb92fbdc89fdbaad947c88


ph-kev (approved):

src/standalone/Vegetation/Canopy.jl
```diff
@@ -1183,10 +1183,7 @@ function initialize_boundary_vars(model::CanopyModel{FT}, coords) where {FT}
         ClimaLand.TopBoundary(),
     )
     additional_aux = map(zip(types, domains)) do (T, domain)
-        zero_instance = ClimaCore.RecursiveApply.rzero(T)
-        f = map(_ -> zero_instance, getproperty(coords, domain))
-        fill!(ClimaCore.Fields.field_values(f), zero_instance)
-        f
+        f = zeros(T, axes(getproperty(coords, domain)))
```
kmdeck, line 1186: With all of the changes that look like this - I am trying to recall why we did the fill! statement with the field_values(f) statement. I think it was because it was convenient to have all the values in the ocean (where the mask = 0) set to a particular value, vs unitialized. Does the zeros(type, axes) know about the mask?

kmdeck: it may not matter, but I am wondering if there is a smalll behavior change here.

src/standalone/Soil/energy_hydrology.jl
```diff
@@ -599,6 +599,9 @@ NVTX.@annotate function horizontal_components!(
     p::NamedTuple,
     z::ClimaCore.Fields.Field,
 )
+    # TODO: ClimaCore v0.16 removes `WeakDivergence` in favour of
+    # `Divergence{WeakForm}`, but `WeakForm` only exists from v0.15.0 and this
+    # package still pins v0.14.5x. Switch when the ClimaCore bound is raised.
```
kmdeck, line 604: I think we are OK? https://github.com/CliMA/ClimaLand.jl/blob/724432622df32e597dc47676c9177a31e59fb450/Project.toml#L52

imreddyTeja: Ah good catch. That was a leftover

src/standalone/Vegetation/Canopy.jl
imreddyTeja: https://github.com/CliMA/ClimaCore.jl/blob/877647c5a5898cab2aa3f54645959695e75a66f4/src/Fields/Fields.jl#L328

I think at some point `zeros` became non-mask-aware, so we should be fine switching.

kmdeck (approved):

src/standalone/Soil/energy_hydrology.jl
kmdeck: sounds good! I think we can update line 605 to this then: Operators.Divergence{Operators.WeakForm}()

PR #1877, opened 2026-09-14
correct J+over_Jmax limit for pmodel and fix bug in SIF

### LINKED ISSUE #1876: Bug: we use light-dependent term in computing dark-condition yield
## Issue
https://github.com/CliMA/ClimaLand.jl/blob/724432622df32e597dc47676c9177a31e59fb450/src/standalone/Vegetation/solar_induced_fluorescence.jl#L319
in the Sif model, we have the following Phi_P0. "dark-adapted" maximum photochemical yield. 

In 
https://pmc.ncbi.nlm.nih.gov/articles/PMC9285887/ and  https://doi.org/10.1111/gcb.12948 we follow the SIF equations. 

This dark adapted yield is computed in eqn (7) as
```
 ϕp0 = kp / max(kf + kp + kd, eps(FT)) 
```

## Possible solution

just swap `kn` -> `kd`

Our current setup is likely a typo, as physically a dark-adapted rate should not have any contribution from `kn` (which represents one of the light-dependent pathways NPQ), while `kd` is a passive heat dissipation that should appear in dark conditions

kmdeck:
## Purpose 
Closes issue https://github.com/CliMA/ClimaLand.jl/issues/1876
The pmodel limit of J/Jmax as Jmax -> 0 is 1; this also fixes that limit. For Farquhar, Jmax should never be zero in vegetated areas so we should not need to worry about this in that case.

## Content

----


src/standalone/Vegetation/solar_induced_fluorescence.jl
```diff
@@ -256,16 +257,12 @@ function compute_SIF_at_a_point_pmodel(
     # The following Jmax and Vcmax are canopy level
     Jmax_c3 = Jmax25_c3 * inst_temp_scaling_Jmax_factor
     Jmax_c4 = Jmax25_c4 * inst_temp_scaling_Jmax_factor
-    J_over_Jmax_c3 =
-        electron_transport_pmodel(ϕ0_c3, APAR_canopy_moles, Jmax_c3)/max(
```
odunbar, line 260: Is this method still needed somewhere? Or can we remove it.

odunbar (approved):
LGTM!

Just wondering if the method bypassed is still needed or should it be removed.

src/standalone/Vegetation/solar_induced_fluorescence.jl
kmdeck: it looks like we still use it in a few places.

PR #1878, opened 2026-09-15
Update to CTS v1

ph-kev:
This PR updates the `Project.toml`s and manifests to use ClimaTimeSteppers v1.


kmdeck (approved):
looks good. any changes we might expect in terms of performance?

ph-kev:
> looks good. any changes we might expect in terms of performance?

I don't think there should be changes to performance.

PR #1882, opened 2026-09-17
Remove ClimaCore type piracy

nefrathenrici:
ClimaCore 1.0 dropped some `Spaces.issubspace` methods on purpose (they would let a center level count as a subspace of a face level, which is wrong). v1.12.2 added them back inside ClimaLand, which is type piracy. This removes them and fixes the broadcasts on our side instead.

Changes
- `Δz_top` and `Δz_bottom` used to live on the top and bottom face levels. They  are surface quantities, so they now live on the surface space, same as `z_sfc`
-  The top-face local geometry used in the Jacobian boundary terms (soil andsoilco2) now gets projected onto the surface space before use.
- The free-drainage bottom BCs use `bottom_center_to_surface` instead of `Fields.level`
- Also restores the plain `Aqua.test_piracies(ClimaLand)`


src/shared_utilities/Domains.jl
```diff
@@ -1170,6 +1170,28 @@ function top_face_to_surface(face_field::ClimaCore.Fields.Field, surface_space)
...
+Given a `face_field` defined on a 3D extruded face finite difference space,
+this would return a 2D field with values equal to the bottommost level.
+"""
+function bottom_face_to_surface(
+    face_field::ClimaCore.Fields.Field,
+    surface_space,
+)
+    sfc_level =
+        ClimaCore.Fields.level(face_field, ClimaCore.Utilities.PlusHalf(0))
+    # Project onto surface space
```
dennisYatunin: "Project" isn't the right word for this context, since it's not doing a change of basis. Something like "reassign" or "reconstruct" would be more accurate

src/simulations/initial_conditions.jl
```diff
@@ -1058,9 +1058,17 @@ function make_set_initial_state_from_atmos_and_parameters(
...
-            ClimaCore.Fields.level(Y.bucket.T, i) .= p.drivers.T
+            # Project each level onto the surface space to set it in place
+            T_level = ClimaCore.Fields.Field(
+                ClimaCore.Fields.field_values(
+                    ClimaCore.Fields.level(Y.bucket.T, i),
+                ),
+                surface_space,
+            )
+            T_level .= p.drivers.T
         end
```
dennisYatunin: Splitting this 2D-to-3D extruding broadcast into `n_levels` separate broadcasts seems pretty inefficient. Better to just grab the `field_values` and launch only one kernel:
```julia
ClimaCore.Fields.field_values(Y.bucket.T) .= ClimaCore.Fields.field_values(p.drivers.T)
```

dennisYatunin (approved):

kmdeck (approved):
This looks good to me - thank you

ph-kev (approved):
Does this change still work on ClimaCore v0.16?

nefrathenrici:
> Does this change still work on ClimaCore v0.16?

I haven't tested it but yes it should, I don't think we need to change the compat. The downgrade test will break if not.

src/simulations/initial_conditions.jl
dennisYatunin: Oh no, looks like my suggestion revealed a bug in ClimaCore. @nefrathenrici, to avoid the materialization error, try the following instead:
```julia
ClimaCore.Fields.field_values(Y.bucket.T) .=
    ClimaCore.Fields.field_values(p.drivers.T) .+
    zero.(ClimaCore.Fields.field_values(Y.bucket.T))
```
The added zero field should make the spaces line up properly, while still keeping everything in one kernel.

nefrathenrici: Yes, sounds good! Thanks for the fix.

PR #1883, opened 2026-09-17
Raise unit test coverage and fix diagnostics found by the new tests

AlexisRenchon:
Codecov reports 55% on main, but that is `src` at 79% plus ~3000 lines of `ext` at 1%. This PR raises the number by testing what should be tested and excluding what Buildkite already covers.

## Codecov exclusion
`codecov.yml` ignores `ext/land_sim_vis/` and `LandSimulationVisualizationExt.jl` (~1230 lines). Plotting and leaderboard code runs in every Buildkite fluxnet and long-run job, so unit tests there would be redundant. Nothing else is excluded; the Fluxnet and neural snow extensions stay counted and are tested below.

## Tests added
- **Neural snow / SNOTEL tools** (`test/standalone/Snow/tool_tests.jl`, 385 assertions): re-enabled after being commented out in #1726. The two Box CSV downloads are replaced by local fixtures in `test/standalone/Snow/data/` via a new `DataTools.df_from_string`, so no network access is needed.
- **Fluxnet extension** (`test/integrated/fluxnet_sim.jl`): forcing construction, data dates, comparison data and observation-based initial conditions for US-MOz.
- **Diagnostics compute methods** (`test/diagnostics/compute_methods.jl`): for each model type, every diagnostic in `get_possible_diagnostics` must be defined and computable on an initialized state.
- **Show methods**, **`LandSimulation` interface** (float/date constructors, `solve!`, crash reporting, output-dir warning), **perturbed ERA5 drivers**, and model/utility fallbacks (`isdivisible`, `check_land_equality`, checkpoint model-hash warning).

## Changes to `src` and why they are needed
All of these were surfaced by the new tests; none change results for configurations that worked before.

- **`define_diagnostics.jl`: `a0d`, `a0a`, `pra`, `sco2_ppm` now use `conditional_add_diagnostic_variable!`.** Every other diagnostic is registered through that function, which adds the variable only if its short name is in the model's `get_possible_diagnostics` list. These four called `add_diagnostic_variable!` directly, so they were registered for every model, including ones whose compute function would throw (e.g. a bucket model got a soil CO2 ppm entry). The new test asserts that the registered set equals the possible set. Models that support them are unaffected: the optimal LAI biomass model already adds the three `a0*`/`pra` names and `SoilCO2Model` already lists `sco2_ppm`. `Val(:all)` still adds everything.
- **`land_compute_methods.jl`: `compute_10cm_water_mass!`** used `land_model.soil.domain` and the top-level `p.sfc_scratch`/`p.subsfc_scratch`, which only exist for integrated models, so the diagnostic crashed for a standalone `EnergyHydrology` even though it is in that model's possible list. It now uses the soil-owned scratch fields and the output field.
- **`land_compute_methods.jl`: new `compute_photosynthesis_net_leaf!` and `compute_respiration_leaf!`.** The `"an"` and `"rd"` diagnostics referenced these functions but they were never defined, so requesting either raised `UndefVarError`.
- **`land_compute_methods.jl`: `compute_soilco2_ppm!` accepts `SoilCO2Model`; snow diagnostics accept `SnowModel`.** Both standalone models list these diagnostics as possible, but the compute methods only accepted integrated models. The cache paths are identical.
- **`perturbed_drivers.jl`: default `time_interpolation_method`.** The default was `LinearInterpolation(PeriodicCalendar(Dates.Year(1), DateTime(Dates.year(stop_date))))`, but `stop_date` is not an argument of either function and no such global exists, so relying on the default raised `UndefVarError`. It now defaults to `LinearInterpolation(PeriodicCalendar())`, matching `prescribed_forcing_era5`. Note the semantic difference: the original expression intended a one-year period anchored to the stop year, which only differs from wrapping over the whole file when the forcing spans more than one year. For the 2008 low-res file used by the two long-run experiments that call these functions, the two are equivalent. If the anchored version is the intent, the alternative is to add `stop_date` as a positional argument like `prescribed_forcing_era5`; happy to do that instead.
- **`ext/data_scraping/WebTools.jl`:** `df_from_url` is split into a download step plus `df_from_string`, so the SNOTEL tests can parse local fixtures with the same code path.

## Wall time (local, Julia 1.12, single process, excluding package load)
| Test | Group | Time |
|---|---|---|
| Neural snow tools | core | ~63 s |
| Show methods | core | 11 s |
| LandSimulation interface | core | 27 s |
| Utilities / variable types additions | core | < 5 s |
| Diagnostics compute methods | simulations | ~4 min (LandModel 1.5 min, SoilCanopy 1 min) |
| Fluxnet extension additions | simulations | ~45 s |
| Perturbed drivers | simulations | ~15 s |

Roughly +2 min on the core job and +5 min on the simulations job. The simulations job needs the `fluxnet_sites` artifact (176 MB, cached in the CI depot).

Not run locally: the complete core and simulations groups (killed for memory while running concurrently); every touched file was run individually and passes, and the unchanged `diagnostics_tests.jl` passes against the src fixes.


test/shared_utilities/show_methods.jl
```diff
@@ -0,0 +1,274 @@
...
+using ClimaLand.Soil.Runoff:
+    NoRunoff, SurfaceRunoff, TOPMODELRunoff, TOPMODELSubsurfaceRunoff
+import ClimaLand.Parameters as LP
+import ClimaCore
+import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput
+
+# Every custom show method should produce a multi-line text/plain form that
+# names the type, a one-line compact form, and `summary` should match the
+# compact form.
+function check_show(x; typename = string(nameof(typeof(x))))
```
kmdeck, line 23: @imreddyTeja Alexis improved our test coverage of src code. What do you think about testing the show methods? it's nice to have a higher % of code covered on our homepage, but Im not sure if testing the show methods really makes sense (besides making sure they execute)?

test/diagnostics/compute_methods.jl
```diff
@@ -0,0 +1,223 @@
...
+
+function initialize_with(set_ic!, model, t0)
+    Y, p, cds = initialize(model)
+    set_ic!(Y, p, t0, model)
+    set_initial_cache! = make_set_initial_cache(model)
+    set_initial_cache!(p, Y, t0)
+    return Y, p
+end
+
+function set_soil_ic!(Y, soil, ::Type{FT}) where {FT}
```
kmdeck, line 46: I dont think the actual IC matter at all for these tests. do you want to use the set_ic functions we already have to simplify things?

test/diagnostics/compute_methods.jl
```diff
@@ -0,0 +1,223 @@
...
+        forcing,
+        LAI,
+        toml_dict,
+        domain,
+        dt;
+        prognostic_land_components = (:canopy, :snow, :soil, :soilco2),
+    )
+    Y, p = initialize_with(model, t0) do Y, p, t0, model
+        set_soil_ic!(Y, model.soil, FT)
+        Y.soilco2.CO2 .= FT(6e-5)
```
kmdeck: if you do keep the set_ic! functions here, for individual components, move these to a set_soilco2_ic! function?

test/integrated/fluxnet_sim.jl
```diff
@@ -347,3 +351,162 @@ end
...
+        lat,
+        long,
+        time_offset,
+        atmos_h,
+        start_date,
+        toml_dict,
+        FT;
+        split_precip = false,
+    )
+    @test evaluate(atmos.snow_precip, t) == 0
```
kmdeck: do we know that there would be snow precip at this time if split_precip was true? Or is it possible it is not snowing at this time?

test/integrated/fluxnet_sim.jl
```diff
@@ -347,3 +351,162 @@ end
     @test Vcmax25 == FT(2.5e-5)
     @test SAI == FT(0)
 end
+
+@testset "US-MOz forcing, data dates and comparison data" begin
```
kmdeck: It may be that, since these tests are mainly just checking that the functions run, they are already covered in buildkite. It might be better to reserve the unit tests for tests that check that something == something we expect, but I am not sure we can really do that here (without looking in the data and figuring out what some values should be -not sure that is worth it, though it could uncover bugs..)

test/diagnostics/compute_methods.jl
AlexisRenchon: Done for LandModel and BucketModel, which now use `make_set_initial_state_from_atmos_and_parameters`. The standalone models have no parameter-based IC helper in src, so they keep small local helpers, with a comment saying so.

test/diagnostics/compute_methods.jl
AlexisRenchon: Done, moved into `set_soilco2_ic!`.

test/integrated/fluxnet_sim.jl
AlexisRenchon: Good catch: there is no precipitation at all at that time in the US-MOz data, so the check passed with or without partitioning. It no longer matters now that these tests are removed (see below).

test/integrated/fluxnet_sim.jl
AlexisRenchon: Agreed. Removed the new Fluxnet tests and excluded the Fluxnet extension from Codecov, since Buildkite already runs the site simulations.

test/shared_utilities/show_methods.jl
imreddyTeja: I think that testing the show methods isn't strictly needed, but it is OK to have. Testing `show` methods in the way done below seems OK to me.  The only downside I can think of for having these is if they add a non negligible amount of time to the ci. I do think some of the tests below only test a very small part of what is shown, but at least they also make sure the methods can execute.

test/diagnostics/compute_methods.jl
```diff
@@ -0,0 +1,225 @@
...
+    @test allunique(names)
+    Diagnostics.define_diagnostics!(model, names)
+    @test Set(keys(Diagnostics.ALL_DIAGNOSTICS)) == Set(names)
+    for name in names
+        diag = Diagnostics.get_diagnostic_variable(name)
+        out = diag.compute!(nothing, Y, p, t)
+        @test out isa ClimaCore.Fields.Field
+        out .= 0
+        diag.compute!(out, Y, p, t)
+        @test out isa ClimaCore.Fields.Field
```
imreddyTeja: nit: duplicate of line above

test/shared_utilities/simulations.jl
```diff
@@ -0,0 +1,155 @@
...
+        solve!(simulation)
+        @test date(simulation._integrator.t) == stop_date
+    end
+end
+
+@testset "LandSimulation output directory warning, FT = $FT" begin
+    start_date = DateTime(2005)
+    stop_date = start_date + Second(round(Int, tf))
+    ClimaLand.Diagnostics.define_diagnostics!(model, ["rn"])
+    tmpdir = mktempdir(".")
```
imreddyTeja: nit: do the dirs used in this test need to be in the base package directory? When they are made in the `ClimaLand.jl/` dir, they can pollute the git status until the process ends. I have accidentally commited tmp dirs and files because of this before.

test/standalone/Snow/tool_tests.jl
```diff
@@ -15,6 +15,7 @@ ClimaComms.@import_required_backends
 
 using Flux, Adapt, JLD2, InteractiveUtils #leftover weak dep for ConstrainedNeuralModelExt
 using Downloads, Statistics, DataFrames #leftover weak dep for SNOTELScraperExt
+using Sockets
```
imreddyTeja, line 18: should this be added to `test/Project.toml`?

test/shared_utilities/show_methods.jl
```diff
@@ -0,0 +1,274 @@
...
+    out = sprint(show, MIME("text/plain"), x)
+    @test occursin(typename, out)
+    @test count(==('\n'), out) <= 12
```
imreddyTeja: magic number for upper bound - it would be good to add a comment explaining why 12 was chosen

imreddyTeja (approved):
thanks! I left some minor nits

kmdeck (approved):
thanks Alexis!

test/diagnostics/compute_methods.jl
AlexisRenchon: Removed the duplicate check.

test/shared_utilities/simulations.jl
AlexisRenchon: Good point, both dirs now come from `mktempdir()` in the system temp dir, so nothing is created in the repo.

test/standalone/Snow/tool_tests.jl
AlexisRenchon: Yes, added `Sockets` to `test/Project.toml`.

test/shared_utilities/show_methods.jl
AlexisRenchon: Replaced it with a `MAX_SHOW_LINES` constant and a comment. The longest current text/plain output is 6 lines, so 12 leaves headroom while still catching an accidental dump of a full field or array.

PR #1888, opened 2026-09-22
Update deps for ClimaCore v1.0.1

imreddyTeja:
This should fix the performance regression
and increased compile times.

## Purpose 
Updates manifests to use new CC patch

## To-do
- verify long runs and ci timings return to CC v0.16 baseline

----


kmdeck:
thank you, Teja! This does fix the compilation time problem, so that is a huge win. The SYPD is lower on this branch for the soil and bucket, though, by a somewhat significant amount, but it's a little hard to tell because "estimated" and "instantaneous" differ a lot. But SYPD for snowy land is about the same compared to CC v1. 
For reference, from the issue:
```
Two days before yesterday, the global land model simulation (2 years long) on GPU took 40 minutes (130SYPD) including setup, bucket took 15 (330-340SYPD), soil alone took 16 (290-300SYPD): https://buildkite.com/clima/climaland-long-runs/builds/5004
If I launch a build on main now, I see:
https://buildkite.com/clima/climaland-long-runs/builds/5006
global land: 120 SYPD (<10% slowdown), but 85 minutes to run! it seems to take a very long time to even make the simulation - even after 40 minutes it is not clear that the simulation is running yet
soil: 270 SYPD (<~10% slowdown) - 20.5 minutes vs 16
bucket: 300 SYPD (~10% slowdown) - 18 minutes total vs 15 total
```

kmdeck (approved):
@imreddyTeja thank you for working on this. I'll approve it and you can merge when it's ready - not sure if you are tackling all the performance issues in this PR/release of CC or if some would be follow-on.

imreddyTeja:
> @imreddyTeja thank you for working on this. I'll approve it and you can merge when it's ready - not sure if you are tackling all the performance issues in this PR/release of CC or if some would be follow-on.

@kmdeck This is probably worth merging in just for the reduction in compilation time / ci time. I'm still looking into the performance regression. I re-triggered 2 more long run builds on this branch, and some of the SYPDs changed

| Model              | Build baseline SYPD| Build 1 SYPD |  Build 2 SYPD|  Build 3 SYPD|
| :---------------- | :------: | ----: | ----: |  ----: |
|Bucket|   334   | 210 | 315 | 295 |
|Soil|   291   | 214 | 273 | 263 |
|Full Land|   133   | 122 | 128 | 128|


I'm a bit bewildered by these results. I'm running the soil model locally to compare CC v0.16.2 to v1 using the latest ClimaLand commit. It seems like without any diagnostics, the time a single step takes is almost exactly the same. I think I need to run the "baseline" (last weeks long run) a few more times to see what the expected variation in SYPD is.

imreddyTeja:
Leaving this comment to document my findings:

I am unable to reproduce the SYPDs from last Wednesday's long run build locally. I used the same land commit and climacore version. My local version is a bit slower, and matches the SYPD seen in this PR's builds. I want to retry this using the exact manifest produced by the baseline build. I also want to retry this when the clima cluster is less congested. I also noticed that any step that reads/writes to disk is an order of magnitude slower.

kmdeck:
> Leaving this comment to document my findings:
> 
> I am unable to reproduce the SYPDs from last Wednesday's long run build locally. I used the same land commit and climacore version. My local version is a bit slower, and matches the SYPD seen in this PR's builds. I want to retry this using the exact manifest produced by the baseline build. I also want to retry this when the clima cluster is less congested. I also noticed that any step that reads/writes to disk is an order of magnitude slower.

maybe a climadiagnostics or climautilities change?

kmdeck:
> This is probably worth merging in just for the reduction in compilation time / ci time.

agreed!!

imreddyTeja:
I ran the long runs again [here](https://buildkite.com/clima/climaland-long-runs/builds/5026#01a0d282-dfe0-4d00-b372-8e634a5997fd) when the clima cluster had fewer people using it, and the full "baseline" SYPD from climacore v0.16 is recovered for the full land and full soil. The bucket is slightly slower, but that may be within the noise.

PR #1889, opened 2026-09-23
benchmarks: Write the timings to a file

nefrathenrici:
`profile_and_benchmark` in `experiments/benchmarks/benchmark_utils.jl` reported its mean and standard deviation only through `@info`. This PR also writes them to timings.toml in the benchmark's output directory:
```
average_timing_s = 0.0794
std_timing_s = 0.00123
```
I am adding ClimaLand benchmarks to ClimaCore's end to end performance pipeline in https://github.com/CliMA/ClimaCore.jl/pull/2655 and this PR will make it easier to parse the performance metrics.

No change to what the benchmarks do. TOML was already a dependency, and the file is only written when the benchmarks run, so the external-profiler path is untouched.


imreddyTeja (approved):
Thank you!

PR #1891, opened 2026-09-23
Use NaN-aware regridding in postprocessing

### LINKED ISSUE #987: Mask-aware interpolation needs in ClimaLand.jl
It's becoming increasingly apparent that mask-aware interpolations will be helpful to land model development in multiple ways, including the following. 

Related to #488, but doesn't solve that issue

#### Interpolating parameters/inputs to model grid
Parameter values are not guaranteed to be physical over the ocean, and may even be NaN. We need to make sure we don't include those values when interpolating parameters/inputs to our model grid.

#### Using parameter/input maps with different resolutions
Similar to the previous point - Different resolutions will lead to mismatch at the coastlines. We need a general solution that allows us to handle input maps with a variety of land/ocean masks.

#### Plotting
Right now, our global plots use regular interpolation everywhere, even at the coastlines. This produces chunks of NaN along all coasts, which inhibits our ability to visually inspect our outputs and determine regions of stability, and also lowers the quality of our output plots. See [this build](https://buildkite.com/clima/climaland-long-runs/builds/2773#0193dc97-3170-48c2-bb0c-15b55fd56576) for examples.

## Proposed solution
We want to have a function `interpolate` with an interface like this:
```
interpolate(target_x::FT, target_y::FT, data::Arr{Nx, Ny}, source_xs::Nx, source_ys::Ny, mask::Array{Nx', Ny'}, mask_xs::FT{Nx'}, mask_ys::FT{Ny'})
```

The first 2 argument specify a point P on our simulation grid that we want to interpolate to.
The next 3 arguments provide information about the data to be regridded (e.g. a parameter map), and its grid
The next 3 arguments provide information about the land/ocean mask to be used in the interpolation, and its grid

### Algorithm outline
- For each of the 4 closest source data points to P, determine if ocean or land
  - Find 4 closest mask points, check if they're ocean or land
  - If >= 2 are land, label source data point as land; otherwise ocean
- If all 4 points land -> bilinear interpolation
- If 3 points land -> ?
- If 2 points land -> linear interpolation between these two points (simplification, loses second dimension)
- If 1 point land -> use this value

## Alternative solution
Temporarily enlarge the land areas of the mask, use this larger mask to interpolate inputs/parameters to the model grid. Then, use the normal mask again to only select input values within the real land.
- Q: what if input values are 0 over ocean and this just becomes enlarged land area? Not sure I understand this approach

ph-kev:
closes https://github.com/CliMA/ClimaLand.jl/issues/987 - This PR uses the NaN-aware regridding feature of the main branch of ClimaAnalysis for the postprocessing pipeline to properly regrid along the coastlines. New functionality in ClimaAnalysis also simplifies how masks are handled in the postprocessing pipeline. I asked a LLM to implement the functionality and I reviewed the changes.

Here's a plot of `FlagshipCarbonMetrics_gpp_bias_plot.png`.

On the `main` branch:
main-FlagshipCarbonMetrics_gpp_bias_plot

On this branch:
nan-regrid-FlagshipCarbonMetrics_gpp_bias_plot

TODO

- [ ] Make a release of ClimaAnalysis
- [ ] Update compat and manifests


docs/src/leaderboard/leaderboard.md
```diff
@@ -19,17 +19,24 @@ Each loader exposes a `Base.get` method that returns a preprocessed
...
 ### Computation
 
-As of now, the leaderboard produces bias plots with the global bias and global
-root mean squared error (RMSE). These quantities are computed for each month
-with the first year of the simulation not considered as that is the spinup time.
-The start date of the simulation is 2008 which means that only the year 2009 is
-used to compare against observational data.
+Before computing the bias and RMSE, the observations are resampled onto the
+simulation grid with NaN-aware resampling, so a cell next to the ocean is kept
+when most of its interpolation weight comes from valid data. Both variables are
```
kmdeck: What does "most" mean?

kmdeck:
I think this doesnt quite close that issue. we also would like to use a mask when interpolating parameter data to the model grid

